PR #7, opened 2016-07-06
Create language_sv.h

xpandedreality:
Swedish translation. I have avoided translating some technical terms.

### DIFF
--- a/Firmware/language_sv.h
+++ b/Firmware/language_sv.h
@@ -1,0 +1,244 @@
+/**
+ * Swedish
+ *
+ * LCD Menu Messages
+ * Please note these are limited to 17 characters!
+ *
+ */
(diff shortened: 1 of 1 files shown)


3d-gussner:
I am translating to German and Dutch and tried to get the 'Umlaute' ÄäÖöÜü on the display, BUT the HD44780U display PRUSA is using has an A00 firmware and not A02. I had to change 'Ä' to 'Ae' 'Ö' to 'Oe' and so on. I have no idea if there is a way in Swedish to write special characters in a different way. If you can they maybe will add it to their firmware. Check this https://www.sparkfun.com/datasheets/LCD/HD44780.pdf

3d-gussner:
@xpandedreality This PR have inactive for long time, can you please close it?

PR #8, opened 2016-07-06
Fix of wrong displaing minutes and seconds

pacaj2am:
Status is displaing wrong minutes and seconds, when printing longer than 1h


michalprusa:
Thanks for noticing, it's fixed!

PR #11, opened 2016-10-02
Update max_jerk in ConfigurationStore.cpp

thinkyhead:
It looks like EEPROM isn't in use, but I figured I might as well offer this patch for consistency.


bubnikv:
Thanks. I may integrate your change, but please keep in mind that we do not consider writing these numbers into the eeprom, so the whole functionality of storing these values into the eeprom may disappear in the future from our firmware.

PavelSindler:
Thank you. It was implemented in https://github.com/prusa3d/Prusa-Firmware/commit/fcce374a14b94ae6dc01466d9262a92d3ae1bf50 together with storing settings in EEPROM.

PR #12, opened 2016-10-02
Jerk is based on change in speed

thinkyhead:
This seems like it should be using the change in speed and not the absolute speed.


bubnikv:
The original code is correct. The change in velocities is handled by the following block. The block you modifies serves as a jerk limitation when starting from a halt.

PR #13, opened 2016-10-02
UNLOAD FILAMENT sequence

Eriobis:
Unload filament will push filament before unloading.
In configuration_prusa.h you need to add another line 

```
// Unload filament commands
#define UNLOAD_FILAMENT_0 "M83"
#define UNLOAD_FILAMENT_1 "G1 E10 F400" 
#define UNLOAD_FILAMENT_2 "G1 E-90 F6000" 
```
- The LOAD filament state machine has been changed to defines, for a cleaner code.
- Other minor indent correction in ultralcd.c


bubnikv:
Thanks. In regard to the push before unload, this is intentional. This is called "ramming" and its purpose is to get the molten material out of the melt zone before removing the filament. This sequence minimizes chances of pulling thin strings after the filament end, which often jam the filament in the extruder during the removal.

3d-gussner:
@Eriobis This PR have been inactive for long time, can you please close it?

PR #31, opened 2016-12-25
E3D PT100 with Amplifier and ENCODER bugfix

ayourk:
The ENCODER steps and pulses for the LCD that is currently being used is off from the physical rotation clicks present.  This causes people to sometimes select a menu option other than what is planned.

Added #defines and temp table information about E3D's PT100 temperature sensor so it can be used with and without their Amplifier.  Allowed for an increased extruder temperature to 410.  Added the results of my PID autotune with the PT100 temperature sensor and made it selectable via the previous #defines.


lbussy:
Please consider merging - the encoder clicks being off is a most annoying behavior.

rob5standingby:
I would love to see the encoder clicks merged in soon. I have the same issue, and it's a bit frustrating at times :)

bubnikv:
I am sorry we cannot react faster to the community requests. Trust me, we
are doing the best we can, but there is just a handful of us and there is
much more on our hands than the firmware.

Currently our new hire Pavel is integrating multi material extensions into
the MK2 firmware. I am sure he will take care of this request after that.

PR #36, opened 2017-01-21
Slow extruder fan for high temperature printing

eyal0:
This PR slows the extruder's cooling fan when the requested temperature is over 200C.

## Background

There are two fans on the print head.  One is the model fan to cool the extruded plastic so that it sets quickly and one is the extruder fan to cool the upper part of the print head so that it doesn't melt the plastic holding it in place nor the plastic feeding in to the print head.

When both the model fan and extruder fan are running at 100% and the print head is near a surface, the breeze from the model fan will be reflected back up to the print head and cool it.  In this situation, the print head can't maintain more than ~230C and will report "THERMAL RUNAWAY" if the target temperature is above ~230C.

## Workarounds

For most people, the solution is to turn down the model fan.  This has the disadvantage of making poor bridges.  It is not possible to control the extruder's fan from outside the firmware.

## Possible solution

A search on the web shows that some people have had success with turning down the extruder's fan.  With the extruder fan at 37.5% (96 of 255), it's possible to print at 255C with the model fan at 100%.  In practice, prints turn out much nicer and bridging is greatly improved.

This PR has no effect if the target temperature is less than 200C.  If the target temperature is more than 200C, the fan is set to just 96/255.  In case one fan is set to cool multiple extruders, the highest target fan speed is used.

For prints at less than 200C, there will be no change in operation of the printer.  For prints above 200C, it should be possible to get much nicer results.  The risk is that the decreased dissipation of heat will lead to wear on printer parts but so far I've not noticed any failure.


bubnikv:
At the first sight this sounds like a dangerous non-systematic proposition.
The purpose of the extruder fan is to control the melt zone length. In that
regard an opposite of your proposal makes sense: to lower the extruder fan
speed at a lower extruder temperatures. Also please keep in mind, that the
speed of the fan is not controlled. Some dust in the fan bearings will slow
the fan somehow, so it makes sense to let it rotate somehow faster up front.

eyal0:
A better solution would be welcome.  Turning down the model's cooling in order to print ABS gives poor results.  There must be a better way!

eyal0:
What if we made it programmable?  If we can come up with a reasonable gcode for it and I implement it, would you consider merging it?

bubnikv:
A better solution would be to insulate the heater with something like this:
https://www.aliexpress.com/popular/3d-hotend-insulation.html
Some customers reported success with it.

or the E3D sock
http://e3d-online.com/Socks-Launch
but it is known to not hold well over the heater block. You may try to use a piece of wire to hold it in place.

I think a piece of thin aluminum foil may work as well, or a piece of fiber glass cloth, if you are able to affix it over the heater block somehow.

eyal0:
We've insulated with a E3d sock and had great results.

PR #45, opened 2017-02-14
Added German/Deutsch and Dutch/Nederlands final

3d-gussner:
- Added German and Dutch languages, to make it easier to non-english speaking persons, especially kids, to operate the MK2.
- Fixed compiling issue with new 3.0.3.10RC1 feature 'MSG_PAPER'.
- Changed baudrate to 250000 to get file transfers via USB bit faster.
- These files contain the original PID settings, but i added my to have them even after a hard reset.
-Deutsch und Niederländisch als Sprachen hinzugefügt, um nicht Englisch sprechenden Menschen, speziell Kindern, den Gebrauch vom MK2 zu erleichtern.

-Nederlands en Duits als talen toegevoegd, om mensen die niet Engels spreken, speciaal kinderen, de gebruik van de MK2 gemakkelijker te maken.


paul999:
To be honest, I think the Dutch translation can see a huge improvement, some of the translations seem to be done with Google translate, and make no sense in my opinion. Also some stuff is not translated at all, while other stuff is over translated.

3d-gussner:
Hey Paul999,
you are more then welcome to improve the translation. I am non-native Dutch speaking person BUT there were few people (Dutch and Belgium) reviewing the translation. Sometimes it is difficult to use the 'best' word, because of the limited characters the LCD can display. I am now busy to compare this translation with that one from Marlin github. Maybe if we have more people reviewing the Dutch messages, Prusa will add it to the official firmware next time. Now they didn't because nobody there is speaking Dutch. So please join and let me know what can be done better. Here the link to my fork https://github.com/3d-gussner/Prusa-Firmware

PR #56, opened 2017-03-07
Bug: M20 - List SD card

alcor6502:
When the SD card is inserted in a Mac computer some special directory are written. The Prusa firmware is unable to navigate those directories and generates an error sent on the serial line.

The BUG is the instruction SERIAL_ECHOLN it does not work with string pointers and therefore garbage is sent on the serial line.

The proposed correction is to use the right instruction: SERIAL_ECHORPGM that works with string pointers.
I suggest to change also the original error message into MSG_SD_CANT_ENTER_SUBDIR with makes more sense and have semicolon and a space to his end.


PavelSindler:
Thank you @acortellini.

Issue #29, opened 2016-12-20
Please update the build instructions


pboschi:
The build instructions need to be more precise.
For example:
1) you need a specific Arduino IDE version (1.6.8) or to do  some changes 
2) the stk500boot_v2_mega2560.hex file is missing
3) the generate_version_header_for_marlin file is missing
4) a bit of explanation on why you are doing each step usually helps

ryanneufeld:
👍

pboschi:
The following may be more reasonable instructions, if you want to update the readme file.

# Original Prusa i3 MK2 Firmware

## General instructions

Pre-compiled hex files for all printers by PRUSA RESEARCH are available in hex_files folder.

Just clone the repo and flash it to the firmware.


## Build instructions




### Step 1 - Download and Install/Unpack Arduino

Download and install Arduino; suggested and verified version is the 1.6.8 from:
https://www.arduino.cc/en/Main/OldSoftwareReleases#previous

If you also use Arduino for other purposes, it is suggested to use a stand-alone version rather than a system-wide install (e.g., use the "Windows" version instead of "Windows Installer"), and use it only for Prusa i3 MK2 firmware generation.



### Step 2 - Download PRUSA i3 MK2 Firmware

Clone the prusa3d/Prusa-Firmware (Branch: MK2) repository to a local path.

If you are not an expert git user, it is suggested to Download it as ZIP from:
https://github.com/prusa3d/Prusa-Firmware/tree/MK2



### Step 3 - Prepare the Arduino toolchain


#### Step 3a - Remove the LiquidCrystal library

Due to name collision, you have to remove the Liquid Crystal library of the original Arduino installation. To do so, you may rename the "Arduino\libraries\LiquidCrystal" directory to "Arduino\libraries\LiquidCrystal_", or remove it altogether.


#### Step 3b - Add RAMBo hardware / library support

Copy the contents of the ArduinoAddons/Arduino_1.6.x directory from the downloaded firmware repository to the root of the installed arduino path.


### Step 3c - Add the bootloader

Download the bootloader image from:
https://raw.githubusercontent.com/arduino/Arduino/master/hardware/arduino/avr/bootloaders/stk500v2/stk500boot_v2_mega2560.hex

Save a copy of it on each of the following directories:
* Arduino/hardware/arduino/avr/bootloaders/stk500v2/
* Arduino/hardware/marlin/avr/bootloaders/

Double-check that the file has the .hex extension and not .txt.


[...]

PR #77, opened 2017-04-16
Better instructions for building and installing firmware

eyal0:
More detailed instructions copied verbatim from https://github.com/prusa3d/Prusa-Firmware/issues/29#issuecomment-294346773

`wget http://api.github.com/repos/prusa3d/Prusa-Firmware/issues/comments/26898529`

Fixes https://github.com/prusa3d/Prusa-Firmware/issues/29


Sebastianv650:
They might merge it some day if it is as perfect as it can be. Maybe we can make it a bit better..

There was the question if we can have this merged into the new "Community Mod" Firmware. For me, there are some questions to answer first:
- Step 1: You are using a quite old (1.6.8) Arduino IDE version. Any reason for this?
- Step 2: I'm missing a note that you get a "nightly" snapshot if you use the MK2 and not a release branch.
- Step 3: I'm not having problems with a liquid crystal library, never deleted one. Outdated information?
Further, I would recommend using the [Board manager](http://reprap.org/wiki/Rambo_firmware#Arduino_1.6.4.2B_Board_Manager_Plugin) instead of copying maybe outdated files for the Rambo board.
Regarding the bootloader, I don't understand this up to now. I'm always using the Arduino IDE to flash the FW and beside it's telling me the stated bootloader isn't found it's running fine. Also new flashes are no problem. Is this only necessary if you flash the .hex file over the Prusa FW Tool?
- Step 4: There are more configuration files than just two. I think you have to clarify here what the differences are.
- Step 6: You pointed out some configuration options the most users will never touch manualy like the PID values. Is there a reason for this? If so, I feel better if there are links to further details. If not, a hint about the extruder step/mm value might be a more often needed value as an example.
- Step 7: Why not use the Arduino IDE for flashing the new FW? I guess there are more ways to break things when you are using a .hex compared to the IDE way of doing it. Step 8 could be deleted then.

I would prefer two ways of flashing a new FW for end users:
- The ones who only wants to try an RC or nightly build: Point them to a precompiled .hex if the PR regarding Travis CI builds is merged. Very low risk to break things.
- If somebody wants to play with settings or doesn't like .hex flashing, use Arduino IDE. No need to do the work-around with creating hex files and use a 3rd tool for flashing.

If you want to update this PR to make it more intresting for Prusa team to get it merged, you might also have a look at the [Marlin installation instructions](http://marlinfw.org/docs/basics/install.html).

If you update this PR, we can merge it over in [my Community Mod fork](https://github.com/Sebastianv650/Prusa-Firmware) :)

fiveangle:
Nice work @Sebastianv650:1st_place_medal: 

Those who wish to effectively print at > 40mm/s on their MK2 salute you :+1:

3d-gussner:
@eyal0 This PR have been inactive for long time, can you please close it? And the build instructions are updated.

Issue #123, opened 2017-06-20
Combine the single material and multi material firmware into one


GurliGebis:
Currently, there is two different firmwares, one for the single material setup, and another for the multi material version.

This causes confusion, and adds to the maintenance for both the user and you guys. (Every time there is a new firmware, it has to be built 4 times instead of 2, and the user has to pick the right one between not only the RAMBo boards, but also the number of materials).

This seems fairly easy to fix in the firmware.

GurliGebis:
@PavelSindler What do you think about the changes in #124 ? (You seems to be the one doing the development of the firmware)

GurliGebis:
Whoever sees this, just ignore it - I won't be keeping this branch up to date anymore, since I cannot get a single word from upstream if this is even a good idea.

PR #124, opened 2017-06-20
Unified mm firmware

GurliGebis:
Here I have changed the multi material logic to be a setting stored in the EEPROM instead of being a different firmware.

Now, in the Settings menu, there is a menu item for changing the printer type to/from MultiMaterial.

All the ifdef/ifndef's has been changed to read the variable storing which mode we are in.


GurliGebis:
If this gets merged, it fixes #123

GurliGebis:
Fixed merge conflicts with upstream

GurliGebis:
Changed the new MK1 config to match the other configs

GurliGebis:
Whoever sees this, just ignore it - I won't be keeping this branch up to date anymore, since I cannot get a single word from upstream if this is even a good idea.

3d-gussner:
I like the idea having one firmware source for SM and MM. But i think it doesn't make that much sens to be able to switch via the LCD screen, because most will have SM or MM and will not switch frequently (too much work). Prusa should develop ONE firmware version and release hex files for SM and MM.

GurliGebis:
How about having it changeable from the reset menu? That way it won't get changed by accident, and it would be more of an advanced system setting.

3d-gussner:
That is an option but the change between SM and MM takes quite long and then the user can also upload the right hex file. Prusa also removed lot of Marlin code in their version, maybe to save memory for additional functions and this would add some bytes to used memory and EEPROM.

It would make more sens if there was a way to 'read' if there is a SM or MM installed so that the firmware would change automatically. But that will be difficult because the two pins 16/17 are used to control the 4 switches/Nema17 extruders and autosens would need another pin...change of the super switch ... etc.

GurliGebis:
I agree, but I don't think there is a way to detect it reliably.
What I fear is alot of people mixing up the different hex files, when it should be possible to have it merged into one.
I don't think we are that low on mempry are EEPROM space, that those few extra bytes for the setting and variable will be a problem.

3d-gussner:
Good point, looking from customer and support view it makes sens to get ONE firmware hex file. Forget too often that some of us (including ME) are used/comfortable with modifying and flashing firmware and we should not expect the regular user to do that.

- Easier to flash firmware for customers just having ONE firmware hex file
  - User just have to choose once and future updates will read the EEPROM to get the right settings active.
- Less support calls
  - and if there is a support call the user just have to switch SM/MM mode and not re-flashing it

I didn't check all your changes so sorry asking if that is already implemented:
The EEPROM value your are adding shouldn't it have tree states
 1. not configured = User should be forced to choose if it is SM or MM. In case the EEPROM get cleared, corrupt, overwritten, etc. you can ask the user via LCD screen what configuration she/he uses.
2. Single Material. I am not sure if that should be like the standard that will become active after a timeout/delay of point 1 ?! Don't know if a Prusa i3 MK(s) MM will work with the SM firmware different 'DEFAULT_AXIS_STEPS_PER_UNIT' and power current for Extruder; 'BOWDEN_LENGTH'; 'FIL_LOAD_LENGTH', etc.? 
3. Multi Material.

What do you think? Is there ANY chance that the EEPROM get cleared, corrupted, overwritten, etc. to do this?

GurliGebis:
It defaults to single material, so everyone flashing their printer with it will see no difference.
So a simple bool is enough, since it defaults to false, which in this case means single material.

If the EEPROM get's cleared, the value will be read as 255, which will make it default to 0, which is false, which in turn means single material.
So if the user does a factory reset, they would have to go back and change it to multi material again.

PavelSindler:
We had discussion about this, but in the end we decided that we do not want to have option for switching from SM to MM in LCD menu. It is also beneficial that we can save some memory using conditional compilation.

GurliGebis:
Thats okay :-) I have seen people (both on youtube and facebook) forgetting and/or mixing the firmwares up when upgrading. Not sure what the best solution would be, if it can be done better at all.
I'm closing this.

PR #135, opened 2017-07-01
Fix/implement M110 command. Allow colons in command - don't treat as EOL

thess:
Here's a couple of minor items affecting operation and building. The redundant definitions in **pins.h** are the cause of about 20% of the compiler warings.

- [Fixed] M110 commands return error: 'Invalid M code'
- [Fixed] Allow colon in commands. I believe them to be allowed and not to be treated as end-of-line. Try sending "M117 July 1,2017 14:30" and watch your print die. ref: Marlin FW.
- [Fixed] Multiple redundant pin definitions (removed unused rather than fixing conditionals)

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -1379,7 +1379,6 @@ void get_command()
         continue;
     if(serial_char == '\n' ||
        serial_char == '\r' ||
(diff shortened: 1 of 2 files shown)


thess:
Closing this - will re-submit without pins.h edits. Compiler warnings to be covered by a separate PR

PR #137, opened 2017-07-03
Fix/implement M110 command. Allow colons in command - don't treat as EOL

thess:
[Fixed] M110 commands return error: 'Invalid M code'
[Fixed] Allow colon in commands. I believe them to be allowed and not to be treated as end-of-line. Try sending "M117 July 1,2017 14:30" and watch your print die. ref: Marlin FW.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -1379,7 +1379,6 @@ void get_command()
         continue;
     if(serial_char == '\n' ||
        serial_char == '\r' ||
-       (serial_char == ':' && comment_mode == false) ||
        serial_count >= (MAX_CMD_SIZE - 1) )
     {
       if(!serial_count) { //if empty line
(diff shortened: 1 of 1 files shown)


Sebastianv650:
That's a good one, finaly an answer why printing over Pronterface always says Invalid M code!

PavelSindler:
Thank you @thess for your work. It really seems that treating colon as end of line has no sence. I've just noticed that in your pull request colon is still considered as end of line if read from SD card.

PavelSindler:
We had discussion about colon as EOL and decided that we would like to keep the way which is currently used. Some users prefer having multiple gcodes on one line separated by colon. But we will add escape character to allow using colon in M117.

GurliGebis:
@PavelSindler Could you merge the M110 part of this pull request?

thess:
Using a colon for end-of-command is not part of what passes for a RepRap gcode standard. There is AFAIK, no documented usage of this. If you are interested in multi-commands without requiring an OK, you should investigate what has been discussed (maybe not implemented yet) for streamlining commands. Forcing an "escaped" usage to allow valid comment characters is not going to help much since a lot of M117 commands are issued by other software / apps.

If it would help, I can re-submit the M110 fix separately.

PR #143, opened 2017-07-11
retrieve settings from eeprom changed for axis steps per unit, translated support menu, long beep in multi material color print fixed

PavelSindler:


lekoder:
@PavelSindler please elaborate on reason behind removing E steps per unit from EEPROM. It makes popular method of calibrating extruder unusable.

Sebastianv650:
@lekoder have a look at this PR: #151

lekoder:
@Sebastianv650 thianks for a pointer.

To summarize - form reading the code I have an impression that 3.0.12 identifies itself as `V2` in EEPROM regardless of single or multi-material flavor. #151 changes that with [this ifdef](https://github.com/prusa3d/Prusa-Firmware/pull/151/files#diff-00c94c8d6304fb8ea59d213e10e1173dR46), so single-material version still uses `V1` EEPROM signature and will load and store settings normally?

GurliGebis:
Correct, 3.0.12-1 will work like always for people with single material printers. It is only when going from single material to multi material this will have the effect of resetting the settings back to default.

PR #151, opened 2017-07-19
improved scrolling in SD card menu, storing to eeprom via M500 improved and uses different eeprom version for multi material

PavelSindler:


GurliGebis:
@Sebastianv650 would it be an idea to merge this one into the community mod? (After reverting the undo of the hack you did) :)

bubnikv:
or wait for our team to merge the "community mod" changes into the official
firmware?

GurliGebis:
Would also be an option, but for people wanting to help test the linear advance feature, it is alot easier than having to setup an arduino environment :)

Sebastianv650:
On a first look on my mobile phone, this isn't changing the "ignore e steps/mm changes" behaviour? But i guess with the two eeprom versions that should be easy to solve now. Why not in this PR?

GurliGebis:
@Sebastianv650 This fixes the problem properly, by loading the default settings if the firmware is changed between SM and MM, and only telling the user this, if they have ever run M500

PavelSindler:
Yes, it works axactly as @GurliGebis said. If M500 was used for storing setting into EEPROM and firmware is than changed between SM and MM, default setting are loaded and user is informed about that (notice that "old" settings remains stored in EEPROM). Message about loading default settings is shown at every printer startup until user saves parameters to EEPROM using M500 again.

GurliGebis:
@PavelSindler would it be possible to have a variant of the M503 gcode, that shows the stored variables, instead of the ones in memory? That way, it would be possible to see the PID variables, so the user don't have to redo the PID calibration :) (Or maybe carry over the PID values for the hotend and bed, so they are still read from EEPROM)

PavelSindler:
It would be possible to implement it, but if EEPROM version changes it also means that different variables may be now stored in EEPROM. For showing stored variables, you would need variant of M503, which would correspond to some previous EEPROM version which was used for storing the values. For each EEPROM version which could be used in the past for storing, there would be slightly different variant of M503 for showing stored variables.

GurliGebis:
I see, that makes sense :-)

Sebastianv650:
@GurliGebis 

> This fixes the problem properly, by loading the default settings if the firmware is changed between SM and MM, and only telling the user this, if they have ever run M500.

You are right, I missed the lines on the small display I guess. Time to update the fork..

PR #162, opened 2017-08-01
Fan kick improvement - long pulse (800ms) when starting, short pulse …

XPila:
…(50ms) every 4s when speed < 25.

Debug macros, debug codes (D0 - reset, D1 - clear eeprom, D2 - get/set pin).


Sebastianv650:
I don't get the reason for having short pulses at low speeds: If you need the puses to keep the fan spinning, say at 15% speed doesn't that mean the fan speed with pulses at 15% now means 20% power (the speed it would run fine without spikes)?

XPila:
Hello Sebastian, the short pulses at low speeds is experimental solution. It prevents the fan from self stop. You are right - the result power is little bit higher (+1.25%).

PR #166, opened 2017-08-07
Ported MarlinFirmware/#3109 - Provide feedback to hosts when busy

GurliGebis:
This is a port of the USB host reporting functionality implemented in Marlin back in 1.1.0
It is based on the code found here: https://github.com/MarlinFirmware/Marlin/pull/3109

It should merge without any issues, and be enabled by default.
Please review the changes and comment/change if something needs to get done.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -151,6 +151,7 @@
 //        Rxxx Wait for extruder current temp to reach target temp. Waits when heating and cooling
 //        IF AUTOTEMP is enabled, S<mintemp> B<maxtemp> F<factor>. Exit autotemp by any M109 without F
 // M112 - Emergency stop
+// M113 - Get or set the timeout interval for Host Keepalive "busy" messages
 // M114 - Output current position to serial port
 // M115 - Capabilities string
 // M117 - display message
@@ -380,8 +381,8 @@ bool cancel_heatup = false ;
   };
 
   static MarlinBusyState busy_state = NOT_BUSY;
-  static long next_busy_signal_ms = -1;
-
+  static long prev_busy_signal_ms = -1;
+  uint8_t host_keepalive_interval = HOST_KEEPALIVE_INTERVAL;
   #define KEEPALIVE_STATE(n) do { busy_state = n; } while (0)
 #else
   #define host_keepalive();
@@ -1321,13 +1322,15 @@ int serial_read_stream() {
 }
 
 #ifdef HOST_KEEPALIVE_FEATURE
+/**
+* Output a "busy" message at regular intervals
+* while the machine is not accepting commands.
+*/
 void host_keepalive() {
   long ms = millis();
-  if (busy_state != NOT_BUSY) {
-    if (ms < next_busy_signal_ms) return;
+  if (host_keepalive_interval && busy_state != NOT_BUSY) {
+    if (ms - prev_busy_signal_ms < 1000UL * host_keepalive_interval) return;
     switch (busy_state) {
-      case NOT_BUSY:
-        break;
       case IN_HANDLER:
       case IN_PROCESS:
         SERIAL_ECHO_START;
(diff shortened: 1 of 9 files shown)


Sebastianv650:
I added it to my Community Mod fork to see if it works. While it basicaly works, it's different than it behaves in Marlin: You port is sending busy messages as soon as you connect to the printer. While I can also see some sense in this way of working, Marlin would only send busy messages after a print has been started. This way, the serial log window will not get that cluttered. And it should be enough, as no host software should wait for an answer if it hasn't sent real print code to the printer before.

While I don't know how Marlin does this, I think it may control it with the motion buffer. If the buffer isn't full, this means all incomming commands have been accepted and no further one is waiting in the host software. In this case, no busy is necessary. So the check to send commands would be the 2s interval + is the buffer full?

GurliGebis:
@Sebastianv650 good to hear - I haven't been able to test it, since I'm in the middle of a print currently :)
I have taken a look at it, and I'm not sure what should be changed for it to behave like you mention (and like it should, since sending busy all the time makes no sense)

Firmware/Marlin_main.cpp
```diff
@@ -2058,6 +2106,8 @@ void process_commands()
   float tmp_motor_loud[3] = DEFAULT_PWM_MOTOR_CURRENT_LOUD;
   int8_t SilentMode;
 #endif
+  KEEPALIVE_STATE(IN_HANDLER);
```
GurliGebis, line 2112: @Sebastianv650 If you have time to test, can you try and comment out this line, and see if that fixes the constant busy signal?

Sebastianv650: No, with this change it won't send anything anymore. Is this PR based on the recent Marlin Codebase or some early test branch?

GurliGebis: @Sebastianv650 it is based on the MarlinFirmware #3109 pull request.
There is a good chance there has been some other changes since then, that needs to be included also, that I have missed :)

GurliGebis:
@Sebastianv650 I have ported the remaining host keepalive commits (that I could find, they should be there all now). Can you give it another test, and see if it works better now?

Sebastianv650:
Sorry, it's not sending anything on my Mk2..

GurliGebis:
Hmm, I don't know what is causing it then, and I'm not that well into it, that I can hook up the Arduino IDE and debug the firmware :)
Anyway, it is a good starting point for the prusa team, since it has the code from the Marlin, so it is just a matter of finding out why it is not running.

GurliGebis:
@PavelSindler Would you be able to use this as a starting point?

PavelSindler:
Yes @GurliGebis, I think this will be definately good starting point. Thank you for your work. I've just started to test it localy.

Firmware/Marlin_main.cpp
```diff
@@ -1361,6 +1411,7 @@ void loop()
   manage_heater();
   isPrintPaused ? manage_inactivity(true) : manage_inactivity(false);
   checkHitEndstops();
+  host_keepalive();
```
PavelSindler, line 1414: It seems that this is the only place where host_keepalive() function is called. Or am I missing something? This is inside void loop() function which is blocked when for example some blocking gcode is executed. host_keepalive() function should be called in function which takes care about managing heaters/steppers.

GurliGebis:
Great, and thanks :)
That might be the reason it isn't working - I don't know how to debug the firmware (just upload it using the Arduino IDE, and attach the debugger?), so not really sure how to go and test. (And I only have one printer ATM, since the wife doesn't think we have room for two - so development has to be done on the main machine, when not printing) :)

Firmware/Marlin_main.cpp
GurliGebis: That sounds likely, I haven't been able to test it though. (See explaination below)

GurliGebis:
Closing, since Pavel has a PR that has more work than this :)

PR #174, opened 2017-08-14
Sorting files in SD card menu; xyz cal. details menu updated show real skew even if axes were orthogonalized during cal.

PavelSindler:
Alphabetical sorting were imported from Marlin firmware. Sorting by time when file was added to SD card was added. Default sorting uses stack and folders are shown on the top. 
SDSORT_CACHE_NAMES and SDSORT_DYNAMIC_RAM currently not supported


3d-gussner:
@PavelSindler The LCD screen has just 4 lines so limited space. I think it would be better to move the SD Card sorting under the settings menu.

PavelSindler:
Thank you for feedback @3d-gussner . It is true, that space in SD card menu is limited, but on the other hand I think that user often enters SD card menu and then if he can't find some file, he would like to sort files. So I think it is handy to have it in the same menu. But we will test it and I will see what the feedback from other people will be.

3d-gussner:
Hi @PavelSindler , you got a point, would it be an option to use long press to de-/activate the sorting when you are in the SD card menu? Didn't check if the long press is used 'globally' to do things.

PavelSindler:
It would be also possible, but it is maybe less intuitive and it would block "move Z" funcion in SD card menu.

PR #175, opened 2017-08-15
run a reasonable LCD message on boot asap, support the second serial …

AndreeeCZ:
…port (enabled only in farm mode)


3d-gussner:
@AndreeeCZ and @PavelSindler 
Where do you connect the second serial port? TX1/RX1 are on connector P1 Tx3/RX3 are on connector P2 which are used for the LCD screen and TX2/RX2 are on connector P3 which are used for the MM upgrade.
I would like to use an ESP8266 or ESP32 with ESP3d with https://github.com/luc-github/ESP3D or https://github.com/probonopd/WirelessPrinting. Or even a C.H.I.P. with octoprint and saving the one USB port for a web cam.

Issue #132, opened 2017-06-26
Question: Why is pin 84 defined twice in 'pins.h'


3d-gussner:
This is a minor question/issue, but still what to know why you defined the pin 84 twice in the 'pins.h' file?

**#ifdef SNMM** 

**#define E_MUX0_PIN 17** This pin is connected to the Prusa Multi Material Super Switch
**#define E_MUX1_PIN 16** Also this pin is connected to the Prusa Multi Material Super Switch
**#define E_MUX2_PIN 84** But this one is NOT connected to the Prusa Multi Material Super Switch which just uses a 4-pin connector VCC,PIN17,PIN16,GND to miniRAMbo P3. So why is it defined?

and later in the 'pins.h' you find

**#define BEEPER 84** Which is connected to the LCD board beeper.

Checking the firmware code i found **E_MUX_2_PIN** in 'ultralcd.cpp' and 'Marlin_main.cpp' where it stays always 'LOW'. The 4 extruders on the Prusa MM Super Switch are just controlled by the E_MUX0_PIN (17) and E_MUX1_PIN (16).
Doesn't it make sense to remove the 'E_MUX2_PIN' from the firmware when it is useless or may even interfere with the beeper?

PavelSindler:
Hello, 
yes it is redundant in SNMM section. I will remove it. Thank you very much.

3d-gussner:
Thanks, close the question

PR #182, opened 2017-08-23
Cleanup E_MUX2_PIN and disabled max end stop

3d-gussner:
Bit while ago i opened an issue https://github.com/prusa3d/Prusa-Firmware/issues/132 but it is still there. So i decided to try it myself. Maybe you like it. PIN 84 is being used to identify which extruder is chosen and at the same time used for the buzzer.

- removed E_MUX2_PIN and updated function
- unmarked #define DISABLE_MAX_ENDSTOPS, Prusa i3 MK2(s) doesn't use these end stops ... M119 shows now just used end stops


PavelSindler:
Thank you @3d-gussner for this PR.

3d-gussner:
Glad to help out.

PR #183, opened 2017-08-23
Updated German translation

3d-gussner:
and modified few things in language.en to get less warnings/errors using langtool.pl script


PavelSindler:
Thank you for your work @3d-gussner . I send it to our german language expert to double check it. I have noticed, that you made also some changes in max. length for some messages. There is for example MSG_FAN_SPEED which used to be max. 14 chars and now it is 20 chars.  But 14 chars was correct, because, there must be also space for cursor and three digit number etc.

3d-gussner:
Hi Pavel,
I don't use the German or Dutch translation that often, but for Kids or people with less english knowledge it is helpful. So i try to keep the translation up to date.

At this moment i am busy to remove all 'length' value from the German/Dutch files. They are useless in these files, because the length has to be defined in the English language file to get rid of the langtool.pl perl script errors/warnings. Using the perl script i got lot of warnings for de,cz,pl,es and it, so i modified the length in the english file to get rid of it. 

Sorry for the MSG_FAN_SPEED value change, that it is a copy/paste error.

I also try to compare the translation with the Marlin version to get them closer.

How are the chances to get a Dutch translation (compared to Marlin and reviewed by native Belgium/Dutch persons) pulled?

PavelSindler:
Be careful with seting max. length to higher value then it was before. It is not only MSG_FAN_SPEED case. It is necessary to check it with LCD. For example menu items are usually limited to 17 chars (MSG_LOAD_FILAMENT etc.) If somebody changes this message in the future, there is danger that there will be no warning from perl script and the message will not be displayed correctly on LCD.

PavelSindler:
Dutch translation is unfortunatelly not planed in near future. One reason is that we would need somebody here who would do translations and maintain it in the future. And also languages consumes quite a lot of memory so with current approach (all languages in one hex), we are unable to add many different languages.

3d-gussner:
Didn't make it on time so i made a new one https://github.com/prusa3d/Prusa-Firmware/pull/218

PR #184, opened 2017-08-23
Changing baud rate from 115200 to 250000

3d-gussner:
I use sometimes Octoprint to transfer files to the SD card and having baud rate set to 250000 helps bit...still slow but much better than 115200. Never had any issues with that baud rate and i am using it quite a long time.

Check for more info http://shop.prusa3d.com/forum/original-prusa-i3-mk2-f23/can-octoprint-connect-to-mk2-at-250k-baud-rate--t2026.html#p22216

### DIFF
--- a/Firmware/Configuration.h
+++ b/Firmware/Configuration.h
@@ -80,7 +80,7 @@
 #define SERIAL_PORT 0
 
 // This determines the communication speed of the printer
-#define BAUDRATE 115200
+#define BAUDRATE 250000
 
 // This enables the serial port associated to the Bluetooth interface
 //#define BTENABLED              // Enable BT interface on AT90USB devices
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -2173,7 +2173,7 @@ void process_commands()
         return;
     } else if (code_seen("SERIAL HIGH")) {
         MYSERIAL.println("SERIAL HIGH");
-        MYSERIAL.begin(1152000);
+        MYSERIAL.begin(250000);
         return;
     } else if(code_seen("Beat")) {
         // Kick farm link timer


Sebastianv650:
I would be very careful with this one. With the experience I made with the marlin branch 250k baud has more disadvantages than advantages.
- Printing over serial takes no advantage, as the used atmega can't handle more print instructions than 115k baud provides. 
- Incoming chars at 250k can easily be skipped due to long running, blocking ISRs. This leads to serial communication errors, which leads to the command being resent (slowing down the command rate) in best case. In worst case, you get all sorts of strange printer behaviour which is very hard to debug. 

While this problems might not show up at the moment, they can at further updates. So the developers would need to check the ISR runtimes after every change, having in mind possible impacts due to the baud rate. I don't think this is worth a little speed up for loading files to SD - in fact you are the first one I now for sure who is using this feature.

3d-gussner:
@Sebastianv650 thanks for the comment, i will close the pull request.
And thanks for the Linear advanced feature, my prints were dimensional quite accurate but now with that it got even better!

3d-gussner:
@Sebastianv650 Hi, just found in Marlin pull requests https://github.com/MarlinFirmware/Marlin/pull/7459
What do you think?

Sebastianv650:
Sounds intresting. But this seems to be only a Marlin-side of the implementation. As long as your PC host software has no implementation of XON/XOFF, you gain nothing. If it get's merged into Marlin, maybe the host side will follow..

thinkyhead:
I'm getting ready to merge MarlinFirmware/Marlin#7459 later today. I suppose I should add a Capability entry for this so hosts can adapt if they want. Although the changes give nothing extra if the host doesn't do XON/XOFF, there are some small optimizations that should make throughput a little more efficient in the usual case.

bubnikv:
Thanks Scott for heads up.

3d-gussner:
@thinkyhead Thanks for the update

PR #201, opened 2017-09-15
Put space between filename and size (fix M20 / lsDive)

thess:
Recent revision broke M20 file lists.


thess:
Updated: Added the output of the long filename if present to M20.

PavelSindler:
Thank you @thess for this PR. I am merging it.

PR #206, opened 2017-09-18
Auto power mode

PavelSindler:


3d-gussner:
Hi @PavelSindler 

what is the advantage of Auto power mode?

fiveangle:
It appears to set stepper current to high when moving steppers, but low when not (to maintain stepper location, perhaps only to the nearest full step, I don't know if the drivers support changing DAC without a deactivate/reactivate).

I always meant to submit feature request for this (since I don't like my printer wasting energy when sitting idle) but never got around to it.

Glad to see it already in the works.  Thanks @PavelSindler  :+1: 

-=dave

PR #209, opened 2017-09-19
Preheat order PLA 1st

3d-gussner:
It is a smal change but i use most times PLA and changing the preheat order to PLA 1st makes it bit easier.


3d-gussner:
It made it into mk3 branch so I close it.

PR #211, opened 2017-09-28
Hyperfine Bed Levelling/correction for Prusa i3 MK2/MK2S

3d-gussner:
**NOTE: PJR** and **JeffJordan** have shared the modifications on the Prusa3d.com Forums (http://shop.prusa3d.com/forum/prusa-i3-kit-building-calibrating-first-print-main-f6/hyperfine-bed-leveling--t4330.html) and **all credit goes to them**. Robrps and i just tried to get it in github to make a pull request possible.

The Hyperfine Bed leveling/correction is an advanced 'Bed level correct' version of the original Prusa code, which allows to manually adjust Left/Right/Front/Rear bed heights. Next to the default Left/Right/Front/Rear it now allows to adjust the four corner points FrontLeft/FrontRight/RearRight/RearLeft.
Best explanation can be found here by **JeffJordan** https://shop.prusa3d.com/forum/prusa-i3-kit-building-calibrating-first-print-main-f6/hyperfine-bed-leveling--t4330-s110.html#p41311

hyperfine bed leveling jeffjordan

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -5769,7 +5769,7 @@ case 404:  //M404 Enter the nominal filament width (3mm, 1.75mm ) N<3.0> or disp
 
   else if(code_seen('T'))
   {
-	#ifdef SNMM
+	
 	  int index;
 	  st_synchronize();
 	  for (index = 1; *(strchr_pointer + index) == ' ' || *(strchr_pointer + index) == '\t'; index++);
@@ -5785,6 +5785,7 @@ case 404:  //M404 Enter the nominal filament width (3mm, 1.75mm ) N<3.0> or disp
 			  tmp_extruder = code_value();
 		  }
 		  snmm_filaments_used |= (1 << tmp_extruder); //for stop print
+#ifdef SNMM
       #ifdef LIN_ADVANCE
         if (snmm_extruder != tmp_extruder)
           clear_current_adv_vars(); //Check if the selected extruder is not the active one and reset LIN_ADVANCE variables if so.


GurliGebis (changes requested):
Since the offset is stored as an int8, we can go from -127 to 128.
So storing a value above 128 would cause it to overflow and become negative, which would be a bad thing.
I don't think it should be changed from 100, since it could break things.
Also, I think it should handle if the value is too low, which it doesn't seems to do.

3d-gussner:
@GurliGebis Thanks for the comment...but isn't the code starting line 3546 of the Marlin_main.cpp preventing values being too high?

GurliGebis:
It is, but the problem is, that if you change the maximum from -100/+100 to -200/+200, you go from a span of 200 to a span of 400, which is a too big a value to store in an int8.

So the problem I see, is that if the user sets it too a value above 128 or below -127, it will overflow/underflow, and store some other value than the user expects (since for instance 129 would be saved as -127)

GurliGebis:
@PavelSindler can you help here, I'm not 100% sure I'm right.
Would it be a problem going above 128, or am I misunderstanding something? :)

3d-gussner:
@GurliGebis in the LCD menu the 'BED_ADJUSTMENT_UM_MAX' was raised from 50 to 99
and in Marlin_main.cpp 'if (fabs(offset) > 0.101f)' to 'if (fabs(offset) > 0.201f)'. I didn't saw that and also didn't test abcdefgh values above +-99. Will test later when the Excessive bed leveling correction: message is triggered.

Thanks again pointing out.

GurliGebis:
If you can try and adjust it to, lets say 160, save it, power cycle the printer, and confirm it is still at 160, I would say we are home free, so to speak :)

3d-gussner:
@GurliGebis The LCD menu is limited to +-99 and this is stored in the EEPROM. With the G80 a b c d e f g h g-code the values are not stored in the EEPROM so i cannot test it with a power cycle of the printer.

Just tested with Pronterface and original Prusa firmware, G80 V10 L102 result:
Error:Excessive bed leveling correction: 0.10 microns
[ERROR] Error:Excessive bed leveling correction: 0.10 microns

Now i am confused ... Via the LCD menu i am allowed to correct +-50 microns and with G80 L R F B just +-0.10 microns? I guess 'SERIAL_ECHOLNPGM(" microns");' means 0.10 millimeters or 100 microns.

Here some other things may help us to understand what the limits are:
In the 'Marlin_main.cpp'
- the 'offset' (abcdefgh) values are used in the 'mbl.z_values[0][0] += offset;'
- and that used in 'z0 = mbl.z_values[0][0] + *reinterpret_cast<int16_t*>(&z_offset_u) * 0.01;'
- and 'uint16_t z_offset_u = eeprom_read_word((uint16_t*)(EEPROM_BED_CALIBRATION_Z_JITTER + 2 * (ix + iy * 3 - 1)));

So if i am not wrong the abcdefgh values can be int16 or am i? These are limited to 200.

In the 'ultralcd.cpp'
- the 'front_left/front_centr/...' values are int8
- but also limited to '#define BED_ADJUSTMENT_UM_MAX 99', that should be okay or not?

macfly1202:
I read on Prusa forum and found this idea really clever! thanks PJR and JeffJordan

About '#define BED_ADJUSTMENT_UM_MAX 99', that should be okay or not?
I don't know if it was a question but 99 perfectly working; using it 2 months before MMU update. 

I made a fork of Sebastianv650 version with some other minor change to add the ability to have bed level compensation at his maximum value: 99! cf here https://github.com/macfly1202/Prusa-Firmware/commit/3601032ac9ff52fb3a2fb44f9c435b7fb7bc2c71. 

(Build available here for mk2 https://travis-ci.org/macfly1202/Prusa-Firmware (thanks Sebastianv650 for Travis setup!).

3d-gussner:
@macfly1202 Thanks for the input!

Jeff-Jordan:
Hi guys, after thinking about the problem with the limits of the int8 value: why not handling every single unit as 2µm instead of 1µm...  so everything you need to do is to multiply it by 2 when using it inside the bed correction routine.
of course this would apply to both, the parsed values through G80 and the stored values inside the eeprom.
and you've to take it into account when finding the perfect correction level....

ok, you'll loose accuracy... but +/- 1µm doesn't really make any difference in the real world ;-).

so we'll end up with a correction range of +/- 198µm which should fit. if not, you've got a problem with the heatbed that you need to correct mechanically.

@3d-gussner & @robrps and all helping to make this amendment "proposable" at the github: a **## BIG THANK YOU !**

3d-gussner:
@Jeff-Jordan That could be a solution.
As you said in real life it should not make a big difference. I am not sure what the max resolution of the z-rods per step is, but assuming the printer is able of 50 microns adjustments by 1 OR 2 microns doesn't make the big difference.
The PINDA is very accurate (except the thermal issue) just checkout TOMs video https://www.youtube.com/watch?v=il9bNWn66BY
Even for the PINDA it will be hard measuring 1 micron differences.
Saying that, we could have the LCD menu and g-Code values in the same range and maybe make it possible to store the values via g-code, and not as it is now 50/99um in LCD and 101/201um via g-code.

3d-gussner:
Checkout https://shop.prusa3d.com/forum/original-prusa-i3-mk2-f23/firmware-3-1-0-rc2-mk2-multi-material-mk2-s-mk1-t6454.html#p45064

PR #217, opened 2017-10-04
First layer calibration: print small rectangular in the end of calibration

PavelSindler:


GurliGebis:
@PavelSindler How far away is 3.0.13? It looks like features keep getting added to 3.0.12-X, but no new hex files gets compiled and uploaded?
There is a lot of interresting features added, that people don't get to test, since they don't know how/want to compile it.

PR #218, opened 2017-10-04
Updated German translation 04 oct 2017 and

3d-gussner:
modified few things in language.en to get less warnings/errors using langtool.pl script.
Even with existing language_en.h and other languages i got lot of warnings and errors.
Also limited the 'length' to the minimum across all used languages. It should still fit in the LCD screen.
Attached files contain the results before
[langtool_warings.txt](https://github.com/prusa3d/Prusa-Firmware/files/1354928/langtool_warings.txt)
and after
[langtool_warnings_after.txt](https://github.com/prusa3d/Prusa-Firmware/files/1354978/langtool_warnings_after.txt)
this change.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -1483,7 +1483,7 @@ static void lcd_support_menu()
   }
   #ifndef MK1BP
   MENU_ITEM(back, PSTR("------------"), lcd_main_menu);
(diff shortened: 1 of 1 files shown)


thess:
@3d-gussner - I really think you should either:
- close this as it is hard to tell what has really changed - too many merges
- rebase and squash your commits so only the changes related to the German translation are present and on top.

Also, don't submit patches from the main (MK2) branch in your repo fork - please!

PR #220, opened 2017-10-04
Add optional 8-point Mesh Bed Correction offsets + LCD and translations support

thess:
Inspired by the forum thread on Hyperfine bed leveling, I did my own implementation of the 8-point mesh bed level correction similar to that of Peter & Jeff (a couple of forum luninaries). See:

http://shop.prusa3d.com/forum/prusa-i3-kit-building-calibrating-first-print-main-f6/hyperfine-bed-leveling--t4330.html

This PR is an alternative proposal to that of @3d-gussner #211

I use a different grid layout which made a bit more sense to me - 8 points of correction (a thru h) starting at the front / home position and progressing left-to-right, skipping the center (X) as follows:
```
      B A C K
   +-----------+
   | f   g   h |
   | d   X   e |
   | a   b   c |
   +-----------+
     F R O N T
```
This alternate id scheme only affects the G80 command previously proposed. The English LCD settings are FrontLeft/FrontCenter/FrontRight, MidLeft/MidRight, etc. 

All language files are updated but it would be nice if the non-english versions were to be refined - Google doesn't help much with abbreviations.

Additionally, I store the offsets as 16-bit integers in order to allow larger range of settings. The currrent limitation is hard-coded for +/- 500um.


3d-gussner:
@thess Hi, i guess your code is more elegant then what we tried and it doesn't matter if it is
f   g   h
d   X   e
a   b   c
or
g  f  e
h  X  d
a  b  c
it could be also (following the mesh leveling sequence)
f  g  h
e  X  d
a  b  c
as long the hyperfine tuning is implemented.

I hope nobody has a bed that needs up to 500 microns correction, that one should be changed.
Do the bed correct values higher than +-99 fit on the LCD screen?

In beginning i was struggling with making the hyperfine tuning optional via the 'Configuration_prusa.h' file and use #ifdef MBC_8POINT, but have to say nicely done by using
#define EEPROM_UNUSED  (EEPROM_BED_CORRECTION_VALID-1)
#define EEPROM_TOSHIBA_FLASH_AIR_COMPATIBLITY (EEPROM_UNUSED-4)
you made sure that the EEPROM values stay in the right order.

For German you can check my translation.

thess:
I did consider the z-probe order for the point names but that seemed even odder than the anti-clockwise notation used previously.

The LCD mods were made to accommodate +/- 3-digits. To that end, the display text labels were limited to 9 (+4 for `[um]`) chars max. I will need all eight translated positions (not just the previous 4). Ex:

```
#define(length=13,lines=1) MSG_BED_CORRECTION_P1                "FrontLeft[um]"
#define(length=13,lines=1) MSG_BED_CORRECTION_P2                "FrontCntr[um]"
#define(length=13,lines=1) MSG_BED_CORRECTION_P3                "FrontRt  [um]"
#define(length=13,lines=1) MSG_BED_CORRECTION_P4                "MidLeft  [um]"
#define(length=13,lines=1) MSG_BED_CORRECTION_P5                "MidRight [um]"
#define(length=13,lines=1) MSG_BED_CORRECTION_P6                "RearLeft [um]"
#define(length=13,lines=1) MSG_BED_CORRECTION_P7                "RearCntr [um]"
#define(length=13,lines=1) MSG_BED_CORRECTION_P8                "RearRight[um]"
```

PR #235, opened 2017-10-14
Fix parsing of M110 command

### LINKED ISSUE #232: M110 support is not functional in the 1_75mm_MK2-RAMBo13a-E3Dv6full.hex file
Just updated my firmware in hopes of adding M110 support. The M110 command does not throw an error, but it also does not set the appropriate line number.

thess:
Not the best fix but it does work. Fixes #232.


PavelSindler:
Thank you @thess .

PR #236, opened 2017-10-15
Remove 'unused-parameter' and 'empty-body' warinings

thess:
One last pass to get a clean build with `-Wall -Wextra` compiler opts. I'm not sure about the removal of `mesh_bed_leveling::get_meas_xy` but it wasn't being called. I'm sure there probably is a lot more dead code which could be removed as the firmware image is getting full.


thess:
Ping @PavelSindler - how about this? Nice and clean for your next release.

``get_meas_xy`` hasn't been missed. If you want to preserve it with a tombstone, I can add a conditional.

thess:
Added cleanup for SNMM and MK1 configs.
Fixed reference to uninitialized var in "SilentMode" selection in stepper.

PR #238, opened 2017-10-15
Fixes power mode in Tune menu...

3d-gussner:
...and changed the filament runout procedure and the baudrate from 1152000 to 115200 (one zero too much)

During the print you can access the Tune menu and the new feature Power Mode [auto] was added, but choosing it it goes to the Settings menu. Here the fix for that in the ultralcd.cpp.

Also changed the filament runout procedure in Marlin_main.cpp to use M600 command (origin idea/update from Marlin 1.1.x) defined in the 1_75mm_MKx_yyyy_E3dv6full.h files


PavelSindler:
Thank you for this PR @3d-gussner. What is actually advantage of using M600 as filament runout procedure over current filament runout procedure?

3d-gussner:
@PavelSindler Hi, I compared the FRS procedure and the M600 procedure and saw that M600 has evolved over time but the FRS stayed.
Also the current Marlin changed to use the M600 procedure for the FRS, which makes kind of sense because FRS is basically a Filament change after an event of runout.
Improving the M600 code in the future don't need to do the same things again in the FRS.

In the future there might be some needs in the FRS that don't apply to the M600, but I think you can implement these needs around the M600 or other using existing g-codes.

Issue #234, opened 2017-10-14
3.1.0-rc LinAdv with retraction causes(?) to accidental stop


Gobol:
Hello,
Setup: 
Prusa i3 MK2 + MMU + 3.1.0-rc (linadv) + Octoprint@NanoPi-Neo.
Previously many printouts - flawless, without problems. 

Started to dial-in linear advance feature on MMU and 3 consequent prints fails in the middle of process - the printer stops, still keeping the temperature. GCode sliced with Slic3r-Prusa 1.37.1-prusa3d-win64, filament PLA, gcode starts with M900 K230 (determined with test pattern prints for linadv). 

Octoprint shows in terminal log: 

`Last lines in terminal:
| Communication timeout during an active resend, resending same line again to trigger response from printer. Configure long running commands or increase communication timeout if that happens regularly on specific commands or long moves.
| Send: N23806 G1 X173.403 Y115.801 E0.01415*98
| Recv: Error:Line Number is not Last Line Number+1, Last Line: 23806
| Recv: Resend: 23807
| Communication timeout during an active resend, resending same line again to trigger response from printer. Configure long running commands or increase communication timeout if that happens regularly on specific commands or long moves.
| Send: N23806 G1 X173.403 Y115.801 E0.01415*98
| Recv: Error:Line Number is not Last Line Number+1, Last Line: 23806
| Recv: Resend: 23807
| Communication timeout during an active resend, resending same line again to trigger response from printer. Configure long running commands or increase communication timeout if that happens regularly on specific commands or long moves.
| Send: N23806 G1 X173.403 Y115.801 E0.01415*98
| Recv: Error:Line Number is not Last Line Number+1, Last Line: 23806
| Recv: Resend: 23807
| Communication timeout during an active resend, resending same line again to trigger response from printer. Configure long running commands or increase communication timeout if that happens regularly on specific commands or long moves.
| Send: N23806 G1 X173.403 Y115.801 E0.01415*98
| Recv: Error:Line Number is not Last Line Number+1, Last Line: 23806
| Recv: Resend: 23807
| Communication timeout during an active resend, resending same line again to trigger response from printer. Configure long running commands or increase communication timeout if that happens regularly on specific commands or long moves.
| Send: N23806 G1 X173.403 Y115.801 E0.01415*98
| Recv: Error:Line Number is not Last Line Number+1, Last Line: 23806
| Recv: Resend: 23807
[...]

PR #245, opened 2017-10-20
LIN_ADVANCE: high e_steps value error fixed

PavelSindler:
When using colorprint (M600) during multi material print it sometimes happened that right after M600 procedure, printer stopped working and also teperature regulation didn't work.
Reason was that in the end of M600 procedure, there is plan_set_e_position() function used, but inside this function, there was `position_float[E_AXIS] = e;` missing. This leaded to high e_steps value and thus eISR_Rate was set to zero value inside isr() function.
This leaded to extremly short TIMER1_COMPA_vect interrupt time where only advance_isr() (not isr()) was executed and eISR_Rate and nextAdvanceISR were never updated so it was stucked in loop.
TIMER0_COMPB_vect interrupt which serves for heater regulation (!) was blocked by this, because of lower priority.
There should be probably some low limit set for counting eISR_Rate/interrupt time.
I have updated advance_isr() function by setting nextAdvanceISR value to ADV_NEVER after all e_steps are executed.
Could you please look on this PR @Sebastianv650? It may also relate to: https://github.com/prusa3d/Prusa-Firmware/issues/234 .

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -5764,7 +5764,9 @@ case 404:  //M404 Enter the nominal filament width (3mm, 1.75mm ) N<3.0> or disp
 
   else if(code_seen('T'))
   {
+	#ifdef SNMM
 	  int index;
+	  st_synchronize();
 	  for (index = 1; *(strchr_pointer + index) == ' ' || *(strchr_pointer + index) == '\t'; index++);
 	   
 	  if ((*(strchr_pointer + index) < '0' || *(strchr_pointer + index) > '9') && *(strchr_pointer + index) != '?') {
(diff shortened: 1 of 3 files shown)


Sebastianv650:
Can you tell me why this is a MMU related problem, I would think M600 filament change is independet from MMU/ single material version?
Beside that, an e-only move should never trigger a move with advance (block->use_advance_lead) enabled:

```cpp
block->use_advance_lead =  block->steps_e
                            && (block->steps_x || block->steps_y) // <-- Only use advance if there is something in X/Y dir.
                            && extruder_advance_k
                            && (uint32_t)block->steps_e != block->step_event_count
                            && de_float > 0.0;
```

Therefore it would be intresting how nextAdvanceISR becomes a small number..

PavelSindler:
I am not sure if this is only MMU problem, but we have observed that only with MMU so far. 

At the beginning of M600 procedure, there are current positions for all axes stored.
In the end of M600 procedure, there is plan_set_e_position() function used. This function sets current_position[E_AXIS] to value which was stored at the beginning of M600 procedure. But in this function, there is `position_float[E_AXIS] = e;` missing. So there arise difference between current_position[E_AXIS] and position_float[E_AXIS]. Printer then continue with printing (moving in "XYE") and when plan_buffer_line is called, wrong de_float value is counted:  
`float de_float = e - position_float[E_AXIS];`.

So printer stoped working after M600 procedure, when it continues in printing and there was some movement in X or Y.

Also on MMU prints we are using higher K values, which leads to higher e_steps values compared to single material which can be one of reasons why MMU is more prone to this.

PR #256, opened 2017-10-29
Fix possible array bounds overflows, remove some unreachable code

thess:
Additional cleanup from static analysis.
Note: setTargetHotend does not check bounds (EXTRUDERS = 1)

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -2190,9 +2190,9 @@ bool gcode_M45(bool onlyZ) {
 	bool final_result = false;
 	if (!onlyZ) {
 		setTargetBed(0);
-		setTargetHotend(0, 0);
-		setTargetHotend(0, 1);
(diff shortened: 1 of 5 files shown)


thess:
@PavelSindler - Not that I believe in "serendipitous engineering" and these changes may eliminate unexplained bugs, but they do eliminate some random memory writes.

wavexx:
Haha, scanning through old PRs to find this. It seems that the duplicate return in SdBaseFile.cpp is still with us with the MK3, the rest are mostly gone.

In similar spirit there's also a single setTargetHotend in ultralcd.cpp:5091 that uses extruder 0 as opposed to active_extruder.

Would you like to update the PR?

thess:
Well, you can lead a horse to water...
3yrs old and not in my version - Haha indeed!

If I get some time in the next week or so, I'll check this out. Otherwise go ahead and close it and make the appropriate changes if they are still relevant.

PR #257, opened 2017-10-29
Add option in menu to change filament notification mode

housepage:


Firmware/language_en.h
```diff
@@ -101,6 +101,9 @@
 #define(length=20) MSG_CHANGING_FILAMENT	"Changing filament!"
 
 
+#define MSG_FILAMENT_NOTIFICATION_MODE_CONTINUOUS					"Fil Notif  [continuous]"
```
3d-gussner, line 104: The MSG is 23 characters long, the limit is 20 on the display as no chars are before or behind.
I think this would fit better
#define MSG_FILAMENT_NOTIFICATION_MODE_CONTINUOUS					"Fil.Notify  [long]"
#define MSG_FILAMENT_NOTIFICATION_MODE_SHORT					        "Fil.Notify [short]"
#define MSG_FILAMENT_NOTIFICATION_MODE_SILENT					        "Fil.Notify [silent]"

housepage: Will incorporate that change.

housepage:
@3d-gussner The other major question I had was about my EEPROM address and making sure picked one that was non-conflicting with that block of calculated ones. Do you know if there is a reference table for EEPROM assignments?

Also, what is the process for translations?

3d-gussner:
@housepage Maybe you should ask @PavelSindler about a EEPROM table. At this moment most of them are calculated and don't have a define spot, don't know the reason for that. It makes sometimes pull requests difficult as different pull requests may overlap in the same space.

Here how I translate to German or Dutch when there is/are a new MSG/s added to the language_en.h

- First see if it makes any sens to translate to other languages. I have a branch which has a Filament Runout Sensor and some MGS for the LCD or serial, like "Fil.Runout S " and this doesn't make much sense to translate to German or other languages. If there isn't a MSG defined in the other language_xx.h files the 'langtool.pl' will automatically use the 'language_en.h' for the chosen language.

- I try to keep the lines/TABs identical between the different languages files, which makes comparing and translating easier.

- Copy the English MSG/s line/s without any '(length=xx, lines=yy)', which are just needed in the language_en.h file

- After editing all of them you need to run langtool.pl perl script to generate the language_all.cpp and language_all.h file, here you get warnings and errors like lengths or lines aren't fitting or a MSG is defined first in another language than english. I try to squeeze the MSG in that specific space. Sometimes you can change the limited 17 chars to 20 chars, but here you have to make sure it fits in the LCD screen menu.
BUT be careful with that, Prusa will not accept your pull request if you have too many of these.

As an example is MSG_FAN_SPEED limited to 14, because in the submenu Settings-Temperature the 1st char is the ">" selection and at the end you need space for ":" + value <xxx> + " " space. Doing the math 20-1-1-3-1=14
Long MSGs (length=20, lines=yy) may need more lines depending on the language and there it isn't a big problem to correct the value for the MSG in the language_en.h file, as it is a scrolling message.

The last thing before publishing it and doing a pull request is to run the bash script le.sh to get all edit files to 'LF' and then test it.

Hope that helps

Firmware/Configuration.h
```diff
@@ -22,6 +22,7 @@
 #define EEPROM_FILAMENTUSED 4081
 // uint32_t
 #define EEPROM_TOTALTIME 4077
+#define EEPROM_FILAMENT_NOTIFICATION 4076
```
3d-gussner, line 25: I think should not use that address 4076 because it is used by the line beneath !!!

3d-gussner:
@housepage Long time no update. can you please close it?

3d-gussner:
@michalxfanta No reaction from owner, guess it is outdated and can be closed.

PR #260, opened 2017-10-31
presort files when print is finished

PavelSindler:


Firmware/cardreader.cpp
```diff
@@ -246,7 +246,7 @@ void CardReader::startFileprint()
   {
     sdprinting = true;
     #ifdef SDCARD_SORT_ALPHA
-	  flush_presort();
+	 // flush_presort();
```
thess, line 249: Why comment out a single line of code under a conditional when removing the entire block would be clearer?

Firmware/ultralcd.cpp
```diff
@@ -4959,9 +4959,9 @@ void lcd_sdcard_menu()
   uint8_t sdSort = eeprom_read_byte((uint8_t*)EEPROM_SD_SORT);
 
   if (presort_flag == true) {
-	  presort_flag = false;
+ 	  presort_flag = false;
```
thess: Whitespace change?

Firmware/ultralcd.cpp
```diff
@@ -4959,9 +4959,9 @@ void lcd_sdcard_menu()
...
 	  card.presort();	  
-  }
+   }
```
thess: Another whitespace error?

Firmware/cardreader.cpp
```diff
@@ -1047,7 +1047,7 @@ void CardReader::printingHasFinished()
       }
       autotempShutdown();
 	  #ifdef SDCARD_SORT_ALPHA
-		presort();
+		//if(!check_file)	presort();
```
thess: See previous comment regarding adding code then removing it!?

PR #273, opened 2017-11-11
Fix for [ISSUE-243] First layer calibration menu

Panayiotis-git:
Fix for the ISSUE-243


thess:
LGTM - Additionally, when LCD v2_calibration commands are executing, serial port activity can interfere with clean execution. Not sure how that can be avoided (yet).

PR #364, opened 2018-01-01
Set command type before parsing line. Skip leading whitespace

thess:
This is an attempt to address Issue #331. I have only done limited testing with this change however, I think it is a correct fix.


wavexx:
This is now done with #3823

Issue #365, opened 2018-01-01
MK3: ENCODER_STEPS_PER_MENU_ITEM set incorrect for Rotary Knob Positions


TheZeroBeast:
Hello,

I have had my MK3 for a few days now and have experienced differences with the rotary LCD knob selecting menu items.
On the MK2S each position of the knob selects the next item or increments one step.

It appears now that the varriable ENCODER_STEPS_PER_MENU_ITEM isn't getting set to 5 on line 254 of ultralcd.cpp instead getting set to 2 on line 261 of ultralcd.cpp.

This results in unreliable selection of menu items. Especially frustration when attempting to increase the speed to an exact figure of say 120% and the position for this being unobtainable.

This is on Firmware 3.1.1 RC4 on MK3

Thank you for all your hard work, keep it up

PR #398, opened 2018-01-14
ENCODER PULSES and STEPS Re-aligned with Clicks

TheZeroBeast:
Tested and confirmed working on MK3. Menu options now are aligned with encoder clicks/positions.

In response to Issue #365


camlesom:
If the MK3 is using the same encoder from the MK2 then this would be correct, the Configuration.h file for the MK2 has the exact same values shown [here](https://github.com/prusa3d/Prusa-Firmware/blob/f91ace6fec924d3c20d0cc4f765da3d64dce7fcb/Firmware/Configuration.h#L481). Can also find the related commit here 93400ea.

Panayiotis-git:
Agree with camlesom.

The case that the encoder's clicks most of the times do not correspond to exactly 4 pulses and the accurate positioning of the encoder, according to clicks, is rather erratic, is another issue!
It would have been better if the encoder had a 1:1 ratio for the pulses/click.

camlesom (approved):

camlesom (changes requested):
Too many changes made in this PR, would recommend rebasing to head of MK3 branch and opening a new PR for other changes (leaving only the encoder changes in this PR).

TheZeroBeast:
Hi Cameron,

Sorry, I’m fixing this up in the next few hours and branching. Still new to github.

PullRequest will be back to the iniatial two lines for encoder step changes.

Regards,
Robert McKenzie

TheZeroBeast:
Recreating pull-request.

PR #457, opened 2018-02-02
Mk3 fast dbg

bubnikv:
Enabled linear advance.

Reworked the stepper routine: 

1) The computational load is spread more evenly along the stepper ISR
ticks by moving some of the timer and linear advance calculation from
the block initialization into the 1st tick of the steady phase
of the trapezoid.

2) Reworked planning of the Linear Advance ISR ticks to fit
the time slot allocated for the main stepper ISR tick. This is achieved
by grouping the Linear Advance extruder ticks by a power of two to tick
the Linear Advance interrupts at a maximum 7kHz. Also some
of the extruder ticks are performed just after the XYZ ticks
and if the remaining time slot for the Linear Advance ticks is too short,
all the Linear Advance steps are ticked inside the main stepper ISR invocation.

3) Added some calls to MSerial.checkRx() if the stepper ISR routine
is delayed for too long by the additional LinearAdvance ticks.

This implementation differs significantly from the original implementation
by @Sebastianv650, as this implementation tries to follow the exact
timing of the XYZ axes with the drawback of possibly ticking the extruder
faster than it could handle, while the original implementation
by @Sebastianv650 ticks the extruder slower with the drawback of possibly
spreading the XYZ ticks, thus introducing jerk in the cartesian movement.

Optimized the MarlinSerial class by declaring all methods static. The MarlinSerial defines no member variable, though the AVR GCC stored a "this" pointer onto stack when calling the non-static methods anyway.

Optimized the stepper routine by calculating the Bresenham algorithm in 16 bits only for short segments (which are the majority with the bed leveling enabled).


Sebastianv650:
> ..as this implementation tries to follow the exact
timing of the XYZ axes with the drawback of possibly ticking the extruder
faster than it could handle..

I hope you do a lot (!) of testing before merging this. The one big problem with LIN_ADVANCE is skipping extruder steps due to not respected jerk limits and bad step pulse timing. If your change makes this worse, I would think it will end in unusable prints.

Beside that, maybe you can explain me the big difference between way of executing steps "today" and with your PR? At the moment I don't see it:

As-is behaviour is:
- Main ISR loop fires XYZ steps
- Instantly after that the first eISR loop is called, ending up executing all normal E steps
- [Delay between Main ISR] / [advance steps] (rounded down) is the time intervall we wait for the next eISR call, except if we have a failure case and this devision results in 0.
- No Main ISR loop can be delayed, except by a very short fraction of time when the last eISR call happens at [Time at which Main ISR should fire] - [Time needed by an eISR loop] or even closer. As the time needed by an eISR loop on it's own is low, this should be unlikely.
- So we end up with:
[Main ISR] - [eISR] - [delay] - [eISR] - [delay] - [Main ISR] depending on the numbers of eISR steps todo.

I understand that you check for the [Time at which Main ISR should fire] - [Time needed by an eISR loop] case and execute all remaining steps together if so. As this isn't supposed to happen, it shouldn't change anything in theory.
I never observed a reduction in XYZ speed due to this, only slow downs I'm aware of is the buffer emtying due to increased CPU load.

bubnikv:
> I hope you do a lot (!) of testing before merging this.

We certainly plan to test it at least for a week.

> The one big problem with LIN_ADVANCE is skipping extruder steps due to not respected jerk limits and bad step pulse timing. If your change makes this worse, I would think it will end in unusable prints.

Sure. Let's see how it will work. My guess is, it will work for the MK3's direct drive setup. For the multi-material bowden setup, the extruder ticks may be too quick, it has to be verified.

> Beside that, maybe you can explain me the big difference between way of executing steps "today" and with your PR? At the moment I don't see it:

My work is based on analyzing the stepper routine timing using a logical analyzer. You will find couple of WRITE_NC(LOGIC_ANALYZER_CHx, xxx) calls in stepper.cpp, toggling some of the unused pins of the Einsy controller board. By following these signals with a logical analyzer, one will clearly see the time spent in various parts of the stepper routine at various phases of the trapezoid block processing. I will try to explain my findings, though it would be much easier with the logical analyzer screen shots. Unfortunately I did not take any.

First let me digress a bit from the linear advance issue. As you know, the Marlin stepper routine fires at a maximum 10 kHz rate. If one of the stepper motor should tick at higher rate than 10 kHz, the stepper motor ticks are grouped by 2 or 4 ticks per interrupt invocation. This is done to conserve CPU, as there is a cost associated to interrupt invocation. Also PWM of the stepper drivers runs at 15 to 25 kHz, therefore I don't think it makes much of a difference, whether for example the step pulses are ticked evenly at 40kHz, or they are ticked at 10kHz and then grouped by 4. I believe the original Marlin without your linear advance algorithm was optimized by Ultimaker to move the XY axes at a travel speed 250 mm/sec, this corresponds to a 25 kHz step rate with a 10um step length. With the 10 kHz ISR limit, the steps are 4x grouped and the interrupt ticks at 6.25 kHz.

Now we extended the stepper ISR routine with the stall guard and filament sensor read outs, and we experienced some weird behavior. Namely, when the stepper ISR routine decelerated and it went from 4x grouping to 2x grouping, suddenly the next tick was planned before the current ISR invocation ended, the planned tick was missed and invoked first after a full cycle of the timer, which is 32.768 ms. This 32.768 ms delay was very noticeable, it was very loud and it lead to layer shifts. I know you had the same issue with missed timer ticks, and you handled this situation by the following workaround: if (OCR1A < TCNT1 + 16) OCR1A = TCNT1 + 16

When going from 4x to 2x grouping, the next tick was planned before the current ISR invocation ended, because the 4x stepping of the Bresenham algorithm took too long. I went to such detail describing the stutter when going from 4x to 2x grouping to demonstrate, that the Marlin's stepper routine timing is quite fragile on its own. One of my optimizations, which is part of this pull request and which I believe makes sense to merge with the upstream Marlin, was to calculate the Bresenham algorithm in 16 bits only, which is quick enough to fix this particular timer miss issue.

Now to linear advance. I admire your original implementation. However, eye balling the logical analyzer captures, I see some irregularities it creates.

When you calculate the LA stepping, you don't take into account the time spent in the main stepper routine. This time can be significant, especially for the main ISR invocation, where a new block is pulled from the planner queue, or when the stepping is grouped by 2 or 4. The first LA tick is performed at the end of the main ISR routine. For short main ISR intervals and multiple inserted LA E steps, the LA time inteval calculated will be quite short, so the 2nd LA tick planned will likely be missed already and the if (OCR1A < TCNT1 + 16) OCR1A = TCNT1 + 16 applies, delaying the other LA ticks and the next main ISR invocation. For short stepper intervals and multiple inserted LA E steps, the planned LA stepping is so short, that the OCR1A = TCNT1 + 16 rule applies to each of them, further delaying the main ISR invocation. How good or bad it is, I don't know.

> Main ISR loop fires XYZ steps
> Instantly after that the first eISR loop is called, ending up executing all normal E steps

I am ticking some of the E steps just after the XYZ steps to make the E stepping more evenly spreaded in time. Indeed, the calculation inside the main ISR after that point may take significant time, so ticking the E steps rather early extends the interval to the 1st LA ISR tick.

> [Delay between Main ISR] / [advance steps] (rounded down) is the time intervall we wait for the next eISR call, except if we have a failure case and this devision results in 0.

You did not account for the time spent in the main ISR routine, which may take a significant part of [Delay between Main ISR].

[Delay between Main ISR] / [advance steps] (rounded down) may be too short, so you often apply the OCR1A = TCNT1 + 16 rule, which adds to the time spent in the LA interrupt, leading to a maximum repeat frequency of your LA ticks roughly 30kHz (I don't remember the number exactly, I would have to measure it with the logical analyzer again).

> No Main ISR loop can be delayed, except by a very short fraction of time when the last eISR call happens at [Time at which Main ISR should fire] - [Time needed by an eISR loop] or even closer. As the time needed by an eISR loop on it's own is low, this should be unlikely.

The main ISR loop is certainly delayed at least by the time of the main ISR invocation, which may be 80us easily for the ticks where a new planner block is picked up from the queue. That's why I moved initialization of the steady part of the trapezoid speed profile into the first tick of the steady part to make the main ISR invocation times more even.

> So we end up with:
> [Main ISR] - [eISR] - [delay] - [eISR] - [delay] - [Main ISR] depending on the numbers of eISR steps todo.

It looks more like
[Main ISR..................................][eISR]delay[eISR]delay[eISR]delay[Main ISR]
how exactly the time line will look like depends on the K factors and the acceleration, and on the time spent in the main ISR, which depends on many factors. The 2nd eISR invocation is often late, and the other interrupts including the next Main ISR are delayed by the same factor. The delay is quite often too short, so you extend it artificially with the OCR1A = TCNT1 + 16 rule. Again, how often it happens and whether it will disturb anything depends on many factors, including the K and acceleration values.

> I understand that you check for the [Time at which Main ISR should fire] - [Time needed by an eISR loop] case and execute all remaining steps together if so. As this isn't supposed to happen, it shouldn't change anything in theory.

That is really just a conservative check and I don't expect it to happen with reasonable accelerations and K factors.

> I never observed a reduction in XYZ speed due to this, only slow downs I'm aware of is the buffer emtying due to increased CPU load.

It is just a couple of main ISR ticks where the linear advance kicks in with high E steps, that are delayed significantly. It is likely not a problem with a direct drive setup, though my college has warned me, that the irregularities in the movements of the X and Y axes may lead to false crash alarms reported by the Trinamic stepper drivers. I find it unlikely, but I would rather play it safe and try to improve the timing.

Your OCR1A = TCNT1 + 16 rule introduces an extruder feed rate limit for the inserted LA ticks. I am not quite sure what were your intentions, when you introduced this rule. Was your intention to introduce a feed rate limit, or did you want to reserve a big enough gap for the other interrupts to be invoked (the serial ISR likely being the most important), or did you just want to avoid the missed timer interrupt?

My intention when reworking your code was to fit the planned extruder steps in between the main ISR ticks for reasonable acceleartion and K values, and to cap the ISR invocation rate in a similar way it has been done in Marlin originally. That means, if the E step rate is too high, to decrease the interrupt rate by grouping the E steps. You implemented the grouping, but with linear advance the grouping of the XYZ steps has to be separated from the E steps IMHO, as when the LA kicks in, there will be often many E steps inserted individually, as the main ISR often does not trigger any XYZ steps grouping. So one is left with too dense LA ticks, each ticking a single E step per ISR invocation, which is not efficient. It is worse now on the MK3 printer, where we step the extruder at 32 micro stepping, while on MK2 the extruder micro stepping rate was 16.

When trying to squeeze the E step grouping into the ISR routine, I had to optimize the E step timing and grouping calculation to the bones, otherwise the main ISR invocation would take too long and I would end up with the missed timer ticks again. So I used some tricks to estimate the timing and grouping quickly and conservatively. Due to the conservative estimate, the E ticks may be grouped at a higher rate than needed, and they may be squashed a bit to the start of the interval, still there is a guarantee, that the LA ISR will not tick at a rate higher than 7 kHz.

I don't find my solution particularly elegant. I sweated it out with a particular purpose, taking into account the limitations of the platform, and I have profiled it with a logical analyzer. It remains now to be tested, how the extruder motor will handle it.

I will be out of office the next week, spending time with my family on the snow, so I will not be able to discuss this topic for a week, I plan to review your write-up on linar advance v2 though. During that week, my colleges will give it a try.

Sebastianv650:
Thanks for the writeup, it adds some more points I'm knowing at the moment to the list of problematic Marlin things..

> Your OCR1A = TCNT1 + 16 rule introduces an extruder feed rate limit for the inserted LA ticks. I am not quite sure what were your intentions, when you introduced this rule. Was your intention to introduce a feed rate limit, or did you want to reserve a big enough gap for the other interrupts to be invoked (the serial ISR likely being the most important), or did you just want to avoid the missed timer interrupt?

That wasn't my invention, it was in Marlin since I work with it. Maybe I added it to the PR just to be on the save side, I can't tell for sure.

The v1.5 PR on Marlin is always somewhat behind my development and at this time point I consider it as faulty. But I think the concept of using a fixed precalculated speed offset is promising and I testing a local version at the moment which combines fixed speed offset and checking for K steps executed as a limit. This one might be the version I searched for..

Good luck on the snow and have a lot of fun!

bubnikv:
> The v1.5 PR on Marlin is always somewhat behind my development and at this time point I consider it as faulty.

Sorry, I don't understand.

> But I think the concept of using a fixed precalculated speed offset is promising and I testing a local version at the moment which combines fixed speed offset and checking for K steps executed as a limit. 

I read your write-up and it revealed some new things, but I have a difficulty to understand some of your proposals.

Anyway, in regard to the smoothness of extruder movement and a gap fill: When I started coding on Slic3r, I implemented the infamous "volumetric slope limit" feature, which I never finished. The feature had the purpose to smoothly modulate the extruder movement speed, but it failed on gap fill (I don't remember the details, I believe the gap fill forced everything to print slowly). There is a "constant volumetric speed" mode in the original Slic3r, which had the same issues with the gap fill as well.

I believe the correct solution for the gap fill would be to extend the firmware to be able to drive the extruder independently from the Bresenham algorithm applied to the XYZ axes to be able to extrude linearly widening or narrowing lines. If there was a support in firmware, I would modify Slic3r to make use of it.

Sebastianv650:
> Sorry, I don't understand.

I tried to say that at the time I wrote this text the code in the branch for v1.5 LIN_ADVANCE was broken and that I had a new version lokaly on my hard drive.

> I believe the correct solution for the gap fill would be to extend the firmware to be able to drive the extruder independently from the Bresenham algorithm applied to the XYZ axes to be able to extrude linearly widening or narrowing lines. If there was a support in firmware, I would modify Slic3r to make use of it.

Yes, I was also thinking about that. Basicaly instead of using "G1 X50 Y10 E1" where E tells Marlin how much filament to extrude during the whole move we would need two "E" options to specify a fill rate at move start and one on move end. One could be even omitted when we say start of next move is always end of last move.
I think implementing a second line tracer for the extruder shouldn't be that bad. But I'm not sure if it's worth the effort. The problematic kind of gap fill like the tip of a triangle is not so often needed, and even when it's needed it's usualy very short. So if we don't start with the perfect pressure for the first move, we might not have the time to adjust the pressure at all. Especialy bowden printers seem to be quite impossible with pressure advance even on normal moves due to their extrem spring aka filament.

But expect from that I consider my v1.5 branch as nearly finished today. I want to add more intelligence for variable width paths, but beside that I'm using it for my daily prints since a week or so with not a single fault. It runs much smoother than my original one.

PR #463, opened 2018-02-05
Mk3

murathanaraz:


3d-gussner:
@murathanaraz Did you test this PR?

TheZeroBeast:
I have tested and have some comments.

The total crash fail menu has an artifact on top left of lcd the flashes quickly.

Filament runout (laser) detection failed by not sensing the runout before filament had run past Extruder gears so it couldn’t be unloaded.
Once manually removed and new filament inserted it didn’t automatically pickup despite having this feature enabled (I double checked).
Once it started resuming it went to last location and printed extra slow for 3mm before raising ~50mm and then printing mid air. After a minute it detected a crash mid axis/mid air and re-homed continued. X Axis offset 1mm right.

Crash detection homing on the X axis doesn’t return to the same home, offset of 1mm compared to initial X axis homing.
May I suggest that the homing procedure to recover from crash has different config/setup to the G28 W initial homing.

Belt status is X 244 Y 274

GurliGebis:
What is the point of this pull request? The MK3 branch cannot be merged with the MK2 branch.
Prusa Research have said that there will be a unified firmware repo later, where one repo will work for both printers, but not for now.

stahlfabrik (approved):
Awesome!!!

fiveangle:
Someone close this monstrosity 😜

3d-gussner:
@murathanaraz Can you please close this pull request as it doesn't work.

PR #485, opened 2018-02-21
Unified the volumetric_multiplier with extrusion_multiply to improve

bubnikv:
Unified the volumetric_multiplier with extrusion_multiply to improve 
numeric accuracy and to reduce compuatitonal load. With this commit, 
the numeric rounding is fixed not only for the M221 G-code 
(as implemented by the preceding commit), but also for the volumetric 
extrusion in general.

Removed the old FILAMENT_SENSOR code, which served the purpose 
to modulate the volumetric multiplayer in real time depending 
on the measured filament diameter. This feature will certainly not be 
used by Prusa Research in the near future as we know of no sensor, 
which would offer sufficient accuracy for a reasonable price.

### DIFF
--- a/Firmware/Configuration.h
+++ b/Firmware/Configuration.h
@@ -831,34 +831,7 @@ const bool Z_MAX_ENDSTOP_INVERTING = true; // set to true to invert the logic of
 //
 //#define NUM_SERVOS 3 // Servo index starts with 0 for M280 command
 
-/**********************************************************************\
- * Support for a filament diameter sensor
- * Also allows adjustment of diameter at print time (vs  at slicing)
- * Single extruder only at this point (extruder 0)
- * 
- * Motherboards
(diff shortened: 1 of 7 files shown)


3d-gussner:
That's a pity that the FILAMENT_SENSOR width didn't get more attention, even filament getting better in tolerances still see a value in it.
But you at Prusa have few more print hours than i have :-)
So guess you know if it is handy or not.

bubnikv:
It is much easier for us to set up our own filament extrusion lines with our filament quality control than to solve the filament quality control at the customer side :-)

TheBrigandier:
I am unable to compile MK3 branch now due to ConfigurationStore.cpp having another "calculate_volumetric_multipliers();" that needs changing to the new function name. Can you confirm?

Thanks!

PR #491, opened 2018-02-22
Safety timer (disable heaters after 15min idle)

XPila:


Blizzke:
And what if someone wanted to pause their print overnight but leave the bed heated to make sure the print doesn't detach?

Panayiotis-git:
It will not affect the paused prints.

dalgibbard:
Similar functionality is available in an Octoprint plugin - if no active job present, but elements are set to heat, they turn off after a preset duration. Helpful if you preheat to change/remove filaments, or warm up the bed to remove a part etc, and forget to turn it off again :)

Nice feature.

PR #503, opened 2018-02-25
Renaming calculate_volumetric_multipliers() to...

TheBrigandier:
calculate_extruder_multipliers(). MK3 firmware will no longer compile when LIN_ADVANCE is enabled in Configuration_adv.h due to this function name change. I am betting nobody noticed it because LIN_ADVANCE is off by default now. Several of us are using it with sdcard only without issues. Thanks!


Sebastianv650:
You might try if this is still necessary with #504?

TheBrigandier:
No longer needed, fixed in #528 which just got merged.

PR #514, opened 2018-03-03
MK3 Fix temperature dependent 1st layer height

### LINKED ISSUE #373: PINDA temperature calibration gone bad
Hi,

my PINDA probe sensor on my MK3 got heavily temperature dependent since a few days - it might have coincided with the latest firmware release. I also did a factory reset and also tried a temperature calibration. To find a solution I did again a factory reset and reflashed the firmware.

But it does not help: the warmer the PINDA gets the more squished the layers get. And when the PINDA is cold, the layer height is way to high.

Is there a way to show the internal temperature calibration table that my EINSY uses? How can I reset this table so that it has default values? What could I do see if it is hardware issue (the PINDA generally works and shows sane temperatures under support/temperatures)

On the picture: left the PINDA was warm after a print (layer slightly to squished). Right it was room-cold (light shines through space between extrusion lines). Of course both use the same z live adjust value!
img_1213

Best regards

stahlfabrik:
There is a substraction of the temperature compensation offset in the mesh bed leveling function that must be changed into an addition in order to achieve stable first layer height in a wide range of PINDA temperatures.

Background:
Replacing the PINDA probe and then replacing the complete MK3 did not solve my problem that I was incapable of achieving a perfect - or sometimes even a working - first layer. With everyprint I had to adjust the live z value. I had to start prints multiple times before the first layer was roughly OK to let the print progress.

Since hardware replacements did not help I started to dig into the firmware to search for a reason or even a solution for the problem.

I started by commenting out that line were the offset_z is set to the PINDA temperature corrected value. Thus the offset was always zero. Instantly the first layer height got far more consistent and much less temperature influenced. Then I knew what I should focus on next.

The problem is that the warmer the PINDA gets, the less negative I have to set the live adjust value. When the printer is cooled down again, the live z value must be set way more negative again.
That is because the PINDA probe triggers lower with increasing temperature.
The problem with the stock firmware is until now, that the PINDA triggers lower and still the firmware SUBTRACTS roughly the same amount of millimeters instead of ADDING that amount in order to compensate the physics of the PINDA probe.

Using this little patch now results in perfect first layers every time for me and I even can fine tune the first layer among multiple prints - that was absolutely impossible for me before.

fixes #373

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3768,7 +3768,7 @@ void process_commands()
 				SERIAL_ECHOLNPGM("");
 			}*/
 //			#endif // SUPPORT_VERBOSITY
-			mbl.set_z(ix, iy, current_position[Z_AXIS] - offset_z); //store measured z values z_values[iy][ix] = z - offset_z;
+			mbl.set_z(ix, iy, current_position[Z_AXIS] + offset_z); //store measured z values z_values[iy][ix] = z + offset_z;
 
 			custom_message_state--;
 			mesh_point++;


TheBrigandier:
Thank god, ready for this to be in. Would explain why my Z offset was so jacked up when switching between PLA and PETG.

alexiri:
I'm having problems calibrating the first layer consistently and I think this firmware patch has helped. Previously I would get very inconsistent results and it would take lots of tweaking to get a good first layer along the entire bed. If I then let the printer cool completely and try to print after an hour or so, the first layer would again be terrible.

I think this patch solves this. Yesterday I flashed this firmware and recalibrated everything until I got a decent first layer. This morning, I turned on the printer again and printed a test and the first layer came out like the night before.

stahlfabrik:
@alexiri thank you for the feedback! That is exactly what I found too: Let's hope Prusa will a)acknowledge and then b) fix this. 
Looking at the code I think the fix is as easy as my pull request for MK3. But if you look at the surrounding code for the MK2 maybe there is a little more to do for Prusa - except if the bug is true for this machine as well:-)

A comment from Prusa devs would be greatly appreciated!

cjo20:
I haven't looked at the code in detail (just searched for what might be similar situation), does the same fix need to be applied here? 
https://github.com/stahlfabrik/Prusa-Firmware/blob/1b19acf5b7991283174efaaf1dda20c52a506213/Firmware/Marlin_main.cpp#L7417

vanvlierden:
I installed the patch and found the first layer to be much more consistent over the width of the bed than before. I would like to see confirmation that this patch is correct.

stahlfabrik:
The link to my write up
https://shop.prusa3d.com/forum/viewtopic.php?f=30&t=14815

stahlfabrik:
I am still 100% sure of my bug fix - the people in the forum acknowledge that it works! Still I have not gotten any feedback from Prusa Research yet @bubnikv @josefprusa  @PavelSindler @XPila @michalprusa It would be nice to get some kind of feedback and outlook on what you will do with my findings.
I can print now PLA, PETG and ABS now without ever touching live Z again. Before I had to restart in average about two to three times in order to get a half way decent layer. Support and I even arranged for an MK3 kit replacement - so I think this is HUGE for you.

I understand it is weekend now: I hope to hear from you the coming week. Would really appreciate it

Best regards stahlfabrik

Crunch69:
Improved first layer, improved xyz calibration, improved mesh bed leveling
still needs a bit of tweeking

stahlfabrik:
Hello @PavelSindler 
Thank you for your reaction about #542 and #543.

What about this Pull Request? For my MK3 it is correct to change the sign of the operation. For other MK3, it may be not the right thing to do? I do not know that of course. What would be an optimal solution in this case would be if the step offsets would not be unsigned integers but signed integers. That way, everybody could store positive or negative ustep values and this pull request would be obsolete. Yet, of course only if this Pull Request does not solve the problem for every machine like it does for mine.

Best regards

stahlfabrik:
@XPila @PavelSindler 
I want to follow up on my previous comment.

These are my manually calibrated (print based) EEPROM values that work perfectly for me:
Temp	LiveZ		usteps (unsigned)
35C	-0,822	0
40C	-0,810	5
45C	-0,795	11
50C	-0,795	11
55C	-0,787	14
60C	-0,790	13

(Ergo I put Live Z to -0.822 and do not have to touch Live Z any more)

The problem is: These values work ONLY because in my local firmware I have changed the sign of the math operation that this pull request is about. So in order to run the stock firmware, I would have to store NEGATIVE ustep values to eeprom. (-5, -11, -11, -14, -13). The firmware on the other hand uses the EEPROM values as unsigned integers.

So there are two possible solutions - that I see:
a) my Pull request is correct and the sign of the operation needs to be changed for all MK3s or
b) the firmware must be changed from unsigned to signed integers in not too many places.

No matter which solution is the best - PLEASE - implement one of them in the upcoming release. For optimal flexibility b) might be the best approach

Best regards.

stahlfabrik:
I also wanted to add this link, where the supplier of the PINDAv2 probe has drawn the temperature response graph of the PINDAv2 probe. It also states that the probe triggers LOWER with HIGHER temperature. Thus my observation is backed by that sketch. So yeah, my PR should be OK for all MK3 https://shop.prusa3d.com/forum/hardware-firmware-and-software-help-f64/pinda-v2-temp-compensation-graph-t15492.html#p76343

egret85:
The values that I get on my MK3 (Black PINDA) seem to indicate that the original sign is correct and that the PINDA triggers higher the hotter it gets.  

My values are:
35C	-0.540	0
40C	-0.545	2
45C	-0.570	12
50C	-0.630	36
55C	-0.710	68
60C	-0.820	112

stahlfabrik:
@egret85 
Yes, my current conclusion is that the sign is just a sign - both are valid options:-)
There are surely MK3s were the one sign works (like yours) and for others the switched sign is working (like mine).
The magic thus needs to happen at the sign of the offsets. Thus I have set my offsets now all to negative values and try to make it work on the official stock firmware.

So with my modified (switched sign) firmware. My optimal values are:
0, 40, 5, 12.50
1, 45, 11, 27.50
2, 50, 11, 27.50
3, 55, 14, 35.00
4, 60, 13, 32.50

Now, as I need to set all offset negative to be able to use the stock firmware, this is what happens:
0, 40, 65531, 163827.50
1, 45, 65525, 163812.50
2, 50, 65525, 163812.50
3, 55, 65522, 163805.00
4, 60, 65523, 163807.50

That is because internally the EEPROM values are stored as UNSIGNED values, which sucks:-) 
But there is hope - the function that delivers the offset for a given temperature works internally using SIGNED integers, so these scary looking huge numbers seem to translate to the correct values:
35 : 0.00 : 0.00
36 : -0.85 : -0.00
37 : -1.74 : -0.00
38 : -2.70 : -0.01
39 : -3.78 : -0.01
40 : -5.00 : -0.01
41 : -6.38 : -0.02
42 : -7.81 : -0.02
43 : -9.15 : -0.02
44 : -10.26 : -0.03
45 : -11.00 : -0.03
46 : -11.29 : -0.03
47 : -11.26 : -0.03
48 : -11.08 : -0.03
49 : -10.93 : -0.03
50 : -11.00 : -0.03
51 : -11.40 : -0.03
52 : -12.05 : -0.03
53 : -12.79 : -0.03
54 : -13.49 : -0.03
55 : -14.00 : -0.03
56 : -14.21 : -0.04
57 : -14.14 : -0.04
58 : -13.87 : -0.03
59 : -13.47 : -0.03
60 : -13.00 : -0.03

So currently I hope that this does not lead to the nozzle to drive through my heatbed unexpectedly:-) I have yet to try with different PINDA temps, but it seems it should work.

I really would appreciate it, if Prusa Research could make the few code changes, to make the offsets signed throughout the firmware.

Attention is needed before each firmware upgrade, so that the seemingly huge offsets suddenly are not going into effect possibly damaging the printer.

stahlfabrik:
@bubnikv @XPila @PavelSindler 

How should we proceed with this Pull Request?

Because if I need to put positive offsets to EEPROM then I need this Pull Request (aka change the sign of the mathematical operation)

And if I put negative values into EEPROM I can only pray that the firmware handles the values correctly and it does not lead to destruction of the MK3:-) The problem is that the offsets are stored as unsigned integers and many places in the firmware treat them as unsigned (e.G. the new M861 gcode or the similar Dcode) I THINK that the function, where I propose the sign changed, handles the EEPROM values as signed (normal) integers. No I mean that function that calculates the temp compensation value for a given temperature. It gives back negative values if the eeprom values are entered negative.

So should we close this pull request? Will you be able to store the EEPROM values signed and treat them signed throughout the firmware for the final release? That would be the best solution and greatly appreciated by all customers who found that this Pull Request works for them!

Regards

Edit: See comment above, too, please

stahlfabrik:
This PR is not needed anymore:

Happy to say that 3.2.0-RC2 SHOULD fix the issue (by supporting negative ustep offset values)

Thank you Prusa Research for adopting my work on that issue and improving it further!

PR #519, opened 2018-03-05
Update language_de.h

Xillius178:
Why do you use not the German "Umlaute" ä ü ö ???

So i have corrected some wrong words.

Hope this ist helpful for you.

Greetings Martin aka Xillius178


GurliGebis:
Might be that the font the LCD is using doesn't have them? (Not sure, just guessing)

Xillius178:
That can be, depending on which library is used,

Xillius178:
The Code on the last of this Homepage could help:
http://www.mathias-wilhelm.de/arduino/beginner/schritt-6-lcd-displays/

AnHardt:
Usually the display on the `REPRAP_DISCOUNT_SMART_CONTROLLER` is a HITACHI `HD44780` in the Japanese font version. (https://www.sparkfun.com/datasheets/LCD/HD44780.pdf page 17)
'ä' = 0xE1, 'ö' = 0xEF, and a mixture between 'ü' and 'Ü' = 0xF5. 'beta' can be used as 'ß' = 0xE2. 'Ä' and 'Ö' are not available.
You could use that but it will break the compatibility with all the graphic displays.
In the language files you have to use the hex numbers. Usually your text editor (including Arduino IDE) will produce 2 byte UTF8 sequences. They will be seen as two strange glyphs.

Firmware/language_de.h
```diff
@@ -127,15 +127,15 @@
 	+#define MSG_SD_CANT_OPEN_SUBDIR             "Kann Unterverz. nicht oeffnen"
 	+ #define MSG_SD_INIT_FAIL                    "SD Init fehlerhaft"
 	+ #define MSG_SD_VOL_INIT_FAIL                "Dateisystem Init fehlerhaft"
-	+ #define MSG_SD_OPENROOT_FAIL                "Zugriff auf Basisverzeichnis misslungen"
+	+ #define MSG_SD_OPENROOT_FAIL                "Zugriff auf Basisverzeichnis Misslungen"
```
benderl, line 130: "misslungen" is not a noun and must not be written with a capital m.

Firmware/language_de.h
```diff
@@ -127,15 +127,15 @@
...
 	+ #define MSG_SD_CARD_OK                      "SD Karte ok"
-	+ #define MSG_SD_WORKDIR_FAIL                 "Oeffnen Arbeitsverzeichnis misslungen"
-	+ #define MSG_SD_OPEN_FILE_FAIL               "Fehler beim Oeffnen der Datei: "
+	+ #define MSG_SD_WORKDIR_FAIL                 "Oeffnen Arbeitsverzeichnis Misslungen"
```
benderl, line 132: Same as above. No capital m here.

Firmware/language_de.h
```diff
@@ -127,15 +127,15 @@
...
+	+ #define MSG_SD_OPEN_FILE_FAIL               "Fehler beim oeffnen der Datei: "
```
benderl, line 133: Here "Öffnen" has a capital O.

Firmware/language_de.h
```diff
@@ -316,7 +316,7 @@
 #define MSG_EXTRUDER_3									"Extruder 3"
 #define MSG_EXTRUDER_4									"Extruder 4"
 
-#define MSG_WIZARD							"Wizard"
+#define MSG_WIZARD							"Erste Schritte Assistent"
```
benderl, line 319: I would suggest just "Assistent".

benderl (changes requested):

3d-gussner:
I did some language adjustments and added some other and what you have to watch out is the limited space and location on the LCD. So you have to watch out even if you translate things right they might not fit and will fail on the screen. Long texts aren't the issue as they scroll but even here you have to remember how it looks on that LCD. Sometimes it makes more sense to split words to make it readable otherwise you get few lines with just one or two words, which isn't pleasant to read and follow the instructions.

It is a pity that Prusa didn't use the LCD with western European ROM as they sell that much printers. In German you can write ö,ü,ä,β as oe,ue,ae,ss and will be understood by German speaking users, but French/Spanish/Portuguese/Swedish often don't have a different way to write it and makes it odd to read.

3d-gussner:
@Xillius178 This PR have been inactive for long time, can you please close it? Or does it still makes sense?

3d-gussner:
@DRracer no reaction from owner, guess it is outdated and can be closed.

3d-gussner:
@michalxfanta No reaction from owner, guess it is outdated and can be closed.

PR #543, opened 2018-03-09
Gcode tool to modify PINDA temp compensation table

stahlfabrik:
To achieve a consistent first layer height it is necessary to have the optimal temperation compensation offsets in the EEPROM. The default values are very, very off for my grey PINDA probe and also the build in automatic temperature calibration does not lead to good values.

With this Gcode the user is able to reset the values to factory default, zero out all values in order to be able to do manual temperature calibration and to set a ustep value for each of the temperature slots of 40, 45, 50, 55 and 60 C.

With that code and my procedure to do a manual temperature calibration I was able to achieve a perfect first layer at PINDA temp of 35 and of 55C - without ever changing live Z.

Please see me forum post (which I will link soon) for more information about my manual temperature calibration procedure.

This gcode is of course just a hack to get users rolling - the ability to change these values MUST come to the menu - please!


stahlfabrik:
https://shop.prusa3d.com/forum/viewtopic.php?f=30&t=14815

The link to my write up

PavelSindler:
It was merged with PR: https://github.com/prusa3d/Prusa-Firmware/pull/612 so I am closing this one. Thank you very much for your effort.

PR #439, opened 2018-01-24
MK3 - PINDA2 Calibration Tuning

TheZeroBeast:
Tuning of the PINDA2 Internal Thermistor table to better align with temperatures seen as they can get upward of 80C.
Also taken into account errors happening causing concern for machine safety and hard limits introduced for these cases. Issues raised under #421

Adjusted the echo temperatures function to show the PINDA2 temp when preforming this calibration.

This is tested and working on my MK3 with good results so far, hopefully a guide that some part or all can be incorporated to make the firmware more stable for all.

Rob "TZB"

PR #552, opened 2018-03-13
New PINDA temperature calibration

lukasmatena:
New PINDA temperature calibration procedure with constant temperature bed and linear/parabolic fit.
Needs several more changes: LCD status messages, etc.


stahlfabrik:
Hi Lukas I PMed you my results from the new calibration process via the Prusa forum PM function. best regards

3d-gussner:
@stahlfabrik Hi can you share the results and how is this PR compared to yours and/or https://github.com/prusa3d/Prusa-Firmware/pull/439

TheZeroBeast:
Good idea @3D-gussner, it would be good to see some results and get this issue sorted once and for all. Eagerly awaiting the outcome from PRUSA & community on this.

stahlfabrik:
@3d-gussner @TheZeroBeast 
I posted my results on the Prusa forum as well. https://shop.prusa3d.com/forum/original-prusa-i3-mk3-f30/1st-layer-problems-in-depth-look-at-software-pinda-t14815-s90.html#p71795

Personally I am not a big fan of the new data model (calculating a linear equation for calculating PINDA temp offsets for a given temp) because my research has shown (to me personally) that my PINDA (or at least my MK3 system) does not work linearly.

That being said, I also think that a semi automatic/ manual print based temperature calibration like I did will work better than any automatic PINDA based approach. Because in the end, with different temperatures, the bed warps, plastic warps, the PINDA triggers in different height etc. So only looking at the PINDA might not yield optimal results.

With my manual calibration I can print PLA, ABS, PETG without ever adjusting live Z.

So I would hope that Prusa Research make it possible to let the offsets be signed (currently they are uint16_t) and that there are two more temperature calibration points at 25 and 30 degrees, so that one does not have to preheat the PINDA.
Then there should be a guided print based temperature calibration process and of course a menu to change the stored offset values in EEPROM.

That is my personal opinion.

3d-gussner:
@stahlfabrik Thanks for the update.
As far i know the grey PINDA is linear and the black PINDA isn't, that's why Lukas made this PR.
I hope @lukasmatena and Prusa will use all feed back to get it working fine.

It would be great if there were a automatic version (for most users) and an advanced option to be able to manually fine tune the PINDAv2.

It exists for the bed leveling (bed level correct), where you can manually adjust few things.
BTW: I like the hyperfine bed tuning more as it can adjust 8 points and not just left,right,front and back.

stahlfabrik:
I am quite confident that Prusa Research will fix this issue for all users. 
I think yes, maybe the grey pinda is linear or maybe even constant with temperature but the whole system, bed, sheet, nozzle, heaterblock, plastic around hotend, plastic holding pinda all react differently to heat. 
That is why a simple: „let’s heat up pinda and take z measurements“ might not be good enough. 
Maybe (I think that) a print based temperature calibration (where the user dials in live z for different pinda heat levels) to gain offsets is better over all.

Crunch69:
lukasmatena fork improved my first layers quite a lot.   stock firmware bed leveling correction too basic does not allow adjustment in the center or mid points of build surface would like to see the addition of 3d-gussner hyperfine bed tuning to get the bed close to 100%

Crunch69:
odd that PRusa is not merging this  bug fix

stahlfabrik:
As Lukas wrote it needs a lot more work. I hope they will tackle the problem once and for all so that all users can start printing without worrying about an ever so temperature sensitive live z offset. For that future I like to give them all the time that is needed to do it right:-)

3d-gussner:
@lukasmatena @stahlfabrik @Crunch69 @TheZeroBeast Do you think we still need to look into this PR as we have the SuperPinda now. I am trying to clean up the PRs.
Same for https://github.com/prusa3d/Prusa-Firmware/pull/735

TheZeroBeast:
Hi @3d-gussner,
I also feel it should be closed. Not that I’ve had a super pinda but it does appear to have negligible temp drift in our application.

PR #575, opened 2018-03-23
Adc pins, pinda temp table

PavelSindler:


Firmware/ultralcd.cpp
```diff
@@ -1824,7 +1824,7 @@ static void lcd_support_menu()
     
   MENU_ITEM(submenu, MSG_MENU_TEMPERATURES, lcd_menu_temperatures);
 
-#if defined VOLT_BED_PIN || defined VOLT_BED_PIN
+#if defined (VOLT_BED_PIN) || defined (VOLT_BED_PIN)
```
Lexx666: This OR makes no sense. Perhaps here is meant VOLT_PWR_PIN? Just guessing from line 1698

PavelSindler: Yes, you are right :-) Thanks.

PR #579, opened 2018-03-26
Fix wraparound from lowest value to highest value in menu value edit.

mkbel:


3d-gussner:
This has been recently changed in pull request https://github.com/prusa3d/Prusa-Firmware/pull/472 ... is there an issue after this 'Menu - encoder position stack' fix you have?

mkbel:
Yes, 'Menu - encoder position stack' 081b385 has broken MENU_ITEM_EDIT
as in #define menu_edit_type
condition:
if ((int32_t)encoderPosition < 0)
is never fulfilled if type encoderPosition is uint16_t and not uint32_t

PR #650, opened 2018-04-21
Update the description for MINTEMP_MINAMBIENT

jwest75674:
EDIT: Whoops, Sorry, I am just learning how github/git works: I accidentally commit more updates to my local repository clone which were NOT meant to be pulled. I didn't realize pull requests stay up to date. As such I undid the changes that were only meant for my own testing. For this reason there are 5 commits. (I only meant to request a pull of the updated description to MINTEMP_MINAMBIENT, which has been fixed now)

=============== Original comments + Questions =================

Update the description for MINTEMP_MINAMBIENT to be closer to grammatically correct.

Questions for those who know better than me:
(I want to learn and contribute as I go)

In these comments what symbols can we use safely? Is there a best practice list?
I kept the character count exactly the same as I feared the implications of changing the file size or memory usage or something. Is this level of care required?


stahlfabrik:
I think your new text does not match the old text. I understand the old text better:-) 

Especially this sentence has a totally different meaning in the original text: "+// Temperature inside the case is around 31C when in use (25C when the printer is powered on for a long time but idle)"

jwest75674:
Edit: Added some more thoughts.

Thanks for the feedback!
To be clear, none of the below is directed at you stahlfabrik just general comments and observations, but please keep letting me know your thoughts. I spent more time writing the below than was probably needed, I'll work on keeping it shorter next time, haha. the below is all meant to be helpful, not harsh. Let's make this as good as possible for everyone to understand. 😄 And thanks for replying! I am eager to keep learning git and github, so thanks for helping!

It's possible I misunderstood the original text. (I re-wrote it specifically because it was broken English that was extremely hard for me to understand.) English is my only language, I only mention as it seems relevant.

(I suspect the original was written by someone with English as a second language as Google translate would not make some of these mistakes, but overall it's well enough to get started. I think it's English second language as the sentence syntax is swapped on a line or two, and at least one word ("temperated") shows that this was likely human translated. On an interesting side note, anyone who learned English as a second language could find that the original (broken) English is easier to read. This is because the English language is the odd language. It swaps sentence order in many instances compared to many other popular languages. --> English is weird, lol)

At a minimum there was:
- Spelling and typo mistakes. ("litlebit" --> "little bit", "termistor"--> "thermistor",  "that"-->"than" )
- 'and is "temperated" from it'... not even sure what to say about that one, but it's wrong.
- There are missing grammatical articles (e.g. "the", "a", etc.)
- There are missing Conjunctions (e.g. "if", "but" etc.)
- There are missing Correlative conjunctions (e.g. "higher" --> "higher than ", "at" --> "at which", etc.)
Each of these is a tricky part of English to learn. 

If you want to see a great poem which makes fun of how hard English is, see: https://www.youtube.com/watch?v=tfRSvTSY0d4
(This is my favourite way to explain the complexity of English.)

I re-read the original text a few times and think I understand better. I see the mistake in that one line having re-read the original a few times. I have corrected it now.
I've made another change or two, can you let me know if it is correct in meaning now, or if anything else is specifically incorrect in meaning?
--> The grammar, spelling, and sentence structure are all correct, I want to make sure I didn't confuse the "meaning" of other lines.

For anyone who wants a third party to confirm the changes see:
Grammar website spell check:
Original: https://www.screencast.com/t/y4h5OO4Ex2kh
jwest75674: https://www.screencast.com/t/NK8QwQMR
--> This site only checks for spelling and punctuation. Other things like "temperated" which does not make sense and may not actually even be a word, need the whole sentence to be rewritten in order to make sense. Additionally, this site cannot detect when two portions of a sentence are in the wrong order very well.


JW

stahlfabrik:
All ok

I am actually from Germany and learned English in school. 

I really do think that also code comments could be corrected in an open source software project by people who care - you obviously care a lot. That is ok. 

I personally care more for the code quality, features and correctness:-) 

But of course code comments are not only for the programmer but also for others that need to understand the code. So if you want to improve their quality it is also not a totally unnecessary thing to do. 

The sentence with the „temepered“: I like the original better. You write that heat is produced. I think the idea is that some components on the board get warm, not necessarily hot and so warm up the termistor. So it cannot measure the ambient temperature but it is influenced by the Temperature the other components generate. Heat is a to strong word. There are no heating elements in the board. :-)

stahlfabrik:
You just need to make sure that your corrections do never change the meaning. That creates potentially bigger problems than a ugly grammar or typos

jwest75674:
I agree completely thank you, it's actually not that critical, being a comment. I agree that the code is the most important thing. I really am using this as an opportunity to learn more about Git and Github. :)

Your ability to edit code is way more important than me editing grammar, for sure, haha.

You're right that changing a meaning is worse than doing nothing, but when someone reads it and understands it wrong, like I did, it's just as bad.

Thanks! The sentence ending with: "the board, which generates heat." is accurate. The board does generate heat, though not much.
( Heat =/= hot. Heat is a type of energy, like light, sound, or electricity. You can generate heat, light, sound or electricity. Unless stated otherwise, none of these implies a quantity of how much is made, just that some is made. Same with the sentence in question. Again, English is weird. lol)

stahlfabrik:
Ok I agree. 

But German is worse;-)

BTW. I read somewhere that this is indeed a good method to dive into GitHub and git. It seems to work:-)

3d-gussner:
@jwest75674 I think this PR is outdated, can you please close it?

PR #654, opened 2018-04-23
New current setting for MK3

XPila:
X, Y: no change
Z: +26.1%
E: -28.5%


jwest75674:
How did you come to these numbers? (Why Z: +26.1%, E: -28.5%?)

stahlfabrik:
I would also like to know. But XPila works @prusa so I think it is based on testing, observation and optimization. :-)

But nevertheless an interesting question

devilhunter84:
Stock current values for the Axes are X16 Y20 Z28 and E36
Applying this would make them X16 Y20 Z35.3 and E28

My math is simplified and perhaps flawed here, but i think that would make the new currents about
X=240mAh
Y=300mAh
Z=540mAh ( divided by two so each stepper gets 270mAh)
E=420mAh
(calculated with a scalar value of max 63)

This sounds about right for a Cartesian. 
Motor was getting around 550mAh before, which was over half of what was allowed by the hard current setting. Motor will get less hot now, but will result in more jamming when printing around 205 degrees.

XPila:
We meassured currents using current probes and compared with MK2. Z and E axis is modified to be nearest values as on MK2. We want reduce extruder motor heating and improve Z stability. 
Previous setting for E axis was based on force meassurement.

PR #656, opened 2018-04-24
Fix knob scroll direction

JoePrints:
Knob direction of MK3 is completely opposite of MK2S in menu direction and value amount

### DIFF
--- a/Firmware/Configuration.h
+++ b/Firmware/Configuration.h
@@ -704,6 +704,7 @@ const bool Z_MAX_ENDSTOP_INVERTING = true; // set to true to invert the logic of
 	#define ENCODER_STEPS_PER_MENU_ITEM 2
   #endif
 
+#define REVERSE_ENCODER_DIRECTION // reverses value and menu directions
 
   #ifdef LCD_USE_I2C_BUZZER
 	#define LCD_FEEDBACK_FREQUENCY_HZ 1000


stahlfabrik:
So MK2S has it reversed? Why not change it’s direction then?

People who only own a MK3 and maybe other marlin printers might prefer it non-reversed.

JoePrints:
There are 2 ways to look at it- the MK3 owner, and the MK2 owner.
Neither of them want the direction changed from what it is currently..... but one needs changed.
I assumed there were less MK3's out in the wild than MK2's and MK2 users have been using their machines quite a bit longer than the new horde of MK3 owners.
Both of my printers are original Prusa- it just doesn't make sense that they are opposite.

Panayiotis-git:
It was a wrong decision from the PR to change the direction of the knob from MK2 to MK3.
I think the only acceptable fix is to add a user configurable parameter to select the preferred direction.

JoePrints:
I am not opposed to a user configurable direction of rotation.
Whichever direction it is, just make them the same- besides..... which direction is actually "reversed"?

vestom:
Well, let us pretend it was a conscious decision by Prusa to change the direction to the "right (tm)" way going forward and deciding to make MK3 the break-in point. Then we can leave all existing MK2 and MK3 owners in blissful peace - besides the "unlucky" few who have both printers and have to cope with all their differences.
Please close.

Panayiotis-git:
I've almost completed my MK3 custom build. So, I power it on, and start navigating to the menus. The inverted knob direction is strange at the beginning, but after a while it seems natural.

Then I started the calibrations, live-Z, move axis, increase/decrease temperature... generally increase and decrease many parameters... by rotating the knob...

Happy enough with the results, I shut down the printer and went to the living room to relax.
To lower the lighting I turned the knob clockwise, but they behave abnormal: the light increased! Quickly I turned the knob at the wrong opposite direction.

Frustrated enough, I went to the amplifier to increase the music volume, so I turned the button anti-clockwise and this happened again: the volume decreased!!!

I thing that it is a universe conspiracy. All the devices I have behave wrong: when you turn a knob clockwise they increase the value. The sound volume, the lighting, the speed of the blender, the voltage of the power supply...
I thing I have to reverse engineer their circuitry and make it work the right way as the only one Original Prusa i3 MK3!

vestom:
Yeah, it is a conspiracy! On my Original Prusa i3 MK3, bought in the Prusa shop, the knob increases values when turned clockwise, so Prusa is clearly part of the conspiracy!

Perhaps when you make a custom build with odd hardware, you could make a custom firmware with the direction changed? Or perhaps you could even customize the direction in your custom build with a soldering iron??

Panayiotis-git:
Well, I feel like it is indeed a conspiracy against me! I mean, how on earth could I imagine that out there exist two versions of LCD units? Both of them look the same, but for one of them:
1. the PCB connectors are reversed and
2. the encoder produces pulses in reversed order

I bought the wrong version. I connected my MK2s LCD unit and everything was fine.

So, conclusion: **Prusa hasn't changed anything.** The following is valid for both Einsy and Rambo pin configuration.
`#define BTN_EN1                72`
`#define BTN_EN2                14`
`#define BTN_ENC                 9  // the click`

@vestom My MK3 is about 60% original. The LCD unit was bought from a site selling original PRUSA parts. There was no indication about not been compatible with the Ultimachine layout. Anyway, see what you can do with a soldering iron.

dsc02350 1

vestom:
@Panayiotis-git I feel your pain (don't ask :-) But Configuration Management is apparently not a big thing in these DIY printer kits, and "original" may mean different things, depending on where you buy.

Sorry, if I was a bit sarcastic. I sincerely hope you manage to sort it out and get a smoothly working printer! I really like my MK3 - although I envy your cool "non-original" black and white display color ;-)
Have fun! Cheers.

Panayiotis-git:
@vestom No problem!
My MK3 is also jealous of the MK2.5 display!

easonjim:
It is woking:
//buttons are directly attached using AUX-2
#define BTN_EN1                14
#define BTN_EN2                72
#define BTN_ENC                 9  // the click

PR #716, opened 2018-05-03
Set calibration status to calibrated also by using live adjust Z.

mkbel:
This fixes problem, that if the first layer calibration is never done, live adjusting Z during printing is not applied after printer reset. Also repeatedly using live adjust Z on uncalibrated printer leads to steadily decreasing meaningless stored live adjust Z value.


GilesBathgate:
@PavelSindler Please can this be backported to the MK2/~MK2.5~ firmware? It cherry-picks cleanly.

mkbel:
@GilesBathgate This doesn't need to be backported to MK2.5 firmware, as MK2.5 is built from MK3 branch. So it is already there.
@DRracer Is there any MK2 release planned?

GilesBathgate:
@mkbel The MK2 firmware built from the MK3 branch doesn't work. (I've tried it) That's why MK3 is on 3.9.0 while MK2 is still on 3.2.3 (See https://www.prusa3d.com/drivers/#driver-mk2)

GilesBathgate:
@mkbel Sorry I realise now you were only referring to MK2.5

PR #719, opened 2018-05-05
G76: fix Z homing after steel sheet removal

dillya:
When the steel sheet is in place, the user is asked to remove the
steel sheet before continuing the temperature calibration process.
After removal acknowledge from user, a Z homing is performed but world
coordinates are not properly updated and the extruder nozzle can crash
into the bed and then damage the bed and the nozzle.


dillya:
img_1135
Here an example of my bed damaged by the extruder after the steel sheet removal: the golden line comes from the brass of the nozzle which rub the bed to reach the heating point for PINDA temperature calibration.
(Another lines can be seen on start point 0:0 due to the self test for homing calibration which was buggy but fixed with fix #678).

3d-gussner:
@dillya I think this has been fixed in the meanwhile, can you review your PR and if it is not needed anymore please close it.

dillya:
I confirm the issue has been fixed by the call to gcode_G28() after steel sheet removal.
I close the PR.

PR #763, opened 2018-05-23
Enhanced TMC2130-related config options, removed hard-coded values

metacollin:
Hi, I don't know if these are acceptable changes or not, but I figured I'd at least make a pull request in case you like them.  

I did a little 'spring cleaning' of some of the TMC2130 setup code to make it easier to play with the config registers.   This is in the context of figuring out the mysterious extrusion bug effecting print quality of the MK3, and I thought this might make it a little easier to track down (if it is even caused in part by something with the TMC2130 - which may or may not be the case).  But I digress.  

Changes:

1.  Removed hard homing values and gave them their own explicit #defines in the MK3 config variant header.  It's also just nice for people playing with custom firmware - I think most people would assume that changing the TMC2130_SG_THRS defines for the X and Y axes would apply to homing, but as the code stands now, its hard coded to '3'.  

2.  Added a small macro that allows 'native' signed values for sgt to be provided by converting it to the unsigned representation of a 7-bit 2's compliment number.  In other words, stallGuard threshold values can be given as -64 to 63, and as long as _sg(<value>) is used, it will turn negative values into their correct 7-bit 2's compliment.  

3.  There is a fair amount of TMC2130-related #defines in the MK3 variant header that don't do anything.  Not just that, but the define doesn't exist *anywhere* else in the entire codebase, except for that single define.  This too is problematic as one might expect if the #define is present, then it must actually do something.  And there were a few duplicate #defines.   Anyway, they've been removed.  

4. Step interpolation can now actually be turned on or off per axis (all the interpolation related defines the code as it is right now don't actually do anything or even appear anywhere else in the codebase), as well as a few other interesting options have been moved into the variant header and given their own #defines 

5.  Comments documenting some settings have been added or clarified. 

6.  The reason for declaring uint8_t variables for some of the new flags in tmc2130.cpp is to make it easy to add custom gcodes to read or set them instead of having no choice but to recompile.  

Anyway, I completely understand if this is not a direction you're interested in and this pull request gets rejected.  I'm just creating it in case any of you think it is useful.

S pozdravem

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -6448,76 +6448,151 @@ Sigma_Exit:
     }
     break;
 
-	case 911: // M911 Set TMC2130 holding currents
+  case 911: // M911 Set TMC2130 holding currents
     {
-		if (code_seen('X')) tmc2130_set_current_h(0, code_value());
-		if (code_seen('Y')) tmc2130_set_current_h(1, code_value());
-        if (code_seen('Z')) tmc2130_set_current_h(2, code_value());
-        if (code_seen('E')) tmc2130_set_current_h(3, code_value());
+    if (code_seen('X')) tmc2130_set_current_h(0, code_value());
+    if (code_seen('Y')) tmc2130_set_current_h(1, code_value());
+    if (code_seen('Z')) tmc2130_set_current_h(2, code_value());
+    if (code_seen('E')) tmc2130_set_current_h(3, code_value());
+    MYSERIAL.println("Holding currents:");
+    MYSERIAL.print("X   ");
+    MYSERIAL.println(tmc2130_current_h[0]);
+    MYSERIAL.print("Y   ");
+    MYSERIAL.println(tmc2130_current_h[1]);
+    MYSERIAL.print("Z   ");
+    MYSERIAL.println(tmc2130_current_h[2]);
+    MYSERIAL.print("E   ");
+    MYSERIAL.println(tmc2130_current_h[3]);
     }
     break;
 
-	case 912: // M912 Set TMC2130 running currents
+  case 912: // M912 Set TMC2130 running currents
     {
-		if (code_seen('X')) tmc2130_set_current_r(0, code_value());
-		if (code_seen('Y')) tmc2130_set_current_r(1, code_value());
-        if (code_seen('Z')) tmc2130_set_current_r(2, code_value());
-        if (code_seen('E')) tmc2130_set_current_r(3, code_value());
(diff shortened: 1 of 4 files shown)


devilhunter84:
I'm all for added gcodes to change some TMC hardcodes values. Would be really nice if homing currents and thresholds can be set via gcode (always a reflash because of tmc2130_current_r_home is annoying)

metacollin:
@johnlaur Thank you, I don't know how I managed to screw that up.  That was definitely not intentional.  I've corrected the values to their defaults.  

@devilhunter84 I actually have said gcodes in my 'personal use' version of the firmware, along with some enhancements of one or two g codes so they will print the current values without changing them if you do not supply anything with the g code.   Since g codes theoretically should be standardized at some point, I didn't really feel I had the authority to just make up new ones willy nilly.  Plus, given the influence the Prusa Printers and thus firmware tend to have on the 3D printing community at large... I figured at best, including new g codes was a way to ensure the pull request was rejected.  

But screw it, I just added them in anyway =P.

Additions:
- `M919`: Print sgt values
- `M922`: Set TMC2130 homing currents, compliment to M912
- `M926`: Set tmc2130_sg_thr_home - compliment to M916

Enhancements:
- M911, M912, M916, M917, M918 as well as M922 and M926 will print back the values after they've been set.  This also means you can simply send, for example, `M912` without any additional codes to see the current running current values.  
- M913 prints homing currents in addition to running and holding currents.
- tmc current set functions now perform a range check and if the value is above 63 (the maximum), it will refuse to set the value or update the motor config, and print an error stating this.
- the small area I worked on has been formatted so it is consistent with the rest of the coding style in Marlin_main.cpp
- Even better documentation in the MK3 variant header.

moggieuk:
Love this and especially the M926 -- I suggested it in another issue request.  I'd like to decrease sensitivity for running (to mitigate false collision detection) but keep it sensitive for accurate homing. Nice work.

JohnOCFII:
> But screw it, I just added them in anyway =P.

If you haven't already, I suggest you add them to the G-CODE Wiki: http://reprap.org/wiki/G-code

XPila:
Gcodes M911 and M926 are not oficial, it is just for debuging (this codes are already used on RepRap for different functions). In feature we will create set of new Gcodes for trinamic, but in this time are the values hardoced, because invalid settings can degrade printing quality and reliability.

3d-gussner:
@metacollin Thanks for the PR but we are closing it. Please see latest MK3 branch and https://github.com/prusa3d/Prusa-Firmware/pull/4191 if your PR is still valid. Feel free to open a new one.

PR #771, opened 2018-05-26
TMC2130 configuration axis separation

NordicGuy:
When debugging visible vertical lines on prints, I discovered that all axis' are configured equally in regard to the TMC2130 drivers as hardcoded values, which is sub-optimal. Ideally each TMC2130 driver has to be tuned to its specific motor and their individual resistances and inductances to ensure optimal performance. 
All motors cannot be expected to be equal, especially due to manufacturing differences. On my MK3 I have discovered one motors resistance to be 7.2 ohms, which is 11% deviated from the motors specified average of 6.5 ohms, hence the default configuration for that axis was not optimal and produced vertical lines corrosponding to the currents zero-crossing. This was fixed by tuning the HEND and HSTRT parameters.
Thus I attempted to separate the configuration of general parameters for each axis  so that these may be configured individually by advanced users seeking to improve their print quality. 
Parameters are now setup on the format of; TMC2130_PARAMETER {X-axis, Y-axis, Z-axis, E-axis}
More advanced options were left hard coded in tmc2130.cpp. 

Ideally a system should be devised so that anyone with access  to a multimeter should be able to set these parameters directly from the display of the printer: The user would measure the resistance of each motor and enter this into a dialogue on the display, then the printer would itself calculate the deviation from spec and adjust each axis' parameters accordingly. Thus users could fine tune their printers without having to flash the firmware.

Until then, I hope my contribution can be of assistance.

PS. Parameters have been tuned to my machine to the best of my ability, these would have to be fine tuned for general distribution and/or for each individual machine.

### DIFF
--- a/Firmware/tmc2130.cpp
+++ b/Firmware/tmc2130.cpp
@@ -11,7 +11,6 @@
 #include "spi.h"
 #endif //NEW_SPI
 
-
 extern LiquidCrystal_Prusa lcd;
 
 #define TMC2130_GCONF_NORMAL 0x00000000 // spreadCycle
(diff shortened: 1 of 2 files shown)


Firmware/tmc2130.cpp
```diff
@@ -73,15 +67,19 @@ uint16_t tmc2130_sg_cnt[4] = {0, 0, 0, 0};
...
+uint8_t tmc2130_pwm_ampl[4] = TMC2130_PWM_AMPL;     //pwm_ampl  
+uint8_t tmc2130_pwm_grad[4] = TMC2130_PWM_GRAD;     //pwm_grad
+uint8_t tmc2130_pwm_auto[4] = TMC2130_PWM_AUTO;     //pwm_auto
+uint8_t tmc2130_pwm_freq[4] = TMC2130_PWM_FREQ;     //pwm_freq
+
+uint8_t tmc2130_wave_fac[4] = {0, 0, 0, 0};
+
 bool skip_debug_msg = false;
 
 #define DBG(args...) printf_P(args)
```
koenkooi: Were the changes in the block below intentional?

Firmware/variants/1_75mm_MK3-EINSy10a-E3Dv6full.h
```diff
@@ -183,85 +183,60 @@ const bool Z_MIN_ENDSTOP_INVERTING = false; // set to true to invert the logic o
...
-#define TMC2130_CURRENTS_R {16, 20, 35, 30}  // default running currents for all axes
-#define TMC2130_UNLOAD_CURRENT_R 12			 // lowe current for M600 to protect filament sensor 
-
-#define TMC2130_STEALTH_Z
 
-//#define TMC2130_DEBUG
-//#define TMC2130_DEBUG_WR
-//#define TMC2130_DEBUG_RD
+//                          {X, Y, Z, E}                           
+#define TMC2130_MRES        {4, 4, 4, 3}                // Default uStep steeting for each axis; X, Y, Z, E. ( 0 = 256 ... 3 = 32, 4 = 16 ... 8 = Fullstep) (ONLY IN TMC2130.cpp)
```
koenkooi: Which setting takes precendence? TMC2130_USTEPS_*  or TMC2130_MRES?

koenkooi:
I've been trying to debug the vertical line issue at home and not having much luck with it. Did your vertical lines disappear after changing HSTRT and HEND?

Firmware/tmc2130.cpp
NordicGuy: Yes, these changes were to bring the PWM settings in line with the format defined in the configuration file.

Firmware/variants/1_75mm_MK3-EINSy10a-E3Dv6full.h
NordicGuy: Thank you for pointing this out. The MRES option was a remnant from debugging and doesn't actually do anything. Will remove this now. 
TMC2130_USTEPS takes precedence then of course.

NordicGuy:
@koenkooi to the most degree yes. I am still tuning on the extruder, as I am still seeing visible extruder pulses on my prints. But on the X and Y the vertical lines have all but disappeared in tuning mainly HSTRT and HEND. Tuning TBL and TOFF also had an effect on audible chopper noise. 

If you've tried to print the E-correct tower, as defined on the prusa wiki, you could see the moire pattern going from 2 parallel lines to 4 lines on higher correction values. This is to due, from my understanding, with sinus curve distortion from the zero crossing (every 2nd and 4th full step) and platouing of the current at the maximum and minimum of the curve (every 1st and 3rd full step) , which would then show as visible lines.
My theory going into this was that the same was happening to an extent on the X and Y axis, leaving behind visible lines on the surface of the print. That in tuning the axis' parameters the vertical (and maybe moire!) pattern would disappear. Why every axis needed its own configuration options to tune out this effect. The tuning proces is a bit involved, because without a high grade osciloscope you can only go by sight and hearing though iterative tests, but so far it seems promissing. 

You can find some explanation of the tuning process in these documents: 
https://www.trinamic.com/fileadmin/assets/Support/Appnotes/AN001-spreadCycle.pdf 
https://www.trinamic.com/fileadmin/assets/Products/ICs_Documents/TMC2130_datasheet_Rev1.10.pdf

NordicGuy:
Made some updates to both the configuration and tmc2130.cpp. But it seems like the online editor now views the whole files as new. Am still learning github but that just seems strange to me.

XPila:
@NordicGuy values HSTR and HEND for MK3 was tuned by Trinamic engineers and also big part of every other settings. It is possible that some motors are different (out of tolerace). I will check.
Some unwanted effects can be also reduced by adjusting linearity correction (you can enable it for XYZ axes).

stahlfabrik:
@XPila one can enable linearity correction for axis different as E? How can that be done?

XPila:
@stahlfabrik not in release, but if you build own version with #define EXPERIMENTAL_FEATURES, it should appear in settings. But calibration process for XYZ not known in this time, because of this we disabled it.

XPila:
@stahlfabrik just one way I know is calibration by hearing, but it is unusable for many users. And the result effect on most normal prints is minimal or not visible. On the other hand - invalid settings can degrade printing quality and reliability...

XPila:
@stahlfabrik sory, my mistake, it is little-bit broken in last version, #define EXPERIMENTAL_FEATURES will not work, I will fix.

stahlfabrik:
Interesting! Thank you. Maybe one could use a frequency spectrum app on a smartphone to minimize certain frequencies? I do not know - just putting it out:-)

So with a gcode that moves only one axis back and forth - and then looking at the spectrum - would it be possible to optimize it that way?

XPila:
@stahlfabrik you are right, i am using spectroid :) it is better to place the motor dismounted from printer on the table, set silent mode and move at different speeds... but as i sed before - effect is minimal or not visible, depends on objects and also slicer settings.

stahlfabrik:
@XPila :-) OMG - dismounting the printer again is not fun at all:-)

So I would prefer taking a multimeter to measure the resistance of the motor (somehow?) as @NordicGuy has explained. That is easy to do - if it leads to better print quality (as he mentions) even better! What do you think, XPila?

XPila:
@stahlfabrik dismounting is necessary just for fine tune, because nonlinearity produces resonnance and in case of dismounted motor it can by better hear. I am working on calibration through stallguard that will be done in selftest and not need any external tools or calibration object, but it is not ready now...

stahlfabrik:
@Xpila: That sounds very nice! Any work into quality improvements are time well spend:-) As you at PR are very aware of:-)

XPila:
@stahlfabrik meassuring resistance is not accurate with most multimeters, because the resistance is relative low. Accuracy with stock hobby multimetr is +-10-20% (+-1ohm). Production tolerance should be lower. Because of this are the values hardcoded. But it is possible that some series of motors differs. Also is possible that some settings of trinamic is not perfect, we will check.

NordicGuy:
@XPila sounds interesting with the stallguard calibration! Looking forward to it. Any ETA?

The experimental features not working might explain why I didn’t see any differences when enabling them. Thus this solution. 

Were there only one unit sent out to Trinamic or more?

XPila:
@NordicGuy calibration with stallguard is in experimental phase, but the results are good. Now we have some other issues (localization, inconsistent extrussion..).
Linearity correction for XYZ is impemented, but some parts of code are disabled and cannot be simply enabled by #define EXPERIMENTAL_FEATURES. I will fix it asap, also other peaple ask me for that.

XPila:
https://github.com/prusa3d/Prusa-Firmware/pull/767
enabled correction for XYZ - commit 15207e4 //#define TMC2130_LINEARITY_CORRECTION_XYZ
this pullrequest will be merged in short time..

stahlfabrik:
@NordicGuy - you might also want to join the Discord channel (Klipper) - as we were discussing there to what values you had to change the HEND and HSTRT parameters after you measured the resistance of the motors. So standard is 6.5 ohms. You measured 7.2 ohm. In what way did you then change the parameters to make the wall artifacts go away? Thanks a lot!!

trevjonez:
> Made some updates to both the configuration and tmc2130.cpp. But it seems like the online editor now views the whole files as new. Am still learning github but that just seems strange to me.

double check the line endings. https://help.github.com/articles/dealing-with-line-endings/

Crunch69:
+1

stahlfabrik:
@XPila this whole thread goes through my mind again and again. The only thing my MK3 really suffers from is vertical lines in walls. I would be SOOO happy to try out your selftest calibration for stepper linearity. If the firmware JUST contained that - if it would spit out raw values that I could put in a custom build firmware to tune my MK3 - that would be wonderful.

How are the plans about that? How can a user without an oscilloscope tune the trinamics perfectly? Maybe there is a print base approach?

stahlfabrik:
Dear @NordicGuy 

May I ask you to describe how the process to get rid of the vertical line artifacts works?

I would Instantly buy a better multimeter if I would just learn how the tuning is done. I have no access to an oscilloscope. 

Please share your experiences of how you did it! 

Thank you very much and best regards
Stahlfabrik

Crunch69:
@XPila I too only have the vertical lines in walls. I would also be SOOO happy to try out your selftest calibration for stepper linearity. Or the instructions to edit the firmware for a custom build firmware to tune my MK3

@NordicGuy please provide some instruction or a starting point to get rid of or reduce the vertical lines
Cheers.
Crunch69

3d-gussner:
@NordicGuy Thanks but we are closing the PR. There have been made changes to TMC. Please check the MK3 if your PR is still valid and feel free to create new one.

PR #794, opened 2018-05-31
Improved M860 - now wait for cooling of PINDA, too

stahlfabrik:
The M860 - wait for PINDA temperature - gcode has been improved. Until now it could only wait until the PINDA has been warmed up to a set temperature. Now it also can wait until the PINDA has cooled down to a set temperature. The calling syntax has not been changed at all. The logic now looks if the bed or hotend have a target temperature different from 0. If so, the code waits until PINDA has warmed up. Otherwise it waits until PINDA has cooled down.

Applications:
- Possibility to start a Print at exactly the right PINDA temperature. (Do not start the heater until PINDA is cooled down)
- Much easier manual temperature calibration (automating the wait for cool down in between calibration runs)
- Possibility for temp calibration verification on one heat bed by printing multiple objects with different PINDA temperatures one after the other

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -6389,10 +6389,10 @@ Sigma_Exit:
 #ifdef PINDA_THERMISTOR
 	case 860: // M860 - Wait for PINDA thermistor to reach target temperature.
 	{
-		int setTargetPinda = 0;
+		int set_target_pinda = 0;
 
 		if (code_seen('S')) {
-			setTargetPinda = code_value();
+			set_target_pinda = code_value();
 		}
 		else {
 			break;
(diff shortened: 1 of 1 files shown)


stahlfabrik:
@PavelSindler hi! Since you last worked on the M861 gcode that I originally came up with, I wanted to kindly ask you if you could please checkout this change and merge into the running MK3 branch?

Best regards!

stahlfabrik:
@XPila @PavelSindler Here is a gcode file that I am currently working on (will update if needed), that uses the modified M860 logic. It tests all PINDA temperature levels of 35, 40, 45,.. 60C by printing a 2cm by 2cm square on the heatbed. It is all fully automatic - the user has just to make sure the ambient temperature is below 30C, so that the PINDA can cool down in between.

I think that this is a very nice example of how can one verify the manual bed calibration - fully automatically - using the improved M860 gcode logic

[temp_cal_veri v4.gcode.zip](https://github.com/prusa3d/Prusa-Firmware/files/2059715/temp_cal_veri.v4.gcode.zip)

EDIT: Corrections were made to the Gcode File - optimizations and hotter bed temps for hotter PINDA temps
EDIT2: Turning the fans on for hopefully quicker PINDA cool down
EDIT3: Yes - the fan is a great addition for cooling the PINDA quicker

stahlfabrik:
Yes! The Gcode is a total success. The six squares haven each been printer with a different PINDA and bed temperature.
img_1594

XPila:
@stahlfabrik great work! we will merge this pull asap. We plan redesign of PINDA calibration, to be more accurate, your opinion is appreciated.

stahlfabrik:
Awesome! I am happy to keep contributing

PavelSindler:
Thank you @stahlfabrik. It looks great. I am merging it.

PavelSindler:
I was looking on gcode file and I think it would be better to add M104 S0 and M140 S0 at the beginning of gcode file to be sure that we will be waiting for pinda probe cooling.

PavelSindler:
And I am not 100% sure if it is the best option to wait for PINDa probe cooling in case that both heatbed and nozzle target temperatures are set to zero. Another option is for example waiting for PINDa probe cooling in case that PINDa temperature is higher than PINDa target temperature and waiting for heating in case that PINDa temperature is lower that PINDa target temperature.

stahlfabrik:
Your idea to set heaters off at the beginning is a good idea! I will add that to the file!

Your latest comment I think sounds like a good idea but it is not really - because if the heaters are on, the PINDA cannot cool down, as it is passively heated by the heaters around it. Do you agree?:-) I cannot see another way to do it.

stahlfabrik:
And thanks for merging my work!

PR #827, opened 2018-06-10
New ML support

XPila:
printf float library enabled - platform.txt must be modified!
+flash usage optimalization - printing to serial line
current flash usage: 253206bytes (MK3 + ML support)

platform.txt original line 20:
compiler.c.elf.flags=-w -Os -Wl,--gc-sections

platform.txt modified line 20 (enable printf float library)
compiler.c.elf.flags=-w -Os -Wl,-u,vfprintf -lprintf_flt -lm -Wl,--gc-sections


XPila:
flash usage after 4a659ed is 251880 bytes.

XPila:
with LANG_MODE 0 235284 bytes.

PR #861, opened 2018-06-22
Minor fix on paragraph #26

kant:


3d-gussner:
@kant I think that the documentation has been updated, can you review your PR and if it is not needed anymore can you please close it?

PR #868, opened 2018-06-25
Fix compile error if STEEL_SHEET is undefined.

zbrozek:
result is declared inside of an #ifdef, but is used outside it. Moved the use inside the block.


3d-gussner:
@zbrozek The MK25 branch isn't used anymore. Can you please close this PR.

PR #869, opened 2018-06-25
Reflect NEW_XYZCAL dependency on HEATBED_V2

zbrozek:
Context: I've got a MK2.5 with an old bed. Noticed that the firmware doesn't compile if you simply unset HEATBED_V2. Minor improvement.


3d-gussner:
@zbrozek The MK25 branch isn't used anymore, can you please close the PR?

PR #880, opened 2018-06-26
Calibration-Selftest: Check hotend heater after mechanical checks

Panayiotis-git:
Currently the Selftest procedure checks the fans, then the hotend heater cartridge, then the mechanical stuff and finally the heated bed.
If a mechanical check fails, then the user has to wait for the hotend to cool before repeat the selftest.

With this PR, the heater checks are executed together at the end.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -6175,13 +6175,6 @@ bool lcd_selftest()
 	
 	if (_result)
 	{
-		_progress = lcd_selftest_screen(3, _progress, 3, true, 1000);
-		_result = lcd_selfcheck_check_heater(false);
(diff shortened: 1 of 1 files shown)


maximlevitsky:
I would really like this to be accepted, it also really annoys me

Dioblieum:
If the nozzle is heated and the noctura is not working it could cause damage to the extruder.  I’m sure that’s why they test the fan first

knutmatten:
would also love this

Crunch69:
+1

GriffinMeyer:
+1!

Panayiotis-git:
I will close this PR since PR has already implemented this.

Issue #764, opened 2018-05-23
MK2.5 3.2.1 XYZ calibration fails


gregsaun:
I have upgrade from 3.1.3 to 3.2.1 and was unable to finish XYZ calibration. First two points where good but the third one (first on back) failed, unable to find it. I tried to lower the PINDA and did not solved the problem. Same results from wizard or from calibration menu.

I have rolled back to 3.1.3 and XYZ calibration is successful with following values :
- Measured Skew:  0.00°
- Y distance from min : 
  - Left : 6.07
  - Right : 6.27

Panayiotis-git:
I had the same problem with any MK2.5 firmware release. Calibration always failed at the 3rd point.
The solution for me was to move the Z frame about 5mm to the front.
Now, I have Left: 9.12mm / Right: 9.09mm / Skew: 0.00

I've also noticed that the 3rd point is a bit crowded. The noise pattern that the new calibration algorithm produces over the test points is a lot different at the 3rd point than the others. The other 3 point produces a very clean pattern.

gregsaun:
Thanks for the info Panayiotis. I read about this 5mm delta but could not find the issue anymore. Observing the calibration makes me feel that it is the same issue than you, all circles are done few mm in the back.

However I could not observe the noise pattern you are talking about, but maybe it is because I associate this to the fact it cannot get the third point.

gregsaun:
Here it is, similar issue since 3.1.4 :
https://github.com/prusa3d/Prusa-Firmware/issues/616

WizardUli:
I seem to have had a same/similar problem. I have a completely different frame https://www.thingiverse.com/thing:1692666 (with MK2.5 parts and 3.2.1 firmware).

When I first upgraded form 'MK2-X' to 'MK2.5-X' I calibrated on the first try just fine on 3.2.0 firmware. Then later after some moving of the printer and upgrade to 3.2.1 I wanted to redo the XYZ calibration and it started to fail. A few times on point 1, a few times on point 4 and countless times on point 3.
So I tried lowering the PINDA probe, I even tried re-crimping its connector.

Then I remembered that when I first calibrated the XYZ, the PSU was not attached yet to the frame and it had been just lying on the table. So I removed the PSU from the right corner and the XYZ calibration started to work much more reliably. First time it failed again on point 1. but three more consequential calibration passed OK.

[...]

Issue #616, opened 2018-04-12
Prusa MK2.5 3.1.4 firmware calibration failed


Drifter4242:
I upgraded my Mk2s to a Mk2.5 and installed the 3.1.4 firmware. The self test went OK, but when I tried the XYZ Calibration it failed with "Calibration Failed". I retried 3 times - all failures.

Without altering anything, I downgraded the firmware to 3.1.3 and tried the calibration again. It succeeded! The first layer test also went smoothly.

I did notice that when the 3.1.4 firmware was installed, during the calibration, the bed attachment point hit the front (when it was detecting the rear calibration points). Not sure if that was the reason, but it certainly didn't sound healthy.

RacingHell:
Try to remove your springsteel sheet before starting the calibration, seems so the information for doing this is missing. (on 3.1.4 alpha MK3)

Drifter4242:
Thanks for the reply, but the springsteel sheet was already removed.

I was using 3.1.4 final firmware (for Mk2.5) which did ask to remove the springsteel sheet before the calibration.

Also I should also mention that when the calibration succeeded (with 3.1.3), the XYZ Calibration details give these results:

Y Distance from min:
Left: 10.02mm
Right 9.99mm

Measured skew: 0.03 (degrees)
Slight skew: 0.12 (degrees)
Severe skew: 0.25 (degrees)

PavelSindler:
We tested new xyz calibration on many printers here in Prusa Research and it seemed to be reliable, however from customers feedback it seems that it is failing to find third point on some printers (both MK2.5 and MK3). We will fix it. Thank you for your feedback.

cobramike:
Same here. Failing not on the 3 point. But also on points. Ill put it here. Tested it 2 hours straid.

1 / 3 / 3 / 3 / 4 / 4 / 1 / 4 / 4 / 2 / 2 / 4 / 3

And this is after the upgrade..

jolab14:
I am experiencing a similar problem. I upgraded from MK2 to MK2.5, with fw 3.1.4. Sometimes, the xyz calibration finds the 4 points on the first pass but stops saying that it failed. On other occasions, it does not find the 2nd point. I tried several time with various pinda heights. No success.

Update : I tried moving the Z frame a couple mm backward and then calibrate xyz, as well as moving it forward, but either way it was worst thant having it at 100mm as specified in the documentation.

[...]

PR #899, opened 2018-07-06
Fix mk42 bed calibration points

zbrozek:
Six of them are off by 2 mm in the y axis. Verified both by measuring gerbers and by testing on a MK2S.


gregsaun:
Would this be the same issue we are seeing since 3.1.4 on third point calibration with MK2.5? When I am looking calibration steps, it looks like the pinda is off on Y axis by similar value.

https://github.com/prusa3d/Prusa-Firmware/issues/764
https://github.com/prusa3d/Prusa-Firmware/issues/616

3d-gussner:
@zbrozek I think that has been fixed, can you please review your PR and if it is not needed anymore please close it.

PR #923, opened 2018-07-12
Fix upper limit for x

cwei:
Use MESH_NUM_X_POINTS for x.
This only works, because MESH_NUM_X_POINTS and MESH_NUM_Y_POINTS have the same value.


DRracer (approved):

wavexx (approved):
Can't believe this still applies!

PR #933, opened 2018-07-17
Revert "Save 80B of flash and 8B of RAM and fix compiler warning:"

PavelSindler:
Reverts prusa3d/Prusa-Firmware#918


3d-gussner:
@PavelSindler Any reason that the merge was reverted? As it fixes lot of compiling warnings as in the past from Thess for the MK2 branch.

PR #917, opened 2018-07-11
Use timer

mkbel:
Do not merge immediately into final release, merge into next release candidate.

PR #951, opened 2018-07-22
Use timer

PavelSindler:


PavelSindler:
Re-adding changes from PR: https://github.com/prusa3d/Prusa-Firmware/pull/917

PR #989, opened 2018-07-31
Add French Language

kuisac:
I propose to add French language. We are a french facebook prusa i3 community and we propose to add French language and we create the lang_en_fr.txt (P.S. : I update the .h library language)


kuisac:
Sorry I propose this update with a unique commit.

PR #1015, opened 2018-08-06
Fix compiler warnings

mkbel:


3d-gussner:
The warnings in 1.6.9 are gone. And with this one even got rid of some in 1.8.5 ... there are still some :-;
Thanks @mkbel great job cleaning up the code

PR #1023, opened 2018-08-08
Mk3 revision for First Layer Calibration Filament Choice (update PR 816)

RetireeJay46:
A second attempt to have the First Layer Calibration Menu allow for different filaments.  (Not being expert in GitHub, I couldn't bring my previous Pull Request into alignment).


RetireeJay46:
Why won't the developers either merge this request or comment on why not???

geobruce:
I would love to see a feature like this get implemented

Crunch69:
this should be implemented for at the least petg and abs

.gitignore
```diff
@@ -1,6 +1,8 @@
 .settings
 .project
 .cproject
+notes.txt
```
everett1992, line 4: You shouldn't add personal files to a projects .gitignore, these belong in your [personal .gitignore](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c)

You can also add and commit specific files, you don't need a .gitignore if you only commit the files you change

```
git add Firmware/ultralcd.cpp
git commit

# instead of:
git add .
git commit
```

Firmware/ultralcd.cpp
```diff
@@ -1070,11 +1070,13 @@ void lcd_commands()
 		}
 		if (lcd_commands_step == 10 && !blocks_queued() && cmd_buffer_empty())
 		{
-			enquecommand_P(PSTR("M107"));
-			enquecommand_P(PSTR("M104 S" STRINGIFY(PLA_PREHEAT_HOTEND_TEMP)));
-			enquecommand_P(PSTR("M140 S" STRINGIFY(PLA_PREHEAT_HPB_TEMP)));
-			enquecommand_P(PSTR("M190 S" STRINGIFY(PLA_PREHEAT_HPB_TEMP)));
-			enquecommand_P(PSTR("M109 S" STRINGIFY(PLA_PREHEAT_HOTEND_TEMP)));
+			if(degHotend(0)<200){  //If not heated, presume PLA is OK
```
everett1992, line 1073: You are proposing that users set the target temperature via the LCD's main 'preheat' menu. IMO that's okay, but I think it would be better to display a material / temperature selection menu in the first layer calibration program.


I also think that your implementation of your workflow is flawed. Here you check if the _current_ hot end temperature is above 200.
 - what if the users material prints below 200 degrees?
 - what if the extruder is above 200 degrees when the temp is checked but it is cooling and has no set temperature?
 - what if the material requires significantly greater than 200 degrees?

You should check that `degTargetHotend(active_extruder) > 0`, then queue a `M109` with no arguments which will wait to reach target temp. You might need a special case when AutoTemp is enabled.

everett1992:
I just ran XYZ calibration and I don't have any PLA on hand to re-run first layer calibration. I exclusively run PETG. Prusa should merge this change, or another implementation of the same feature.

At-least document a gcode script with custom temperatures. A script could handle material temperatures and nozzle diameters.

.gitignore
RetireeJay46: Thank you everett.  I am still a total newbie at Git.

Firmware/ultralcd.cpp
RetireeJay46: Excellent points.  I'm glad someone is looking at this closely (at last).  
I'm really not a C++ programmer, and struggled mightily to find the appropriate variables.  Your suggestion of "degTargetHotend(active_extruder)" is perfect.
To be honest, I'm hoping someone better qualified than me can take this ramshackle code and make it into something much nicer.  My code is more of a "proof of concept".

everett1992:
I'm not familiar with this code base, or c++ really. I think I could find
time to prioritize this in the next half year, as I finish other projects.

In the mean time, I've noticed that you can print without calibrating the
first layer. It is feasible to slice a single layer object and use live
adjust z to set the first layer offset.

Firmware/ultralcd.cpp
everett1992: I'm not familiar with this code base, or c++ really. I think I could find
time to prioritize this in the next half year, as I finish other projects. This feels like an easy change for someone that _is_ familiar with the codebase.

In the mean time, I've noticed that you can print without calibrating the
first layer. It is feasible to slice a single layer object and use live
adjust z to set the first layer offset.

3d-gussner:
@RetireeJay46 This have been implemented in 3.8.1,  can you please review your PR and if it is not needed anymore please close it.

3d-gussner:
I think this is obsolete and closing it.

PR #1024, opened 2018-08-09
Make MMUv2 compatible with MK3 Einsy and MK2.5 miniRAMbo boards.

3d-gussner:
The defined MMU_RST_PIN pin 76 is only available on the P3 connector on the Einsy board BUT not on the miniRAMbo.
As the MMUv2 should be supported on the MK3 and MK2.5, which has a miniRAMbo controller, you gonna need to use another pin on the miniRAMbo 1.0 and 1.3 for MMU_RST_PIN.
Like 'X-max' is used for Noctua tacho signal I propose to use 'z-max' for MMU_RST_PIN. Prusa also gonna need a split cable for the miniRAMBO MMUv2 as for the MK2.5 Noctuca and PINDAv2.

### DIFF
--- a/Firmware/mmu.cpp
+++ b/Firmware/mmu.cpp
@@ -21,7 +21,6 @@ extern char choose_extruder_menu();
 #define MMU_TIMEOUT 10
 
 #define MMU_HWRESET
-#define MMU_RST_PIN 76
 
 
 bool mmu_enabled = false;
--- a/Firmware/pins_Einsy_1_0.h
+++ b/Firmware/pins_Einsy_1_0.h
@@ -143,7 +143,7 @@
 #define WRITE_LOGIC_ANALYZER_CH4(value) if (value) PORTK |= (1 << 0); else PORTK &= ~(1 << 0) // PK0
 #define LOGIC_ANALYZER_CH5		16				// PH0 (RXD2)
 #define LOGIC_ANALYZER_CH6		17				// PH1 (TXD2)
-#ifndef SNMMP
+#ifndef SNMM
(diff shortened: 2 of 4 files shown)


3d-gussner:
Hi, @PavelSindler you addressed the mmu signal cable issue with PIN 9 on P3 connector by deactivating it on the MK25 miniRAMbo https://github.com/PavelSindler/Prusa-Firmware/commit/bc247ca02c0c4557be8afeeb1b21c55290c9d14e. Why don't you use as described above another y-cable for that function?

PavelSindler:
Hello @3d-gussner, pin 9 on P3 connector (which is MMU2 RST) is connected directly to MiniRambo board Reset. So pressing reset button on MK2.5 printer resets MMU2 and Printer directly (HW reset). To obtain correcrt initialization after powering up the MK2.5+MMU2 printer, there is one more software reset (sending X0 code to MMU2) in printer setup function. No HW changes are needed.
On Einsy board, the difference is that MMU2 HW reset is not connected to printer HW reset, but it is connected to pin 76, so when printer is reseted, MMU2 is reseted indirectly by pin 76 in printer setup function (so MMU2 reset happens little bit later).

3d-gussner:
@PavelSindler Thanks for answering. saw your last implementation in 3_4_2 branch and mm-control-01 to fix the issue that the mmu didn't initialize due missing reset.
I was thinking of the y-signal cable because if the MM-control board stalls, crashes or doesn't answer on time (timeout) via the serial line the user has to manually reset the MMU, as the software reset will not work.
If there is a failure during change which needs a MMU reset it will also reset the MK25 printer and so the hole print.

PR #1042, opened 2018-08-15
Revert "menu data shared memory re-design"

mkbel:
This fixes: Unable to edit values in menu/calibration/Bed level correction.


XPila:
@mkbel solved different way, closing

PR #1065, opened 2018-08-24
Mmu v2: continue loading command

PavelSindler:


mkbel (changes requested):
Please update version check of MMU unit firmware, so it is not possible to use firmware with this change with old incompatible MMU firmware.

PR #1113, opened 2018-09-07
time remaining fix (PFW-554)

PavelSindler:


mkbel (changes requested):
Did you test it?
For the first sight it looks like expression
100ul * print_t
would overflow, if time remaining is larger than 10.93 hours. Before change it would overflow if time remaining is larger than 5.46 hours.

PR #1158, opened 2018-09-16
Fix units reported in excessive bed leveling correction error

vsergeev:
This error was reporting an excessive bed leveling correction value as microns instead of millimeters.

The `correction` variable is in units of microns, before it's scaled down by a factor of `0.001f` to the variable `offset`, which then has units of millimeters. On a excessive correction error, `offset` is reported as having units of microns instead of millimeters.


vsergeev:
Bump

vsergeev:
Fixed in recent merges

PR #1192, opened 2018-09-24
Simplify eeprom m500

mkbel:
Saves 1828 B flash and 68 B RAM. Saved RAM is suspicious - now explained in next comment. New code is less error prone, as it is not needed to remove/add variable in more places exactly in the same order. Also it is less probable to change some variable size or array size without noticing damage to EEPROM layout.


mkbel:
80 bytes of RAM is saved by removing snipped

```
    float tmp1[]=DEFAULT_AXIS_STEPS_PER_UNIT;
    float tmp2[]=DEFAULT_MAX_FEEDRATE;
    long tmp3[]=DEFAULT_MAX_ACCELERATION;
    float tmp4[]=DEFAULT_MAX_FEEDRATE_SILENT;
    long tmp5[]=DEFAULT_MAX_ACCELERATION_SILENT;
    for (short i=0;i<4;i++) 
    {
        axis_steps_per_unit[i]=tmp1[i];  
        max_feedrate_normal[i]=tmp2[i];  
        max_acceleration_units_per_sq_second_normal[i]=tmp3[i];
        max_feedrate_silent[i]=tmp4[i];  
        max_acceleration_units_per_sq_second_silent[i]=tmp5[i];
    }
```
from void Config_ResetDefault()

As compiler allocated such big arrays in global variables instead of on stack.

PR #1193, opened 2018-09-24
Enable fast PWM and higher scaling for print fan. Closes #460

### LINKED ISSUE #460: Hi Pitch noice coming from print fan
When the print cooling fan come on there is a high pitch noise. I believe this is cause by the high frequency PWM signal going to the FAN.

trevjonez:
Co-authored-by: surrealseven <surrealseven@users.noreply.github.com>
Co-authored-by: lenisko <lenisko@users.noreply.github.com>

### DIFF
--- a/Firmware/Configuration.h
+++ b/Firmware/Configuration.h
@@ -467,7 +467,7 @@ your extruder heater takes 2 minutes to hit the target on heating.
 
 
 // Increase the FAN pwm frequency. Removes the PWM noise but increases heating in the FET/Arduino
-//#define FAST_PWM_FAN
+#define FAST_PWM_FAN
 
 // Temperature status LEDs that display the hotend and bet temperature.
 // If all hotends and bed temperature and temperature setpoint are < 54C then the BLUE led is on.
--- a/Firmware/temperature.cpp
+++ b/Firmware/temperature.cpp
@@ -1063,7 +1063,7 @@ void tp_init()
   #if defined(FAN_PIN) && (FAN_PIN > -1) 
     SET_OUTPUT(FAN_PIN);
     #ifdef FAST_PWM_FAN
-    setPwmFrequency(FAN_PIN, 1); // No prescaling. Pwm frequency = F_CPU/256/8
+    setPwmFrequency(FAN_PIN, 5); // Prescale to 1024. Pwm frequency = F_CPU/256/8
     #endif
     #ifdef FAN_SOFT_PWM
     soft_pwm_fan = fanSpeedSoftPwm / 2;


Mike4U:
Have any tests been performed to verify fan speed before and after mod?

trevjonez:
nothing scientific. do you know if there is a gcode I can run to print the current fan RPM? m106 and m107 dont return currents and I don't see anything else on the wiki.

i'll dig into what self test checks and I'll post back. worst case I'll manually verify until we can get a plan together.

trevjonez:
Ok so follow up on this fan speeds. I just did some manual observation and took some notes on it.

| ~ % | S Value | LCD Rpm | Human Observation|
| ------------ | ------------ | ------------- | -------------|
| 10-12% | 25-30 | 0 | 60 - 120 RPM, sine wave like speed fluxuations |
| 29% | 74 | 0 |  |
| 30% | 75 | 960-1080 | first value to produce a number on the LCD |
| 40% | 102 | 1500-1740 |  |
| 50% | 127 | 2160-2220 |  |
| 60% | 153 | 2820-2880 |  |
| 70% | 179 | 3540-3600 |  |
| 80% | 204 | 3840-3960 |  |
| 90% | 230 | 3660-3960 |  |
| 100% | 255 | 2820-3100 | Audibly you can tell this is spinning fastest |

Based on what I can see in the code base it looks like the gcode `PRUSA FAN` does print out the RPM of the fans. which is the value that is printed on the LCD.

image
[interactive graph](https://docs.google.com/spreadsheets/d/e/2PACX-1vQ-ugOezcHFXJwVUCtdaM-VpysJVSbs1Lu8mFPhXLAo1D8YOCDm6PC0XTvvKD-wwLlyuSu0_E_dBWQc/pubchart?oid=617730818&format=interactive)

trevjonez:
anyone know why it would drop at the higher end of the chart? looks like fan_speed[1] is RP-Second * 60 to get the RPM values listed here. is the third pin on the fan a type of tach or a pwm input?

metacollin:
> is the third pin on the fan a type of tach or a pwm input?

Only 4 wire fans have PWM input - on 3 wire fans, the third (yellow) wire is always simply a tachometer pulse output.  Speed control of 3 wire fans can only be realized by either varying the supply voltage (directly, using a dropping resistor, using diodes, etc.) or by varying PWM duty cycle.  

> anyone know why it would drop at the higher end of the chart?

**TLDR:  You can either increase the PWM frequency but lose the ability to measure the fan's RPM, or you can measure the fan's RPM at the price of not being able to make the PWM go any faster.  It's one or the other, there is no way to have our cake and eat it too, unfortunately. 
Yes.  The problem isn't really the drop at the end though, the problem is all of the tachometer readings are completely incorrect, except for the reading taken at 100% duty cycle.**

The stock fan PWM frequency is limited to 325Hz (resulting in that annoying whine that we all know and hate with a fiery passion) because this is the fastest PWM frequency the fans can be driven at and still correctly measure the RPM using the tachometer line.  

The problem is 3 wire fans are not designed nor intended to be speed controlled, especially via PWM.  Only 4 wire fans, in a perfect world, should ever actually be PWM speed controlled.  A 3 wire fan's tachometer circuitry can only produce a tach pulse if it is actually powered, and PWM control is constantly turning off power to the entire fan, tachometer included.  This causes the tach pulses to get 'chopped' due to the PWM cutting power, and this of course will vary with duty cycle.  Here, this might help visualize what I mean.  The bottom graph is what an ideal tach pulse train would look like with the top PWM control applied to the fan:


Chopping

Additionally, the PWM signal is effectively injected into the tachometer pin for the same reasons this chopping occurs, causing pulse triggers to happen for each PWM cycle in addition to the actual 'true' tachometer pulses.  The Prusa firmware keeps track of where in a given PWM cycle it is when a pulse is detected, and if one is detected when the PWM cycle is switching, it knows that the pulse is not a real one and discards it.  

Increasing the PWM frequency would result in both tachometer pulse chopping (which makes the measured RPM be a value largely unrelated to the actual RPM of the fan 😄 ) and erroneous pulses that won't be accounted for.  

There is an entire interrupt (INT7) dedicated to dealing with this in Marlin_main.cpp:

```
// The fan interrupt is triggered at maximum 325Hz (may be a bit more due to component tollerances),
// and it takes 4.24 us to process (the interrupt invocation overhead not taken into account).
ISR(INT7_vect) {

	//measuring speed now works for fanSpeed > 18 (approximately), which is sufficient because MIN_PRINT_FAN_SPEED is higher

	if (fanSpeed < MIN_PRINT_FAN_SPEED) return;
	if ((1 << 6) & EICRB) { //interrupt was triggered by rising edge
		t_fan_rising_edge = millis_nc();
	}
	else { //interrupt was triggered by falling edge
		if ((millis_nc() - t_fan_rising_edge) >= FAN_PULSE_WIDTH_LIMIT) {//this pulse was from sensor and not from pwm
			fan_edge_counter[1] += 2; //we are currently counting all edges so lets count two edges for one pulse
		}
	}	
  EIMSK &= ~(1 << 7);
	EICRB ^= (1 << 6); //change edge
  EIMSK |= (1 << 7);
  
}
```

Simply increasing the PWM frequency without also increasing the CPU time devoted to this interrupt will render any tachometer readings incorrect and useless.  

There is a dirty trick called pulse stretching that lets you avoid all this while running a 3-wire fan at very high (25kHz+) PWM frequencies but still get accurate pulse readings from the tachometer.  Pulse stretching works by periodically holding the duty cycle at 100% for just long enough to pick up two tachometer pulses and get the timing correct, then resuming high frequency PWM control.  

And if the ATMega2560 had the additional timer peripherals, clock speed, and timer interrupts available to implement either clock stretching or simply brute force a higher frequency PWM pulse removal ISR, I'm sure Prusa's Firmware developers would leap at the chance to eliminate that soul-crushing 325kHz fan whine that makes us all die a little more inside every additional second that we are forced to hear it.

But the needed timer peripherals simply don't exist, and the CPU has no processing resources left to devote to any of this anyway.  

 I applied essentially this same change myself to the firmware on my own printer.  I've never wanted to know the RPM of the print fan, I've never needed to know, and in my opinion, the tachometer feature on the printfan has been useful to approximately zero people ever in the entire operational history of the MK3.  So I say, please just sacrifice the tach feature for the print fan if we can lose that horrible whine.  The hotend fan, we need to be able to detect if it stops spinning for safety reasons.   But if the print fan fails, it means any bridges on your print might get a little crappier, but it isn't going to kill your cat and burn down your house.  

I totally support this pull request - I think it would be an excellent change, not that my endorsement really means anything.  And II mean no disrespect to @trevjonez, @surrealseven, and @lenisko...

... but I don't really see a way this pull request could be accepted, at least as is.  It removes an annoyance at the cost of reducing the fan RPM values to nonsense values that have little to do with what the fan's actual RPM is.  Not through any fault with your code, it's just the crappy reality of the situation :(.

At the minimum, you would need to also remove the tachometer functionality (probably using some #ifdefs and a descriptive #define, like `#define FAN_PWM_THAT_HASNT_CAUSED_ANY_SUICIDES_YET`) at the same time it enabled the higher pwm.  And then convince the Prusa devs that its a good idea :). 

Ok, maybe a more adult #define.  But you know what I mean.  

I really hate that whine.

maximlevitsky:
@metacollin Great comment!
Since its already possible to ignore pwm from fans, fast pwm mode can be made conditional on that.
Plus, it can be enabled only on print fan, for which failure isn't critical, as it is for cooling fan.

hoppke:
@maximlevitsky a compromise might be possible too - to only read the tach when pwm is at 100% duty? (assuming that 100% duty is equivalent to constant for the purpose of reading the tach).
Would still let you verify that the fan is there if you needed to (self-test? before each print?), maybe even that it gets the expected RPMs.

trevjonez:
let me take a look at these suggestions. I like the idea of fan monitoring setting being a toggle for this. although it will mean the config won't be via ifdefs so it will probably touch a lot of places.

I also want to do more comparisons of rpm. I get a strong feeling that the part fan spins much faster while flashing firmware which makes me think it is perhaps not running at 100%. need to test though as the steppers are silent during the flash so it could just be in my head. might be time to purchase an optical tachometer so I can get real readings.

r-a-i:
Hi, I recently received my Prusa Mk3 and this noise is killing me a little every day. 

Can we just get a 4-pin fan and solve everything? 
This is a first find (after 10 minute search) https://www.digikey.com/product-detail/en/sanyo-denki-america-inc/9BMB24P2G01/1688-1381-ND/6192097

trevjonez:
@r-a-i that thing is way too big given the stock cooler and mount points. certainly anyone could design around it though. 

I ended up ditching prusa FW completely for klipper which doesn't exhibit this issue. although I think a 24v pwm controlled fan would be the best solution there may not be any in the 50x15 blower form factor.

Edit:  yes I lose some of the prusa FW features, but I also gain other options. one of them isn't failing a test or just getting bad data because a non safety critical fan rpm signal isn't where expected. I feel like it comes down the prusa fan being designed to run 100% and that is sometimes ok, but outside of that use case their implementation feels like hell on earth with the pwm whine.

r-a-i:
@trevjonez thanks for sharing about Klipper; looks very interesting and probably the future of these systems. 

I just saw an official Prusa RC for firmware 3.5.2-RC1.  A quick review of the changes going into this RC looks like they are doing the "pulse stretching" trick that @metacollin described above. 

I'll give the new FW a shot and report if this fixes the fan noise over the next couple of days.

r-a-i:
I had implemented a workaround to the hum of death by configuring Slic3r to always set the print fan at 100%, but I still wanted to see if I could revert to the original setting. 

I ran a quick experiment to compare the noise level between firmware 3.5.1-1778 and 3.5.2-RC1-1993 and qualitatively I can say the new version is better.  Instead of the constant annoying hum at ~350 Hz, there is now a low-level crackling coming from the fan which may be something I can accept.  You can tell they are trying to do some form of pulse stretching to read the Tach signal as the fan sounds like it fully speeds up once every ~3 seconds when presumably they are setting the PWM at 100% to read a clean tach.

Trying to do a more quantitative comparison I used an app (Decibel X) to measure the noise level by setting my phone close to the printer and sweeping the fan speed setpoints. Following is a table of the results. 

Fan Speed | Noise for 3.5.1_1778 | Noise 3.5.2_RC1_1993 | Notes
----| ------------- |------------- |------------- 
0 | 40.2 dB |  | Base background noise (distant fridge fan)
64 | 43.0 dB | 44.8 dB | 
76 | 45.2 dB | 44.7 dB | This was the setpoint for PETG that drove me crazy
128 | 45.7 dB | 45.3 dB | 
192 | 48.0 dB | 48.3 dB | 
255 | 47.8 dB | 51.0 dB | 

Although the data does not show a substantial reduction of noise, perhaps this was due to the imperfect setup and microphone position between trials. Specifically, I was surprised by the sample with full fan speed (255) being ~3 dB louder in the latter case, because it should have been the same. This makes me think that all the samples for firmware 3.5.2 may have a higher offset because the microphone was in a different position. 

Overall, the humming noise of dead seems to be gone, but I'm not convinced this was the best way to resolve it as there is still some crackling sound that may drive us crazy and the way they are sampling the RPMs seems a complete hack.  For the time being, I will keep the 3.5.2 firmware and see if I can live with it. 

I've been thinking of a way of simply de-convolving the tach signal as samples by the PWM raising edges but I doubt that I'll have the time and energy to try implementing this. Perhaps @metacollin has some insights into similar attempts to read a tach signal convolved by the PWM pulses that we have on this 3-pin fan.

maximlevitsky:
@r-a-i I hate to be the guy whom they throw out of the window, but why do we need a reliable tach readout from the print fan? Just sense that it is spinning and thats all. I bet that if the fan stop spinning, this will be detected.
Plus an option to not read it too, since print fan (as opposed to cooling fan) isn't a critical part to stop the print if fails anyway.

r-a-i:
@maximlevitsky that is a really good point.  Coming from a background of implementing tight PID loops for servo motors, it feels very unsatisfying to not do that here. But in this case, having precise RPM control on the fan will probably not affect the print considerably so it is not a worthy pursuit.  

Now that you have highlighted that, the current implementation, which every few seconds runs a "pulse stretching", is a hacky overkill.

r-a-i:
I've been running the 3.5.2-RC1-1993 firmware for a few days now and printing PETG with the original printer profiles and have forgotten about this issue.  Today I realized, hey, the fan has not been driving me crazy. I've also seen the RC1 has been officially released (virtually identically).

I think this issue should now be closed.

PR #1227, opened 2018-10-04
Fix (some) compiler warnings

noradtux:
I noticed a lot of compiler warnings when using -Wall on avr-gcc 5.4.0. I removed some unused variables, added initialization to others and adjusted signedness where comparisons between signed and unsigned was used. My Mk3 seems to work fine with these adjustments. I hope you can use some of this.
There are still a lot of unused functions and some unused variables left, but I guess those are used to create tiny delays while executing.


3d-gussner:
@noradtux Can you please review this PR as lot has changed? If it not needed anymore please close it.

Issue #1378, opened 2018-12-03
[3.5.0] M601 results in an infinite pause/resume loop


TheHammockGuy:
A bug was introduced in the 3.5.0 firmware release related to the M601 gcode command when printing from an SD card.  When this command is executed it is taking the print's current location on the sd card and rolling backwards equal to the length of data in the command buffer and planner.  This results in a new location to be used upon resume.  This new location is immediately before the M601 command.  This results in the long pause operation being called again after resuming from a long pause.

Also, M601 now implicitly pauses as part of the command, where before you would need to use an M1 command to wait for user input before using the M602 resume command.

The bug appears to come from this commit:
https://github.com/prusa3d/Prusa-Firmware/commit/1585dc66b0f96c63dd65b57644326180761ef3c5

The issue resides in Marilin_main.cpp in the function "stop_and_save_print_to_ram"

I have made a workaround, but this should really have a more proper fix:
https://github.com/TheHammockGuy/Prusa-Firmware/commit/867efed04353a11877742bba19288d3d5bfad35a

Current Behavior:
M601 Turns off nozzle heat, moves to a holding location, and triggers an implicit pause.
User selects Resume from LCD menu and it re-executes the above M601 command resulting in a loop

Expected(previous) behavior:
M601 Turns off nozzle heat, moves to a holding location, and then continues executing gcode to allow for
M1 to wait for user input
M602 to reposition back to original location, reheat, and continue printing

Behavior with my workaround:
M601  Turns off nozzle heat, moves to a holding location, triggers an implicit pause and sets a flag to ignore further M601 commands until after M602 is called
User selects Resume from the LCD menu and it skips the above M601 due to the flag setting
M602 Resets the flag allowing M601 to be executed again in the future

TheHammockGuy:
I have also created a pull request in case you want to use my workaround until a better fix can be scheduled:

Pull request https://github.com/prusa3d/Prusa-Firmware/pull/1379

PavelSindler:
Thank you for reporting this bug.

thenickdude:
~Looks like this got fixed in 3.5.1 by 16e5fb50ddc72a294c7eb173622d0ec2961210cc~

kubiani:
> Looks like this got fixed in 3.5.1 by [16e5fb5](https://github.com/prusa3d/Prusa-Firmware/commit/16e5fb50ddc72a294c7eb173622d0ec2961210cc)

[...]

PR #1379, opened 2018-12-03
Work around for M601 bug introduced in 3.5.0 firmware release

TheHammockGuy:
The M601(long pause) command will result in an infinite loop, because it resets the sd card location to before the M601 command.  This causes M601 to be executed again immediately after resuming.

This is more of a work around to skip any M601 commands until the corresponding M602(resume from long pause) has been triggered.  The proper fix would be to correct the behavior of M601 so that is doesn't execute itself again after a resume, but if that isn't something that will make it into the next release this could be a work around for anyone who relies on the M601 command.

See issue https://github.com/prusa3d/Prusa-Firmware/issues/1378

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -394,6 +394,8 @@ static bool saved_extruder_relative_mode = false;
 static int saved_fanSpeed = 0; //!< Print fan speed
 //! @}
 
+bool allow_M601 = true;
+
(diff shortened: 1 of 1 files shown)


3d-gussner:
@TheHammockGuy Thanks. Closing PR as M601/602 changed. Please check MK3 branch and feel free to open a new PR.

PR #1412, opened 2018-12-18
Repeat last MMU command after timeout

XPila:


Firmware/Marlin_main.cpp
```diff
@@ -2796,7 +2796,12 @@ bool gcode_M45(bool onlyZ, int8_t verbosity_level)
 		// Move the print head close to the bed.
 		current_position[Z_AXIS] = MESH_HOME_Z_SEARCH;
 
-		bool endstops_enabled  = enable_endstops(true);
+		bool endstops_enabled  = enable_endstops(false);
+        current_position[Z_AXIS] -= 1; //move 1mm down with disabled endstop
```
PavelSindler, line 2797: This movement should be performed before setting current_position[Z_AXIS] = MESH_HOME_Z_SEARCH; Currently possible crash is not detected during movement from Zmax to MESH_HOME_Z_SEARCH-1.

PR #1421, opened 2018-12-21
Fix warning during language build

3d-gussner:
Deleted an é as it generates a warning during lang-build.sh script.


3d-gussner:
No reaction from Prusa so i close it

PR #1437, opened 2019-01-01
set array size correctly to 4

craiglink:
the code may access 4 elements of the array, so allocated arrays of length 4 to avoid potential invalid memory access


craiglink:
Is there anyway to help getting simple C/C++ bug fixes into the code base?  @PavelSindler  @XPila

jurriaan:
cc @mkbel @MRprusa3d

mkbel:
Thank you for your contribution. Even better fix would be to change parse_version() second parameter to exact size array packed in struct so such error would be found by compiler next time.

PR #1438, opened 2019-01-01
fix scoping of filament variable

craiglink:
use the outer filament variable in all conditional blocks


craiglink:
underlying code was fixed elsehwere

PR #1439, opened 2019-01-01
use global static extruder_autofan_last_check var

craiglink:
use the existing global static variable instead of a locally scoped variable


craiglink:
no longer needed

PR #1444, opened 2019-01-05
use fastio macros over slow Arduino gpio functions

craiglink:


xchwarze:
Another improvement that is lost in time...

wavexx:
LA1.5 does make changes in stepper.cpp obsolete.
I could review usage in fsensor.cpp as I did some work on it.
A lot has been improved in the lcd handling too.

We're getting there

3d-gussner:
@gudnimg Please have a look and if it still something to consider.

gudnimg:
Fastio is used everywhere now. I think we changed a lot of it in 3.12

Closing this PR for now because Fastio is used everywhere in the current release

PR #1445, opened 2019-01-05
optimize lcd using fastio macros

craiglink:
- free up cpu cycles using the fastio macros vs the slower Arduino digitalRead/Write functions.
- use the _delay_us() function instead of delayMicroseconds() for more accurate delays.
- unroll the LCD pin writing loop to write directly to pins allow use of fastio
- ifdef 4bit / 8bit mode code paths

### DIFF
--- a/Firmware/lcd.cpp
+++ b/Firmware/lcd.cpp
@@ -58,11 +58,11 @@
 
 FILE _lcdout; // = {0}; Global variable is always zero initialized, no need to explicitly state that.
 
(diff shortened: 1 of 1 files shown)


3d-gussner:
@gudnimg Please check if still valid PR.

gudnimg:
@3d-gussner I think we should close this one as well. Though I'm not sure if there is benefit in changing `delayMicroseconds(100);` into `_delay_us(100);`

But all the other changes for fastio are already in 3.13.2

PR #1446, opened 2019-01-05
remove unused swspi

craiglink:
these files aren't used for any builds


MRprusa3d:
... software SPI library is needed for some future usage ...

craiglink:
The code doesn’t even compile because of some syntax errors. If it’s going to stick around, those should be fixed

MRprusa3d:
... we have now corrected (yet unpublished) version - see "PAT9130" PR for example ...

DRracer:
@craiglink would you mind closing this PR?

PR #1448, opened 2019-01-08
don't recursively manage_inactivity by calling delay_keep_alive from …

craiglink:
…within itself


3d-gussner:
@gudnimg Please check if still valid PR.

gudnimg:
Sound functions don't call manage_inactivity as far as I can tell by looking at our current release. And the slow `digitalWrite()` function has been replaced with `WRITE()`

I am closing this PR as resolved.

Issue #1454, opened 2019-01-15
Revamp of Pinda Temp Calibration


CCS86:
After watching the 3.5.x temp calibration run, I believe it is introducing error into the resulting calibration data.

It is probing the same spot on the bed to see how the pinda changes its triggering height, as the sensor changes temp. This method is based entirely on the assumption that the bed surface at this point remains in the exact same position. 

The problem is that the calibration routine is also changing the bed temp, and because of thermal expansion of all the bed/platform components, that same spot physically changes height. This gets erroneously rolled into the probe calibration data.

My thought for a more ideal temp calibration:

- Raise print head to near max height and turn on bed to the highest temp needed for pinda calibration

- Allow a long enough dwell so that physically all bed components stabilize. Pinda is still cool. The amount of dwell time could easily be established by attaching a dial indicator to the print head, indicating on the intended spot for probe calibration (I think bed center is better than the front corner used), then turn on the bed to the max temp and wait until indicator movement stops (time lapse would make this even easier).

- Now drop the head and probe the bed, creating a point set of height and probe temp.

- Wait close to the bed, as the pinda heats up, and continue probing when the pinda reaches specified probe temps.

ghost:
This is really great idea!

Tinker999:
I agree a simple but elegant approach which could eliminate the whole need for pre heating the Pinda. It needs to be tested but it looks easier and potentially more accurate than other methods use now. I know there is a wiki that has some other approaches. But I can’t find it. It would be great if we could get some of the  developers that came up with the other approaches to comment.

stahlfabrik:
The other methods we have described here: https://github.com/PrusaOwners/prusaowners/wiki

CCS86:
I dug around a bit in the firmware and found the temp calibration routine. Unfortunately, between my schedule and general faded comprehension of C++, I will not be the man for the job, hahaha

I think this is a relatively minor rewrite though. It looks like the probing location can be easy moved to bed center. Additional code added to move near the Z+ limit and dwell while the bed heats. Some code removed for the changes in bed temp.

dougisfunny:
> 
> 
> I dug around a bit in the firmware and found the temp calibration routine. Unfortunately, between my schedule and general faded comprehension of C++, I will not be the man for the job, hahaha
> 
[...]

PR #1458, opened 2019-01-16
Added 7x7 for FW 3.1.0 changes to the original files

aNiclas:
Based the changes done to 3.5.1 FW by @stahlfabrik on https://github.com/prusa3d/Prusa-Firmware/compare/MK3...stahlfabrik:HD-MBL

### DIFF
--- a/Firmware/Configuration.h
+++ b/Firmware/Configuration.h
@@ -43,7 +43,7 @@
 #define EEPROM_BED_CORRECTION_REAR  (EEPROM_BED_CORRECTION_FRONT-1)
 #define EEPROM_TOSHIBA_FLASH_AIR_COMPATIBLITY (EEPROM_BED_CORRECTION_REAR-1)
 #define EEPROM_PRINT_FLAG (EEPROM_TOSHIBA_FLASH_AIR_COMPATIBLITY-1)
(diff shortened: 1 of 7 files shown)


mionut:
Saving/restoring of all 49 values to eeprom on power failure must be implemented, and tested to see how much time it takes, if it si enough until the power runs out.

PavelSindler:
I would like to thank all people who worked on 7x7 mesh bed leveling for your effort. We are not ignoring you and we would like to implement this feature. We started testing some versions last week, but we didn't have convincing proof that it works the way we want. so now we are doing another tests with @stahlfabrik, @mionut and our versions with accurate micrometer mounted on extruder (measuring distance between extruder and steel sheet surface) and hopefully we will clearly see how MBL is improved by higher number of points.

stahlfabrik:
Hello @pavelsindler

Thank you very much! Feedback is greatly appreciated! I am looking forward to your implementation. Getting a consistent first layer across the whole bed is imho really quite an issue for PrusaFirmware in the moment

Tinker999:
Thank you, I’m really looking forward to this change.

 I was  wondering if you would also consider looking at implementing. A better Pinda calibration which I think would fix the other part of the first layer issue. The approach below appears to be a simple approach which I think will be very accurate. While not implement it looks like it has tremendous potential.  

https://github.com/prusa3d/Prusa-Firmware/issues/1454

And again Thank Thank Thank

3d-gussner:
@aNiclas Thanks. Closing PR as there have been changes made. Please review MK3 branch and feel free to open a new PR.

PR #1501, opened 2019-02-04
Unload filament after MMU load failed so MMU unit can be controlled b…

mkbel:
…y its buttons.


Firmware/mmu.cpp
```diff
@@ -1353,49 +1353,61 @@ void mmu_eject_filament(uint8_t filament, bool recover)
 	}
 }
 
-void mmu_continue_loading() 
+void load_more()
```
PavelSindler, line 1356: use static function instead?

PR #1525, opened 2019-02-11
Improve mmu load failed 2

mkbel:
There is no need to update MMU firmware version dependence, as printer works also without MMU support. But of course print can not be resumed by MMU button.


mkbel:
There is no need to update MMU firmware version dependence, as printer works also without MMU support. But of course print can not be resumed by MMU button.

PR #1549, opened 2019-02-18
7x7 mesh bed leveling with power failure save + pinda temperature filter

mionut:
Hello, i use 7x7 mesh bed leveling for about 8 months. It is equivalent with old implementation if no parameter is given to G80.

Usage: `G80 N7[3], R[1..10]` - N is number of probing points, R is number of retries for probing function.
Default values for N is 3, for R also 3.

I tested power failure save for all 49 points, witch I added recently, and it works, for 220v mains voltage.

I also did some modifications to some functions not compiled for MK3 and MK2.5 (only MK2, GCODE M48) because i eliminated `bed_ref_points` constant and replaced it with  `BED_X0, BED_Y0, BED_Xn, BED_Yn`. I removed  `mesh_bed_leveling::get_meas_xy` because is not used.

Pinda temperature is averaged over 4 consecutive reads to make a more stable temperature.


GWdd:
I also have been using the 3.5.1xx version of this 7x7 code for months without issue on my Mk3; and can also report no difficulties with the 7x7 level feature.  It does a better job correcting the bed than a tedious manual level plus 3x3 auto level with bed correction using the four points.

Firmware/mesh_bed_calibration.cpp
```diff
@@ -2428,8 +2401,11 @@ BedSkewOffsetDetectionResultType find_bed_offset_and_skew(int8_t verbosity_level
 					refresh_cmd_timeout();
 					// Go to the measurement point.
 					// Use the coorrected coordinate, which is a result of find_bed_offset_and_skew().
-					current_position[X_AXIS] = pgm_read_float(bed_ref_points + mesh_point * 2);
-					current_position[Y_AXIS] = pgm_read_float(bed_ref_points + mesh_point * 2 + 1);
+					uint8_t ix = mesh_point % MESH_MEAS_NUM_X_POINTS; // from 0 to MESH_NUM_X_POINTS - 1
+					uint8_t iy = mesh_point / MESH_MEAS_NUM_X_POINTS;
+					if (iy & 1) ix = (MESH_MEAS_NUM_X_POINTS - 1) - ix;
+					current_position[X_AXIS] = BED_X(ix, MESH_MEAS_NUM_X_POINTS);
+					current_position[Y_AXIS] = BED_Y(iy, MESH_MEAS_NUM_X_POINTS);
```
ddaniinfo: Y axis with MESH_MEAS_NUM_**X**_POINTS constant? Deliberate or copy paste?

mionut: i will modify, copy paste

PavelSindler:
Thank you @mionut and all for your work and patience. I am merging this PR. It will in next release together with other improvements and also we will add some menu item for switching between different mesh bed leveling precision levels.

mionut:
Thank you. You should also change the definition of EEPROM_UVLO_MESH_BED_LEVELING_FULL to be after the last one, without -1000.

PavelSindler:
yes, this has already happened when merging PR https://github.com/prusa3d/Prusa-Firmware/pull/1578

schneems:
What would it take to make 7x7 the default instead of 3x3? I know many users recommend changing the default setting.

PR #1567, opened 2019-02-21
7x7 gcode MBL workaround by @minout

lowfat32:


mionut:
Hello, i removed the part with lround((z/+mbl.get_z(x, y)/)*cs.axis_steps_per_unit[Z_AXIS], you can take my latest modifications from branch 7MBL, or see the #1549 .
Pinda temperature averaging i think is essential for increasing precision of measuring. Also the latest version doesn't have division, only multiplication and shifting.

PR #1570, opened 2019-02-22
mmu stealth mode

PavelSindler:


kdomanski:
Why replace variable checking with preprocessor logic? Why the DEBUG mode?

PavelSindler:
Debug mode is already used in MK3 and MK3S and it is helpful for debugging mmu issues and there is no downside in my opinion. Only thing that is excluded from debugging output on serial line is FINDa reading as this occurs very frequently (300ms).

PavelSindler:
It is not prefered to use PRINTER_TYPE in code unless it is really necessary. It is better in my opinion to have  use preprocesor logic for different features and define it in variant files/config files. This allows to easily add new printer type with its own variant file/config file without changing the code.

kdomanski:
That makes sense, thanks!

Firmware/mmu.cpp
```diff
@@ -181,6 +181,18 @@ bool check_for_ir_sensor()
 #endif //IR_SENSOR
 }
 
+static bool activate_stealth_mode()
```
mkbel, line 184: #define MMU_STEALTH_MODE has misleading name, it sounds like it is enabling stealth mode support, but in reality it is forcing it, I would propose to rename it to something like MMU_FORCE_STEALTH_MODE

There are three possible alternatives of how activate_stealth_mode() behave, but it is not clear from the source code at the first sight and in case both defines are enabled, first action is redundant.
I would propose something like:

```cpp
#if defined (MMU_STEALTH_MODE)
	return true;
#elif defined (SILENT_MODE_STEALTH)
	return (eeprom_read_byte((uint8_t*)EEPROM_SILENT) == SILENT_MODE_STEALTH);
#else
	return false;
#endif
```

PR #1584, opened 2019-03-01
Fix compiler warnings

mkbel:


3d-gussner:
Also during MK25 compiling i get warnings about redefined `#define E0_MS1_PIN` and others

PR #1607, opened 2019-03-08
7x7 mbl and switching mmu modes

PavelSindler:


stahlfabrik:
Hi @PavelSindler,

very interesting approach that you take to work around the influence of the magnets.

Please see the attached picture - I have questions:-):
7x7_magnets

1) Point A seems to not be over a magnet - why did you choose to mark it as a invalid point? Also point A is part of the 3x3 probing for ever.
2) Point B is invalid because it is over a magnet. Why though, is Point C valid? Under the thermistor gold foil there is a magnet too. So I think that point C might not be "magnet-free"
3) Point D is invalid - But this point has been part of the 3x3 probing for ever - I find that noteworthy (and relates to my 1) question)

Best regards
stahlfabrik

PavelSindler:
Thank you for feedback. I made this mask from heatbed *.dwg drawings to test this concept and wanted to double check it properly with real heatbeds later. I will work on it today (there is one bug and also whole Z point coordinate approximation should be improved) and provide more information in the evening. From your input it really seems that some points may be marked incorrectly. I just want to point out that whole mesh has moved little bit (10mm) to the left side of heatbed comapered to standard/old mesh bed leveling.

PavelSindler:
Heatbed plan with current measurement points which I was using for deciding which point is valid and which could be affected by magnet is below. Mesh points were moved 10mm to the left compared to original 3x3 mesh bed leveling. Pink is restricted area.
mesh_plan

3d-gussner:
Hi Pavel did you see my comment in your repository https://github.com/PavelSindler/Prusa-Firmware/pull/1 ?

PavelSindler:
Sorry I haven't noticed it before. It seems promising. Thank you.

stahlfabrik:
What I am not a huge fan of is not having the three middle probe points (Forming an "I") in the exact mechanical middle of the bed (vertically). Also the three middle points that form the "-" should be right in the middle (horizontally).

The middle of course, is defined by the 9 screw positions where the heated is mounted too. In order to do mechanical adjustments the PINDA should really hit the two middle axes defined by the screws. IMHO.

Just by eyeballing it seems that this might not be considered yet? I am not sure! It is hard to tell because the drawings do not show the screw holes. 

Spacing of the probe points should be the second thought then - The goal would be there to lessen the influence of most magnets in order to get many "valid" measurements

3d-gussner:
@stahlfabrik I think it is more important to get as accurate readings from the PINDA as possible and as less interference with the magnets compared to the mechanical properties. At the end the hole bed flat-/warpness will be calculated.

PavelSindler:
So I have used @3d-gussner 's mesh and three points are now marked as "invalid". For most printers I think all points will be probably measured accurately enough. MK2.5 may be assembled more inaccurately so there is higher risk that PINDa will be affected by magnet on some point, but I think what we have now may be good starting point.

mionut:
Hello, i think `MESH_MIN[MAX]_X[Y]` values should be modified to be dependent on `BED_X[Y]0[n]` value, or the other way around.

3d-gussner:
Hi @PavelSindler, i tried few other options and maybe you find one of these more promising. I don't know how far the magnetic fields are going and what a save distance is (before PINDA get faulty reads), do you have some data from the micrometer you have in the code? PINDA vs Micrometer?

- Most promising / min lost points + highest distance to magnet cutouts is with nozzle starting point x0 y0 and spacing in x 34mm and y 30mm.
  Pro:
  - just 2 probe points need to be ignored
  - minimum distance of non interfering probe points is ~2.02mm vs. 1st attempt ~0.25mm
  - x spacing is quite optimal as the left and right probe points have equal distance ~19mm to the build area
  Cons:
  - x and y spacing are not the same
  - front probe points are ~1mm from the build area and the back are ~21mm away

7x7 MBL points x0 y0 xd34 yd30 
7x7 MBL points x0 y0 xd34 yd30 v6

Here some others i tried:
7x7 MBL points x0.5 y0.5 xd34 yd34
7x7 MBL points x0 5 x0 5 v5
7x7 MBL points x1 y1 xd34 yd34
7x7 MBL points x1 y1 v5
7x7 MBL points x0 y1 xd34 yd34
7x7 MBL point x0 x1 v5


Here a Fusion 360 share link https://a360.co/2TrpuLp

leptun:
Tried the modifications on my modded Mk2.5-Mk3 and it is working like it should. It did not however do this with 0.15 z lift. Now with 0.2mm z lift it works perfectly

mkbel (commented):
removing bed_ref_points probably breaks MK2 mesh bed leveling

mkbel (commented):
MK2 mesh bed leveling would be probably broken by removing/changing bed_ref_points
BED_X, BED_Y macros probably can be replaced by functions and this could save some flash

Firmware/Marlin_main.cpp
```diff
@@ -4515,9 +4516,20 @@ if((eSoundMode==e_SOUND_MODE_LOUD)||(eSoundMode==e_SOUND_MODE_ONCE))
...
-				kill_message = _i("Bed leveling failed. Sensor disconnected or cable broken. Waiting for reset.");////MSG_BED_LEVELING_FAILED_PROBE_DISCONNECTED c=20 r=4
-				break;
+			if (MESH_HOME_Z_SEARCH - current_position[Z_AXIS] < 0.1f) { //broken cable or initial Z coordinate too low. Go to MESH_HOME_Z_SEARCH and repeat last step (z-probe) again to distinguish between these two cases.
+
+				current_position[Z_AXIS] = MESH_HOME_Z_SEARCH;
+				plan_buffer_line(current_position[X_AXIS], current_position[Y_AXIS], current_position[Z_AXIS], current_position[E_AXIS], Z_LIFT_FEEDRATE, active_extruder);
+				st_synchronize();
+
+				if (!find_bed_induction_sensor_point_z((has_z && mesh_point > 0) ? z0 - Z_CALIBRATION_THRESHOLD : -10.f, nProbeRetry)) { //if we have data from z calibration max allowed difference is 1mm for each point, if we dont have data max difference is 10mm from initial point  
+					kill_message = _T(MSG_BED_LEVELING_FAILED_POINT_LOW);
```
mkbel: Unused variable kill message.

Firmware/mesh_bed_calibration.cpp
```diff
@@ -3065,4 +3065,28 @@ void mbl_mode_init() {
 	uint8_t mbl_type = eeprom_read_byte((uint8_t*)EEPROM_MBL_TYPE);
 	if (mbl_type == 0xFF) e_mbl_type = e_MBL_OPTIMAL;
 	else e_mbl_type = mbl_type;
+}
+
+bool mbl_point_measurement_valid(uint8_t ix, uint8_t iy, uint8_t meas_points) {
```
mkbel: It would be nice to state input values valid ranges.

Firmware/mesh_bed_calibration.cpp
```diff
@@ -3065,4 +3065,28 @@ void mbl_mode_init() {
...
+	    //"human readable" heatbed plan
+		//magnet proximity influence Z coordinate measurements significantly (40 - 100 um)
+		//0 - measurement point is above magnet and Z coordinate can be influenced negatively
+		//1 - we should be in safe distance from magnets, measurement should be accurate
+		uint8_t valid_points_mask[7] = {
+					//[X_MAX,Y_MAX]
```
mkbel, line 3109: It would be nice do decorate with point numbers as it is hard to guess indexes.

Firmware/mesh_bed_calibration.cpp
```diff
@@ -3065,4 +3065,28 @@ void mbl_mode_init() {
...
+			0b1111101,
+			0b1110111,
+			0b1111111,
+			0b0111011,
```
mkbel: 0b0111011
--*------
Looks like we have invalid point in 3x3 mesh bed plan, looks suspicious.

Firmware/mesh_bed_calibration.cpp
```diff
@@ -3065,4 +3065,28 @@ void mbl_mode_init() {
...
+			0b1110111,
+			0b1111111,
+			0b1110111,
```
mkbel, line 3112: 0b1110111
-----*---
Looks like we have invalid point in 3x3 mesh bed plan, looks suspicious.

Firmware/mesh_bed_calibration.cpp
mkbel: Ahh, fixed in later commit.

Firmware/mesh_bed_calibration.cpp
mkbel: Ahh, fixed in later commit.

Firmware/mesh_bed_calibration.cpp
```diff
@@ -3067,11 +3067,13 @@ void mbl_mode_init() {
 	else e_mbl_type = mbl_type;
 }
 
-bool mbl_point_measurement_valid(uint8_t ix, uint8_t iy, uint8_t meas_points) {
+bool mbl_point_measurement_valid(uint8_t ix, uint8_t iy, uint8_t meas_points, bool zigzag) {
 	    //"human readable" heatbed plan
 		//magnet proximity influence Z coordinate measurements significantly (40 - 100 um)
 		//0 - measurement point is above magnet and Z coordinate can be influenced negatively
 		//1 - we should be in safe distance from magnets, measurement should be accurate
+		if ((ix >= meas_points) || (iy >= meas_points)) return false;
```
mkbel, line 3106: I can still read out of array bounds, if i set meas_points higher then 7. But as it is reading it is not dangerous.

Firmware/mesh_bed_calibration.cpp
```diff
@@ -3089,4 +3091,24 @@ bool mbl_point_measurement_valid(uint8_t ix, uint8_t iy, uint8_t meas_points) {
...
 		else return (valid_points_mask[6 - iy] & (1 << ix));
+}
+
+void mbl_single_point_interpolation(uint8_t x, uint8_t y, uint8_t meas_points) {
+		uint8_t count = 0;
+		float z = 0;
+		if(mbl_point_measurement_valid(x, y+1, meas_points, false)) { z += mbl.z_values[x][y+1]; count++; }
+		if(mbl_point_measurement_valid(x, y-1, meas_points, false)) { z += mbl.z_values[x][y-1]; count++; }
+		if(mbl_point_measurement_valid(x+1, y, meas_points, false)) { z += mbl.z_values[x+1][y]; count++; }
+		if(mbl_point_measurement_valid(x-1, y, meas_points, false)) { z += mbl.z_values[x+1][y]; count++; }
```
mkbel: Sign error valid(x-1 VS values[x+1]

Firmware/mesh_bed_calibration.cpp
```diff
@@ -3089,4 +3091,24 @@ bool mbl_point_measurement_valid(uint8_t ix, uint8_t iy, uint8_t meas_points) {
 		}
 		if((iy%2) == 0)	return (valid_points_mask[6 - iy] & (1 << (6 - ix)));
 		else return (valid_points_mask[6 - iy] & (1 << ix));
+}
+
+void mbl_single_point_interpolation(uint8_t x, uint8_t y, uint8_t meas_points) {
+		uint8_t count = 0;
+		float z = 0;
+		if(mbl_point_measurement_valid(x, y+1, meas_points, false)) { z += mbl.z_values[x][y+1]; count++; }
+		if(mbl_point_measurement_valid(x, y-1, meas_points, false)) { z += mbl.z_values[x][y-1]; count++; }
```
mkbel: You are passing negative number to mbl_point_measurement_valid() second parameter. It works as intended but this feature should be at least documented.

Firmware/ultralcd.cpp
```diff
@@ -6632,10 +6634,46 @@ static void lcd_tune_menu()
 	MENU_END();
 }
 
-static void mesh_bed_leveling_menu()
+static void mbl_magnets_elimination_set() {
```
mkbel, line 6664: Misleading function name. It is called set, but in reality it is toggle.

Firmware/ultralcd.cpp
```diff
@@ -6632,10 +6634,46 @@ static void lcd_tune_menu()
...
+	bool magnet_elimination = (eeprom_read_byte((uint8_t*)EEPROM_MBL_MAGNET_ELIMINATION) > 0);
+	magnet_elimination = !magnet_elimination;
+	eeprom_update_byte((uint8_t*)EEPROM_MBL_MAGNET_ELIMINATION, (uint8_t)magnet_elimination);
+}
+
+static void mbl_mesh_set() {
```
mkbel, line 6670: Misleading function name. It is called set, but in reality it is toggle.

Firmware/ultralcd.cpp
```diff
@@ -6632,10 +6634,46 @@ static void lcd_tune_menu()
...
+static void lcd_mesh_bed_leveling_settings()
 {
+	
+	bool magnet_elimination = (eeprom_read_byte((uint8_t*)EEPROM_MBL_MAGNET_ELIMINATION) > 0);
+	uint8_t points_nr = eeprom_read_byte((uint8_t*)EEPROM_MBL_POINTS_NR);
 
-	SETTINGS_MBL_MODE;
+	MENU_BEGIN();
+	// leaving menu - this condition must be immediately before MENU_ITEM_BACK_P
+	if (((menu_item == menu_line) && menu_clicked && (lcd_encoder == menu_item)) || menu_leaving)
```
mkbel, line 6685: I think first part of condition is redundant, as far as I know each menu is guaranteed to be called once as it is left with menu_leaving flag set, independent of reason for leaving the menu.

mkbel: I checked that and it is no more true. But I am 100%, that in history menu_goto() called previous menu with menu_leaving flag, someone must have broken that feature. And now I understand that ticket, that something is not saved (Lin. correction).

Firmware/mesh_bed_calibration.cpp
mkbel: Already fixed in later commit.

PR #1615, opened 2019-03-10
Mk3 German lang fix1

3d-gussner:
Fixed few translations like "Severe skew" and "Slight skew" as the translations are too long to be displayed correctly on LCD.
Also added few replacements for German as well for 'Ä', 'Ö', 'Ü' and 'ß' to prevent issues after using PO-Edit to translate.


3d-gussner:
Outdated and no reaction for Prusa so i close it.

PR #1616, opened 2019-03-10
MK3S Selftest Print Fan Fix

ojrik:
MK3S Selftest Print Fan Bug  FAN_SOFT_PWM


PavelSindler:
We have manual check there conciously. After change to software pwm, automatic selftest check for swapped fans wasn't reliable so we solve this temporarily by manual check. It is better to have reliable manual check, then unrealiable automatic check in my opinion. We are aware that it is not very user friendly and we will fix it.

PR #1624, opened 2019-03-13
Update build environment to fix maximum flash size check.

mkbel:


3d-gussner:
Hi @mkbel , can you explain why you lowered the max upload value for the Rambo? Is it to prevent strange behavior of the controller reaching 97% of its flash we had in the past? I remember that when I was adding few more languages back in MK2  branch, as it hit the 97% mark it was unreliable. Also bit later was an issue with 'too much' code and was fixed to go under the 97% mark.
At this moment the miniRAMbo can handle the code but the Einsy with the TMC drivers barely fits.
As the Einsy got some extra flash would it be possible to move some code there? 
Like the 1st layer calibration uses quite a bit of internal flash and is barely used.
I was also trying to disable farm mode, so it cannot be switched between farm mode and user mode. Also saved few bytes. If you need farm mode be active you can compile that version, missing other things.
At this moment it looks like we reached the limits and new features may be hard to implement, especially on the Einsy.

3d-gussner:
Just tried some quick/dirty un-comment for the 1st layer cal and saved around 412 bytes. I don't know how easy/hard it would be to move things/code to the extra flash and how to use it when needed.
At this moment the extra flash is just used to store the translations and there is internal space reserved for it to be copied.

mkbel:
> Hi @mkbel , can you explain why you lowered the max upload value for the Rambo?

I explained it upstream. https://github.com/ultimachine/ArduinoAddons/pull/3 Limit set there was wrong. In discrepancy with bootloader size and bootloader related fuze bits. And you only noticed, that something is wrong too late - firmware verification fails, because bootloader refuses to rewrite itself, or even worse with some our ancient bootloader it rewrites itself leading to bricked board.

mkbel:
> At this moment it looks like we reached the limits and new features may be hard to implement, especially on the Einsy.

We are hitting flash limit quite regularly, forcing us to optimize some code or update compiler etc. There is still lot of area which can be optimized, like user interface, stored gcode fragments, utilization of external flash.

3d-gussner:
@mkbel thanks for the answer didn't see that you opened an pr at ultimachine.
Yeah the portable mode seperates it quite well. You can see my changes made to Arduino IDE in my release notes https://github.com/3d-gussner/PF-build-env/releases

PR #1627, opened 2019-03-13
print parameters checking

MRprusa3d:
nozzle diameter


Firmware/Marlin_main.cpp
```diff
@@ -1039,6 +1039,9 @@ void setup()
 		//disabled filament autoload (PFW360)
 		fsensor_autoload_set(false);
 #endif //FILAMENT_SENSOR
+          // ~ FanCheck -> on
```
mkbel, line 1066: It looks like it is not related to nozzle diameter.

mkbel: It is for farm mode only.

Firmware/ultralcd.cpp
```diff
@@ -6631,6 +6746,12 @@ static void lcd_sd_updir()
 
 void lcd_print_stop()
 {
+//-//
+     if(!card.sdprinting)
+          {
+          SERIAL_ECHOLNPGM("// action:cancel");   // for Octoprint
```
mkbel, line 6752: How it is related to parameters checking?

Firmware/util.cpp
```diff
@@ -327,3 +327,46 @@ void update_current_firmware_version_to_eeprom()
...
+{
+uint16_t nDiameter_um;
+
+nDiameter_um=eeprom_read_word((uint16_t*)EEPROM_NOZZLE_DIAMETER_uM);
+if(nDiameter==nDiameter_um)
+     return;
+switch(eCheckMode)
+     {
+     case e_CHECK_MODE_warn:
+          lcd_show_fullscreen_message_and_wait_P(_i("Nozzle diameter doesn't match! Press the knob to continue."));
```
mkbel, line 365: There is no timeout.

mkbel (approved):

PR #1644, opened 2019-03-17
Mk3 lang fix2

3d-gussner:
Does it makes sense that i put more time in that kind cleanup for a pull request?

This is not completely tested yet, just a beginning.

Why do i do that?
- There are lot of messages which aren't in the 'messages.c' and 'messages.h' even they have been in the MK2 branch. 
- Lot of the translations are made do not fit correctly on the LCD screen and generate unpleasant results. Issue is that there not all messages have c=x and r=y limits defined.

Possible improvement:
We are hitting the EINSY resource limits quite often, so I was thinking about a different approach to use the internal and external flash for the languages.
- Now all the english messages are stored in the flash
- for the additional language translation a reserved space for secondary language (12032 bytes) of the internal flash is used.
- Wouldn't it be better to have just **one** reserved 'language space' in the internal flash?
  - Few messages have to stay in the internal flash to be able to switch between the languages
  - All other english messages are stored in the reserved 'language space' and used as it is done now with the other languages
  - As soon  another language is chosen by the user, the language from the external flash is copied to the 'language space' replacing all existing messages.
     - All languages **including** english need to be flashed to the external flash during firmware update
  - this way i guess we could save 8-10kb of internal space


3d-gussner:
@mkbel @PavelSindler 
Short update. Compiling on windows PF-build-env for the MK3 with english only uses 
```
Sketch uses 234452 bytes (92%) of program storage space. Maximum is 253952 bytes.
Global variables use 6110 bytes of dynamic memory.
```
compared to multi-languages
```
Sketch uses 252004 bytes (99%) of program storage space. Maximum is 253952 bytes.
Global variables use 6107 bytes of dynamic memory.
```

That are 17552bytes less! 

As I wrote before, wouldn't it make sense to point to ONE language space and have english next to the other languages in the external flash, which could be copied to 'language space' if needed?

mkbel:
> As I wrote before, wouldn't it make sense to point to ONE language space and have english next to the other languages in the external flash, which could be copied to 'language space' if needed?

Definitely makes sense for me and sounds like easiest way of freeing some space in flash.

3d-gussner:
@mkbel Thanks for the answer. Hope it will be implemented.
Do you think I should continue with the messages cleanup? Would be pity if all the time would be wasted if there is very low chance to get the PR merged.
Also do you think i should split these in smaller parts?

Firmware/cardreader.cpp
```diff
@@ -183,23 +183,23 @@ void CardReader::initsd()
   {
     //if (!card.init(SPI_HALF_SPEED,SDSS))
     SERIAL_ECHO_START;
-    SERIAL_ECHOLNRPGM(_n("SD init fail"));////MSG_SD_INIT_FAIL c=0 r=0
```
mkbel, line 186: I don't think any of change of kind _n("") --> MACRO is right. _n means not to be translated, as those are messages outputted to serial line.

Firmware/ultralcd.cpp
```diff
@@ -763,24 +763,24 @@ void lcdui_print_status_line(void)
...
+					lcd_puts_P(_T(WELCOME_MSG));////WELCOME_MSG c020 r=0
+					lcd_setstatuspgm(_T(WELCOME_MSG));////WELCOME_MSG c020 r=0
 					custom_message_type = CUSTOM_MSG_TYPE_STATUS;
 				}
 				if (custom_message_state > 3 && custom_message_state <= 10 )
 				{
 					lcd_set_cursor(0, 3);
 					lcd_puts_P(PSTR("                   "));
 					lcd_set_cursor(0, 3);
-					lcd_puts_P(_i("Calibration done"));////MSG_HOMEYZ_DONE c=0 r=0
```
mkbel, line 787: I don't thing most of changes of kind _i("") --> MACRO are right. _i("") is preferred form. There is currently one limitation with _i("") form, that if the same string is used in two places it drains twice as flash, so for such strings it should be avoided until this limitation is fixed.

Firmware/cardreader.cpp
3d-gussner: Thanks for the answer...and sorry that my is so late, was on vacation.
I close this PR and try another one

3d-gussner:
Outdated so i close it

PR #1653, opened 2019-03-19
Phase correct PWM for bed, frequency 40KHz.

XPila:


Panayiotis-git:
I tried this PR. Even though the printer now reports lower environment temperature (50 degrees instead of >60), the MOSFET is still too hot to touch and there is a high frequency sound produced by the bed.

PavelSindler:
We will probably use it in next fw version, because temperature reduction is now highest priority and then we will work on switching modes dependending on measured ambient temperature.

PavelSindler:
So in the end we are reverting this PR and will work on better solution. On some printers there was really annoying high frequency noise going from power supply.

eamaclean:
> 
> 
> So in the end we are reverting this PR and will work on better solution. On some printers there was really annoying high frequency noise going from power supply.

I would suggest you remove the high frequency PWM driving entirely for the time being, seeing as it is capable of damaging hardware..

PR #1658, opened 2019-03-19
Allow save of uStep resolution with M500

Harpalyke:
Allow saving of the micro-stepping resolution to EPROM using the M500 command.

Changing this setting is useful for folks who are running geared extruders like skelestruder with a motor that requires a very high number of esteps such as the LDO pancake 0.9° stepper. 

At the moment changing the e axis resolution requires a recompile of the firmware. With this change, a user can run an "M350 Exxx" command to set the resolution for the e axis and then run an M500 to persist the changes.


mkbel:
And what about using M350 command to change E resolution?

mkbel:
Thank you for your contribution.

PR #1661, opened 2019-03-20
Clarify the sound settings

DavidWAbrahams:
I found the sound settings confusing. The name didn't make sense until I read the source code. I suggest some new names here that might be more clear.

Old: Sound [loud, once, silent, mute]
My confusion is that the words "silent" and "mute" are too similar. "once" is vague. "loud" is inaccurate, since it doesn't effect the volume.

(modified) proposal: two separate settings. One for the menu sounds and one for alert sounds.
Menu beep [on, off]
Alerts [repeat, once, off]
https://photos.app.goo.gl/5KcHWpXttWe7o8Uh9

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -660,7 +660,7 @@ static void factory_reset(char level)
                    
         // Level 0: Language reset
         case 0:
-if((eSoundMode==e_SOUND_MODE_LOUD)||(eSoundMode==e_SOUND_MODE_ONCE))
+if(eSoundMode==e_SOUND_MODE_ON)
             WRITE(BEEPER, HIGH);
             _delay_ms(100);
             WRITE(BEEPER, LOW);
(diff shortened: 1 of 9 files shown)


PavelSindler:
"Once" actually makes sence as it dont repeat periodic beeping (which is present on filament runout for example) but just beeps once in that case.

jurriaan:
@PavelSindler shouldn't that be a separate setting then? To me it is kinda weird that some options affect the menu and some effect filament runout.

DavidWAbrahams:
I agree, there are really two behaviors being controlled by one setting. My ideal would be something like:
Alarms [repeating, once, off]
Menu beep [on, off]

DavidWAbrahams:
Ok, I've done the much harder change of separating beeps and alerts.

Menu beep [on, off]
Alerts [repeat, once, off]
https://photos.app.goo.gl/5KcHWpXttWe7o8Uh9

jurriaan:
@PavelSindler What do you think?

jurriaan:
cc @mkbel @XPila

jurriaan:
@PavelSindler I see you just released the RC of the MK2 firmware including sound settings. Does it make sense to backport this PR as well?

DavidWAbrahams:
BTW I've been running this mod on my MK3s for weeks now. Things are working normally. I have muted the menu but left alarms repeating. When my filament runs out, the alarm repeats as expected.

larsrc:
So none of these settings affect the sound level of the M300 gcode output?

leptun:
Nope. M300 only changes the duration and frequency. Sound intensity is always the same.

wilybrace:
Does this need a new advocate to update/rebase?

DavidWAbrahams:
Yes if you want this to move forward, someone will need to rebase and get a code review. I don't know the process to get a official code review so I gave up.

3d-gussner:
@DavidWAbrahams Thanks for this PR and the idea. At the moment I will close this PR as
- It would need a re-base
- It would differ from the 32-bit firmware.

PR #1677, opened 2019-03-27
7x7 polishing

PavelSindler:


Firmware/mesh_bed_calibration.cpp
```diff
@@ -678,11 +678,13 @@ void reset_bed_offset_and_skew()
...
 // offsets of the Z heiths of the calibration points from the first point are saved as 16bit signed int, scaled to tenths of microns
-{
-    for (int8_t i = 0; i < 8; ++ i)
-        if (eeprom_read_word((uint16_t*)(EEPROM_BED_CALIBRATION_Z_JITTER+i*2)) == 0x0FFFF)
-            return false;
-    return true;
+// if at least one 16bit integer has different value then -1 (0x0FFFF), data are considered valid and function returns true, otherwise it returns false
+{	
+	bool data_valid = false;
+	for (int8_t i = 0; i < 8; ++i) {
```
mkbel, line 684: It would be nice, if EEPROM_BED_CALIBRATION_Z_JITTER would be struct instead of preprocessor macro, so you would use its size as iteration limit.

Firmware/ultralcd.cpp
```diff
@@ -6688,32 +6688,38 @@ static void mbl_mesh_set() {
 	eeprom_update_byte((uint8_t*)EEPROM_MBL_POINTS_NR, mesh_nr);
 }
 
+static void mbl_probe_nr_set() {
```
mkbel, line 6691: Suggest rename set to toggle.

PR #1683, opened 2019-04-01
Revert "Phase correct PWM for bed, frequency 40KHz."

PavelSindler:
Reverts prusa3d/Prusa-Firmware#1653

### DIFF
--- a/Firmware/timer02.c
+++ b/Firmware/timer02.c
@@ -11,11 +11,19 @@
 
 uint8_t timer02_pwm0 = 0;
 
-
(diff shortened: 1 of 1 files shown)


GurliGebis:
So in other words, 3.7.0 is still able to fry the motherboard @PavelSindler ?

leptun:
That is correct

GurliGebis:
Okay, maybe there should be a big, fat, warning telling people to not use it, since it likely can fry the board.

PR #1691, opened 2019-04-03
show estimated time

PavelSindler:
- cherry-picked commits from MK3 branch
- plus M27 update

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,14 @@
 /Firmware/Firmware.vcxproj
 /Firmware/Firmware.sln
 /Firmware/Firmware - Shortcut.lnk
+.settings
+.project
+.cproject
+Debug
+Firmware/Configuration_prusa.h
+Firmware/Doc
+/Firmware/.vs/Firmware/v14/.suo
+/Firmware/Firmware.sln
+/Firmware/Firmware.vcxproj
+/Firmware/Firmware.vcxproj.filters
+/Firmware/__vm
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -349,7 +349,7 @@ extern uint16_t print_time_remaining_silent;
 
 #define PRINT_TIME_REMAINING_INIT 0xffff
 #define PRINT_PERCENT_DONE_INIT   0xff
-#define PRINTER_ACTIVE (IS_SD_PRINTING || is_usb_printing || isPrintPaused || (custom_message_type == 4) || (lcd_commands_type == LCD_COMMAND_V2_CAL) || card.paused || mmu_print_saved)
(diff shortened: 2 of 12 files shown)


3d-gussner:
Nice to see some development on MK2. What are we missing to merge MK2 branch with MK3?

PavelSindler:
They differ a lot as there were menus reworked, stepper routines, selftest, calibration xyz and much more. So we choosen some nice features together with important safety updates and will add them to MK2. Another option was to start building MK2 release from MK2.5 version. But MK2 and MK2.5 versions differ more then is obvious or expected. Latest MK2 fw is working reliably in my opinion and I think that more safe and controlled way is to continue MK2 develompent there and just pick certain features and merge/cherry-pick it to MK2 branch.

cimoalpacino:
I've used MK3 fw and ported it to an Arduino RAMPS board.
It is compatible with the MK42 heated bed from the MK2 and PINDA v1. It has new xyz calibration, pid tuning, remaining time... 
Due to lack of time i had to disable:
- filament sensor for which i will use the old mechanical switch 
- PINDA v2 (which i don't have)

So in theory and practice it is possible to adopt the MK2.5 firmware to a MK2/S. 

Here's my repo:
https://github.com/cimoalpacino/Prusa-Firmware/tree/MK3_3.6.0_ARDUINO_RAMPS/Firmware

cimoalpacino:
I would just like to mention that i also had mechanical filament sentor and power panic (UVLO) working on a previous mod of 3.4.0-RC1 (https://github.com/cimoalpacino/Prusa-Firmware/tree/MK3_RAMPS/Firmware)

leptun:
Why do you keep developing on this branch when you can still use the latest branch with some changed for MK2?

I have been using a slightly moddified 3.4.1 firmware on my MK2 reliably since it was released.

Also it would be nice if NEW_XYZ_CALIBRATION was made functional for the mk2. In my testing with changed I was able to get it to try to find the point where it should, but when the spiraling motion got close to the endstop it crashed into it without detection resulting in invalid location for second point. At that point I gave up since I didn’t have time. Ended up using old xyz calibration and at some point even added a pinda v2 which worked perfectly with pinda thermistor enabled.
Maybe I can help backporting Mk2 to latest Mk3 firmware.

cimoalpacino:
NEW_XYZ_CALIBRATION is working normally on MK2 with fw 3.6.0.
It just needs a bit more tweaking.

PavelSindler:
I 've already answered above why I chose to develop on MK2 branch. Both ways has pros and cons. Last MK2 release was working reliably and merging new pull requests to this branch is more controlled way (you can see and check which changes are you doing). Releasing MK2 from MK3 branch can be done, it would work, but there would be probably many hidden bugs, which we wouldn't be able to find before release and it would need much more time in my opinion to find and fix them.

Issue #1680, opened 2019-03-31
Self-test nozzle crash on MK3 without steel sheet


wavexx:
I did the self-test several times so far, and I re-did another one recently to test the 7x7 leveling on 3.7.0-RC1, both before and after the XYZ calibration.

During the self-test, the nozzle is lowered until some triggering point. I noticed already early on that there's quite a difference if you have the steel sheet on or not: without the steel sheet, the nozzle is brought *very* close the bed. With the steel sheet, there's significant margin left before it's stopped.

In my last self-test I didn't have the sheet and the nozzle pushed down on the bed for roughly 1mm. This was actually done just after a XYZ calibration. I have no reason to believe this has anything to do with 3.7.0-RC1 specifically. No hardware change was done. The pinda distance was (and still is) set using the zip-tie method.

I probably did the self-test more than 10 times for sure, and never had a crash so far, but I noticed there's already a report about this: #605. For reference, my live-z-adjust is around -0.9, and XYZ calibration never gave me the impression of getting *this* close. Definitely, overshooting by 1mm is odd.

michalxfanta:
Hello, for 7x7 mesh bed leveling, the steel sheet must be placed on the bed.

wavexx:
This is about the self-test, not the leveling.

michalxfanta:
OK, does this problem remain with the final FW 3.7.0?

wavexx:
On Thu, Apr 04 2019, Michal Fanta wrote:
> OK, does this problem remain with the final FW 3.7.0?

This was reproduced on v3.7.0-RC1, by looking at the commits to the
final 3.7.0 I don't see any change in the self-test procedure in
ultralcd that could have fixed the issue in any way.

By briefly looking at the code in lcd_selfcheck_axis, we move Z in steps
of 1mm, checking for the endstop value at each iteration.

By what I've seen, if we start to move just at the boundary before the
pinda triggers while the distance is lower than a full millimiter we
will hit the bed.

Given that Z calibration still hovers correctly without touching, we
should maybe choose a better XY position on the plate that triggers the
probe earlier or reduce the stepping towards the end.

wavexx:
See the discussion on #1703. The proposed PR fixes the issue for me and has an improved behavior overall.

wavexx:
I think I can close my own bug report ;)
Didn't have PINDA offset issues in a long time.

Issue #605, opened 2018-04-06
MK2.5 Self Test will crash nozzle in bed if no sheet on


SaijSaij:
Printer: MK2.5
Firmware: 3.1.3 official release
Issue: When you run the self test without the steel sheet on the nozzle will crash into the bed on testing z-endstop as the probe is not above one of the new points (looks like the position is for the old MK42 bed points).
Steps:
1. Remove steel sheet
2. Start self test via menu
3. Wait (and keep a metal thing at hand to trigger the probe endstop)
Repeatable: yes

Martorias:
I built mine 2 days ago and didn't have this issue. Didn't put the sheet on until asked by the calibration wizard.

SaijSaij:
Well just run the selftest yesterday and it did the descripted thing.

gertlind:
I just completed the 2.5 upgrade.
Had no issues during the calibration process.

clearbucketLabs:
self-test is a different thing from calibration....

gertlind:
Idid the self test also. No issues.

SaijSaij:
Just downgraded back to MK2S or else I could upload a video.
Maybe you need todo a factory reset to trigger this problem (this is what I ever do when I'm doing such an upgrade)

SaijSaij:
Sorry for not writing for so long.

I rebuild the MK2.5 and tested the selftest again. This time I did not a factory reset and I did a short video of the whole self test process.

Link to video: https://www.youtube.com/watch?v=q1546BJgfIQ&feature=youtu.be

As you can see, the probe is not above the point on the heatbed.

Prusa-Support:
Hello! I am absolutely sorry for not intervening on this sooner. Was the issue resolved in the meantime? From the video and description, I believe the main contributor was the PINDA probe position being too high. Closing the issue for now, but in case you still have similar problems, can you please let me know via email? info@prusa3d.com mentioning me. 

Alessandro Pantaleo
Prusa Research

PR #1703, opened 2019-04-06
Fix bed crash during selftest

leptun:
This PR fixes crashing of the head during selftest when it lowers Z to find point.
Changes:
-Lower stepping of z to avoid touching when pinda is mounted higher in holder (Live Z < 1000)
-Does not use hard coded value for the bed point location.
The point was still off on my MK3 on Y axis, so I added an offset of 4mm on Y to the point. This fixed the issue.

This needs testing on the MK25(S) and MK3S.
Issue #1680


wavexx:
Is the pinda expected to be exactly above the mark while lowering? (I didn't pay attention to this - I assumed it wouldn't matter except for calibration)

leptun:
This is so that without steel sheet it triggers higher and does not touch/push the bed with the nozzle

leptun:
Pinda gets triggered not only by the points, but also the traces on the bed. However it is lower on the traces and it might slightly crash and push the bed.

Attached are hex files for testing. Not sure if this would work on MK2 though. If anyone can test that would be great.
[FW370-Build2201_fix_selftest_Z_crash.zip](https://github.com/prusa3d/Prusa-Firmware/files/3050490/FW370-Build2201_fix_selftest_Z_crash.zip)

wavexx:
If you look at the earliest report I've found (#605, linked video: https://www.youtube.com/watch?v=q1546BJgfIQ&feature=youtu.be) that definitely looks like the case.

I'm waiting for a print to finish and I'll give this a test later on.

Firmware/ultralcd.cpp
```diff
@@ -7386,9 +7393,6 @@ static bool lcd_selfcheck_pulleys(int axis)
 			((READ(Y_MIN_PIN) ^ Y_MIN_ENDSTOP_INVERTING) == 1)) {
 			endstop_triggered = true;
 			if (current_position_init - 1 <= current_position[axis] && current_position_init + 1 >= current_position[axis]) {
-				current_position[axis] += (axis == X_AXIS) ? 13 : 9;
```
wavexx, line 7653: what was the original intention here?

leptun: Those were the hard coded values for mk2 after it checked for loose pulleys. The selftest does as follows:
MK2/MK2.5(S) : Home X -> Check pulleys and move X axis to Z probe position -> The same thing for Y axis -> Raise Z -> lower Z to point
Newer printers : Home X -> Check Axis length X-> Check Axis length Y -> move to point position while raising Z -> lower Z to point
With this PR the head is moved to the point while being raised AFTER checking X and Y also on older printers without using hard coded values.

wavexx:
I gave this a shot in two runs: one without adjusting the position (just to test the half-stepping) and one as-it-is: much better in every respect.

Z stops at a safe distance now in any case.

With the position correction Y alignment is perfect, while X looks slightly off (probe is at +1mm in my case). Did you simply correct those 4mm by measuring? I wonder if the constants are wrong to begin with?

leptun:
The values used are the uncorrected values used in XYZ calibration. Without XYZ calibration the offset is unknown, so the point location is off. I could use corrected values, but they are unavailable right after you assemble the printer or you run wizard. What I could do and probably will is checking weather stored data is valid, and if yes use the corrected values. If not then use this 4mm offset

wavexx:
What about using the calculated end-stop distance for XY and the know bed size instead?

This should be better than a simple offset from the bottom-left and does not make assumptions about the build.

wavexx:
Just to make the it clear, I mean calculating the offset from the center of the axis instead of the left offset only.

leptun:
The problem is that the bed can be offset by assembly by the belt holder.
On Mk2 - Mk2.5S (and in between) the min position can be moved by a few mm and MAX position can't be found
On Mk3 and Mk3S the MIN and MAX positions can be shifted by the same piece while still having the same axis length
Edit: Might not be the same on MK3S, but the same idea applies. I'll try to implement it as i suggested before. I'll push the changes if it compiles, but can't test them. Can you do it?

leptun:
Also on mk3/s the X axis min/max position can be affected by the size and position of the bundle of wires.

wavexx:
I didn't think about this.

Then I wouldn't change anything really: I wouldn't use any stored value for a self-test procedure. It's much better to have a consistent test in any scenario.

leptun:
With what I proposed to change it would work as follows:
-If printer is new, then no data available so use 4mm offset
-if you run wizard the data will be erased before selftest; 4mm offset
-if printer is assembled and calibrated and you run selftest you shouldn’t have a problem with using those stored values.
-if you made some changes to the printer you should run wizard anyway to recalibrate the printer, thus not using what might be wrong data. 4mm offset

If for some reason you run selftest and the values are wrong, you will simply probe next to the point and push the bed a little. Undesirable, but not that problematic.

There is a bigger issue though. If you run selftest from menu after succeeding the printer will autohome. This should be disabled as there is no point where it is probing. Will try to address this issue as well. If you find the piece of code responsible for this, please let me know

wavexx:
The way I see it, a self-test (and more generally a calibration procedure) should never rely on pre-recorded data for consistency. You know the test is going to behave always in the same way, irregardless of the values stored. This makes debugging a *lot* easier. It's a little off, but it can never be more off by any other reason.

wavexx:
Isn't the homing done just a few lines later in ultralcd.cpp:7030?

leptun:
Hmmm, that might be it. It is somewhat weird that it is in the middle of the selftest. Now thinking about it, maybe it is queued to be processed after it finishes. Making the changes now.

leptun:
I suppose you have a MK3 printer. Could you please test how this hex file performs selftest. It should do as I said above. It would be best if you could test the scenarios I listed.

I will also continue to test tomorrow morning.
[FW370-Build2201-1_75mm_MK3-EINSy10a-E3Dv6full-EN_ONLY.zip](https://github.com/prusa3d/Prusa-Firmware/files/3050705/FW370-Build2201-1_75mm_MK3-EINSy10a-E3Dv6full-EN_ONLY.zip)

wavexx:
It's an MK3, but it's not so "original" anymore ;). Can I just rebuild with the current PR?
I wonder if the homing could be needed during the wizard, after the self-test part has finished?

leptun:
I have made changes that I am not sure work. I'll do it, but be warned I am not sure how it is going to do this.
As for the homing, it was done as to home all if wizard was not active. If it was active, it wouldn't home. Now it doesn't home at all and is just lifted more before checking hotend, bed and filament sensor.

Edit: meant to say wouldn't

wavexx:
I'm not too keen on f09d719a8ea75efad90e2f165594021ffe2e4717. Overall, think about it: I would have never discovered the issue in the first place: it would have never crashed in my case. I'd rather not use the existing offsets during the self-test.

leptun:
Well you would have crashed anyway before any of these changes since the values did not use existing offset data

wavexx:
On Sat, Apr 06 2019, Voinea Alex wrote:
> Well you would have crashed anyway before any of these changes since
> the values did not use existing offset data

Not necessarily. In my case the self-test crashed on me exactly once. It
was running unusually close to the bed, but still not crashing.

I've ran it multiple times mostly for checking the "belt status" and
electrical connections while doing maintenance.

If it didn't happen at the wizard at the first install I would probably
have never seen it again, despite the issue being there anyway. I ran
the wizard for the second time ever only for resetting state for the 7x7
leveling.

I think that repeatability is an important feature here.
Dumb is good for a self-test.

leptun:
I’ll revert the last commit then. It is true it should be dumb, but I think the first two commits should still be applied

wavexx:
On Sat, Apr 06 2019, Voinea Alex wrote:
> I’ll revert the last commit then. It is true it should be dumb, but I
> think the first two commits should still be applied

Agree also. The first two are perfect. I didn't test the second yet, but
will do so later on. I don't see anything wrong with it.

wavexx:
Ok, sorry for the delay but I could test it now. No homing performed, just Z being raised.
I think it's perfect.

leptun:
@PavelSindler @mkbel  do you plan to merge this PR or should I just close it?

leptun:
@DRracer  Do you consider merging this PR? If not, let me know and I'll close it.

DRracer:
@leptun yes, we'll look at it, please keep it open for now

leptun:
Fixed an issue where the current Z position after Selftest would be negative. Now it's spot on :)

yschroeder:
After upgrading our MK2.5 to MK2.5S it also fails the selftest with this behaviour. However, Y position looked fine, X coordinate is a few millimeters off (too far to the right). It worked fine with the metal sheet mounted.

DRracer:
@leptun since #2264 is a follow up and and improved version of this PR, would you mind closing this PR?

leptun:
Closed on request @DRracer

PR #1704, opened 2019-04-06
PFW-1049: Enable print fan while waiting for cooldown

leptun:
As the title says, this PR enables the print fan when waiting for nozzle and bed to cool down.
After they are cooled it goes back to its initial speed.


wavexx:
Does it make it faster in your case?
I timed this a couple of times and couldn't see any difference.

leptun:
In my case it cools the hotend much faster this way. It depends on ambient temperature, as a cold environment makes it faster. In my case it is <20*C

Panayiotis-git:
Are you using hotend silicon socks?

leptun:
I am using stock Mk3. No sock.

wavexx:
On Mon, Apr 15 2019, Voinea Alex wrote:
> I am using stock Mk3. No sock.

I'm using the sock, so that might be the reason.

leptun:
@PavelSindler do you plan to merge this PR or should I just close it?

leptun:
@DRracer  Do you consider merging this PR? If not, let me know and I'll close it.

DRracer:
@leptun yes, we'll look at it, please keep it open for now

DRracer:
@lepun Are you sure that setting the print fan to full speed after the print has finished is a good idea for "warpy" plastics like ABS or ASA (even thought the head is in parking position after the print)? I'd rather prefer having these ABS printout cooled down slowly without too much airflow around. Such approach has given me the best results so far. What is your experience?

leptun:
@DRracer You misunderstood what this PR is doing. 

This PR changes the lcd_wait_for_cool_down() function. This is used for waiting for nozzle/bed cooldown before starting a calibration. For example, this way if you do a selftest and then you run xyz calibration it cools down and starts the next calibration faster, as the fan blows on the nozzle till it cools it to 50deg.

This does not affect normal printing, only calibrations.

leptun:
Testing: Preheat printer, start XYZ calibration. (any hotend temp over 50 should me enough)

DRracer (approved):

PR #1719, opened 2019-04-10
Configure extruder height: useful for MK3S firmware and Bondtech or Bear extruders.

ruedli:
The MK3S and MK2.5S firmware is not fit for use with a Bondtech BMG extruder kit or a Bear extruder.

This is because during XYZ calibration, the lower MK3S extruder (compared to Bondtech and Bear) causes the Z calibration to fail prematurely: when the Pinda is triggered "earlier" then expected.

This fix configures a #define Z_MAX_POS_XYZ_CALIBRATION_CORRECTION, to be entered in Configuration_Prusa.h to easily compile and use extruders with a different height. In the example configurations, this define is NOT set, so the firmware is not affected in ANY way.

For Bondtech and Bear owners, compiling the firmware is just a matter of uncommenting the new Z_MAX_POS_XYZ_CALIBRATION_CORRECTION #define and compiling it.

FOR DISCUSSION

I feel it is best to instrument the M45 gcode with an option to specify the value for Z_MAX_POS and store this in EEPROM when specified (and initialized depending on MK3/MK3S printer type). This would allow users to use the standard compiled firmware and issue a XYZ calibaration by "just" entering a gcode like: 

M45 O2.0 
M500

to set the offset to 2.0mm (in EEPROM) Bondtech users need to store the eesteps anyhow to reflect the extruder reduction ratio.

Subsequent XYZ calibrations would then use the value stored in EEPROM. 

In my opinion this would be the preferred solution: no compilation needed, just some GCODE commands.

Before implementing this, please provide some feedback. Also, I feel that when EEPROM memory is used coordination with Prusa Dev is needed.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -2118,12 +2118,16 @@ bool calibrate_z_auto()
 	plan_buffer_line(destination[X_AXIS], destination[Y_AXIS], destination[Z_AXIS], destination[E_AXIS], feedrate / 60, active_extruder);
 	st_synchronize();
 	enable_endstops(endstops_enabled);
+#ifdef Z_MAX_POS_XYZ_CALIBRATION_CORRECTION
+    current_position[Z_AXIS] = Z_MAX_POS + Z_MAX_POS_XYZ_CALIBRATION_CORRECTION;
+#else   
 	if (PRINTER_TYPE == PRINTER_MK3) {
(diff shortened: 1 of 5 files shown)


toto99303:
I hope the Prusa guys are going to take this into consideration.

3d-gussner:
@ruedli Thanks for the idea and closing it in favor of https://github.com/prusa3d/Prusa-Firmware/pull/4687

PR #1742, opened 2019-04-16
Bed high frequency PWM disabled

XPila:


leptun:
Do you plan to change bed switching frequency in the future?

Issue #1196, opened 2018-09-25
MK3 3.4.0 Trying to exit Z adjust with a long-press prevents exiting at all


bozzzzo:
Turn on MK3. long press the button. Z adjust appears. long press the button again. cannot exit from the Z adjust anymore. probably related to #1195

stahlfabrik:
I have had the exact same issue happening on 3.4.1

Panayiotis-git:
Why I think that after adjusting the Z height, a simple press to the button exits the screen?

stahlfabrik:
@Panayiotis-git of course, if the bug is not in affect, that is how it SHOULD work.

When the bug happens, you can't leave the menu. You can only press reset.

Panayiotis-git:
Finally, I got it!

ultralcd.cpp, line 7297, replace the content of menu_lcd_longpress_func with:

    void menu_lcd_longpress_func(void)
    {
        if (menu_menu == lcd_move_z) {
            lcd_long_press_active = 0;
            return;
        }

        move_menu_scale = 1.0;
        menu_submenu(lcd_move_z);
    }

Panayiotis-git:
The commit 95b0506, changed the scope of the lcd_long_press_active, so the above fix is no longer valid.

Panayiotis-git:
So, new fix:
lcd.cpp, method void lcd_buttons_update(void), line 803
partially uncomment the else block:

           if (lcd_long_press_active == 0)
            { //button released before long press gets activated
                newbutton |= EN_C;
            }
            else lcd_quick_feedback();

maximlevitsky:
Still happens on 3.5.1. 100% reproducible.
Just do a long press to enter the Z axis mode and then another long press.

sgoll:
The same happens with recent firmware version 3.6.0.

It also happens when long-pressing the knob during a print. In that case, why does the Adjust Z Height show up anyway? Does this make sense, or will the next G-Code command just override any changes that I make? Should it be possible to enter Adjust Z Height only when the print is paused?

The printer does not react to any user commands until restarted, even when the Adjust Z Height dialog disappears eventually.

(In my case, I ended up there because I wanted to do Live Z Adjustment and wasn't sure how to reach that. I would have expected to go there by long-pressing during a print.)

ochm:
Yes, i found the same behavior in FW 3.6.0-2069 .
When the button is long pressed during print, it displays Z level change.
[...]

PR #1759, opened 2019-04-23
Default to "Live Adjust Z" when longpressing during a print (closes #1604)

### LINKED ISSUE #1604: Adjust long-press during print to go to Live Z Adjustment
This is related to my comment https://github.com/prusa3d/Prusa-Firmware/issues/1196#issuecomment-470453972

I accidentally long-pressed the knob during a print. In that case, the regular Adjust Z Height showed up. Does this make sense, or will the next G-Code command just override any changes that I make? Should it be possible to enter Adjust Z Height only when the print is paused?

I wanted to do Live Z Adjustment and wasn't sure how to reach that. I would have expected to go there by long-pressing the knob during a print.

wavexx:
Small but ergonomic change:

- Disable longpress action when homing or calibrating
- Default to Live Adjust Z during a print
- Retain "Move Z" otherwise

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -8252,8 +8252,22 @@ uint8_t get_message_level()
 
 void menu_lcd_longpress_func(void)
 {
-	move_menu_scale = 1.0;
-	menu_submenu(lcd_move_z);
+    if (homing_flag || mesh_bed_leveling_flag)
+    {
+        // disable longpress while homing or calibration
+        return;
+    }
+
+    if (moves_planned() || IS_SD_PRINTING || is_usb_printing)
+    {
+        lcd_clear();
+        menu_submenu(lcd_babystep_z);
+    }
+    else
+    {
+        move_menu_scale = 1.0;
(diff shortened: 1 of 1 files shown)


bozzzzo:
Does this also fix #1196 ?

wavexx:
No, I never noticed it before. I'll roll the fix up in this PR too. Thanks for letting me know

wavexx:
@bozzzzo  In the last commit I also fixed #1196. I also now do quick notification that the longpress is rejected (for example, during calibration - which is helpful as it was silent before).

leptun:
@wavexx Is this PR compiling on your PC? I am asking this since it failed travis because of network issues.

wavexx:
Yes it builds fine here. I noticed if failed, but AFAIK the only way to trigger travis as a contributor is  to push something. This happens so often with travis in all the projects where I contribute that I stopped doing this just for the sake of the green tick :/

DRracer:
@wavexx I can trigger the rebuild manually in Travis, stand by ;)

DRracer (approved):

PR #1760, opened 2019-04-23
Filament ramming and z improvements

### LINKED ISSUE #1095: Filament gets stuck during unloading
Hello,
  I have found that filament can easily get stuck in the hot end PTFE tubing during the unload process.  Occasionally a small blob breaks off in the hot end PTFE tubing requiring disassembly of the extruder.  I've tested with a short loading function prior to the unload, and this eliminates this issue entirely.  It would be great to have this short ramming function built in to the unload function to ensure an easy removal.  Thanks

### LINKED ISSUE #1748: Z raised above maximum when issuing "G28" (autohome) or filament autoload from maximum height
"G28" will home X, Y, then move the plate while raising Z before homing Z itself.

If you issue G28 from the condition in which the printer is already at max height, Z is just pushed against the Z-tops in two separate moves.

I assume that checking the current Z height is intentionally avoided while homing, but then this is one condition where I would expect crash detection on Z to work.

Is it possible to perform crash-detection on Z?

The same issue is true if you autoload a filament from a high Z. The filament autoload procedure will raise Z the first time, even if Z hasn't been homed yet.

wavexx:
This PR combines a fix for #1095 and #1748 due to space requirements (FIY this PR will also make room for PR #1759 and #1758). Please let me know if you whish to split this further.

First, we replace the filament unloading procedure with the mmu_filament_ramming() code to fix #1095.
If ramming is good enough for the mmu2, well, it *must* be good enough for plain unloading ;).
I didn't touch the custom filament unload procedure which is triggered by the filament runout, because *that* sequence could be triggered also by a clogged extruder and ramming the clog might not be a good idea. Ramming will thus be done for the Menu -> unload filament, for M702 and M600. I tested all cases with PLA and PETG.

I introduce a new function, raise_z_above to move Z carefully when the current position is potentially unknown, using stallguard. The following conditions will now use raise_z_above:

* during filament loading/unloading, for blob clearance. This was already done during loading, but now it's also done while unloading, since the ramming will extrude a little bit.

* extruder spacing when preheating. This is done to avoid buildplate marks on the PEI sheet, which is something I have, as I often preheat from the menu and sometimes come back late.

* before homing to avoid scratching the build plate. Again, this was already done, but badly (see #1748).

In the original sequence, G28 will raise, home XY, then raise again unconditionally. If the printer was already at max height, then reset, then a G28 is issued, the carriage is essentially crashed twice against the Z tops.

We now use raise_z_above before homing XY, and during home Z. raise_z_above is a no-op if the position is already above the target, which means that no move is ever issued if current position is already beyond the target. This results in G28 in a single raising move, which is faster to booth. If Z was already raised by a preheat or filament loading sequence, then there won't be any Z movement.

As a plus, stallguard is used when the position is unknown, making raise_z_above useful in all cases when Z is unknown (basically, all the above). raise_z_above will move the axis carefully and set the current position to max_height if the endpoint is hit. This will effectively prevent/fix #1748.

Note that #1748 will happen during regular printing too: if a tall print is completed and leaves the extruder at maximum height, a G28 issued by a subsequent print will currently crash the axis. A simple check is not enough: if the printer is reset while the extruder is at maximum height, then both homing and filament unloading will crash the axis.

### DIFF
--- a/Firmware/Configuration.h
+++ b/Firmware/Configuration.h
@@ -336,7 +336,7 @@ your extruder heater takes 2 minutes to hit the target on heating.
   #define Y_PROBE_OFFSET_FROM_EXTRUDER -29
   #define Z_PROBE_OFFSET_FROM_EXTRUDER -12.35
 
-  #define Z_RAISE_BEFORE_HOMING 4       // (in mm) Raise Z before homing (G28) for Probe Clearance.
+  #define Z_RAISE_BEFORE_HOMING 5       // (in mm) Raise Z before homing (G28) for Probe Clearance.
                                         // Be sure you have this distance over your Z_MAX_POS in case
 
   #define XY_TRAVEL_SPEED 8000         // X and Y axis travel speed between probes, in mm/min
@@ -536,6 +536,12 @@ enum CalibrationStatus
     // Currently the G86 sets the calibration status to 
     CALIBRATION_STATUS_UNKNOWN = 0,
 };
+
+// Try to maintain a minimum distance from the bed even when Z is
+// unknown when doing the following operations
+#define MIN_Z_FOR_LOAD    50
+#define MIN_Z_FOR_UNLOAD  20
+#define MIN_Z_FOR_PREHEAT 10
 
 #include "Configuration_adv.h"
 #include "thermistortables.h"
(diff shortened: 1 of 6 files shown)


wavexx:
No idea why the MK3 variant fails on travis, I tried to rebuild on my end without issues (I'm also currently checking the code on a MK3).

wavexx:
Closing in favor of #1768 which isolates only the Z movement improvements

Issue #1668, opened 2019-03-22
MMU2S loading to nozzle should raise the Z axis


Dutchy-79:
Firmware 3.7.0-RC1-2175 MK3S MMU2S 1.0.5

I loaded filament 1 to nozzle, It preheated and the started loading. Though it did not raise the Z axis and now there's a big blob after asking me if it's loaded okay.

I think the Z should go up just like normal load on the MK3

3d-gussner:
As a workaround try to move z up before you load filament.
But I agree that the initial mmu2 filament load should be like the MK3 standalone lifting z by 2mm or if it is blow 25mm it moves to z height 25mm.

Eddiiie:
Dupe of #1410  ?

PR #1767, opened 2019-04-25
Raise the Z axis before loading filament to the nozzle on MMU2S

NotaRobotexe:
Bugfix for https://github.com/prusa3d/Prusa-Firmware/issues/1668

### DIFF
--- a/Firmware/mmu.cpp
+++ b/Firmware/mmu.cpp
@@ -1367,6 +1367,7 @@ bFilamentAction=false;                            // NOT in "mmu_load_to_nozzle_
 	manage_response(true, true, MMU_TCODE_MOVE);
 	mmu_continue_loading(false);
 	mmu_extruder = tmp_extruder; //filament change is finished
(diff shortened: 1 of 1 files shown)


Firmware/mmu.cpp
```diff
@@ -1367,6 +1367,7 @@ bFilamentAction=false;                            // NOT in "mmu_load_to_nozzle_
 	manage_response(true, true, MMU_TCODE_MOVE);
 	mmu_continue_loading(false);
 	mmu_extruder = tmp_extruder; //filament change is finished
+	if (current_position[Z_AXIS] < 20) current_position[Z_AXIS] += 30;
```
mkbel, line 1370: Please add 
```
plan_buffer_line(...
st_synchronize();
```
calls after this line to make it more clear what is going on there.

mkbel: Or even better to avoid copy pasting code, find original occurrence of this functionality (lift nozzle when loading filament), extract the code to meaningfully named function and call this function from both places.

Issue #1410, opened 2018-12-18
MMU2: Load to Nozzle not lifting Z to a reasonable height before flooding nozzle


kagenokurei:
I have just built my MMU2 unit and upgraded my MK2S to MK2.5, so I'm not sure if this issue applies to MK3 users as well.

During my tests, I happened to do a "Load to Nozzle" command via the LCD + knob, while my nozzle is homed in all axes. My Z axis is actually lifted up to 0.35mm after homing since I don't want the heated nozzle sitting too close to my new steel sheet w/ PEI without moving for quite some time and end up making a pit mark at that spot. Due to this distance between the steel sheet and the nozzle, the Load to Nozzle functionality created a blob of plastic instead of just building pressure in the nozzle and possibly causing the gears to strip filament. I was able to lift my Z up further via OctoPrint and release more of the plastic and pressure building up due to the flooding of the nozzle.

I believe this would be a real issue in the event that the Z axis is really homed, which I believe doesn't go to 0, but rather sits at around 0.15mm off the steel sheet. I think that the Z axis should be lifted up a certain height before actually flooding the nozzle.

Mjankor:
I also see this when starting a single material print (3.5/1.03). Prior to the prime-line, the extruder extrudes a big blob as the MMU does its loading routine. Filament is already loaded to the nozzle.

mgg4:
I have confirmed this is also an issue with 3.7.0-RC1.  Load to nozzle doesn't move the print head to a safe place before extruding.  With the printer homed, it produces a blob of filament which has the potential to glom on to the nozzle and heater block.

I would expect this of a G-CODE command, which presumes that other codes would have been issued to move the head to a safe place.  But from the menus, I expect the printer to take whatever actions are necessary to put the head in a safe location.

JohnnyDeer:
This issue is fixed by mentioned pull request #1768 , I´m closing.

PR #1768, opened 2019-04-25
PFW-1052: "Unknown z" improvements

### LINKED ISSUE #1748: Z raised above maximum when issuing "G28" (autohome) or filament autoload from maximum height

wavexx:
This PR fixes #1748

I introduce a new function, raise_z_above to move Z carefully when the current position is potentially unknown, making it useful for several lcd functions which can be issued just after starting the printer.

raise_z_above uses stallguard when the Z position is unknown and a regular move otherwise. 
The following conditions will now use raise_z_above:

* during filament loading/unloading, for blob clearance.

* extruder spacing when preheating. This is done to avoid buildplate marks on the PEI sheet, which is something I have, as I often preheat from the menu and sometimes come back late.

* before homing to avoid scratching the build plate. This was already done, but incorrectly.

In the original homing sequence, G28 will raise Z, home XY, then raise again unconditionally. If the printer was already at max height, a G28 **will** crash twice against the Z tops. The same was true during filament unloading if the printer was reset and issued an unload from the LCD.

raise_z_above is a no-op if the position is already above the target, which avoids several useless moves also when homing the axes independently.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -2104,6 +2104,52 @@ bool check_commands() {
 	
 }
 
+
+// raise_z_above: slowly raise Z to the requested height
+//
+// contrarily to a simple move, this function will carefully plan a move
+// when the current Z position is unknown. In such cases, stallguard is
+// enabled and will prevent prolonged pushing against the Z tops
+void raise_z_above(float target, bool plan)
+{
+    if (current_position[Z_AXIS] >= target)
+        return;
+
+    // Z needs raising
+    current_position[Z_AXIS] = target;
+
+    if (axis_known_position[Z_AXIS])
+    {
+        // current position is known, it's safe to raise Z
+        if(plan) plan_buffer_line_curposXYZE(max_feedrate[Z_AXIS], active_extruder);
+        return;
+    }
+
+    // ensure Z is powered in normal mode to overcome initial load
+    enable_z();
+    st_synchronize();
+
+    // rely on crashguard to limit damage
+    bool z_endstop_enabled = enable_z_endstop(true);
+#ifdef TMC2130
+    tmc2130_home_enter(Z_AXIS_MASK);
+#endif //TMC2130
+    plan_buffer_line_curposXYZE(homing_feedrate[Z_AXIS] / 60, active_extruder);
+    st_synchronize();
+#ifdef TMC2130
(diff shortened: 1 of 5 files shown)


wavexx:
Now also fixes #1668 and #1410 (which seems identical, both pointing to mmu_lcd_load_to_nozzle), superseding pr #1767. I cannot verify the MMU2 behavior myself though.

wavexx:
@mkbel Noticed dad2065da623b1f1eb704de2223fcaed9b96c281 now conflicts with this. Since lcd_mmu_load_to_nozzle can probably be called without Z being homed yet, I assume the above still applies.

It doesn't raise while extruding since SG can be on, but I could definitely also include this option by combining both moves if Z is known, and doing them separately if not.

wavexx:
I updated the PR to combine raising and slow extrusion, as "rise_z" did.
The moves are only combined only when Z height is known.

Firmware/Marlin_main.cpp
```diff
@@ -3166,10 +3200,7 @@ void gcode_M701()
 		plan_buffer_line(current_position[X_AXIS], current_position[Y_AXIS], current_position[Z_AXIS], current_position[E_AXIS], 400 / 60, active_extruder); //fast sequence
 		st_synchronize();
 
-		marlin_rise_z();
-		current_position[E_AXIS] += 30;
```
leptun, line 3184: Why didn't you keep this line? With these changes my filament only loads to the nozzle and doesn't extrude any filament. I have to click NO and only then does it start extruding

wavexx:
On Fri, Jul 12 2019, leptun wrote:
> -		marlin_rise_z();
> -		current_position[E_AXIS] += 30;

Good thing you noticed. Probably shuffled during one of the rebases with
the following commit.

wavexx:
Updated

DRracer (approved):

PR #1771, opened 2019-04-26
build.sh: update Configuration_prusa.h when needed

wavexx:
I know PF-build.sh is the new hot build script, but build.sh is more convenient for testing.

- build.sh now accepts a first optional parameter to specify the build variant
- Always update Configuration_prusa.h when the variant file has changed in order to trigger a correct rebuild

We use cmp and not simply a timestamp check in order to handle *also* a rebuild for a different variant with a pre-existing build tree.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,6 +1,8 @@
 #!/bin/bash 
 BUILD_ENV="1.0.6"
 SCRIPT_PATH="$( cd "$(dirname "$0")" ; pwd -P )"
+DEFAULT_VARIANT="1_75mm_MK3-EINSy10a-E3Dv6full.h"
+VARIANT="${1:-$DEFAULT_VARIANT}"
 
 if [ ! -d "build-env" ]; then
     mkdir build-env || exit 1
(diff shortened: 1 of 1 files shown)


3d-gussner:
You can use PF-build.sh with arguments
 - variant
 - language
 - build variant

wavexx:
On Sat, May 18 2019, 3d-gussner wrote:
> You can use PF-build.sh with arguments
>
> * variant
> * language
> * build variant

I know, but I like the simplicity of build.sh.

3d-gussner:
@wavexx Hi, don't forget that build.sh is used to verify pull requests via travis so your changes don't break it :-)
Like the 'cmp' comparison BTW

wavexx:
> @wavexx Hi, don't forget that build.sh is used to verify pull requests via travis so your changes don't break it :-)

It's fully backward compatible, and I also expect easier to modify, as the default variant is well written at the top.

> Like the 'cmp' comparison BTW

Got bitten by this a few times before deciding to fix it ;)

wavexx:
I now also cleanup the build directory

DRracer:
@wavexx the Travis CI script first copies a variant_xxxx.h into Configuration_prusa.h and then runs the build.sh without arguments. With the proposed changes, the copied Configuration_prusa.h will get immediately overwritten by the default MK3 specified in the build.sh script
So unfortunately yes, this will break the Travis CI (just checked in the logs)

wavexx:
Ok, didn't expect this. Wouldn't it make sense to change travis to call build.sh with the variant instead of doing the copy manually? Feels cleaner from an outsider's approach.

DRracer:
@wavexx makes sense ... let's postpone this PR for a while, it is not really a blocker for 3.9. Please keep it open, we'll merge it later.

wavexx:
With cmake now working, I finally no longer need this. Let's torch it.

PR #1776, opened 2019-04-28
Remove an unused define

wavexx:
``TMC2130_UNLOAD_CURRENT_R`` has no use currently. It was effectively removed in commit 04d0178150cbe2cca89756663a67fbb57f939b5f.


wavexx:
``TMC2130_UNLOAD_CURRENT_R`` is now commented in the variant files.
I don't think there's any plan to resurrect it so I updated the PR to remove the comment as well, but I would be fine also closing the PR.

leptun:
@DRracer This PR could be closed since we've commented these out during the Hackathon
Oh, wait. I just noticed it's by @wavexx. Should we really remove them? I mean they are not used. Anyway, we'll let @DRracer decide

DRracer (approved):
ok, let's clean this up

Issue #1753, opened 2019-04-19
TMC2130 Stallguard TH HOMING XY not taken from config


NielschB:
in FW 3.7.0 on MK3, the homing stallguard thresholds for XY in file tmc2130.cpp are hardcoded instead of taken from configuration_prusa.h. See file tmc2130.cpp line 41, harcoded value 3. configuration_prusa.h has the homing threshold defined around line ~250.  (Would be nice to split the X,Y,Z homing SG thresholds in the config)

jenda122:
I think this should be made configurable not even at compile time, but in the menu/eeprom.

Additionally, I'd really appreciate some better diagnostics of the stallGuard than displaying the average value during selfcheck. The average may not tell you everything if there are sudden glitches (the minimum value would be more interesting), and it is also cumbersome to run the selfcheck just because of this. For example I would appreciate to get this diagnostics on every homing.

PR #1789, opened 2019-05-01
PFW-1053: Use Configuration_Prusa.h Sensorless Homing Values (Fix issue  #1753)

ingybing:
Use the value that have been defined in configuration_prusa.h file instead of the magic number?


DRracer:
closing in favour of #2346

DRracer:
Thanks for your PR ;)

PR #1800, opened 2019-05-06
return with exitcode 1 if any language not generated properly

XPila:
fix - include system_timer.h

### DIFF
--- a/lang/lang_en.txt
+++ b/lang/lang_en.txt
@@ -76,12 +76,6 @@
 #MSG_BED_LEVELING_FAILED_POINT_LOW c=20 r=4
 "Bed leveling failed. Sensor didnt trigger. Debris on nozzle? Waiting for reset."
 
-#MSG_BED_LEVELING_FAILED_PROBE_DISCONNECTED c=20 r=4
-"Bed leveling failed. Sensor disconnected or cable broken. Waiting for reset."
-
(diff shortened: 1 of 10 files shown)


lang/lang_en.txt
```diff
@@ -268,9 +244,6 @@
 #MSG_FSENSOR_ON c=0 r=0
 "Fil. sensor  [on]"
 
-#MSG_RESPONSE_POOR c=20 r=2
```
3d-gussner, line 271: Does that means that the PAT9125 and so the MK3 aren't supported anymore? or is that just the translation?

lang/lang_en.txt
```diff
@@ -703,12 +649,6 @@
 #
 "Print FAN"
 
-#WELCOME_MSG c=20 r=0
-"Prusa i3 MK2.5 ready."
```
3d-gussner, line 707: Why? Isn't that firmware also for MK2.5 and MK3 aswell for MK2.5S and MK3S?

lang/lang_en.txt
```diff
@@ -802,12 +739,6 @@
 #
 "Sensor state"
 
-#
-"Sensors info"
-
-#
-"Show pinda state"
```
3d-gussner, line 809: Why?

3d-gussner (commented):
PAT1925 Translation gone, why? No translation for MK2.5 and MK3 anymore?

PR #1803, opened 2019-05-06
check if firmware is on right type of printer

NotaRobotexe:


Firmware/Marlin_main.cpp
```diff
@@ -865,6 +865,29 @@ void show_fw_version_warnings() {
...
 }
 
+//try to check if firmware is on right type of printer 
+void check_if_fw_is_on_right_printer(){
+  #ifdef FILAMENT_SENSOR
+    swi2c_init();
+    uint8_t pat9125_detected = swi2c_readByte_A8(PAT9125_I2C_ADDR,0x00,NULL);
+    uint8_t ir_detected = !(PIN_GET(IR_SENSOR_PIN)); //will return 1 only if IR can detect filament in bondtech extruder so this may fail even when we have IR sensor
+    
+    if ((ir_detected) || (pat9125_detected)) {
```
mkbel, line 875: It looks like this condition is redundant.

PR #1804, opened 2019-05-07
PFW-688

ondratu:
* Deleted bad cols/rows definitions for translation.
* Check translation dictionary files to display definition

### DIFF
--- a/lang/lang-check.py
+++ b/lang/lang-check.py
@@ -1,0 +1,75 @@
+#!/usr/bin/env python3
+"""Check lang files."""
+from argparse import ArgumentParser
+from traceback import print_exc
+from sys import stderr
+
+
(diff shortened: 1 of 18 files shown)


lang/lang-check.py
```diff
@@ -0,0 +1,75 @@
...
+            if cols is None and rows is None:
+                if not no_warning:
+                    print("[W]: No display definition on line %d" % lines)
+                cols = len(translation)     # propably fullscreen
+            if rows is None:
+                rows = 1
+
+            if len(translation) > cols*rows:
+                stderr.write(
+                    "[E]: Text %s is longer then definiton on line %d\n" %
```
3d-gussner, line 43: small typo. it should be 'is longer than definiton on line' or 'is longer than the definiton on line'

PR #1805, opened 2019-05-07
PFW-844

DRracer:


Firmware/ultralcd.cpp
```diff
@@ -1773,6 +1774,7 @@ void lcd_commands()
 			pid_tuning_finished = false;
 			custom_message_state = 0;
 			lcd_setstatuspgm(_i("PID cal. finished"));////MSG_PID_FINISHED c=20 r=1
+			setTargetHotend(0, 0);  // reset the target hotend temperature including the number displayed on the main screen
```
mkbel: Use setAllTargetHotends(0) instead to be more safe for custom mods.

DRracer: Are you sure we want to disable all hotends after PID calibration and not just the one we were calibrating?

PR #1808, opened 2019-05-08
Update ultralcd.cpp

JoePrints:
Change loading bed temps to 0, no need to heat bed for filament load.


3d-gussner:
@JoePrints Thanks. Closing as we have an option to not preheat the bed in current version.

PR #1809, opened 2019-05-09
 Fix a number overflow to last line of LCD when doing large extruder move

DRracer:


mkbel:
Pull request title doesn't correspond to included commits.

PR #1824, opened 2019-05-13
rev-4-mk3-with-Finda-at-extruder-w/-HyperFine-G80

thoffpau0:


3d-gussner:
As you might know i like the Hyperfine Bed Leveling on the MK2/s, but I don't think that this PR as it is will be merged:
- way too many unnecessary changes have been generated by your editor...It makes seeing the essential changes very difficult.
- You changed in `Configuration.h` `#define FW_DEV_VERSION, #define FW_REPOSITORY and #define STRING_CONFIG_H_AUTHOR`
- The `Configuration_prusa.h` should not be part of the change as it is a copy from the variants
- in `eeprom.h` you "disable"/reuse the `FARM_MODE and FARM_NUMBER`. This alone is a no go  for this PR. If somebody use `G98 and G99` it will cause issues.
- You added other changes to the same PR see `fsensor.cpp/.h and mmu.cpp`
- You should also think about the language files.

I am quite happy with the 7x7 MBL results at this moment, could you see improvements with Hyper Fine Bed Leveling? If so can you provide some pictures stock 7x7 vs 7x7 + Hyperfine Bed Leveling?

If there is significant improvement I may try it my self on the MK3s and discuss the outcome with several people.

3d-gussner:
BTW thanks for the credits BUT these are going to [@jeffjordan and PJR](https://github.com/3d-gussner/Prusa-Firmware/releases?after=v3.1.4-RC1)

3d-gussner:
@thoffpau0 Thanks, you know i like the PR but too many files been changed and impossible to review actual changes. Closing

PR #1827, opened 2019-05-13
Disable menu on MINTEMP errors

DRracer:
Prevent the user from entering the main menu when there is a MINTEMP error. When the error fades out, the user must reset the printer to regain control.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3091,7 +3091,7 @@ static void gcode_M600(bool automatic, float x_position, float y_position, float
         plan_buffer_line(current_position[X_AXIS], current_position[Y_AXIS], current_position[Z_AXIS],
                 current_position[E_AXIS], FILAMENTCHANGE_EXFEED, active_extruder);
     }
-
+	
     //Move XY back
     plan_buffer_line(lastpos[X_AXIS], lastpos[Y_AXIS], current_position[Z_AXIS], current_position[E_AXIS],
             FILAMENTCHANGE_XYFEED, active_extruder);
@@ -3526,9 +3526,13 @@ void process_commands()
                enquecommand_P(PSTR("M24")); 
 		}	
 #ifdef FILAMENT_SENSOR
+		else if (code_seen("fsensor_recover_IR")) //! PRUSA fsensor_recover_IR
+		{
+			fsensor_restore_print_and_continue_IR();
+		}
 		else if (code_seen("fsensor_recover")) //! PRUSA fsensor_recover
 		{
-               fsensor_restore_print_and_continue();
+			fsensor_restore_print_and_continue();
 		}	
 #endif //FILAMENT_SENSOR
 		else if (code_seen("MMURES")) //! PRUSA MMURES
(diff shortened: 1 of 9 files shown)


Firmware/temperature.cpp
```diff
@@ -2021,7 +2090,16 @@ void check_min_temp_heater0()
...
+		menu_set_serious_error(SERIOUS_ERR_MINTEMP_HEATER);	
 		min_temp_error(0);
+	} else if( menu_is_serious_error(SERIOUS_ERR_MINTEMP_HEATER) ) {
+		// no recovery, just force the user to restart the printer
+		// which is a safer variant than just continuing printing
+		// The automaton also checks for hysteresis - the temperature must have reached a few degrees above the MINTEMP, before
+		// we shall signalize, that MINTEMP has been fixed
+		// Code notice: normally the aam instance would have been placed here as static alert_automaton_mintemp aam, but
+		// due to stupid compiler that takes 16 more bytes.
+		aam[0].step(current_temperature[0], minttemp[0] + 50); // @@TODO tady pak dat TEMP_HYSTERESIS
```
mkbel, line 2102: @@TODO, it looks like something has been forgotten here.

Firmware/temperature.cpp
```diff
@@ -2013,6 +2035,53 @@ void check_max_temp()
 
 }
 
+//! number of repeating the same state with consecutive step() calls
+//! used to slow down text switching
+struct alert_automaton_mintemp {
+	enum { ALERT_AUTOMATON_SPEED_DIV = 5 };
```
mkbel, line 2041: Doesn't need to be public, as it is never accessed from the outside.

Firmware/temperature.cpp
```diff
@@ -2013,6 +2035,53 @@ void check_max_temp()
...
+	uint8_t state, repeat = ALERT_AUTOMATON_SPEED_DIV;
```
mkbel, line 2042: Doesn't need to be public, as it is never accessed from the outside.

Firmware/temperature.cpp
```diff
@@ -2013,6 +2035,53 @@ void check_max_temp()
...
+	
+	void substep(uint8_t next_state){
```
mkbel, line 2044: Doesn't need to be public, as it is never accessed from the outside.

Firmware/temperature.cpp
```diff
@@ -2013,6 +2035,53 @@ void check_max_temp()
...
+		} else {
+			--repeat;
+		}
+	}
+	
+	void step(float current_temp, float mintemp){
+		static const char m2[] PROGMEM = "MINTEMP fixed";
+		static const char m1[] PROGMEM = "Please restart";
+		switch(state){
+		case 0: // initial state - check hysteresis
```
mkbel, line 2057: Use meaningful names for states instead of comments. e.g. enum class

mkbel:
Contains unrelated work - commit Fix for resume printing after filament runout. The previous

Firmware/Marlin_main.cpp
```diff
@@ -1,9645 +1,9649 @@
-/* -*- c++ -*- */
```
mkbel, line 1: It looks like you have changed 19294 lines in Marlin_main.cpp this is probably not intended, check line endings CR+LF vs LF

DRracer:
superseeded by another pull request, closing

PR #1832, opened 2019-05-14
Refactor filament menu

mkbel:
to conserve flash usage. 350B saved.


DRracer (approved):
Ok. German translation may be: Auswerfen Fil. x

PR #1838, opened 2019-05-15
Windows subsystem CRLF issues fix

OskarLeirvaag:
Add extra tip to linux sbsystem, where users get the error "$'\r': command not found" where windows uses CRLF and linux uses only LF


mkbel (approved):

mkbel:
Thank you for your contribution/

Issue #1775, opened 2019-04-28
[Question] Why 15°C for min temp?


tobymurray:
Is the 15°C threshold based on anything? E.g. temperature sensor specs or practical experience? 

I regularly hit the `Err: MINTEMP` in the room my printer sits in, as the nozzle temperature appears to rest ~5°C below ambient (I have a copper heat block, maybe that makes a difference?). I've otherwise had no issues with my thermistor, and I have thousands of hours of prints with it so I would be surprised if the sensor itself was faulty. My current workaround is to use a portable heater to preheat the printer until it gets above 15°C, then carry on. I've recently become fed up with this as the nozzle temperature dropping to 15°C has caused print failures due to cooldown after a jam has been detected or something.

I'm considering modifying the configuration in the firmware to lower it to something that seems more reasonable to me from what I've observed experimentally, but I don't want to do it if the 15°C has some safety significance.

JustAnother1:
The safety significance is, as far as I know, to detect sensor failures. Such as short circuit and disconnected. So Check the temperatures that your settings create if the sensor is disconnected and if you short circuit the sensor. One value should trigger max temp error the other value should trigger min temp error. As long as that still works there should be no issue to adopt the values as long as these two hold true.

tobymurray:
I just tried it out - a short sets off the MAXTEMP alarm and displays 739°C. An open circuit shows 0°C (and the MINTEMP). Looks like I'll be lowering the threshold then!

nevercast:
Are you saying you can't start printing when the ambient temperature is under 15C? I'm running V3.7.1 firmware on my MK3S and the ambient was 7°C when I started printing today. I did run a preheat though, perhaps preheat ignores the min temp that you are talking about? When does MINTEMP occur?

tobymurray:
Addressed in https://github.com/prusa3d/Prusa-Firmware/pull/1855

PR #1855, opened 2019-05-21
Drop min nozzle temperature to 5 degrees on MK3S

tobymurray:
Follow up to #1775. My hot end thermistor generally reads ~5° below ambient. It is regularly 15° - 25° ambient temperature in my house (where I use the MK3), so this means I regularly get the MINTEMP error when the printer is perfectly fine. Normally this is just an annoyance and I have a small heater I use to preheat the machine, but with MMU prints I am now regularly failing prints while the printer waits for me to come sort out why the filament isn't loading.

I'm not sure if having a [plated copper heat block](https://e3d-online.com/v6-plated-copper-heater-block) is impacting it, but everything else seems to work fine to me. I have ~5k hours on the printer with the thermistor and it doesn't seem to otherwise have issues.

From a safety perspective, I tried shorting the sensor (read over 700° and set off an alarm), I tried opening the circuit and it showed 0°. I'm not sure why this value is higher than the others so I'm totally open to this being a conscious decision that I wasn't aware of.

### DIFF
--- a/Firmware/variants/1_75mm_MK3-EINSy10a-E3Dv6full.h
+++ b/Firmware/variants/1_75mm_MK3-EINSy10a-E3Dv6full.h
@@ -280,7 +280,7 @@
  *------------------------------------*/
 
 // Mintemps
-#define HEATER_0_MINTEMP 15
+#define HEATER_0_MINTEMP 5
 #define HEATER_1_MINTEMP 5
 #define HEATER_2_MINTEMP 5
 #define HEATER_MINTEMP_DELAY 15000                // [ms] ! if changed, check maximal allowed value @ ShortTimer
--- a/Firmware/variants/1_75mm_MK3S-EINSy10a-E3Dv6full.h
+++ b/Firmware/variants/1_75mm_MK3S-EINSy10a-E3Dv6full.h
@@ -282,7 +282,7 @@
  *------------------------------------*/
 
 // Mintemps
-#define HEATER_0_MINTEMP 15
+#define HEATER_0_MINTEMP 5
 #define HEATER_1_MINTEMP 5
 #define HEATER_2_MINTEMP 5
 #define HEATER_MINTEMP_DELAY 15000                // [ms] ! if changed, check maximal allowed value @ ShortTimer


rennne (approved):
Since an appropriate test was done: the code looks fine too me

nevercast:
Is this of any concern, @tobymurray? https://github.com/prusa3d/Prusa-Firmware/blob/6aface42d8ff4ff2280d3b64f1c680560db8f074/Firmware/variants/1_75mm_MK3S-EINSy10a-E3Dv6full.h#L144-L147

> this is because 15C is end of scale for both used thermistors (bed, heater)

Perhaps the MK3S uses different thermistors. 

Update: I traced that one define all the way back to Initial Commit, rip. 
https://github.com/prusa3d/Prusa-Firmware/blob/30f0528aba00e5949496f4fe7adcbef5b5a18b5c/Firmware/variants/1_75mm_MK2-RAMBo10a-E3Dv6full.h#L70

No idea why it is the value it is.

tobymurray:
@nevercast - the short answer is "I don't know". The comment is just English-y enough to seem concerning ("15C is end of scale for both used thermistors (bed, heater)"), but I don't fully understand it. The bed thermistor has a value of 5°C for the min, so either the comment is wrong or the existing code is wrong. It seems to be saying there is an ambient temperature sensor that is actually _on_ the Einsy, I was not aware of this if there is. My understanding was that the heat bed temperature was determined by the heat bed thermistor, the nozzle temperature was determined by the thermistor in the hot end (what I was attempting to change here), and then there was a third thermistor in the P.I.N.D.A. itself which was used for temperature compensation. 

The behavior I've observed is that my nozzle temperature is _lower_ than ambient (as measured by the bed temperature or by an external thermometer), which is what prompted this change. I (unfortunately) never had a Prusa before my MK3 (now MK3S), so I would have to look at the history to see if the determination of ambient temperature has changed through the models. Looking at the current Einsy documentation, I don't see a reference to an on-board thermistor, so I don't know what the comment is referring to.

I have not seen the behavior described in the comment - i.e. that there is some temperature being taken within the Einsy case and that temperature is higher than ambient.

nevercast:
I'm pretty sure if I go to Support -> Temperatures on my MK3S it gives me four temperatures, perhaps you want to try that yourself too?

If I recall from when I was debugging Temperature Runaway errors, the temperatures were 
Nozzle: ~200
Bed: ~60
Ambient: ~26
PINDA: ~17

Which would indicate that there may be a 10 degree offset on ambient, but it also may be 10 degrees warmer on the EINSY than ambient, since it has high current FETs

tobymurray:
Yeah, I'm definitely seeing that there is a 4th temperature sensor - I have an ambient temperature and it is a little over 10°C higher than e.g. the P.I.N.D.A. (which seems to be much closer to the actual ambient temperature). Any idea physically where it is? I can look at my board at home, but I'm in the middle of a long print. 

My understanding was that `HEATER_0_MINTEMP` corresponded to the nozzle temperature, not to the ambient temperature sensor. I'm not familiar with Marlin, so I don't really know where to look to say definitively what the mapping is. If my proposed change impacts the ambient temperature sensor, it's not intended. Experimentally, it doesn't _seem_ like it does - I've been printing with this modified firmware and the nozzle temperature no longer correlates with MINTEMP errors (although I have not encountered any MINTEMP errors).

tobymurray:
Ok, taking this back to the start I see that the ambient temperature sensor is baked right into the board, PF6 in the Heater Outputs diagram in the [Einsy schematic](https://github.com/ultimachine/Einsy-Rambo/blob/1.1a/board/Project%20Outputs/Schematic%20Prints_Einsy%20Rambo_1.1a.PDF). So at least now I understand where that temperature is coming from!

I'm assuming my thermistor is just slightly out of calibration, so it's reading colder than ambient. While I could replace it, I still want to address being able to print in the winter when it legitimately drops below 15°C at some points (e.g. overnight if the printer sensed an error, now the MINTEMP makes the print irrecoverable).

With that said, I was getting all turned around - I'll dump my understanding here in hopes I can keep it straight in the future. Working back from the UI (e.g. [here](https://github.com/prusa3d/Prusa-Firmware/blob/d3dd0ad271f28f3540eafdae583127c8926884ad/Firmware/temperature.cpp#L2099)) it looks like the MINTEMP error I'm seeing can be triggered by either `HEATER_0` or the bed. Ignoring the bed for now because it's unrelated to the proposed change, this MINTEMP is related to `HEATER_0`. The EinsyRambo documentation shows what it considers the 3 thermistors [here](https://reprap.org/wiki/File:EinsyRambo1.1a-connections.jpg). According to the [Prusa assembly instructions](https://manual.prusa3d.com/Guide/8.+Electronics+assembly/1052?lang=en#s18475), T0 is the extruder and T2 is the heatbed. T1 is left unattached to anything. If you look at the [EinsyRambo pinout](https://www.reprap.org/wiki/EinsyRambo_development) it shows that PE5 goes to "Heat 0". This is mapped in the Einsy's [Marlin configuration](https://github.com/prusa3d/Prusa-Firmware/blob/327e6cd39bcbac9f0ee497daf167575a9e181370/Firmware/pins_Einsy_1_0.h#L65) to `HEATER_0_PIN` - so this is all consistent. `HEATER_0` is the nozzle, and the nozzle is set to 15°C for some reason.

> this is because 15C is end of scale for both used thermistors (bed, heater)

Looking at the [MK3S configuration](https://github.com/prusa3d/Prusa-Firmware/blob/84ac20922262d6d2a8f14334106adf2feefa06e5/Firmware/variants/1_75mm_MK3S-EINSy10a-E3Dv6full.h#L561) (what I changed), the temperature sensor is allegedly an [ATC Semitec 104GT-2](https://github.com/prusa3d/Prusa-Firmware/blob/84ac20922262d6d2a8f14334106adf2feefa06e5/Firmware/variants/1_75mm_MK3S-EINSy10a-E3Dv6full.h#L530). That specific sensor looks to be obsolete now, but the [datasheet for it](https://www.mouser.com/datasheet/2/362/semitec%20usa%20corporation_gtthermistor-1202937.pdf) lists the "Category temp. range" from "-50~+300". I'm by no means a thermistor-ologist, so I welcome any insight, but it doesn't look like there's anything about the specific sensor that leads it to unusable in the 5°C - 15°C range. 

To me it seems like any limitations like that would have to be on how the thermistor is read. The function to read the `HEATER_0` (extruder) temperature looks to be [here](https://github.com/prusa3d/Prusa-Firmware/blob/d3dd0ad271f28f3540eafdae583127c8926884ad/Firmware/temperature.cpp#L903) which uses [this temperature table](https://github.com/prusa3d/Prusa-Firmware/blob/6940b13507b7bb55c23a7f49a6c3dc514b016416/Firmware/thermistortables.h#L173). This is mapped from 0°C - 350°C, but the lowest 3 temperatures are:

```
   {978*OVERSAMPLENR, 22},
   {1008*OVERSAMPLENR, 3},
   {1023*OVERSAMPLENR, 0}
```
I'm thinking that's where the 15°C comes from - it's a pseudo-arbitrary value that's not at the very end of the temperature scale. It looks like there's plenty of resolution at the low temperature end of the thermistor, so the thermistor should be particularly sensitive to the difference between 3°C and 5°C compared to 303°C and 305°C. With that said, I don't think 15°C would be any "safer" than 5°C in this case. Both fall into the same interval between explicitly mapped temperatures. I'm honestly not even really sure why it would be higher than 0°C. So far as I'm aware this is protecting against an open circuit, so the chances of a failure such that the resistance is perfectly not-high-enough to overflow but so high that it registers in the single digit temperatures seems _extremely_ unlikely. I'm quite open to having fundamentally misunderstood something, but so far as I can tell this is a safe change because both the thermistor and the firmware are similarly capable of sensing 5°C as they are 15°C

Disclaimer: this is my first time looking at the Marlin firmware at all, and I am absolutely not fluent in electronics.

tobymurray:
Out of curiosity, I graphed the temperature table. The part that is relevant for this change is in the upper left. Visually it's easier to see that a small change in the temperature (x axis) has a much larger change in the reading (y axis) at the lower end of the scale compared to the higher end of the scale.
image

rhys100:
who can be the 2nd reviewer?

leptun:
I think there are some board differences between mk2 and mk3 that affect the temperature mintemp. On mk2 you can’t get lower than 14*C even with thermistor disconnected. I think that’s why 15*C was chosen as mintemp and was just kept on the new board.

tobymurray:
That seems reasonable to me, unplugging the thermistor on my MK3 showed 0°C and it doesn't waver.

nevercast:
Can anyone offer testing of this on a MK2, per @leptun comment? I only have a MK3.

leptun:
@nevercast https://photos.app.goo.gl/hPQuVAoJn9kLx2SL9

Here is video on mk2 board with no thermistor connected

nevercast:
Thanks @leptun, that confirms that the sensor cannot operate below 15C on the MK2, but seemingly not the case on the MK3, is that right, @tobymurray ?

Because of this [comment](https://github.com/prusa3d/Prusa-Firmware/issues/2006#issuecomment-512181182) I'd like @michalxfanta to weigh in with his opinion too. Why is the minimum operating temperature 18C? 

I personally have had successful prints at 7C room temperature without an extruder sock on my MK3S.

leptun:
I don’t think the mimimum operating temperature has anything to do with hotend or bed temperature handling, but with the printing material itself. Idk. I think you could go as low as 0deg environment. The only problem I see is that the top surface of the bed might be colder than usual and result in warping. As for the electronics, i think they would only benefit from a lower ambient temperature as they get quite hot (Especially the extruder motor).

Edit: what I said about fillament only applies to PLA. The other materials can’t be printed at these temperatures at all, especially ABS and PC

nevercast:
The bed is heated. Ambient doesn't matter. As far as I know, as long as the extruder temperature and bed temperature can be sustained, the ambient temperature will have little effect on the print. This is my experience.

This issue is about the nozzle temperature being less than 5 degrees when the print tries to start, which shouldn't matter as long as it can be heated to 220C and sustained. Right?

tobymurray:
> but seemingly not the case on the MK3, is that right, @tobymurray ?

Correct - I have a print going for the next ~10 hours right now, but I can provide a video if need be. My MK3 (I can pull the board revision at the same time) shows 0°C in the same circumstance as @leptun showed for the MK2.

nevercast:
After thinking about it a bit more, I suspect this test is to flag a bad sensor reading, if the temp is below what is minimum, then the temperature sensor must be faulty. 

However, since the MK3S temperature probe does give readings below 15C, I do think it needs to be changed to whatever its real minimum is. @tobymurray Can you get the probe to go all the way down to 0C? Will it go negative?

leptun:
@nevercast I think the value is constrained to the positive range.

tobymurray:
> Can you get the probe to go all the way down to 0C? Will it go negative?

Disconnecting the sensor on the MK3 (now MK3S), i.e. opening the circuit for the sensor, sends the temperature to 0°C. It doesn't waver, it just sits at 0°C. If the software would allow it to go negative, I'm not readily able to test that now. We should be able to test that pretty easily - I linked the method that calculates the value above. To test it practically I'd have to disassemble pieces and put them in the freezer or something - it's summer here, so there's no negative ambient temperature readily available.

nevercast:
I could freeze spray my own MK3S, but if it reads 0C when unplugged, I'd say that is sufficient evidence. Then perhaps the real limit would be 1 or 2C giving some tolerance for error. 5C should be fine. 

I'd like to see this merged with the setting at 5C for the MK3S.

nevercast (approved):
Because I believe that the temperature sensor has been sufficiently tested on the MK3S, and that 5C is not out of bounds nor an error value of the sensor. I approve this change to allow the minimum temperature of the nozzle to be 5 degrees so that prints do not stall when waiting for a filament change or otherwise.

It is my understanding that the min and max temps are to indicate a sensor/reading error, and sufficient margin has been provided to be safe.

leptun:
@tobymurray I suggest you also change the value for the Mk3, not only Mk3S.

tobymurray:
@leptun - I've updated the value for the MK3 as well in 14351aa.

leptun:
https://github.com/johnnyr/Mini-Rambo/issues/8
Take a look at this. 😐

That's why the miniRambo is showing 14deg at idle with no thermistor connected. If only they knew that in time...

tobymurray:
Is there anything left preventing this from being merged? I'm anxiously awaiting not having to compile my own firmware.

tobymurray:
@leptun - any thoughts on what happens with this now? I was hoping this would be an uncontentious change after we got to the bottom of it - anything I can do to help it get merged? 

Any chance it can get targeted to 3.9?

leptun:
@toburray I’ll look into the safety aspects of it this Monday. If all tests pass successfully, then I’ll propose this further. It should be ok in terms of hardware support. I’ll check this PR for formatting errors when I get to a PC.

tobymurray:
> I’ll look into the safety aspects of it this Monday. If all tests pass successfully, then I’ll propose this further.

@leptun - were you able to get around to this? It looks like 3.9 is taking shape but this hasn't been flagged for that release. As winter is coming where I am this is increasingly useful for me.

leptun:
@tobymurray Sorry for not answering. It's been forwarded. I'll let you know if there are any news about it.

tobymurray:
Is there any way to get more visibility for this issue? Are there any outstanding concerns, or is it just a matter of prioritization? 

It seems like it should be trivial to validate, and it causes giant headaches for people with MMU units. It's been 6 months wait for a change to these two number values that look to be well understood given the discussion above. Many users are using a hair dryer or heat gun or similar silly workaround to preheat their printer, just so they can get by the inappropriate MINTEMP error (not at all related to actual faulty or inconsistent thermistor readings).

As an example, here are 3 different recent Reddit threads of people running into this (not a comprehensive list, just been popping up):

- https://www.reddit.com/r/prusa3d/comments/egb8tg/can_i_disable_mintemp_on_the_mk3s/
- https://www.reddit.com/r/prusa3d/comments/eq7hqg/is_there_some_way_to_change_the_mintemp_bed_error/
- https://www.reddit.com/r/prusa3d/comments/emv186/mintemp_error_while_garage_printing_and_custom/

I know Prusa is doing lots of stuff, and I love most of it, but this is a big QoL issue that is the only reason I have to build my own firmware after every release. I hate to think of the amount of frustration this issue has caused to the many Prusa owners (MMU owners in particular, see issue #2299).

DRracer:
@tobymurray approved after internal discussion, thank you for your PR.
This PR is now planned for the 3.9 RC

tobymurray:
Cold climate dwellers and garage printers alike rejoice, thanks for taking a look at it!

DRracer:
@tobymurray after a series of measurements and internal evaluation we decided to lower the MINTEMP to 10°C for MK3 and MK3S. Hopefully it solves your problem too.
This is due to the resolution of the thermistor (and the span of tolerances) / AD, please see #2470 .

Would you mind closing this PR in favor of #2470? Don't worry, your contribution won't be forgotten ;) .

PR #1856, opened 2019-05-21
pause printer by flag

ondratu:


Firmware/Marlin_main.cpp
```diff
@@ -156,6 +156,9 @@
 //=============================imported variables============================
 //===========================================================================
 
+#ifdef FANCHECK
+extern volatile bool fan_check_error;
```
mkbel: This should go into header file.

ondratu: done

mkbel (approved):

PR #1875, opened 2019-05-27
Added thermistor table 67

tafypz:
Added thermistor table 67 for Slice engineering 450 C thermistor. Table was retrieved from the marlin bugfix-2.0.x branch and edited to match prusa3d firmware syntax


SiwatS (commented):
there are some un necessary diff on the file

tafypz:
> there are some un necessary diff on the file

All these diffs were trailing white spaces that I thought I would clean up to be consistent with the rest. I can add the trailing white spaces if necessary.

PR #1888, opened 2019-05-30
Always cut independent

mkbel:
This is experimental feature requested by our test department. There is no known use for ordinary user. If enabled by MMU_ALWAYS_CUT preprocessor macro (not enabled by default) and enabled from printer menu (not enabled by default). It cuts filament every time when switching filament from gcode.


DRracer (approved):
I agree with merging this experimental piece of code (guarded with #ifdefs as proposed). It may come handy in future testing sessions.

PR #1889, opened 2019-05-31
Fan check error fix

DRracer:
When the user didn't resolve the fan error and tried to resume printing, the printer could get into an undefined state.

### DIFF
--- a/Firmware/pins_Einsy_1_0.h
+++ b/Firmware/pins_Einsy_1_0.h
@@ -99,8 +99,7 @@
 
 //#define KILL_PIN            32
 
-// LCD backlight pin may interfere with something, this is yet to be found out correctly
-#define LCD_BL_PIN          5   //backlight control pin
+//#define LCD_BL_PIN          5   //backlight control pin
 #define BEEPER              84  // Beeper on AUX-4
 #define LCD_PINS_RS         82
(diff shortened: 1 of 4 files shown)


Firmware/pins_Einsy_1_0.h
```diff
@@ -99,7 +99,8 @@
 
 //#define KILL_PIN            32
 
-//#define LCD_BL_PIN          5   //backlight control pin
+// LCD backlight pin may interfere with something, this is yet to be found out correctly
+#define LCD_BL_PIN          5   //backlight control pin
```
mkbel: Is not related to fan check error fix

mkbel (changes requested):

Firmware/pins_Einsy_1_0.h
DRracer: reverted...

Firmware/temperature.cpp
```diff
@@ -522,6 +528,19 @@ void checkFanSpeed()
 	}
 }
 
+void fanSpeedErrorBeep(const char *serialMsg, const char *lcdMsg){
```
mkbel: It doesn't need to be global function.

mkbel (approved):

PR #1893, opened 2019-06-05
Filament runout mk3

DRracer:


Firmware/pins_Einsy_1_0.h
```diff
@@ -99,7 +99,8 @@
 
 //#define KILL_PIN            32
 
-//#define LCD_BL_PIN          5   //backlight control pin
+// LCD backlight pin may interfere with something, this is yet to be found out correctly
+#define LCD_BL_PIN          5   //backlight control pin
```
mkbel: It looks like it has nothing to do with filament runout.

Firmware/Marlin_main.cpp
```diff
@@ -3553,10 +3553,6 @@ void process_commands()
                enquecommand_P(PSTR("M24")); 
 		}	
 #ifdef FILAMENT_SENSOR
-		else if (code_seen("fsensor_recover_IR")) //! PRUSA fsensor_recover_IR
-		{
-			fsensor_restore_print_and_continue_IR();
```
mkbel, line 3558: Prototype of non-existent function extern void fsensor_restore_print_and_continue_IR(void);
not removed from fsensor.h

Firmware/Marlin_main.cpp
```diff
@@ -6628,8 +6624,13 @@ if((eSoundMode==e_SOUND_MODE_LOUD)||(eSoundMode==e_SOUND_MODE_ONCE))
...
           #ifdef FILAMENTCHANGE_ZADD
-            z_shift= FILAMENTCHANGE_ZADD ;
-            if(current_position[Z_AXIS] < 25) z_shift+= 25 ;
+			static_assert(Z_MAX_POS < (255 - FILAMENTCHANGE_ZADD), "Z-range too high, change the following code from uint8_t to uint16_t");
+			// avoid floating point arithmetics when not necessary - results in shorter code
+			uint8_t ztmp = uint8_t( current_position[Z_AXIS] );
+			if(ztmp < uint8_t(25)){
+				z_shift = uint8_t(25) - ztmp; // make sure to be at least 25mm above the heat bed
+			}
+			z_shift += FILAMENTCHANGE_ZADD ; // always move above printout
```
mkbel: It would be nice to extract this snipped into function, so it can be checked, if it does what is intended.

mkbel:
File pins Ensy is still marked as changed because of its mode.

PR #1899, opened 2019-06-05
PFW-1054: Fix the LCD pause function

wavexx:
Do not call long_pause recursively!

long_pause() is called before resetting the lcd_command_type.  As long_pause uses st_synchronize() internally, there could be time to schedule another call to long_pause() within itself.

I'm actually wondering why this doesn't crash as often as it should!

We fix this by calling long_pause exactly after saving. Then we also remove an useless check and bogus comment (no: the command type cannot be changed during long_pause!).

We modify long_pause to schedule the moves so that the LCD is immediately usable, and also embed the heater control inside long_pause() itself, where it makes more sense. lcd_pause_print() is the only caller anyway.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -9492,10 +9492,9 @@ void long_pause() //long pause print
 	current_position[Y_AXIS] = Y_PAUSE_POS;
 	plan_buffer_line_curposXYZE(50, active_extruder);
 
-	// Turn off the print fan
+	// Turn off the hotends and print fan
+    setAllTargetHotends(0);
 	fanSpeed = 0;
-
-	st_synchronize();
 }
 
 void serialecho_temperatures() {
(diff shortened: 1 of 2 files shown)


thess:
This looks like it might fix a problem I encountered when resuming a print after an MMU load/unload failure was cleared. Namely, after the lcd button is clicked to resume and the nozzle re-heats, the display rapidly clears and re-draws until you click the button to toggle menu and back to status. Miraculously, things resumed and the display was stable.

I will test this in a custom firmware build and let you know if the problem I saw still persists.

wavexx:
It's a nasty one which was quite annoying to track down. I was able to find a setup that reliably caused the pause function to break while testing LA1.5

wavexx:
Ping @DRracer since we're talking LCD stuff. That's an easy one.

DRracer:
@wavexx thanks for pinging me ;) , we know about this PR, its planned to get merged.

DRracer (approved):

Issue #1898, opened 2019-06-05
M1 fails in many situations


GWdd:
Adding M1 to a gcode seems to almost never work; it is extremely position dependent; and having more than one can corrupt later insertions that will normally work.

For example this segment of code pauses, but displays nothing (filename is present)
```
M115 U3.7.1 ; tell printer latest fw version

G91
G0 X5 Y5 Z.1 F1000
M1 FIRST BASE
```

However this segment from a larger piece of code works (sometimes)
```
G0 X0 Y200 Z100 F3000 ; this is a good waiting point for bed flip

M1 Flip sheet then press OK 
```

And if the M1 First Base command is in the sequence, the second M1 fails to produce text.
```
M115 U3.7.1 ; tell printer latest fw version

G91
G0 X5 Y5 Z.1 F1000
M1 FIRST BASE

G90 ; use absolute coordinates

M83 ; extruder relative mode
G28 W ; home all without mesh bed level

; heat things up
M104 S150; [first_layer_temperature] ; set nozzle temp
M190 S40 ; [first_layer_bed_temperature] ; wait for bed temp
M109 S150 ; [first_layer_temperature] ; wait for extruder temp

G80 ; mesh bed leveling

G0 X0 Y200 Z100 F3000 ; this is a good waiting point for bed flip

M1 Flip sheet then press OK 
```

Not being able to produce meaningful text messages on the LCD prevents users creating useful scripts to handle odd print conditions.

This is similar to M117 defect, also where desired text is NOT being display when it should be.

I tried many variations, including add G0 commands prior to the M1 - in various places in the gcode, and honestly now believe it was simple chance I found a place where the command actually works, as shown here:
```
M115 U3.7.1 ; tell printer latest fw version

G90 ; use absolute coordinates

M83 ; extruder relative mode
G28 W ; home all without mesh bed level

; heat things up
M104 S150; [first_layer_temperature] ; set nozzle temp
M190 S40 ; [first_layer_bed_temperature] ; wait for bed temp
M109 S150 ; [first_layer_temperature] ; wait for extruder temp

G80 ; mesh bed leveling

G0 X0 Y200 Z100 F3000 ; this is a good waiting point for bed flip

M1 Flip sheet then press OK 
```

vojtech-pavlik:
I wrote a patch that fixes M0, M1 and M117 to work reliably. Hopefully that fixes it for you.
[...]

PR #1900, opened 2019-06-05
Fix M0/M1/M117 commands when printing from SD.

vojtech-pavlik:
As reported in issue #1898 , M0/M1/M117 don't work under many scenarios. One is when an M0/M1 string contains a 'G' or any other substring decoded before M commands. This was already fixed fro M117 some time ago by moving it up to be decoded first. Another scenario is when printing from a SD card. Then the SD status takes priority over STATUS messages and the text from M0/M1/M117 is never shown. I've created two new CUSTOM_MSG_TYPEs, one that overrides the SD status for the M0/M1 message and another that overrides it temporarily for the M0/M1 resume message and the M117 message.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3432,6 +3432,52 @@ void process_commands()
 	  lcd_setstatus(strchr_pointer + 5);
   }
 
+  else if (code_seen("M0 ") || code_seen("M1 ")) { // M0 and M1 - (Un)conditional stop - Wait for user buttn press on LCD
+
+      char *src = strchr_pointer + 2;
+
+      codenum = 0;
+
+      bool hasP = false, hasS = false;
+      if (code_seen('P')) {
+        codenum = code_value(); // milliseconds to wait
+        hasP = codenum > 0;
+      }
+      if (code_seen('S')) {
+        codenum = code_value() * 1000; // seconds to wait
+        hasS = codenum > 0;
+      }
+      starpos = strchr(src, '*');
+      if (starpos != NULL) *(starpos) = '\0';
+      while (*src == ' ') ++src;
+      if (!hasP && !hasS && *src != '\0') {
+        lcd_setstatus(src);
+      } else {
+        LCD_MESSAGERPGM(_i("Wait for user..."));////MSG_USERWAIT
+      }
+
+      lcd_ignore_click();				//call lcd_ignore_click aslo for else ???
+      st_synchronize();
+      previous_millis_cmd = _millis();
+      if (codenum > 0){
+        codenum += _millis();  // keep track of when we started waiting
+		KEEPALIVE_STATE(PAUSED_FOR_USER);
+        while(_millis() < codenum && !lcd_clicked()){
+          manage_heater();
+          manage_inactivity(true);
+          lcd_update(0);
+        }
+		KEEPALIVE_STATE(IN_HANDLER);
+        lcd_ignore_click(false);
+      }else{
+        marlin_wait_for_click();
+      }
+      if (IS_SD_PRINTING)
+        LCD_MESSAGERPGM(_T(MSG_RESUMING_PRINT));
+      else
+        LCD_MESSAGERPGM(_T(WELCOME_MSG));
+    }
+
 #ifdef TMC2130
 	else if (strncmp_P(CMDBUFFER_CURRENT_STRING, PSTR("CRASH_"), 6) == 0)
 	{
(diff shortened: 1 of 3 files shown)


mvidner:
Links to Marlin FW docs:
- [M0, M1 - Unconditional stop](http://marlinfw.org/docs/gcode/M000-M001.html)
- [M117 - Set LCD Message](http://marlinfw.org/docs/gcode/M117.html)

Firmware/Marlin_main.cpp
```diff
@@ -3430,8 +3430,62 @@ void process_commands()
 	  if (starpos != NULL)
 		  *(starpos) = '\0';
 	  lcd_setstatus(strchr_pointer + 5);
+      custom_message_type = CUSTOM_MSG_TYPE_MSGUPD;
   }
 
+  else if (code_seen("M0 ") || code_seen("M1 "))
```
mvidner: This would be much easier to review if there was a separate commit moving the original piece of code in this place, keeping it intact, even with syntax errors (`case` instead of `else if`). @vojtech-pavlik could you do that please?

vojtech-pavlik: I split it out and provided better commit messages. I hope that's what you were looking for. I wasn't keen of causing intermediate syntax errors, though, I prefer to keep the code bisectable.

mvidner:
Regarding indentation:
- Arduino IDE renders the TAB character as 2 columns wide
- GitHub renders TAB as 4 columns wide
- the terminal renders TAB as 8 columns wide

Unsurprisingly, [the Marlin style guide says](http://marlinfw.org/docs/development/coding_standards.html#indentation) to indent with 2 spaces and avoid tabs. (I could not find a style guide for Prusa-Firmware)

vojtech-pavlik:
> Regarding indentation:
> 
> * Arduino IDE renders the TAB character as 2 columns wide
> * GitHub renders TAB as 4 columns wide
> * the terminal renders TAB as 8 columns wide
> 
> Unsurprisingly, [the Marlin style guide says](http://marlinfw.org/docs/development/coding_standards.html#indentation) to indent with 2 spaces and avoid tabs. (I could not find a style guide for Prusa-Firmware)

The Prusa source files use 2 columns of indentation and randomly either 4-wide tabstops or just 4 spaces instead. I just went with the flow and have no aspiration to fix the coding style of the file while fixing a functional issue.

mvidner:
> I just went with the flow and have no aspiration to fix the coding style of the file while fixing a functional issue.

Fair enough for me, I just wanted to summarize the current state and leave the decision up to the maintainer.

mvidner:
Relation to upstream: Marlin (1.1.x) is [using a parser class](https://github.com/MarlinFirmware/Marlin/blob/5d96a6d983c065403216ed52a8c92ed22dc268ba/Marlin/Marlin_main.cpp#L12649) which eliminates the parsing part of this issue/fix. I haven't checked the status-overriding part.

Code: Looks Good To Me.

Style: the indentation is a mess, but that's not a fault of this PR.

Build: My own build with `build.sh` passes, and Travis is passing too.

Automated tests: The existing tests are very sparse. I wouldn't want to invest the effort to add them for this PR.

Manual tests: We should quote a minimal gcode that demonstrates the bug and the effectiveness of the fix (just auto home, move Z a bit, message, move Z again). I could then flash my MK3S with this fix and verify that it works. If you are busy, I will create such gcode tonight (UTC+2).

This last part is why I am not giving an Approval yet :)

Firmware/Marlin_main.cpp
mvidner: Great, thanks!

vojtech-pavlik:
> Relation to upstream: Marlin (1.1.x) is [using a parser class](https://github.com/MarlinFirmware/Marlin/blob/5d96a6d983c065403216ed52a8c92ed22dc268ba/Marlin/Marlin_main.cpp#L12649) which eliminates the parsing part of this issue/fix. I haven't checked the status-overriding part.

That's so much nicer. But bringing the Prusa Firmware up to speed with current Marlin seems to be a rather huge job.

> Manual tests: We should quote a minimal gcode that demonstrates the bug and the effectiveness of the fix (just auto home, move Z a bit, message, move Z again). I could then flash my MK3S with this fix and verify that it works. If you are busy, I will create such gcode tonight (UTC+2).

Keep in mind that for the issues to manifest, some conditions need to be met:

1. The M0/M1/M117 commands need to be far enough from the end of the file so that the "SD print" is not complete by the time they get executed from the command buffer. Once the "SD print" is done, the filename will no longer be overriding the user messages. So a short example will work fine even on the current firmware. That's why this bug hasn't been fixed for so long.

2. The other case is when the text message contains the letter capital '**G**' or substrings like "**PRUS**". In this case the command will be misparsed as a G0/G1/G117 or a PRUSA custom command instead.

> This last part is why I am not giving an Approval yet :)

I can look into making a test case in the evening (CEST), too.

wavexx:
> > Unsurprisingly, [the Marlin style guide says](http://marlinfw.org/docs/development/coding_standards.html#indentation) to indent with 2 spaces and avoid tabs. (I could not find a style guide for Prusa-Firmware)
> 
> The Prusa source files use 2 columns of indentation and randomly either 4-wide tabstops or just 4 spaces instead. I just went with the flow and have no aspiration to fix the coding style of the file while fixing a functional issue.

It's a mess. I also try to follow whatever style is being used nearby, but there's hardly any.

See: https://github.com/prusa3d/Prusa-Firmware/pull/1752
and: https://github.com/arduino/Arduino/issues/8804

vojtech-pavlik:
This is a test case that exercises the M117 and M0 commands and triggers the bugs this pull request is trying to fix. Assuming you name it 'test.gcode', the status should show, in sequence:

- M117 test
- test.gcode
- Wait for user ...
- Resuming print
- test.gcode
- G Press button!
- 0 hours 0 minutes

On a firmware with the bugs, it'll just display

- test.gcode
- 0 hours 0 minutes

and will not wait for the button press.

`M115 U3.7.1`
`M117 M117 test`
`G28 W`
`M400`
`M0 P1500`
`G28 W`
`M400`
`M0 G Press button!`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M400`
`M84`

vojtech-pavlik:
And the test shows we're not quite there yet even with this pull request. So back to the drawing board.

hilsonp:
Please Prusa, come up with a solution to display custom messages with the M0, M1 and M117 commands ! Pleaaaase.

PatB42:
hello all 
So what is solution . finally ! i only use the M117 , added G4 and P 500 but still same issue , the text wont stay on display !! i tried to read all reply and solution but can't find the correction !! if any ..thx

PR #1901, opened 2019-06-06
Provide Platform.io project config & use it for CI

zavorka:
Compared to the current solution, I find this a ridiculously trivial & transparent way to provide reproducible builds across host platforms.

It looks that the u8g library is not actually being linked to in which case it can be dropped. It is however included in the [original UltiMachine's Arduino Addon repository](https://github.com/ultimachine/ArduinoAddons/tree/master/rambo/libraries/U8glib), so I included it for parity.


3d-gussner:
@zavorka Thanks, but we switch to cmake. Closing.

PR #1904, opened 2019-06-10
PFW-861 sound refactoring, delete mute mode, add blind mode

NotaRobotexe:


Firmware/sound.cpp
```diff
@@ -84,13 +115,43 @@ switch(eSoundMode)
...
      default:
-          ;
+          break;
      }
 }
 
+static void Sound_DoSound_Blind_Alert(void)
+{
+     _tone(BEEPER,100);
+     delay_keep_alive(50);
```
mkbel, line 138: This is called from lcd_update() and it calls lcd_update() leading to possible stack overflow.

PR #1915, opened 2019-06-13
Fix length false positives in lang-check.py

3d-gussner:
- Just reading the length of the translation line gives false positives as the " are included in the length BUT not displayed on the LCD.
- added missing argument 'pl'


mkbel (approved):

mkbel:
Thank you for your contribution.

3d-gussner:
Thanks, that was fast :-)

PR #1932, opened 2019-06-18
Retry nozzle load check

mkbel:
MMU: Filter IR sensor signal when detecting whether filament tip passes to heat break. Increase number of tries when loading filament to 3. Based on https://github.com/prusa3d/Prusa-Firmware/pull/1920 by KDanielK. This should decrease amount of false positive MMU Load errors.

### DIFF
--- a/Firmware/mmu.cpp
+++ b/Firmware/mmu.cpp
@@ -1451,8 +1451,8 @@ bFilamentAction=false;                            // NOT in "mmu_fil_eject_menu(
 
 //! @brief Fits filament tip into heatbreak?
 //!
-//! If PTFE tube is jammed, this cause filament to be unloaded and no longer
-//! detected by pulley IR sensor.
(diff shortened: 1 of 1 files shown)


Firmware/mmu.cpp
```diff
@@ -1447,26 +1449,69 @@ bFilamentAction=false;                            // NOT in "mmu_fil_eject_menu(
 	}
 }
 
+//! @brief Fits filament tip into heatbreak?
+//!
+//! If PTFE tube is jammed, this cause filament to be unloaded and no longer
```
DRracer: causes

Firmware/mmu.cpp
```diff
@@ -1447,26 +1449,69 @@ bFilamentAction=false;                            // NOT in "mmu_fil_eject_menu(
...
+//! detected by pulley IR sensor.
```
DRracer: no longer being detected by the pulley IR

DRracer (approved):

PR #1937, opened 2019-06-19
PFW-1060: added new function belt test under calibration menu

arons:
it takes quite some times to check belt status passing trough self test.
That is why I've added a specific function for that.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -152,7 +152,10 @@ static void mmu_cut_filament_menu();
 static void lcd_menu_fails_stats();
 #endif //TMC2130 or FILAMENT_SENSOR
 
+#ifdef TMC2130
 static void lcd_belttest_v();
+#endif //TMC2130
+
 static void lcd_selftest_v();
 
 #ifdef TMC2130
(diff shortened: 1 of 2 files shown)


arons:
If you want I can also move the functionality to other menu places.
For example directly under Support->Belt status.
Let me know your wishes.

F0x06:
Good job, this is awesome ! Works perfectly, this really needs to be merged by Prusa.

I implemented that myself some weeks ago, you did a better job than me with real time display etc, thanks :)

DRracer:
@arons good job, we'll merge this into 3.9

Firmware/ultralcd.cpp
```diff
@@ -6941,6 +6948,63 @@ void lcd_sdcard_menu()
...
+{
+    lcd_belttest();
+    menu_back_if_clicked();
+}
+void lcd_belttest_print(const char* msg, uint16_t X, uint16_t Y)
+{
+    lcd_clear();
+    lcd_printf_P(
+              _N(
+                 ESC_H(0,0)
```
leptun: @DRracer If you merge this and also the lcd improvements changes you need to get rid of that escape code. It is not needed in this case as lcd_clear() will also set the cursor to 0,0. Just remove line 6962.

DRracer (approved):

vintagepc:
This feature does not work on my printer. If I try to use it, Z moves up ~20mm, and I get a "Selftest error" on screen instructing me to check X axis length. X never attempts to move, and the updated status value is 0. I filed a bug report #2356

PR #1939, opened 2019-06-21
Fan error bug

NotaRobotexe:


Firmware/temperature.h
```diff
@@ -255,11 +255,11 @@ void check_fans();
...
-
+void fanSpeedErrorBeep(const char *serialMsg, const char *lcdMsg);
 
 #endif
 
 extern unsigned long extruder_autofan_last_check;
 extern uint8_t fanSpeedBckp;
 extern bool fan_measuring;
-
+extern bool resume_print;
```
DRracer: document variable usage - it is not completely clear, what it is supposed to do/guard/set. It looks like a flag saying "the print is being resumed right now", but not sure...

Firmware/temperature.h
```diff
@@ -255,11 +255,11 @@ void check_fans();
 
 void check_min_temp();
 void check_max_temp();
-
+void fanSpeedErrorBeep(const char *serialMsg, const char *lcdMsg);
```
DRracer: why do you need this function non-static?

Firmware/temperature.cpp
```diff
@@ -554,16 +557,18 @@ void fanSpeedError(unsigned char _fan) {
...
 	else {
 			setTargetHotend0(0);
 			SERIAL_ECHOLNPGM("// action:pause"); //for octoprint
+      heating_status = 0;
+      resume_print = false;
 	}
 	switch (_fan) {
 	case 0:	// extracting the same code from case 0 and case 1 into a function saves 72B
-		fanSpeedErrorBeep(PSTR("Extruder fan speed is lower than expected"), PSTR("Err: EXTR. FAN ERROR") );
+    fanSpeedErrorBeep(PSTR("Extruder fan speed is lower than expected"), PSTR("Err: EXTR. FAN ERROR") );
```
DRracer, line 568: misleading change of indentation

Issue #1928, opened 2019-06-17
Slow LCD refresh may affect printing


wavexx:
Every 30 seconds we fully refresh the LCD. When running some complex GCODE (a good example would be a large area with >30% gyroid infill), the full lcd refresh is an operation that may delay enough to starve the planner.

I have found some reports in slic3r that mention this.
One example is this: https://github.com/prusa3d/PrusaSlicer/issues/2065

Note that while the resolution of gyroid is high, it's not actually *too* high. You can easily reach the same level of complexity with some lattices, so I wouldn't "tune-down" gyroid.

If I manually disable the lcd refresh then I cannot easily manage to stop the planner anymore (gyroid keeps flowing without hiccups). I didn't look deeply into why the full lcd refresh is slow. Speeding it up, if possible, would be a good idea.

However a good interim solution would be to check for queued blocks before attempting to refresh the lcd. If the queue is less than half-full, skip the full refresh for a better time.

leptun:
Hi,

Can you please post a gcode / the settings which cause this behavior for testing purposes. Maybe even a video so I can be sure I know what I'm seeing.

So far I began rewriting some of the parts of the lcd library and maybe managed to make it faster. One problem I stumbled upon is what frequency the lcd chip is running at. In firmware is set to the slowest speed with extra safety delays. If prusa can confirm what frequency all the lcds run at, maybe we can make the timing tighter without breaking compatibility with slower lcds.

I also made changes so that when a command is sent to the lcd, it doesn't wait for the command to complete and continues to run the code, than if another lcd command is to be sent, it just waits till the previous command finishes processing and sends the next one and the cycle continues.

As for preventing the buffer from emptying, I made it so that if the ring buffer is at least half filled it reloads the lcd, but if it is not it just retries next update cycle. An exception to this is when the buffer is empty. If it weren't for that, then no lcd reset would happen and when not printing static electricity might kill the display.

I will post the changes when I come back (~2 weeks) and test them on my printer as there is a high chance I broke something in the lcd library.
As for code reduction, the code has gone from 234396 to 234184; 212 bytes saved, if you are curious.

One more thing I want to know: does the code use VT100 escape codes? If not, these could be removed to save flash space.

wavexx:
The part that affects the planner while printing is lcd_refresh() and lcd_refresh_noclear(). lcd calls are slow in general, but lcd_refresh* call lcd_begin() internally which is the real killer.

Did you manage to improve this as well?

[...]

Issue #1406, opened 2018-12-14
MK3 3.5.0: Very short interruptions during fast prints


haarp:
Hello,

during fast and frequent head moves while printing from the SD card, I can often hear the head pausing for the fraction of a second. This seems to coincide with LCD updates. I suspect that the LCD update routine takes long enoguh on the tiny CPU to delay print commands for a couple of milliseconds.

inquisitiveKangaroo:
Hi, 
Thank you for feedback. Is it possible to provide us the gcode?
Kangaroo

hoppke:
In case it's maybe related -- I observe sth like this even without printing. If I turn the printer on, connect from octoprint, then issue a longer move command (like move Y 150mm, default speed) via console, then the printer makes a small initial move (lasting a fraction of a second), stops, and then immediately makes the rest of the move in one motion. I can actually hear it pause and resume - sounds like 2 separate moves in sequence. I think it affects all axes, not just Y. Weird to see a single gcode stutter like that.

leptun:
@hoppke I've seen this bug for ages and it only happens sometimes. I don't know how to reproduce the issue, but what I can tell you is that it happens when "printing" over usb (octoprint not required).

I think the first time a saw it was in 3.1.0 with the introduction of linear advanceon the MK2.
I can still see the issue on my custom 3.4.1.

Panayiotis-git:
If I remember correctly it starts happening after the printer calibration.
Try to do a full printer reset. WITHOUT calibration try the long move. It will be executed in one movement.

haarp:
Unfortunately I can not reproduce this anymore. Not sure what was going on. But as others have reported similar problems, I'll keep this open for now.

haarp:
Ok, I managed to reproduce this over USB. It has nothing to do with LCD updates.

When many G-codes follow each other in very quick succession, the printer eventually starts pausing briefly (milliseconds). I highly suspect a buffer underrun occuring in the command buffer.

This can happen for example when printing Slic3r's Archimedean Cords Top/Bottom infill, which consists of very short G1 moves (why don't they use G2 for that?!?). If the print speed becomes high enough, USB seems incapable of sending commands fast enough, causing the printer to very briefly halt.

leptun:
@DRracer @mkbel @haarp I have also experienced this issue from when I started 3d printing on the MK2 2 years ago. It is really bad when you try to reproduce it intentionally, especially at higher speeds. I think it needs more attention.
[...]

PR #1940, opened 2019-06-21
Limit LCD refresh while printing to avoid planner stalls

wavexx:
Limit LCD reinitialization when the planner is busy

LCD operations, and to be precise the lcd_begin() call is extremely costly. When printing complex geometry or infill (such as gyroid) it can starve the planner and cause a stall.

This is also frequently visible during the 7x7 mesh leveling, where each move is individually planned and a single lcd refresh becomes noticeable as it prevents further moves to be scheduled.

We calculate a very rough estimate of the time taken by queued moves in ``planned_time``. This is just overall length at the average nominal speed. If we have less than 250ms available, we skip the current refresh and wait for the next cycle. If a full reset is pending though we keep on probing every second, as it's likely we're printing complex geometry and we should attempt more frequently in order to catch an available time slot.

planned_time() is usually quite optimistic as acceleration is completely ignored, but it's otherwise accurate while printing high-resolution curvy geometry such as gyroid. We only need a true lower-bound anyway, and the function should be fast to avoid overloading the planner. This means we will be refreshing the LCD mostly during slow moves, outer/straight perimeters or long travels.

See issue #1928 for more details. Using just the queue size and/or planned length is not sufficient to prevent stalls and ensure the LCD is still refreshed regularly. LCD functions can still be sped up, so the issue cannot be closed completely with this PR yet.

### DIFF
--- a/Firmware/planner.cpp
+++ b/Firmware/planner.cpp
@@ -1369,3 +1369,23 @@ uint16_t planner_calc_sd_length()
 	}
 	return sdlen;
 }
+
+float planned_time()
+{
+    unsigned char _block_buffer_head = block_buffer_head;
+    unsigned char _block_buffer_tail = block_buffer_tail;
+
+    unsigned char n = 0;
+    float mm = 0;
+    float speed = 0;
+
+    while (_block_buffer_head != _block_buffer_tail)
+    {
+        mm += block_buffer[_block_buffer_tail].millimeters;
+        speed += block_buffer[_block_buffer_tail].nominal_speed;
+        ++n;
+
+        _block_buffer_tail = (_block_buffer_tail + 1) & (BLOCK_BUFFER_SIZE - 1);
+    }
+    return mm * ((float)n / speed);
+}
--- a/Firmware/planner.h
+++ b/Firmware/planner.h
@@ -237,7 +237,6 @@ void set_extrude_min_temp(float temp);
 #endif
 
 void reset_acceleration_rates();
-#endif
 
 void update_mode_profile();
 
@@ -254,3 +253,9 @@ extern void planner_queue_min_reset();
 extern void planner_add_sd_length(uint16_t sdlen);
 
 extern uint16_t planner_calc_sd_length();
+
+// Return a rough, optimistic estimate of the total time (s) for the planned moves
+// This function takes in the order of 50us
+float planned_time();
+
+#endif
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -936,6 +936,45 @@ void lcdui_print_status_screen(void)
 
 }
 
+
+// Attempt to occasionally reinit to revive the LCD if static electricity killed it
+#define LCD_REINIT_FULL 30 // time between full reinitializations
+#define LCD_REINIT_FAST 10 // time between fast reinitializations
+#define LCD_REINIT_OVR   1 // check frequency for overdue full refreshes
+
+static_assert((LCD_REINIT_FULL % LCD_REINIT_FAST) == 0, "FULL should be a multiple of FAST");
+
+static void lcd_attempt_refresh()
+{
+    ++ReInitLCD;
+
+    if ((ReInitLCD % LCD_REINIT_FAST) == 0)
+    {
+        // limit LCD reinitialization while printing as it can starve the planner in tight situations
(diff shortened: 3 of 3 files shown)


thess:
Funny you should do this -- I determined the ReinitLCD logic and support vars was totally unnecessary a couple of years ago. I still remove this re-draw hack from their firmware in custom builds and features. The constant display flickering from the erase-and-redraw procedures was annoying.

leptun:
I am currently working on a better way to handle the lcd refresh in general. I will also tinker with the frequency at which the lcd is refreshing. The changes are not pushed yet as I haven’t really finished them. The refreshes shouldn’t block the loop function with my changes, avoiding planner starving.

wavexx:
Note that it's not the refresh in general that creates big problems for the planner. No question that they're slow, but it's only the full re-init that takes enough time to stall.

As of why the lcd is fully re-initialized frequently, is that their internal state does get corrupted. I've seen this on other printers, although never on the mk3 so far.

I'm not entirely sure if that's because of the ribbon cable picking up noise or just low quality parts.

If you say that you never had this, I could also try to run with the reinit disabled.

leptun:
It was the lcd reinit I was talking about :) 

I don’t know why the lcd is reinitialized so frequently. Once a minute is enough and the reiniting when changing menus is kinda useless. I think only when entering (like when it times out) and exiting the status screen should it be reinited.

Still can’t test it though...

wavexx:
On Mon, Jun 24 2019, Voinea Alex wrote:
> It was the lcd reinit I was talking about :)

Oh ok then =)

> I don’t know why the lcd is reinitialized so frequently. Once a minute
> is enough and the reiniting when changing menus is kinda useless. I
> think only when entering (like when it times out) and exiting the
> status screen should it be reinited.

I never watched the LCD long enough to see how/when it corrupts exactly.
What I can easily do is disable reinit completely and see how long it
takes to happen.

> Still can’t test it though...

As I happened to hit this problem by chance while doing other stuff with
gyroid, I can wait :)

Since you did a lot on this already you're in a better position to
continue. Let me know if you have something you want some testing on.

leptun:
I pushed to MK3_LCD_FAST branch.
My changes do as follows:

All lcd communications were moved to a separate planner. It has 8 blocks which can be filled with commands or characters for the lcd. These blocks are added to the ring buffer by the lcd_plan_data function. If the buffer is full, it will wait like before, but this should impact the mail loop much less since it only gets filled when lcd characters are printed to the lcd. lcd_begin() should stop the lcd from being updated untill the reset is finished, resulting in no downtime during reset.
The buffer has its own lcd_isr() ran alongside the stepper isr() that reads from the buffer, executes the content (prints data/commands to the lcd) and calculates when next command can be executed and saves it to the next block, regardless if it's filled or empty.

As for other improvements: 

- VT100 codes support has been disabled to save flash and sram and can be enabled with define in Configuration.h. It is not properly implemented and only uses precious flash.
- RW support was removed since it isn't used on prusa printers and on most other lcds. 
- 8bit bus support is still there, but it is only enabled when LCD_PINS_D4-7 are defined in pins_...board....h
- The lcd library no longer uses arduino pinMode and digitalWrite functions and uses fastIO macros instead.

If the changes I made still work (which I hope they do), it should fix the issue #1928 
I would be greatful if @wavexx could test these and report back. Be prepared for some debugging :)

Will address lcd refreshes at a later time

wavexx:
Can you open issues on your fw fork? I'd like to make a few comments and I have no idea where to put those. I would really avoid putting the lcd isr in timer1. At low speeds with LA any delay will kill advance steps, on fast moves when dual/quad stepping kicks in the timing gets really tight. And it's not like serial I/O that needs to drain a buffer quickly.

I think timer2 should be fast enough for this.

leptun:
here you go:
https://github.com/leptun/Prusa-Firmware/issues/1

haarp:
Interesting, so #1406 is related to the LCD after all.

@leptun You might also wanna take a look at #1445

leptun:
@haarp It’s interesting that other people attempted this before. I looked at #1445, but I like my approach more. Anyway, who am I to judge... :)

wavexx:
I'm closing this for now. The current LCD implementation is vastly superior and doesn't delay the planner enough to warrant extra checks.

PR #1952, opened 2019-06-28
new sound

NotaRobotexe:


mkbel (approved):

mkbel (approved):

Firmware/sound.cpp
```diff
@@ -108,7 +169,7 @@ for(nI=0;nI<10;nI++)
 static void Sound_DoSound_Prompt(void)
 {
 WRITE(BEEPER,HIGH);
-delay_keep_alive(500);
+delayMicroseconds(500);
```
leptun, line 172: @NotaRobotexe Why did you change from delay_keep_alive that accepts time in **ms** to delayMicroseconds that accepts time in **us**? Same in the next lines.
issue #1995

NotaRobotexe: It is mistake, I didn't notice It is going to be fixed in new pull request. Also I replaced delay_keep_alive because if you call it from lcd_update() it can call again lcd_update() which can lead to stack overflow.

PR #1962, opened 2019-07-03
Planner for lcd data

leptun:
This PR addresses issue #1928 and is an alternative to #1940 by @wavexx.

The issue is that when lcd_refresh() is called the code hangs so long that the motion planner is drained during gyroid infill and results in stuttering. With this PR the way the lcd is handled is changed completely.

- Slow arduino pinMode and digitalWrite are replaced with the corresponding fastio macros.
- RW support is removed since no prusa printer uses it, no other printers use it and is not implemented completely in firmware.
- 8bit lcd support is only enabled when the required pins are defined.
- the timing has been tightened a bit according to the documentation.

The above changes only make it a bit faster, but do not resolve the fundamental issue: the lcd functions block the main loop and use long delays.

My approach to this issue was implementing a LCD planner similar to the motion planner. All commands and data are queued in this buffer and ran when the previous command has finished. This way when long running commands such as the ones from lcd_begin() are queued, the code can continue freely and start printing other characters only when the init has finished. When it is time to print characters on the lcd, the buffer gets filled, but that's ok since it is a lot faster than the full refresh of the lcd.

The planned data is sent to the lcd using an ISR and the dedicated timer3. You can read more about why this timer was chosen in the changes made to "Marlin_main.cpp". The isr is executed with the correct delay required by each command to the lcd and is disabled once all data has been printed.
The ring buffer that stores all the queued data has only 8 chunks. It could be expanded, but I think this is enough for now. It would have been best to have it big enough to store the entire lcd, but this is a lot better than waiting for the refresh (The entire refresh fits in the buffer).

This still needs testing regarding the effectiveness of the changes though. The 7x7 mesh bed leveling seems smoother and with less stops as noted by @wavexx in the initial issue.

Hope this gets merged. If you want any changes, do not hesitate to ask. I am open to suggestions.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -775,12 +775,8 @@ int uart_putchar(char c, FILE *)
 
 void lcd_splash()
 {
-//	lcd_puts_at_P(0, 1, PSTR("   Original Prusa   "));
-//	lcd_puts_at_P(0, 2, PSTR("    3D  Printers    "));
-//	lcd_puts_P(PSTR("\x1b[1;3HOriginal Prusa\x1b[2;4H3D  Printers"));
-//    fputs_P(PSTR(ESC_2J ESC_H(1,1) "Original Prusa i3" ESC_H(3,2) "Prusa Research"), lcdout);
-    lcd_puts_P(PSTR(ESC_2J ESC_H(1,1) "Original Prusa i3" ESC_H(3,2) "Prusa Research"));
-//	lcd_printf_P(_N(ESC_2J "x:%.3f\ny:%.3f\nz:%.3f\ne:%.3f"), _x, _y, _z, _e);
+	lcd_clear(); // clears display and homes screen
+	lcd_puts_P(PSTR("\n Original Prusa i3\n   Prusa Research"));
 }
 
 
@@ -910,7 +906,7 @@ void update_sec_lang_from_external_flash()
 		uint32_t src_addr;
 		if (lang_get_header(lang, &header, &src_addr))
 		{
-			fputs_P(PSTR(ESC_H(1,3) "Language update."), lcdout);
+			lcd_puts_at_P(1,3,PSTR("Language update."));
 			for (uint8_t i = 0; i < state; i++) fputc('.', lcdout);
 			_delay(100);
 			boot_reserved = (state + 1) | (lang << 4);
@@ -984,8 +980,8 @@ void list_sec_lang_from_external_flash()
 
 static void w25x20cl_err_msg()
 {
-    lcd_puts_P(_n(ESC_2J ESC_H(0,0) "External SPI flash" ESC_H(0,1) "W25X20CL is not res-"
-            ESC_H(0,2) "ponding. Language" ESC_H(0,3) "switch unavailable."));
+	lcd_clear();
+	lcd_puts_P(_n("External SPI flash\nW25X20CL is not res-\nponding. Language\nswitch unavailable."));
 }
 
 // "Setup" function is called by the Arduino framework on startup.
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -1011,7 +1011,8 @@ void setup()
 	ultralcd_init();
 
 #if (LCD_BL_PIN != -1) && defined (LCD_BL_PIN)
-	analogWrite(LCD_BL_PIN, 255); //set full brightnes
+	//analogWrite(LCD_BL_PIN, 255); //set full brightnes
+	WRITE(LCD_BL_PIN, HIGH); //LCD_PLANNER needs a dedicated timer. On the MiniRambo timer3 is free, but timer5 is used by the stepper drivers. On the einsyRambo timer3 could be used by the lcd backlight pwm, however this is not implemented fully as of yet. Timer5 on the other hand is free on this board, so this could be a sollution to the conflict. For now we use Timer3 on all boards and write directly to the pin.
 #endif //(LCD_BL_PIN != -1) && defined (LCD_BL_PIN)
 
 	spi_init();
--- a/Firmware/lcd.cpp
+++ b/Firmware/lcd.cpp
@@ -55,9 +55,11 @@
 #define LCD_5x10DOTS 0x04
 #define LCD_5x8DOTS 0x00
 
-#define LCD_RS_FLAG 0
-#define LCD_HALF_FLAG 1
+// bitmasks for flag argument settings for lcd_planner
+#define LCD_RS_FLAG 0x01
+#define LCD_HALF_FLAG 0x02
 #define LCD_WAIT_FLAG 0x04
+#define LCD_WAIT_UPDATE_BACKUP_FLAG 0x08
 
 FILE _lcdout; // = {0}; Global variable is always zero initialized, no need to explicitly state that.
 
@@ -120,17 +122,20 @@ ISR(TIMER3_COMPA_vect)
 	if(lcd_current_block == NULL)
 		return;
 	uint8_t flag = lcd_current_block->flag;
-	if (flag & LCD_WAIT_FLAG) lcd_update_enabled = 1;
+	if (flag & LCD_WAIT_FLAG) {
+		if (!lcd_update_enabled) // in case the update was enabled while the lcd was initializing so not to disable it
+			lcd_update_enabled = lcd_current_block->flag & LCD_WAIT_UPDATE_BACKUP_FLAG;
+	}
 	else
 	{
-		WRITE(LCD_PINS_RS, flag & (1 << LCD_RS_FLAG));
+		WRITE(LCD_PINS_RS, flag & LCD_RS_FLAG);
 		_delay_us(5);
 		lcd_writebits(lcd_current_block->data);
 	}
 #ifndef LCD_8BIT
-	if (flag & (1 << LCD_HALF_FLAG)){
+	if (flag & LCD_HALF_FLAG){
 		lcd_current_block->data = lcd_current_block->data<<4;
-		flag &= ~(1 << LCD_HALF_FLAG);
+		flag &= ~LCD_HALF_FLAG;
 		lcd_current_block->flag = flag;
 		OCR3A = lcd_plan_calculate_timer_delay(LCD_COMMAND_DELAY);
 	}
@@ -151,7 +156,10 @@ void _lcd_plan_data(uint8_t data, uint8_t flag, uint16_t command_delay_us) {
 	block->data = data;
 	block->flag = flag;
 	block->command_delay_us = command_delay_us;
-	if (flag & LCD_WAIT_FLAG) lcd_update_enabled = 0;
+	if (flag & LCD_WAIT_FLAG) {
+		block->flag ^= (-lcd_update_enabled ^ block->flag) & LCD_WAIT_UPDATE_BACKUP_FLAG;
+		lcd_update_enabled = 0;
+	}
 	lcd_block_buffer_head = next_buffer_head;
 	if ((TCCR3B & 0x02) == 0) {
 		OCR3A = lcd_plan_calculate_timer_delay(command_delay_us);
@@ -170,9 +178,9 @@ void lcd_plan_data
 	,uint8_t flag
 )
 {
-	flag |= (RS << LCD_RS_FLAG);
+	flag ^= (-RS ^ flag) & LCD_RS_FLAG;
 #ifndef LCD_8BIT
-	flag |= (half << LCD_HALF_FLAG);
+	flag ^= (-half ^ flag) & LCD_HALF_FLAG;
 #endif
 
 	_lcd_plan_data(data, flag, command_delay_us);
@@ -267,7 +275,7 @@ static void lcd_begin(uint8_t clear)
 	// set the entry mode
 	lcd_command(LCD_ENTRYMODESET | lcd_displaymode, 60);
 	lcd_escape[0] = 0;
-	lcd_plan_data(0x00, LOW, 0, 0, LCD_WAIT_FLAG); // special lcd begin flag
+	lcd_plan_data(0x00, LOW, 0, 0, LCD_WAIT_FLAG); // special flag that disables lcd_update() to not block the loop with the status screen update during lcd_begin()
 }
 
 int lcd_putchar(char c, FILE *)
--- a/Firmware/lcd.cpp
+++ b/Firmware/lcd.cpp
@@ -16,6 +16,12 @@
 //-//
 #include "sound.h"
 
+#define LCD_BLOCK_BUFFER_SIZE 8
+#define LCD_COMMAND_DELAY 100
+
+#if (defined(LCD_PINS_D0) && defined(LCD_PINS_D1) && defined(LCD_PINS_D2) && defined(LCD_PINS_D3))
+	#define LCD_8BIT
+#endif
 
 // commands
 #define LCD_CLEARDISPLAY 0x01
@@ -73,6 +79,19 @@ uint8_t lcd_escape[8];
 
 void lcd_writebits(uint8_t value);
 
+typedef struct {
+	uint8_t data;
+	uint8_t flag;
+	uint16_t command_delay_us;
+} lcd_block_t;
+
+#define ENABLE_LCD_TIMER() do {\
+	TCNT3 = 0;\
+	TCCR3B |= 0x02;\
+} while(0)
+
+#define DISABLE_LCD_TIMER() TCCR3B = (TCCR3B & ~(0x07<<CS30))
+
 lcd_block_t lcd_block_buffer[LCD_BLOCK_BUFFER_SIZE];          // A ring buffer for lcd data blocks
 volatile unsigned char lcd_block_buffer_head;                 // Index of the next block to be pushed
 volatile unsigned char lcd_block_buffer_tail;                 // Index of the block to process now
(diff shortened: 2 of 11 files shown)


leptun:
These were tested on a mk3 hybrid that had the lcd from a mk2s. It is from a batch after the mk3 was released. As far as I know, the mk3 has a different display that supports backlight control. Don’t think the timing between these two is different, but something worth checking.

leptun:
There is currently a small issue with this PR on the latest branch. When you edit the name of a sheet the second character is the “H” custom character from the heatbed temperature. I’ll try to fix it soon. I think it’s related to the way I disable refresh during a reset and reenable it after the reset.

EDIT: Nevermind. Also happens on stock firmware

DRracer:
@leptun please note that the sheet profiles are subject to test right now and especially the name edit will be changed.

leptun:
@DRracer I am starting to shrink the code now. I am thinking of removing all VT100 codes from messages and stuff like that. After that disable the VT100 feature of the lcd library with defines and see what I can do after that. Should I push changes here or make a new PR with that? I would prefer if I merged them here since some changes might require the lcd_planner.

DRracer:
@leptun Please feel free to continue adding changes to this PR. I'm looking forward to seeing how much FLASH you can save with this ;)

leptun:
@DRracer  With the latest push the code shrank by 638B of flash. I also tried to see if the linefeed changes worked on other languages, but can't switch to any other language. Can maybe someone else test if for example the crash detection warning displays correctly?

DRracer:
@leptun good job optimizing the code. Are you sure no VT100 features were used elsewhere in the firmware?
We'll test your PR in ca. 14 days, right now we are rolling out the 3.8 alpha for internal testing. Please stay tuned ;) .

leptun:
They were used, but I made sure all were handled in another way

DRracer:
@leptun ok, thanks for the info

F0x06:
@leptun 
Hi, i have issues with fast perimeters on circular shapes 70mm circle in this case using **STOCK** firmware, it's like the printer is lagging resulting marks on top layers of perimeters, can this PR solve the issue ? I'm printing with a Pi3 from octoprint or it's due to USB-Serial connection and printing from SD will solve the issue ? Thanks

PS: Issue not present with slower speeds, i'm using PrettyPETG

leptun:
I don’t think this is related to my change. I also experienced this (without my changes) when I was printing with simplify3d and the skirt would stutter. Also fast perimeters. This was only on usb for me, sd card resulting in smoother “circular” motion.

wavexx:
On Thu, Jul 11 2019, Kevin Velickovic wrote:
> @leptun Hi, i have issues with fast perimeters on circular shapes 70mm
> circle in this case, it's like the printer is lagging resulting marks
> on top layers of perimeters, is this related ? I'm printing with a Pi3
> from octoprint or it's due to USB-Serial connection and printing from
> SD will solve the issue ? Thanks

I'd really love a video on that.
Please record both the extruder and the LCD in the same shot.

F0x06:
@leptun 
> I don’t think this is related to my change

Oh i talk about Stock firmware not your changes :), i want to know if your modifications can solve my issue. If i understand correctly yes ? I'll give it a shot.

@wavexx I'll restart my print and take a video of that :), i'll perform the test with SD card to exclude serial issues.

F0x06:
> 
> On Thu, Jul 11 2019, Kevin Velickovic wrote: @leptun Hi, i have issues with fast perimeters on circular shapes 70mm circle in this case, it's like the printer is lagging resulting marks on top layers of perimeters, is this related ? I'm printing with a Pi3 from octoprint or it's due to USB-Serial connection and printing from SD will solve the issue ? Thanks
> I'd really love a video on that. Please record both the extruder and the LCD in the same shot.

Gotcha ! Here is the video with sound (OctoPrint@PI3B+): https://f0x.me/20190711_172847_1.mp4

EDIT: I'll restart the print from SD later to see.

DRracer:
@F0x06 it is very likely, that @leptun 's LCD patch and @wavexx 's Linear Advance 1.5 may help to speed up print when there are too many short segments. But please first test a print from an SD card to really elliminate serial communication issues

F0x06:
> 
> 
> @F0x06 it is very likely, that @leptun 's LCD patch and @wavexx 's Linear Advance 1.5 may help to speed up print when there are too many short segments. But please first test a print from an SD card to really elliminate serial communication issues

Yes of course I'll test from SD first.

wavexx:
On Thu, Jul 11 2019, Kevin Velickovic wrote:
>     @F0x06 it is very likely, that @leptun 's LCD patch and @wavexx 's Linear Advance 1.5 may
>     help to speed up print when there are too many short segments. But please first test a
>     print from an SD card to really elliminate serial communication issues
>
> Yes of course I'll test from SD first.

Also please attach the config bundle and STL file you're using, if
possible so that I could try to reproduce that.

Removing micro-stutters especially when printing from the serial is
something I'd like to investigate.

F0x06:
> 
> On Thu, Jul 11 2019, Kevin Velickovic wrote: @F0x06 it is very likely, that @leptun 's LCD patch and @wavexx 's Linear Advance 1.5 may help to speed up print when there are too many short segments. But please first test a print from an SD card to really elliminate serial communication issues Yes of course I'll test from SD first.
> Also please attach the config bundle and STL file you're using, if possible so that I could try to reproduce that. Removing micro-stutters especially when printing from the serial is something I'd like to investigate.

Here you go, you can find the `3mf` file with used settings and model (PETG).
[bath_hair_filter.zip](https://github.com/prusa3d/Prusa-Firmware/files/3382894/bath_hair_filter.zip)

F0x06:
@leptun @wavexx @DRracer Same print from SD card, there is less stutters but still present especially on the first portion of first perimeter (right side): https://f0x.me/20190711_183456_1.mp4

Now time to test your changes @leptun then yours @wavexx 

@wavexx Can you merge your  Linear Advance 1.5 with the latest commits of MK3 branch please ?, do i need specific setting for Linear Advance 1.5 or it's a drop-in replacement ?

Thanks !

wavexx:
@F0x06 Can we attach the videos directly here to ensure they're not going to get lost?
I'll be away for a few more days, I'll do an update then. But you can use this build meanwhile:

https://github.com/vertigo235/Build-Prusa-LA-15/releases

You need to update the K factor for LA 1.5, everything else is drop-in.

F0x06:
> 
> 
> @F0x06 Can we attach the videos directly here to ensure they're not going to get lost?
> I'll be away for a few more days, I'll do an update then. But you can use this build meanwhile:
> 
> https://github.com/vertigo235/Build-Prusa-LA-15/releases
> 
> You need to update the K factor for LA 1.5, everything else is drop-in.

Thanks for the info, how can i convert my actual values for example `45` to new `mm/mm/s` values ? Or i need to manually calibrate linear advance and don't use Prusa's ones ?

The videos:

[20190711_183456_1.zip](https://github.com/prusa3d/Prusa-Firmware/files/3383546/20190711_183456_1.zip)
[20190711_172847_1.zip](https://github.com/prusa3d/Prusa-Firmware/files/3383547/20190711_172847_1.zip)

Thanks !

leptun:
I did some tests using a spare mini rambo with an (unfortunately) undersized LCD and an sd shield on the side + a stepper motor (X). Firmware was compiled with MK2-rambo13a variant. Did tests on MK3 branch and on LCD_PLANNER branch with USB and SD. Gcode used is a 50mm single wall spiral vase cylinder printed at 100mm/s. Feedrate multiplier on screen should indicate the speed tested at.
image

Results are as follows: (Use headphones for video and listen at max)

- MK3 SD:
https://photos.app.goo.gl/bnXqv4zgqYuWN3DA7
Stuttering can be seen at: 1:15 1:26 1:39 1:51 It coincides with lcd_refresh();
Max speed without constant stuttering is 250mm/s (could be 260mm/s)

- MK3 USB:
https://photos.app.goo.gl/m7ettLQiHnSxm9WNA
Stuttering can be seen at 1:37 2:00 It coincides with lcd_refresh();
Max speed without constant stuttering is probably 230mm/s. I have only tried 200 on this run for some reason.

- LCD_PLANNER SD:
https://photos.app.goo.gl/WGyXbxaySSp5Sd63A
There is no stuttering at the ~200mm/s speed like on the MK3 branch. The only stutter that apperars is constant and caused by top speed of the loop being reached. (aka can't process faster) and is probably optimistic since on a MK3 the loop takes longer because of all the features that have to be updated.

- LCD_PLANNER USB:
https://photos.app.goo.gl/DV2CBREd86C8d6Si6
Same as on SD, the only difference being that the "max speed" is now bottlenecked by the USB-Serial connection. Maybe with higher baud it could be faster, but I doubt it. Max speed: 230mm/s

Couldn't test more today. maybe tomorrow. Sorry for the bad video quality and smaller lcd. This is the best I could do for now.

I might test tomorrow @wavexx 's limit_lcd_refresh branch and see how it performs.

[Cylinder_MKB.gcode.txt](https://github.com/prusa3d/Prusa-Firmware/files/3383665/Cylinder_MKB.gcode.txt)

wavexx:
On Thu, Jul 11 2019, Kevin Velickovic wrote:
> Thanks for the info, how can i convert my actual values for example
> `45` to new `mm/mm/s` values ? Or i need to manually calibrate linear
> advance and don't use Prusa's ones ?

This needs to be calibrated per-material for optimal results, but as a
ball-park measure, use K=0.05 for PLA, K=0.12 for PETG (see
http://marlinfw.org/tools/lin_advance/k-factor.html).

F0x06:
> 
> On Thu, Jul 11 2019, Kevin Velickovic wrote: Thanks for the info, how can i convert my actual values for example `45` to new `mm/mm/s` values ? Or i need to manually calibrate linear advance and don't use Prusa's ones ?
> This needs to be calibrated per-material for optimal results, but as a ball-park measure, use K=0.05 for PLA, K=0.12 for PETG (see http://marlinfw.org/tools/lin_advance/k-factor.html).

@wavexx 
Thanks :)

@leptun
I agree with `The 7x7 mesh bed leveling seems smoother and with less stops`, payed particular attention to that before patching and also noticed that some time ago. I'll come back with results soon

wavexx:
On Thu, Jul 11 2019, DRracer wrote:
> @leptun Please feel free to continue adding changes to this PR. I'm
> looking forward to seeing how much FLASH you can save with this ;)

Note that I also have several commits that shave bytes here and there by
doing some cleanup.

See https://github.com/wavexx/Prusa-Firmware/commit/2bd72510c49140ba280fdfb2696e410b8bfa6b27
or https://github.com/wavexx/Prusa-Firmware/commit/d75ad438a1a5352c3d7cb7441f74e6616d0ce1e6

I didn't push any PR on those yet as they build up on top of https://github.com/prusa3d/Prusa-Firmware/pull/1768

leptun:
Soooo... ignore the last commit. I’ll revert it soon since it doesn’t work correctly. The problem is the planner is kinda useless for now during lcd_refresh since I misunderstood how the lcd_update works. The benefits you see from this PR for now are only because it works faster, but still waits. If I start spamming the knob and do many full cleans and redraw of the lcd it becomes evident that it struggles to keep up.

I’ll try to rework how lcd_update works and avoid holding the loop for too long with anything.

leptun:
I am currently almost out of ideas on how to approach this. Does anyone have an idea that could improve how lcd_update works or some other way not to hold the loop for too long? I am currently testing a potential solution but it doesn’t seem that effective. It does lcd_refresh without jerking, but changing menus and especially going into high refresh menus (eg. the ones in support) makes the loop go very slow and inconsistent. I will try something else tomorrow, but in the meantime I am listening for suggestions.

wavexx:
On Fri, Jul 12 2019, leptun wrote:
> I am currently almost out of ideas on how to approach this. Does
> anyone have an idea that could improve how lcd_update works or some
> other way not to hold the loop for too long? I am currently testing a
> potential solution but it doesn’t seem that effective. It does

I didn't look at the entire LCD update/redraw cycle yet.
I assume you're talking about the lcd look taking too long before
returning control to the main loop, or vice-versa?

> I will try something else tomorrow, but in the meantime I am listening
> for suggestions.

Brutal approach: did you look at marlin 2 yet?
There are many things that I'd like to backport from there.

The handling of tunable settings is so much better upstream it's not
even funny :/

DRracer:
@leptun @wavexx could you look more deeply into the problem with LCD update/planning? Making a better (faster) approach would definitely help to reduce print lagging with short segments (I know it's a difficult task). Right now we'd like to avoid backporting a larger part of Marlin 2 when absolutely not necessary.
I was thinking about "disconnecting" the update from the rest of the code - making a similar approach like an old VGA text mode. Just let the main loop code write characters into some kind of frame buffer (4x20 bytes) and do all the costy LCD protocol timing in a finite automaton running async under an interrupt.
@leptun can you prepare a PR with just the first part of improvements you write about in your first post here? I.e. 
- Slow arduino pinMode and digitalWrite are replaced with the corresponding fastio macros.
- RW support is removed since no prusa printer uses it, no other printers use it and is not implemented completely in firmware.
- 8bit lcd support is only enabled when the required pins are defined.
- the timing has been tightened a bit according to the documentation.
That may save a few CPU ticks and a possibly a few bytes of FLASH, woldn't it?

leptun:
@DRracer Already made a planner-less version, posting another PR soon.

leptun:
I am closing this PR because:

1. It is not compatible with the branch anymore. I woulds have to do some changes.

2. @DRacer ‘s approach might be better than this. I’ll make a PR when it’s done

mkbel:
Hello @leptun,
there is problem with both this pull request and already merged https://github.com/prusa3d/Prusa-Firmware/pull/2008 that after external flash (applies only to MK3/S (Einsy board)) is flashed during flashing process LCD freezes and starts operating only after printer reset.

video: http://www.drracer.eu/prusa/lcd_planer_fw_update.mp4 http://ge.tt/6JfJD8x2

maximlevitsky:
Yep. Also language change behaves the same. If I wait after language change for some time, and then reset the printer, other language is broken (mostly each string misses 2 characters from start)

wavexx:
> video: http://ge.tt/6JfJD8x2

Cannot get this to download.

leptun:
@mkbel I can't download the video.

Does the printer reset by itself at the end of flash?

mkbel:
> @mkbel I can't download the video.
> 
> Does the printer reset by itself at the end of flash?

No, you need to reset it manually.

DRracer:
... just placed the video onto my site just until we find a solution:
http://www.drracer.eu/prusa/lcd_planer_fw_update.mp4

mkbel:
The problem is caused by first commit in this pull request https://github.com/prusa3d/Prusa-Firmware/pull/1962/commits/0800bed02fed2f6a49d83c04a97cb5e8c8b5ae27 and the first commit in https://github.com/prusa3d/Prusa-Firmware/pull/2008 https://github.com/prusa3d/Prusa-Firmware/pull/2008/commits/8b1c6876290f6524bbb8a4691d53eb66ea53006d

mkbel:
There is also problem that MK3 flashing from PrusaSlicer fails frequently under linux. This pull request and https://github.com/prusa3d/Prusa-Firmware/pull/2008 is affected:
```
avrdude-slic3r: prusa_init_external_flash(): MK3 printer emitted incorrect start code: `echo: `
avrdude-slic3r: arduino_open(): Failed to initialize MK3 external flash programming mode
```

mkbel:
Problems have probably something to do with some pins shared between external SPI flash and LCD display.

DRracer:
@leptun and other community members? Could you please look into the issues @mkbel and others found? It looks to me like the SPI communication with the LCD and the secondary FLASH may have got too affected. We'd really like to have your LCD improvements merged, but we must temporarily suspend them ( #2018 ) because of this. Thank you ;)

leptun:
@DRracer @mkbel  I've found what was causing the lcd not to init after language update. Forgot to set data lines as output. :). I'll push more changes to the lcd_improvements branch including this fix.

As for the broken languages... I have no idea what might cause them. Could it be something similar to #1980 ?

DRracer:
@leptun great to hear that, your explanation sounds very reasonable! Please make a standalone PR, I'd really like to have this fixed and running. We'll test it right away.

leptun:
@DRracer @mkbel I think I found out what is causing the broken translations. It's the translated crash detection strings. If I revert them to the VT100 codes, the translations work (except the VT100 codes being printed on screen). I've tried multiple ways to fix this, but I can't figure it out. I'm still trying though, maybe I'll figure it out :|

DRracer:
@leptun those are the only strings in lang_en.txt (for example), which have a "\r" inside. What happens if you put just a space " " instead the "\r" (besides broken formatting)?

leptun:
@DRracer If I do that then the languages work, but the formatting is broken. My problem now is how to add a charcter that will be interpreted as \n by the code. Tried use \r before, but with same result. Tried to use \x0A as a newline character, but languages didn't works again. I tried the same with \x1A which can't be used on the display since it isn't in the charmap of the IC: same result.

Right now I am using '~' (0x7E) as another newline character. The problem with this though is that you can no longer use the builtin arrows of the display. There are still problems with this since it conflicts with octoprint and short file names. It also conflicts with the sheets naming, this is however easily fixable.

Also there is the problem that if I have \x0A and it is imediately followed by characters that it could interpret as hex it breaks again as it converts those as well. ex: "can\x**0Abe** turned".

I've seen other strings that are translated and are very long that have no \n in them, yet they seem to be working fine.

DRracer:
@leptun it seems there is much more into the language build process. Since we are talking about just 2 strings causing this, let's put spaces instead of "\r". I'll look into the language build process tomorrow. It should work with \x0a, but one must regenerate the dictionary, the .po files and then build the firmware.

leptun:
@DRracer Is there any reason for me to keep this PR open? It is superseded by #2064 anyway.

DRracer:
@leptun we had this PS as a reference and discussion forum for LCD improvements. You are right, FW 3.8 has significant LCD improvements and your buffer WIP supersedes both this and FW 3.8. I agree with closing this PR.

leptun:
@DRracer Ok. Closing

PR #1963, opened 2019-07-04
Multiple sheets

NotaRobotexe:


DRracer (approved):

F0x06:
@NotaRobotexe @DRracer I noticed missing characters on "Sheet" menu, for example `Select` is shown as `lect` is thix fixed on recent commits of `MK3` branch ? Otherwise this is awesome and perfectly working

Thanks

leptun:
@F0x06 I think I saw a commit that addressed this issue, so yes.

DRracer:
Guys, please wait a few moments with these recent commits for the upcoming FW 3.8 ;) . It may get changed in the near future (a week or so), the UI is still not finished, but we had to merge it into the trunk. But this feature will definitely be in 3.8.

PR #1979, opened 2019-07-08
Heatbed noise suppression

DRracer:
heatbed audible noise suppression using short fast PWM pulses with variable duty


F0x06:
Great Job, clean and well documented code :)

DRracer:
@F0x06 thank you ;) . Pleases note you'll need our new Arduino board definition because we had to define our own ISR(TIMER0_OVF). We are working on it right now (you might have noticed that in my repo). That will affect all builds of Prusa firmware from now on as soon as we roll out the FW 3.8

mkbel (approved):

F0x06:
@DRracer working very well, just a little question, is it normal that these changes increase power consumption of heat-bed during heat-up ? For example during PLA prints before with bed at 65°, when the bed kicked to stabilize (not first heating but during the print) i had 80W~ consumption, now i have spikes between 150-200W for 10s~. This is not an issue, just to understand, Thanks ;)

PR #1985, opened 2019-07-09
Save 8B of RAM and 4B of FLASH memory. Make code more readable.

mkbel:
State transition directly from RISE to FALL and opposite transition is not handled, as we are multiplying soft_pwm_bed by 2, so we expect minimum duration of pulse to be at least 2 so there should be no direct transitions between RISE and FALL.

### DIFF
--- a/Firmware/heatbed_pwm.cpp
+++ b/Firmware/heatbed_pwm.cpp
@@ -27,20 +27,6 @@ enum class States : uint8_t {
   FALL = 3
 };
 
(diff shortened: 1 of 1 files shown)


DRracer (commented):
The proposed changes are not more readable for me, moreover the new code does not retain the original functionality in all cases.

DRracer:
@mkbel Not applicable anymore. Would you mind closing this PR?

PR #1996, opened 2019-07-13
Incorrect ADC-Input for Ambient in Einsy 1.0 include

froks:
Ambient ADC in the Einsy board include was incorrectly set to 5 (A5) instead of 6 (A6), which is used in the schematic https://github.com/ultimachine/Einsy-Rambo/blob/1.0a/board/Project%20Outputs/Schematic%20Prints_Einsy%20Rambo_1.0a.PDF

The value 5 - even though it's wrong - works, because the `ADC_PIN_IDX` macro returns 5 in both cases (due to ADC5 not being included in `ADC_CHAN_MASK`).

### DIFF
--- a/Firmware/pins_Einsy_1_0.h
+++ b/Firmware/pins_Einsy_1_0.h
@@ -71,7 +71,7 @@
 #define HEATER_2_PIN        -1
 #define TEMP_2_PIN          -1
 
-#define TEMP_AMBIENT_PIN     5 //A5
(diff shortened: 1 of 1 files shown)


MRprusa3d:
... this task is coincidentally now in process - in couple of days we will know the final result ...
thank you for your interest ;-)

3d-gussner:
@froks Thanks closing it at it has been implemented 4 years ago

PR #2003, opened 2019-07-15
Heat hysteresis

mkbel:
Restore print fan speed earlier then nozzle temperature.
Revert "plan move to filament change position before enqued M600".
This was workaround to problem that restore_print_from_ram_and_continue() did 3s heating pause before resuming print. 

wait_for_heater() in restore_print_from_ram_and_continue() only if saved_extruder_temperature differs from target hot-end temperature.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -9476,7 +9476,8 @@ void restore_print_from_ram_and_continue(float e_move)
 //	    current_position[axis] = st_get_position_mm(axis);
 	active_extruder = saved_active_extruder; //restore active_extruder
 	fanSpeed = saved_fanSpeed;
-	if (saved_extruder_temperature) {
+	if (degTargetHotend(saved_active_extruder) != saved_extruder_temperature)
+	{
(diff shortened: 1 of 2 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -9474,15 +9475,16 @@ void restore_print_from_ram_and_continue(float e_move)
 //	for (int axis = X_AXIS; axis <= E_AXIS; axis++)
 //	    current_position[axis] = st_get_position_mm(axis);
 	active_extruder = saved_active_extruder; //restore active_extruder
-	if (saved_extruder_temperature) {
+	fanSpeed = saved_fanSpeed;
+	if (degTargetHotend(saved_active_extruder) != saved_extruder_temperature)
```
DRracer, line 9479: Is it necessary to have the saved temperature and new target one equal? Wouldn't some hysteresis of e.g. 3°C help prevent meaningless waiting for the extruder to reach exact temperature?

DRracer (approved):
all comments and questions explained

PR #2008, opened 2019-07-16
Lcd optimization - 1K of flash saved

leptun:
This PR is a followup to #1962 

It incorporates all changes that improve lcd functionality, but lacks the lcd_planner integration.
In total 1K+ of flash was saved by this and it should run a bit faster also.

You can read more about the changes in my original PR.
@DRracer Done :)

Please note that some of the support menus change a bit in formatting. It just makes the code a bit smaller and gives a more unified look to the menus.

### DIFF
--- a/Firmware/lcd.cpp
+++ b/Firmware/lcd.cpp
@@ -16,6 +16,13 @@
 //-//
 #include "sound.h"
 
+#define LCD_DEFAULT_DELAY 100
+
+#if (defined(LCD_PINS_D0) && defined(LCD_PINS_D1) && defined(LCD_PINS_D2) && defined(LCD_PINS_D3))
+	#define LCD_8BIT
+#endif
+
+// #define VT100
 
 // commands
 #define LCD_CLEARDISPLAY 0x01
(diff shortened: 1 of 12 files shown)


DRracer:
@leptun well done, thank you. I just skimmed though your PR and the changes are looking well written and consistent, even though there is obviously plenty of them. 
Could you please point us to the menus, that shall be especially checked for possible formatting changes (multiple languages)? I can see the changes themselves, but if you know about possible other menus that might have gone wrong. I'd give that list to our testing team as soon as we merge this.

leptun:
@DRracer 

- The temperatures menu has a slight change that should fix the two degree signs issue when temperature of nozzle or bed drops from 100 to 99
- The Voltages menu has a slight change.
- Belt status menu was shifted a bit to the left to match the formatting of other support menus
- WARNING: Crash detection disabled in Stealth mode
- Crash detection can be turned on only in Normal mode

Just realized I made a mistake in the translation files. Fixing now.

Will also check the changes on a real printer soon since I only have the MK2.5 "Test bench" at home.

Firmware/lcd.cpp
```diff
@@ -141,141 +138,98 @@ void lcd_no_autoscroll(void);
...
+	
+	#ifdef VT100
 	lcd_escape[0] = 0;
+	#endif
 }
 
 int lcd_putchar(char c, FILE *)
 {
 	lcd_write(c);
-	return 0;
```
mkbel, line 227: sketch/lcd.cpp:196:1: warning: no return statement in function returning non-void [-Wreturn-type]

mkbel (changes requested):

Firmware/lcd.cpp
```diff
@@ -55,73 +62,63 @@
...
+	WRITE(LCD_PINS_D5, value & 0x20);
+	WRITE(LCD_PINS_D6, value & 0x40);
+	WRITE(LCD_PINS_D7, value & 0x80);
+	
 	lcd_pulseEnable();
 }
 
-// write either command or data, with automatic 4/8-bit selection
-void lcd_send(uint8_t value, uint8_t mode)
+void lcd_send(uint8_t data, uint8_t flags, uint16_t duration = LCD_DEFAULT_DELAY)
```
mkbel: Can be made static, as is not used outside lcd.cpp nor has prototype i lcd.h

Firmware/lcd.cpp
leptun: @mkbel How do you think it would be more efficient: Re-add the return 0; or make function of type void?

leptun: Made it void for now. Can change later if you want.

Firmware/lcd.cpp
```diff
@@ -190,7 +190,7 @@ static void lcd_begin(uint8_t clear)
 	#endif
 }
 
-int lcd_putchar(char c, FILE *)
+void lcd_putchar(char c, FILE *)
```
DRracer: This function must return int, because it is passed as a function pointer into those streams - expecting to have signature of: int (*putc)(char c, FILE *) and shall return the character c if succeeded or EOF if failed. Nevertheless, the return value is not compared anywhere, so a simple "return c;" would do just fine.

DRracer:
@leptun please have a look at the lcd_putchar and related comments. Have a look at a related PR https://github.com/prusa3d/Prusa-Firmware/pull/2010 by @mkbel .

leptun:
@DRracer I see mkbel did the changes requested. I am cherrypicking the changes to this branch

mkbel (approved):

leptun:
@mkbel The last commit breaks the status screen. Fixing now.

GurliGebis:
@leptun you might want to do that change in a new PR, since this one has been merged (not sure a PR can be merged twice in github) 🙂

leptun:
@GurliGebis Apparently you can :) It’s already merged two times

Panayiotis-git:
I tried to compile the latest MK3 branch code. After uploading to the printer, my status screen has been broken:
DSC03229_p


The commits that seem to broke the status screen belong to this PR. Remove them and everything is ok. Add them and the status screen breaks.

It seems that at the ultralcd.cpp, the method 
`void lcdui_print_temp(char type, int val_current, int val_target)`
`{`
`	int chars = lcd_printf_P(_N("%c%3d/%d%c"), type, val_current, val_target, LCD_STR_DEGREE[0]);`
`	lcd_space(9 - chars);`
`}`

uses the return value of the lcd_printf_P, to detect the number of printed characters and fill the gap with spaces.
The problem is that the lcd_printf_P always return 0.

If I fix the size of the printed string to 9 characters and remove the call to the lcd_space, it is getting better.

leptun:
I will check this out. I thought this was fixed in #2012
I’ll try again soon with latest MK3. Shouldn’t make a difference, but maybe something broke it again. 🤷‍♂️

leptun:
@Panayiotis-git I checked the latest MK3 branch and I have no problem with the status screen.

leptun:
@Panayiotis-git Also did you reverse the polarizer on that lcd? it looks very nice :)

DRracer:
@leptun @mkbel I just looked at int fputc(int c, FILE *stream) in AVR LIBC:
The key lines are:
```
	if (stream->put(c, stream) == 0) {
			stream->len++;
			return c;
		} else
			return EOF;
```
So contrary to our expectation, lcd_putchar shall really return 0. Then printf routines are not confused by the underlying stream interface return values and behave correctly - i.e. the status screen looks as it should

leptun:
@DRracer Why does it work if it's void and no return then? Anyway, should I change it to int and return 0 in #2023 ?

DRracer:
@leptun I'd say its just an unreliable coincidence, which can be easily broken in the future, when the compiler decides to do some other optimization. Yes please, change it to int and return 0. You may also reference our conversation here. I just checked the code on my machine, the status screen behaves correctly with lcd_putchar returning 0.

leptun:
@DRracer Done.

Panayiotis-git:
> 
> 
> @Panayiotis-git I checked the latest MK3 branch and I have no problem with the status screen.

I changed the LCD module to one with black background. It is looking nice. With OLED it would be much better!

GurliGebis:
@Panayiotis-git OLED would have a high risk of burn-in, since it is showing the same (or close to) most of the time it is powered on 🙂

DRracer:
@leptun good job ;) . We'll merge it soon, right now we are in the process of testing ;) .

Panayiotis-git:
I confirm, that reverting the signature of the lcd_putchar in lcd.h and lcd.c and returning 0, fixed the problem.

Respect guys, keep going!

Panayiotis-git:
@DRracer I believe that something went wrong, because the changes of the commit ced514c to the lcd.cpp are not included in the current MK3 branch. Thus, the problems at the status screen still exist.

leptun:
@Panayiotis-git Are the issues fixed here? https://github.com/leptun/Prusa-Firmware/tree/MK3_fix_lcd_warning

I didn't have any problems with the status screen. It is possible that your avr-libc is acting up and failing to return the correct value from vfprintf_P() in lcd_printf_P().

Panayiotis-git:
@leptun Yes, the source at the linked github branch works fine.

I do not get it. I use the sloeber eclipse plugin for arduino projects. And compiling the project with sloeber, makes the status screen to fail.
I know another guy that had the same problem, but I don't know which IDE he uses.

I installed the Arduino IDE and compiled the same sources without the problem. So, I spent a couple of hours after the work comparing the two environments and it seems that sloeber uses different command line switches that the arduino IDE. Perhaps there is the problem.


Thanks anyway. If I manage to find something I'll inform you.

PR #2019, opened 2019-07-18
Fix lcd not initing after WDT reset

leptun:


DRracer:
@leptun thank you, we'll test it right away

DRracer (approved):

PR #2020, opened 2019-07-18
lcd_improvements fix languages

leptun:
@DRracer @mkbel This PR fixes corrupt languages caused by lcd_improvements.

The crash detection messages are changed to match the other non-blocking long messages by using `lcd_display_message_fullscreen_nonBlocking_P()` instead of `fputs_P()`. As a side effect the formatting gets a bit shifted in the "WARNING: ...." message and is compensated in every language in order for the rest of the text to be on the next row. Not the cleanest approach, but it works. I've checked other languages and they seem to display without broken words, but what I have noticed is that the menus in French language are quite weird and misaligned (that's how the translation is). It might be worth looking also into this.


leptun:
@DRracer Found a better way to handle linefeed in translations. I can use tab characters or in strings and have it work correctly. Much cleaner than using spaces to force the text on the next line.

Tried using \t instead of TAB, but the text shifts.

leptun:
@DRracer I've found a better way to do this. I've "created" (it's in the ANSI standard) a new escape code "\x1b[E" and it means linefeed. This is the only escape code allowed when VT100 is disabled and it's just skipped. With VT100 enabled it's parsed the normal way, but with the same outcome. I'm testing the languages themselves now since I've seen that these two messages are messed up in some languages (more than 20 chars per line). I'll push the changes once I'm sure everything is working. 

Btw I tried release 3.7.2 and my changes and both SD menus run at the same speed. Can you maybe make a video of the slow speed so I can compare?

leptun:
@DRracer The menus appear to be working and it only cost 48 bytes. But what I have noticed is that most languages are very inconsistent. I think the next thing on the TODO list is fixing the languages.
eg: 
- text overflows on next line. XYZ cal. details pages. Can be seen in italian.
- [On] and [on] one under the other.
- Inconsistent spacing in "Something       [STATE]" toggles

What I just mentioned is not caused by my changes. It's also in the 3.7.2 release and possibly earlier.

DRracer:
@leptun good job, I'll keep my eye on this in case my approach fails. 
I'm looking into the lang scripts ... I may have found a way to put simple \n into the source code and have it correctly transformed into the languages. Give me a few hours, I may fix it completely correctly in the scripts ... and we'll have a clean and predictable solution ;) ... and saving 48B ;)

DRracer:
@leptun ok, I've got my machine with correct CZ translation (no characters missing) and \n in the source codes. Now I must replicate the procedure and document properly, stay tuned ;) .
Btw.: lang_en.txt must include \x0a instead of \n . The same applies for lang_en_*.txt files. There are some nasty copy/conversion routines which rely on the fact, that the whole text is just a single line.

leptun:
That’s great to hear. Can you push the changes you made to this branch? This way everything is unified.

DRracer:
@leptun yeah, I will when I'm finished. I made a couple of script changes and hand-refinements in the temporary files during lang generation and I'm not completely sure which one solved the problem ;)

DRracer:
@leptun looks like my commit landed in your repo correctly, please have a look at it, the changes are minimal after all...

leptun:
@DRracer Can I force push to this branch without all the tries or should I just commit after your commit?

DRracer:
@leptun or make a new clean PR without all the history, that's a better solution.
Please test my commit on your machine too to confirm. Thanks

leptun:
Creating new PR.

PR #2022, opened 2019-07-19
Updated doxygen docs to work better with confluence.

kuhnmarek:


Firmware/Marlin_main.cpp
```diff
@@ -4935,54 +5046,74 @@ if(eSoundMode!=e_SOUND_MODE_SILENT)
             
         }
         break;
-            /**
+            /*
```
DRracer: why there is no /** here?

DRracer (approved):

PR #2023, opened 2019-07-19
Some more minor improvements to lcd

leptun:
@DRracer @mkbel This PR does the following:

- Fix compile warning introduced in #2012 
- Do not create custom █ character since it already exists in the lcd as character 0xFF. Used for sd card sorting bar
- Make lcd_createChar_P() static since it's only used in lcd.c
- Revert lcd_putchar to type int and retrun 0. Discussion: https://github.com/prusa3d/Prusa-Firmware/pull/2008#issuecomment-514500738

As a side note, is the LCD_STR_REFRESH character ever used in the code? It has value 0x04.


leptun:
@DRracer  Are you considering merging the two lcd related PRs?

DRracer:
@leptun yes, got interrupted with a higher priority task last week .

leptun:
@DRracer I did another small change. Now lcd_home function works faster. You can look at the changes and see why it works faster. I made sure lcd_scrollDisplay* functions still work, but I don't think these can be used on this firmware. There just isn't any way these can be useful. They are still there though and are enabled by LCD_SCROLL_SUPPORT define.

DRracer:
@leptun good job. We might have merged the LCD improvements too early into the 3.8, there are still some regressions. I'll be looking into the language generation process today to see if I can fix the "crash detected" messages without just replacing control characters with spaces.
Please have a look, why SD card file browser has gone notably slower. Would your recent fix solve it too?

leptun:
@DRracer I fixed the Crash detected messages. Please note that now it's a TAB character, not spaces. It's a placeholder for \n.
I'm now doing some work on the status screen. should make it faster. After this I'll go to 3.7.2 and take a picture of all menus in english and another language and make sure that with my changes all looks the same.
I didn't notice that the SD menu got slower. When did it get slower? I'll check that also. Yes, the "faster lcd_home()" commit should make each line 1.6ms faster.

EDIT: Nevermind. Can't make the status screen faster without problems. Also I am wrong about the sdcard menu being faster. It is NOT 1.6ms faster. The only menus that benefit from this are the ones that use lcd_home(), the sdcard menu not being among them.

DRracer:
@leptun yes, I'm aware of you TAB fix. But there should be a way to encode these special characters into the lang files without directly inserting TAB characters into the source code. Relying on a TAB in the source code is not the best practice - readability and stability is questionable and some IDE like to change TAB->spaces without notice. But the main problem are the generation scripts and external translations.
SD menu: its a few days back, we still haven't tracked down whether it was our fault or some regression of your LCD refactoring. Could you please check it? Try long filenames (what PrusaSlicer does) on the SD card and 20+ files in 5 directories each.

leptun:
@DRracer The thing is I need something in the string that doesn't trip the lang scripts and also that isn't a printable character. TAB is an example of this, but this is where the examples end. For the scripts to not fail you have to use a character from the keyboard or use \xAA. The problem with the first is that TAB is the only character that doesn't overlap with the lcd character map. \xAA has its own problems, mainly that it also takes \xAAAA and \xAAAAAA. You see where I'm going with this. Some translated strings also cause the next char to be taken as well in the conversion. I'll try some more ideas I have. Let's see how it goes. Btw, when are you planning to release 3.8? Just so I know how much time I have left :)

DRracer:
@leptun yes, I understand your motivation completely with the TAB character ;)
3.8 RC1: yesterday :) ... as usually

leptun:
@DRracer I am closing this for now since I am working on the lcd_buffer and some of the commits will change. All of the changes mentioned will be included, but not from this PR. Anyway these changes aren't that significant and don't need to be rushed, so they can wait for a bit.

PR #2032, opened 2019-07-23
detection of faulty RAMBo (incorrect capacitor on fan tach pin) MK3/S

DRracer:


mionut:
Hello i tested your code on head of MK3 branch yesterday, and the values for tachMeasure  with micros() is sometimes negative(if i remove the unsigned prefix), maybe because timer 2 is now used for timing and the function micros() was replaced with micros2(). You can use a macro named _micros() wich uses the correct timer. With _micros() it works ok for me.

DRracer:
@mionut thanks for pointing out the _micros() macro, but please read very carefully the comments at the function - this is a code for testing the RAMBo's and we are still trying to confirm, that the CPU pins survive switching from OUTPUT to INPUT so hard. 
**Only if confirmed, that the CPU goes unharmed**, it will get merged into 3.7.3 as a release patch.
Btw. the timer0->timer2 switch is for the upcoming FW 3.8, where this was originally developed

PR #2038, opened 2019-07-24
Add a threshold distance to Z calibration

Eriobis:
Allow a distance threshold when Z calibration is finding the bed

### DIFF
--- a/Firmware/Configuration.h
+++ b/Firmware/Configuration.h
@@ -291,6 +291,8 @@ your extruder heater takes 2 minutes to hit the target on heating.
 
 #define Z_HEIGHT_HIDE_LIVE_ADJUST_MENU 2.0f
 
+#define HOME_Z_SEARCH_THRESHOLD 0.15f             // Threshold of the Z height in calibration
+
 //============================= Bed Auto Leveling ===========================
 
 //#define ENABLE_AUTO_BED_LEVELING // Delete the comment to enable (remove // at the start of the line)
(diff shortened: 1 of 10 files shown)


leptun:
Umm... @Eriobis I don't see how your changes affect anything in the firmware. `HOME_Z_SEARCH_THRESHOLD` isn't used anywhere.

Eriobis:
> Umm... @Eriobis I don't see how your changes affect anything in the firmware. `HOME_Z_SEARCH_THRESHOLD` isn't used anywhere.

Yeah ... while fixing the commit I omitted the part of the code that works with that ... I'll push a fix

DRracer:
@Eriobis thank you for your PR.
@leptun good catch, I was wondering where should the new constants be used in the code. Now it makes sense. May be we could save some cycles (and code size) if the coordinates weren't compared as floats.

Eriobis:
I didn't want to go with a big threshold, but it make sens to have a +/-1mm threshold and only use integers.
Even at 4mm I don't think the nozzle could touch the bed and damage something. The goal here, if i'm not mistaken, is only to make sure the probe works before moving and damaging the bed.

I'm sorry I don't have time to take a deeper look at this issue, I just found really odd that my printer was failing calibration after my mk2.5 upgrade and that I've been printing for >2 years without any issue :P.

DRracer:
@Eriobis thats ok, we'll look into it ;)

MRprusa3d (approved):

Issue #2045, opened 2019-07-25
M300 reset


bmuessig:
Trying the following command will emit an approx. 4s tone (which the buzzer struggles to do) and in turn crash and reset the printer:
`M300 P10000 S7000`

I suspect that there could either be an integer overrun or - much more likely - a WDT reset due to the too long tone. Much larger time values only yield a short tone.

leptun:
@bmuessig Is this happening in one of the releases or in the latest Mk3 branch?

bmuessig:
@leptun This is happening in release 3.3.0. I unfortunately cannot test any more recent version.

bmuessig:
I guess that this has not been fixed in the meantime though.

Panayiotis-git:
The issue exists also to the latest version

leptun:
Found the problem. I'll test tomorrow and maybe even make a PR.

techfreek:
Still seeing this on 3.9.1 (albeit with a 5 second beep, not a 4 second beep)

Jaxx005:
@techfreek  [edit: this line added in case not seen, as post was already closed :)  ]

Latest release FW (V3.10.0) on MK3S has the problem - it seems related to the P period:

M300 P2000 S0  ; silence for 2sec as expected
M300 P5000 S0  ; causes printer restart!

Pronterface test confirms this line as problematic - note the ERROR after reboot (??).
Here's the Pronterface USB/serial output for M300 with P2000 and then P5000:

_**>>>M300 P2000 S0**_
SENDING:M300 P2000 S0
**_>>>M300 P5000 S0_**
SENDING:M300 P5000 S0
start
echo: 3.10.0-4481
echo: Last Updated: May  6 2021 13:52:17 | Author: (none, default config)
Compiled: May  6 2021
echo: Free Memory: 2054  PlannerBufferBytes: 1760
echo:Stored settings retrieved
adc_init
Extruder fan type:
NOCTUA
CrashDetect ENABLED!
E-motor current scaling enabled
E-motor current scaling enabled
E-motor current scaling enabled
E-motor current scaling enabled
E-motor current scaling enabled
E-motor current scaling enabled
FSensor ENABLED (sensor board revision: 0.4 or newer)
E-motor current scaling enabled
Sending 0xFF
**_Error:volume.init failed
[ERROR] Error:volume.init failed_**

LCD status changed
MMU not responding - DISABLED


NOTES:  
[...]

PR #2046, opened 2019-07-26
Use delay_keep_alive if sound is long

leptun:
@DRracer This fixes issue #2045 as it prevents a M300 or any other sound that is long to trigger the WDT reset. Also manages heaters and updates screen if the sound is longer than 100ms. Also restructured Sound_MakeCustom().
I chose 100ms since it is the default refresh rate of the lcd. Can be made higher to avoid the use of this function on some of the error sounds. eg: fan error

### DIFF
--- a/Firmware/sound.cpp
+++ b/Firmware/sound.cpp
@@ -62,34 +62,25 @@ Sound_SaveMode();
 }
 
 //if critical is true then silend and once mode is ignored
-void Sound_MakeCustom(uint16_t ms,uint16_t tone_,bool critical){
-     if (!critical){
-          if (eSoundMode != e_SOUND_MODE_SILENT){
-               if(!tone_){
-                    WRITE(BEEPER, HIGH);
-                    _delay(ms);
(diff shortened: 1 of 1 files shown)


DRracer:
@leptun good idea, however potentially dangerous. Are you sure there are no scenarios in which delay_keep_alive (or any of its callees) don't end up in lcd_* and UI functions? That's a nasty habit in the firmware (not saying its your fault ;) ), that's causing a lot of UI-related bugs and even stack overflows. Can you please check, that delay_keep_alive is "safe"?

leptun:
@DRracer If that is so, then I make sure lcd_update() is disabled during the sound and reenabled afterwards if needed (I use LcdUpdateDisabler class). This should prevent any lcd_* functions from being run during the sound.

leptun:
@DRracer Will you look into this or do I close the PR?

leptun:
Closed since it might introduce more problems than it solves. People shouldn't do long beeps anyway

PR #2056, opened 2019-07-29
Correct bug in util.cpp is_whitespace_or_nl function.

BoCoKeith:
Change 'r' to '\r'.


DRracer:
@BoCoKeith a correct fix, thank you. The "good news" is that the function isn't used anywhere ;)

DRracer (approved):

PR #2057, opened 2019-07-29
PFW-1055: M350 for all axis - MK3

### LINKED ISSUE #2024: M350 non-functional for X/Y/Z
Running 3.7.2 firmware on a Mk3.

Observed:
M350 commands are not stored/working. 
Via pronterface, I issued M350 X8 Y8 to halve the microsteps. Following this with an M503 reveals the values are still fixed at 16. (Tried both with M500 and without printer reboot in between. M350 and M503)

However, M350 E16 does, in fact, work to change the extruder microsteps from the default 32 as reported by M503.

This is confirmed by inspecting the code; the M350 handler only looks for an E-value.

Expected:
M350 X/Y/Z to function the same as M350 E

leptun:
Closes #2024

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -7538,27 +7538,34 @@ Sigma_Exit:
     case 350: 
     {
 	#ifdef TMC2130
-		if(code_seen('E'))
+		for (int i=0; i<NUM_AXIS; i++) 
 		{
-			uint16_t res_new = code_value();
-			if ((res_new == 8) || (res_new == 16) || (res_new == 32) || (res_new == 64) || (res_new == 128))
+			if(code_seen(axis_codes[i]))
 			{
-				st_synchronize();
-				uint8_t axis = E_AXIS;
-				uint16_t res = tmc2130_get_res(axis);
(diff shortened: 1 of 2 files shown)


vintagepc:
I've applied this to a local build and it does indeed work as expected.

leptun:
Testing: Try setting a different resolution for X or Y (eg 32) and see if everything still works. Also test saving the values to eeprom

Firmware/Marlin_main.cpp
```diff
@@ -7538,27 +7538,34 @@ SERIAL_PROTOCOLPGM("\n\n");
...
 				{
-					uint16_t fac = (res / res_new);
-					cs.axis_steps_per_unit[axis] /= fac;
-					position[E_AXIS] /= fac;
+					
+					st_synchronize();
+					uint16_t res = tmc2130_get_res(i);
+					tmc2130_set_res(i, res_new);
+					cs.axis_ustep_resolution[i] = res_new;
+					uint16_t fac;
```
DRracer: even though extracting this var away from the if-else subtree may look like correct programming paradigm, however it may also prevent optimization

DRracer (approved):

PR #2060, opened 2019-07-30
Make M109 S not wait for cooldown

### LINKED ISSUE #1474: M109 S always waits for cooldown
Hello,

I'd like to file a feature request. Very frequently I see start g-codes that first heat up the nozzle and bed to an intermediary temperature, then start mesh bed levelling, then wait for the final printing temperature to be reached. This is pretty clever as it saves some time and prevents oozing during bed levelling.

It doesn't work so well when you are printing several jobs in a row. The nozzle and bed will still be hot, so M109 and M190 will wait for them to cool down before continuing, only for the job to heat them up again immediately afterwards.

My proposal is to add a parameter to M109 and M190 that tells the printer to wait for both to heat up, but if the current temperature exceeds the requested temperature, simply continues without pausing.

Thank you!

haarp:
Make the Prusa firmware follow Marlin specs, as per http://marlinfw.org/docs/gcode/M109.html

fixes #1474

Flash: +78
RAM: +1

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -235,6 +235,7 @@ uint8_t newFanSpeed = 0;
   #endif
 
 static bool cancel_heatup = false;
+bool CooldownNoWait;
 
 int8_t busy_state = NOT_BUSY;
 static long prev_busy_signal_ms = -1;
@@ -6046,17 +6047,19 @@ Sigma_Exit:
     */
     case 109:
     {
+      CooldownNoWait = false;
       LCD_MESSAGERPGM(_T(MSG_HEATING));
-	  heating_status = HeatingStatus::EXTRUDER_HEATING;
+      heating_status = HeatingStatus::EXTRUDER_HEATING;
       prusa_statistics(1);
 
 #ifdef AUTOTEMP
         autotemp_enabled=false;
-      #endif
+#endif
       if (code_seen('S')) {
-          setTargetHotend(code_value());
-            } else if (code_seen('R')) {
-                setTargetHotend(code_value());
+        setTargetHotend(code_value());
+        CooldownNoWait = true;
+      } else if (code_seen('R')) {
+        setTargetHotend(code_value());
       }
       #ifdef AUTOTEMP
         if (code_seen('S')) autotemp_min=code_value();
@@ -6078,7 +6081,7 @@ Sigma_Exit:
       wait_for_heater(codenum, active_extruder); //loops until target temperature is reached
 
         LCD_MESSAGERPGM(_T(MSG_HEATING_COMPLETE));
-		heating_status = HeatingStatus::EXTRUDER_HEATING_COMPLETE;
+        heating_status = HeatingStatus::EXTRUDER_HEATING_COMPLETE;
         prusa_statistics(2);
         
         //starttime=_millis();
@@ -6102,7 +6105,7 @@ Sigma_Exit:
     case 190: 
     #if defined(TEMP_BED_PIN) && TEMP_BED_PIN > -1
     {
-        bool CooldownNoWait = false;
+        CooldownNoWait = false;
         LCD_MESSAGERPGM(_T(MSG_BED_HEATING));
 		heating_status = HeatingStatus::BED_HEATING;
         prusa_statistics(1);
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -9722,6 +9722,7 @@ static void wait_for_heater(long codenum, uint8_t extruder) {
     cancel_heatup = false;
 	while ((!cancel_heatup) && ((residencyStart == -1) ||
 		(residencyStart >= 0 && (((unsigned int)(_millis() - residencyStart)) < (TEMP_RESIDENCY_TIME * 1000UL))))) {
+		if ((CooldownNoWait == true) && !target_direction) break;
 #else
 	while (target_direction ? (isHeatingHotend(tmp_extruder)) : (isCoolingHotend(tmp_extruder) && (CooldownNoWait == false))) {
 #endif //TEMP_RESIDENCY_TIME


DRracer:
postponing to future release, this can be a breaking change of behavior, needs broader discussion

bwright86:
I would like to see this feature added. I have been looking for a way ignore the "Blocking" `M109` command, if my extruder is already above that temperature. This would allow my prints to start quicker, versus having to wait for the extruder to cool down to a midway temperature, then return to the full temperature.

This would be a breaking change, but I don't think it would be a negative one.
People are probably not using the "R" switch, since it has the same effect as "S", and "S" is the first switch in the command.

The scenario for using the `M109` gcode is to block the queue before bed calibration, to ensure the extruder is to temperature. This is likely to prevent a cold blob on the extruder from causing issues during calibration.

I do not see this having a negative effect the other way around. If people were actively using the "R" switch, I don't believe it would negatively affect their print if it didn't block until the extruder came down to that desired temperature.

Here is an example of the current Prusa Start gcode:
(Original Prusa i3 MK2.5S MMU2S Single - PrusaSlicer v2.2.0)
```gcode
Tx
G28 W ; home all without mesh bed level
G80 ; mesh bed leveling
```

The scenario I use in my print jobs is to:
- Set the heat bed to full temperate (Non-Blocking)
- Set the extruder temp to partial temp 180C (Blocking)
- After reaching the temp, set the full extruder temp (Non-Blocking)
- Initiate the mesh bed leveling command while the extruder and heat bed come up to full temp.
- After calibration, I set the heat bed full temp (Blocking) and the extruder full temp (Blocking), then finally start the print job.

This has the benefit of performing the mesh bed calibration while everything is still heating up. But preventing issues with a cold blob from a cold extruder.

Here is an example of my start Gcode:
```gcode
M109 S180 ; wait for extruder temp to reach 180 degrees.

G28 W ; Home all Axis
G80 ; Automatic mesh bed leveling

G1 Z20 F800  ; Raise Z-Axis while heating
```

Not having the "R" switch available causes issues with my scenario, when the extruder is already to full temperature (say 240C), and I start a print job. I must now wait for the extruder to cool to 180C. Rather than immediately starting the bed calibration, because the extruder is already hot enough to prevent a cold blob issue.

Please consider adding this to an upcoming firmware release.

mmirate:
@bwright86 Why not (a) install a SuperPINDA and then just run the G80 on a cold hotend, and/or (b) adjust your end-of-the-print gcode to retract filament all the way out of the hot zone? (without unloading it from the extruder gears)

bwright86:
Hey @mmirate, that is a good point, and I just upgraded to an MK3S recently, so it has the SuperPINDA.

But it is a common scenario to still have a blob of filament on the end of the extruder after a print (no matter the retract logic), so usually the extruder is heated to make it soft.

I did mention this near the end of the comment, as a cold blob issue. If the extruder is cold and the blob hits the print bed, this can cause calibration issues. So SuperPINDA wouldn’t solve for that scenario.

Thanks for the comment!

mmirate:
> Hey @mmirate, that is a good point, and I just upgraded to an MK3S recently, so it has the SuperPINDA.
> 
> But it is a common scenario to still have a blob of filament on the end of the extruder after a print (no matter the retract logic), so usually the extruder is heated to make it soft.
> 
> I did mention this near the end of the comment, as a cold blob issue. If the extruder is cold and the blob hits the print bed, this can cause calibration issues. So SuperPINDA wouldn’t solve for that scenario.
> 
> Thanks for the comment!

Oh, sorry, I wasn't precise enough.

If you retract all the way out of the melt zone at the end of the print; then at the beginning of the next print, you should be able to heat the hotend all the way to the first-layer print temperature - softening the blob and probably increasing accuracy on PINDAv2 - without any more plastic oozing out from inside the nozzle, because it's too high to be melted by the hotend.  
Hence my confusion.

leptun (changes requested):
This PR does not do enough to be able to use M109 S and M109 R. This is because `CooldownNoWait` only gets set properly in M190. With M109, R and S have the same meaning at the moment and never update `CooldownNoWait`. If that part is fixed, the PR should work properly. Until I get to fix that, I'm blocking merging of this PR

leptun (approved):
I've synchronised this PR with the upstream changes and fixed the issues in my last review. I also tested the PR and everything seems to be working according to the spec (both bed and hotend).

wavexx (approved):

3d-gussner:
@haarp Can you please rebase the PR (again sorry) for the last time and fix the conflicts before we can merge it.

bwright86:
Good to see this code going into the firmware. It is still relevant and
would reduce time during the initial bed leveling process.

3d-gussner:
@wavexx @leptun @gudnimg Please review it after my rebase and conflict fixes.

Firmware/Marlin_main.cpp
```diff
@@ -235,6 +235,7 @@ uint8_t newFanSpeed = 0;
   #endif
 
 static bool cancel_heatup = false;
+bool CooldownNoWait;
```
wavexx, line 238: After re-reading this, it seems that wait_for_heater is also called from restore_extruder_temperature_from_ram, so the last value of M109 and M190 will affect the behavior of wait_for_heater, which is incorrect.

I don't see a reason why we need a global for this. I would add a parameter to wait_for_heater, with the default being false to match the previous behavior.

Firmware/Marlin_main.cpp
```diff
@@ -6102,7 +6105,7 @@ SERIAL_PROTOCOLPGM("\n\n");
     case 190: 
     #if defined(TEMP_BED_PIN) && TEMP_BED_PIN > -1
     {
-        bool CooldownNoWait = false;
+        CooldownNoWait = false;
```
wavexx, line 6108: As written before, this would cause M190 to affect the wait for extruder temperature during recovery.. which is not good. I would keep this one as-is.

wavexx (changes requested):

PR #2064, opened 2019-08-01
⚡️LCD Buffer implementation

leptun:
This is a work in progress version of the lcd buffer implementation. So far the buffer exists and data can be put/read from it correctly. In the current stage of the PR there is no way for the data to be pushed to the lcd. The ISR isn't currently implemented.

Here is a video where I made it so it prints to serial continuously: https://photos.app.goo.gl/tAMGANgmKMPW1fTDA
This is not how it is in the current PR, but I thought it was pretty cool and wanted to make sure it doesn't corrupt or something.
Also the M998 command can be used to get a full lcd buffer sent over serial. It is a bit buggy on the client side since it cannot render the custom lcd characters and displays '~' for '->' and the other custom characters are simply skipped.

@3d-gussner @wavexx @DRracer @mkbel Is anyone interested in implementing the required ISR for the display? I tried doing it, but it was a fail and it didn't work. I'll try it again another time and maybe I'll get it right that time.

Any suggestions are appreciated.

Also it is important to keep in mind that all those functions of the lcd with the cursor and scrolling cannot be used at all with this PR because the cursor moves continuously. They have been removed.

Also I might create another PR with the final changes when this is complete since I see quite a lot of commits from a long time ago. Also it includes the changes from #2023 

One more important thing I forgot: If you make the changes for the MK2.x, use TIMER3, if you make the changes for MK3(S) use TIMER5 or TIMER3 with the analogWrite in `void Setup` removed! (or changed to WRITE(LCD_BL_PIN,HIGH)) !!!

PFW-977


DRracer:
@leptun good job ;) . I'll definitely want to help you with this, just give me some more time. We really need to roll FW 3.8 out.

maximlevitsky:
I can help too on this weekend!

leptun:
@DRracer Sure. This is not targeted for 3.8 anyway. Take your time to release it and after everything is sorted out with it we can start implementing this in the next major release.

@maximlevitsky Sure :) everyone is welcome.

GurliGebis:
Just a thought for the ISR routine - would it be an idea to have a "changed" flag be set every time it should be updated, and then only refresh the screen if it is set (and then reset it) ?
AFAIK (I might be wrong), there is no need to refresh the screen if what is already on the screen matches what will be drawn.

leptun:
That is an option, not sure if there is enough sram for everything. There could be another array where all changed stuff reside (aka a bit map).

I’m not sure how it should be handled when lcd_update triggers. I think the best option would be for lcd_write to compare the new value with the value in the buffet and write it and set the appropriate bit if it doesn’t match.

DRracer:
@GurliGebis a short buffer for recording "damaged" characters ... i.e. positions. If there are too many (i.e. buffer full) since last update, do a full refresh. Otherwise try to send only the changes into the LCD. I've done that 19 years ago on a similar semigraphic system and it worked very fast and it was also very reliable ;)
Triggering update shall be done in the ISR - say every 100ms. Just check, if there are any changes to play with... and once upon a time do a full lcd refresh
And the ISR shall primarily handle all the waiting - just plan next step for another tick, must not block the cpu unnecessarily.

leptun:
I was thinking there could also be a variable that keeps track of the number of changed characters, so the ISR would know when all characters have been printed without the need to check the bitmap.

GurliGebis:
Since it is 20x4, it is 80 bits or 10 bytes, so the load on memory should be pretty compact, if you want a bitmap of what needs to be redrawn.

DRracer:
a bitmap is another option, I agree. However, you'd need some function to check if there are any changes at all. So you end up with another variable saying all clean / something dirty.
A 5 item buffer (x,y may be encoded like nibbles, so 1 byte holds x and y coords) will be enough. If there are more changes we'd probably want a full refresh ...
Anyway, if there are more volunteers, we may test multiple approaches and pick the fastest one ;)

GurliGebis:
Why store coordinates for on the screen?
Shouldn't the position in the array define where it is located on the screen?
(something along the line of this, in pseudo code)

x = position / 10
y = position ^ 10
changed = position == 1

so if it is the 26th bit, it should be in x = 2, y = 6 and changed should be true.
That way all those three informations are stored in just 10 bytes.
It has been a while since I have written anything close to C code, but IIRC there is a fast and easy way to check if a range of bytes are all 0 (which would be the case if all 80 bits are "off") :)

EDIT: Yep, there is with memcmp https://www.thecodingforums.com/threads/memcmp-compare-buffer-for-0.754748/ :)

leptun:
I just pushed a commit where you can find some defines for timer registers on different boards.

leptun:
> Anyway, if there are more volunteers, we may test multiple approaches and pick the fastest one ;)

I agree with DRracer. Can everyone branch from this one and make PRs to my branch on https://github.com/leptun/Prusa-Firmware . I've created MK3_LCD_Buffer_leptun where I'll do my approach. I'll also make a PR when I make the first commit. Is this ok with everyone?

wavexx:
I don't know where exactly to comment, so I'm plugging this here =)

After a round of performance tests, aside from the bed PWM regression, the LCD handling is vastly improved compared 3.7.1. I still see stutters from lcd_refresh(), but it's nowhere as impactful as before. 

@leptun If I remember correctly in #1962 the idea was to effectively remove the delay and split the init into subsequent stages so that there's no hard-coded pause. Am I right?

If the chunks are guaranteed to be small enough also for init, and given we're almost there already, I would consider already closing #1940 in favor of this. The cost of planned_time would be simply wasted under most circumstances.

DRracer:
@wavexx are you referring to the lcd improvements already merged upstream or this LCD buffer @leptun is working on?
Subsequent changes in #1962: exactly - that was the original idea to do as little as possible in small chunks leaving the costy CPU cycles for more important work.
PWM regression: we'll get that fixed, I have a couple o ideas ;) . We got rid of our one and only blocker issue for the 3.8 today, so hopefully the tests tomorrow go fine and we'll have time looking into future speed and code size improvements.

wavexx:
@DRracer I'm referring to what's already in master, which is great!

leptun:
@DRracer Sooo... I'm a bit confused. Which lcd improvement would you like to keep working on?

Lcd planner: the one with the ring buffer for the display. Needs to be rebuilt because of incompatibility with the main branch. This one seemed to improve the stuttering, but as noted some time ago, it doesn't really improve much.

Lcd Buffer: What I am currently working on. Full display buffer in memory, keeps track of changes. Pr is here: https://github.com/leptun/Prusa-Firmware/pull/2 in case you didn't see it. It still needs some more time to fix all bugs, but it looks as if it might work. Video of it in action: https://photos.app.goo.gl/wQtUttd8Chh11byU7

I really need to know on which approach to focus on.

DRracer:
@leptun that's a bit of a misunderstanding ;) . 
1. Your work on the LCD planner is already merged and highly appreciated. The improvements are already visible. No doubt about that, @wavexx was only confirming the improvement.
2. Please keep on working on the LCD buffer, I'm convinced this advanced approach will free even more CPU cycles and will help us to simplify the UI logic in a second step. @wavexx will benefit from this even more when doing stepper speed optimization.

leptun:
It is possible that I've fixed all and every issue on my PR. Did anyone else try another approach. I'm interested in comparing the results :)

Also, if it's possible, can other people confirm that [my approach to the ISR](https://github.com/leptun/Prusa-Firmware/tree/MK3_LCD_Buffer_leptun) also works for them? My tests were done on an ebay display with MK2.5 firmware.

As a heads-up, I'll be gone from this Sunday to next Sunday and won't be able to work and test anything. This doesn't mean I can't answer on github, just that it will be slower.

DRracer:
@leptun good job, enjoy your holiday ;)

leptun:
@DRracer @GurliGebis I've merged my ISR implementation into this branch. I'll continue testing my approach and squash any bugs that I find.

GurliGebis:
Code looks clean and well written to me 🙂

leptun:
Hi everyone. I was recently looking in the datasheet of an OLED lcd that is mostly compatible with the one on the Prusa Printers and realized something from the example code. There might be a way to double the command speed while also making it more reliable.
Right now I am waiting a minimum of 100us for each nibble, but I might be able to send the first nibble and wait only 1us for it and after that send the second nibble and wait at least 100us.
As for reliability, I think if the nibbles follow each other so quickly, they will self-align if they get misaligned after a few commands. This is huge since I can simplify the ISR a lot by doing everything in only one isr cycle. I’ll test this during the weekend and report back the results.

DRracer:
@leptun I'd be very carefull waiting in the ISR for anything. If I'm not mistaken, one CPU cycle takes ~63ns, waiting a whole 1us look like much. Anyway, I'm curious about your results ;)

leptun:
@DRracer I think it's more efficient waiting 16 clock cycles rather than splitting the ISR in two parts and handling each of them separately. Also the code will get much more readable.

gudnimg:
Could you please update the PR with a quick rebase and resolve any merge conflicts?

```
git remote add upstream https://github.com/prusa3d/Prusa-Firmware.git
git fetch upstream
git rebase upstream/MK3
git push -f
```

I ask so I can test the firmware with all the latest changes :)

GurliGebis:
> 
> 
> Could you please update the PR with a quick rebase and resolve any merge conflicts?
> 
> ```
> git remote add upstream https://github.com/prusa3d/Prusa-Firmware.git
> git fetch upstream
> git rebase upstream/MK3
> git push -f
> ```
> 
> I ask so I can test the firmware with all the latest changes :)

Rebasing when the origin branch has been merged in previously can be quite a mess, so I would recommend just merging MK3 from upstream instead.

GurliGebis:
@leptun why close the PR?

leptun:
I’ve abandoned the idea. The benefits it brings are not worth it compared to the resources it would use. And besides, I have another plan for the LCD that would conflict with this PR. At least that PR has a chance of making it (#3383)
I might return to this PR at a later time, but in the meantime I’ll leave it closed along my other unmaintained PRs.

GurliGebis:
That makes sense

PR #2082, opened 2019-08-07
Fix for false detection of THERMAL RUNAWAY

egilkv:
The idea is that if it is detected that the extruder cannot keep
the temperature, but well before reaching the alarm limit, the cooling
fan is throttled. If that does not help, and the alarm limit is reached,
the THERMAL RUNAWAY alarm is triggered as usual.
This avoids the frustration of having large prints ruined by a falsely
detected thermal runaway when the only problem was that the cooling fan
speed was set too high.

This fix is tested on the MK2, where I believe this problem is really painful for
many. I would expect it may also be a problem on other Marks, and it should not
be too hard also to merge here.


wavexx:
At which hotend temperature did you experience this?

egilkv:
Sorry for late response, I did not see this mail before now.

It happens at normal PLA temperature, around 200, but really not 
terribly dependent upon printing temperature.

My printer as delivered is some sort of hybrid. The printer itself is a 
MK2 while the print head is MK2S. The heater is as delivered.

Perhaps this particular combination is especially prone to the problem, 
I do not know.

Before my fix, the Thermal Runaway was a true pita that caused a lot of 
frustration, since it would typically fatally stop a perfectly nice 
print after, say, 20 hours. I could "fix" it by adding some tape over 
the fan entry to reduce airflow, or setting the fan value to a low 
value, both rendering the cooling feature much less efficient.

Since I have noticed others have the same problem, it occurred to me 
that the implementation of Thermal Runaway is really too crude, and that 
a reduction of fan speed when tendencies of not properly being able to 
maintain head temperature is detected would be a perfectly valid safeguard.

Since adding this fix, I have never had any issues, but I do see the 
text that this feature has been triggered from time to time.

Cheers,

Egil

wavexx:
After working on the temperature model, I wonder what now happens with the current 3.12 (and later) FW since it's now much more sensitive to underperforming heaters.

If that has been updated from a MK2, I assume this is a 12V heater. @leptun since you have some experience with MK2, does a true 40W 12V heater perform worse than a 24V one?

leptun:
Yes, with the printers I had, 12V40W behaved more poorly than 24V40W, even if the power is the same. This part always made little sense to me. What might be happening is either that the resistance increases with temperature, hence limiting the effective power of the lower voltage heaters, or there are some differences in the hotends themselves (eg, more conductive heatbreak, worn out components, not perfectly straight fan shroud, etc). I didn't have time to look at this issue more in-depth.

leptun:
Also, a very important note: MK2 printers had both 30W and 40W heaters during the lifetime of the product. Keep that in mind as it might cause weird results if 40W are assumed.

wavexx:
Huh :(

The problem is that if you get a runaway, which is already pretty substantial, there's no way this will pass model check. Currently the fan is not allowed to overpower the heater (during calibration the heater+fan is not allowed to go down). I think this is a problem even on the MK3 with a a stronger LDO part fan.

I feel like that if your build can't take it, this should really be fixed by changing your slicing settings more than attempting to limit the cooling.

Phetdanai:
- [ ooh]

PR #2084, opened 2019-08-08
Fix for Toshiba FlashAir (or other) SD card initialization

rhounsell:
Add clock cycles between sending CMD0 and CMD8.

I found that in Sd2Card.cpp::init, after CMD0 is sent to my Toshiba FlashAir W04 64Gb card and gets the proper response, the CMD8 that is sent next is receiving back R1_ILLEGAL_COMMAND instead of the R7 response. After much googling, I added code after the CMD0 is sent and before sending CMD8 to repeatedly send 0xFF until the card sends back 0xFF (or it times out). This is apparently sometimes needed to give extra clock cycles to the card to allow it to do work to get ready for the next command. Some (but not all) cards require this to function correctly.

According to google, it's suggested to send 0xFF before every command, but at least for my card it was necessary only between CMD0 and CMD8.

See related issue https://github.com/prusa3d/Prusa-Firmware/issues/1685

Issue #1685, opened 2019-04-02
[3.7.0] FlashAir support brocken


m-cas:
If I select card type FlashAir the card will not be recognised. It will if I set it to normal.

m-cas:
Tested again, even if I switch off or reset the printer with the front X button, the card (if the setting is FlashAir) will not be recognized. Switching to normal everything seems to be working ok.

What is supposed to be different when Flashair is selected anyway?

kazoob:
For what it's worth, my FlashAir card is working on 3.7.0 on an MK3.

m-cas:
did you do anything special after the fw installation? In my case selecting in the settings “FlashAir” makes the card unreadable. Better question yet: how have you configured your FlashAir? That could be a difference...

kazoob:
I did not have to do anything special. My printer is set to FlashAir (not normal). I am not home so I can’t verify my FlashAir card config right now.

ochm:
My Toshiba FlashAair works corectly, printer is set to FlashAair. MK3 3.7.0-2201

m-cas:
Interesting, is it a v4? How is it configured?

-
From my phone while mobile, please excuse any errors.

ochm:
Sure it is v4. Config done exactly step by step by [guide](https://josefprusa.cz/bezdratovy-3d-tisk-snadno-a-rychle-sd-karty-toshiba-flashair/). 

Insert card to your PC/reader
On the card find and backup (just for sure) file named CONFIG, then edit them to something like:

[Vendor]
APPSSID=your-wifi-name
APPNETWORKKEY=your-wifi-password
CID=02544d535733324740e35b5979010b01
VERSION=FA9CAW3AW3.00.01
APPMODE=5
APPNAME=3D_printer
LOCK=1
PRODUCT=FlashAir
STA_RETRY_CT=0
UPDIR=/G-Code
UPLOAD=1
VENDOR=TOSHIBA
WEBDAV=2

Save the file.
Don't forget create directory used in UPDIR direction on your card.
Reinsert the card in your reader, or insert card to printer.

Check if the card is accessible by //ipaddress_of_card (you must reserve IP in your DHCP server for card connection)

> Interesting, is it a v4? How is it configured?

m-cas:
Thanks, here my V4 config file… we are pretty much on the same page.
I need to test more…
Btw you are behind with your sw update :D

[WLANSD]
ID=MACMK3S

[Vendor]
[...]

Issue #1225, opened 2018-10-04
FlashAir: firmware is unable to detect IP address


crazyquark:
I have exactly the same flash air version as the one in the tutorial on the blog.
I can find the card's IP address via my router's logs but it does not show up in the Support menu.
I am running FW 3.4.0 but I will update to 3.4.1 today.

crazyquark:
It's the same on 3.4.1

JoePrints:
What size is the card? There have been instances that users reported that the 64GB version does not work, but the 32GB works fine. I use the W-04 version on MK3 and MK2 and it works just fine for me- however I do not need to se what the IP is after reserving that IP on my router.

crazyquark:
It's a W-03 32GB card, I just upgraded it to the latest available firmware from Toshiba and the problem persists.
Yeah, sure it's not a BIG issue but maybe it's a fixable one. If need be I can provide some debug logs if you let me know who to do that(I suspect some SDIO command fails?)

SaijSaij:
I have the same problem. Card is a W-03 32GB and in the support menu there is no IP. But I can ping it and access it through HTTP and WebDAV. Card Firmware is FA9CAW3AW3.00.02 and printer firmware is 3.4.1-rc1-1354

stephanr:
Same here on MK2.5 with Firmware 3.4.1-1356
The Card is an W-03 with 8GB and Firmware FA9CAW3AW3.00.02
According to the Toshiba Website the Firmware V3.00.02 from April 24, 2017 is the latest available Version.
The Card can be used without Problem but the IP-Address is not shown in the Support menu.

nikst4:
Same here.
Toshiba Card W-04 with 32 GB.
Worked with firmware 3.4.0, now quit workling.

ptr727:
Similar problem.
W-04 64GB, currently the only model available on US Amazon.

Card comes formatted exFAT, Mk3 does not read exFAT, would be nice if you support exFAT.

I updated the card firmware.
I re-partitioned and reformatted the card to 31GB FAT32.
I reinitialized the card.
I edited the config to put it in STA mode.
The Mk3 can now read the SD card, and print files from the card.
I can connect over WebDAV and copy files to the SD card.

But, the Mk3 is not recognizing it as a FlashAir card, just a regular SD card.
After copying files to the card, I need to remove and re-insert the card, else the Mk3 does not see the newly copied files on the card.

nikst4:
[...]

PR #2088, opened 2019-08-09
Fix for Toshiba FlashAir (or other) SD card initialization

rhounsell:
Add clock cycles between sending CMD0 and CMD8.
I found that in Sd2Card.cpp::init, after CMD0 is sent to my Toshiba FlashAir W04 64Gb card and gets the proper response, the CMD8 that is sent next is receiving back R1_ILLEGAL_COMMAND instead of the R7 response. After much googling, I added code after the CMD0 is sent and before sending CMD8 to repeatedly send 0xFF until the card sends back 0xFF (or it times out). This is apparently sometimes needed to give extra clock cycles to the card to allow it to do work to get ready for the next command. Some (but not all) cards require this to function correctly.
According to google, it's suggested to send 0xFF before every command, but at least for my card it was necessary only between CMD0 and CMD8

See related issue https://github.com/prusa3d/Prusa-Firmware/issues/1685

Note: this pull request is a re-do of request #2084, but on the MK3 branch as suggested.

PFW-1071

### DIFF
--- a/Firmware/Sd2Card.cpp
+++ b/Firmware/Sd2Card.cpp
@@ -319,6 +319,21 @@ bool Sd2Card::init(uint8_t sckRateID, uint8_t chipSelectPin) {
       goto fail;
     }
   }
+
+  // send 0xFF until 0xFF received to give card some clock cycles
+  SERIAL_ECHOLNRPGM(PSTR("Sending 0xFF"));
+  spiSend(0XFF);
+  while ((status_ = spiRec()) != 0xFF)
+  {
+    spiSend(0XFF);
+    if (((uint16_t)_millis() - t0) > SD_CARD_ERROR_FF_TIMEOUT)
+    {
+      error(SD_CARD_ERROR_CMD8);
+      SERIAL_ECHOLNRPGM(PSTR("No 0xFF received"));
+      goto fail;
+    }
+  }
+
   // check SD version
   if ((cardCommand(CMD8, 0x1AA) & R1_ILLEGAL_COMMAND)) {
     type(SD_CARD_TYPE_SD1);
--- a/Firmware/Sd2Card.cpp
+++ b/Firmware/Sd2Card.cpp
@@ -321,6 +321,7 @@ bool Sd2Card::init(uint8_t sckRateID, uint8_t chipSelectPin) {
   }
 
   // send 0xFF until 0xFF received to give card some clock cycles
+  t0 = (uint16_t)_millis();
   SERIAL_ECHOLNRPGM(PSTR("Sending 0xFF"));
   spiSend(0XFF);
   while ((status_ = spiRec()) != 0xFF)
--- a/Firmware/Sd2Card.h
+++ b/Firmware/Sd2Card.h
@@ -105,6 +105,8 @@ uint8_t const SD_CARD_ERROR_SCK_RATE = 0X18;
 uint8_t const SD_CARD_ERROR_INIT_NOT_CALLED = 0X19;
 /** crc check error */
 uint8_t const SD_CARD_ERROR_CRC = 0X20;
+/** no response to sent 0xFF */
+uint8_t const SD_CARD_ERROR_FF_TIMEOUT = 0X21;
 
 /** Toshiba FlashAir: iSDIO */
 uint8_t const SD_CARD_ERROR_CMD48 = 0x80;


rhounsell:
Could reset t0 before sending 0xFF but seems to work fine as is.

grzlus:
Did you tried 3.8.0 firmware? Now my FlashAir works fine.

leptun:
@grzlus I don’t think there was any change regarding sd/flashAir. It might be just a coincidence. More testing is needed.

grzlus:
> @grzlus I don’t think there was any change regarding sd/flashAir. It might be just a coincidence. More testing is needed.

I've also changed LCD to Black one. Maybe I had faulty board at first, but now SD works fine. I don't need to restart (not power off) printer after every power off.

leptun:
@grzlus You are the second person I hear changing their display to black. Where did you get it from?

I also have a faulty sd holder on one of my display. It rarely reads correctly from a normal SD and you have to push it so it isn’t all the way in. Maybe that’s what fixed your problem. What I think is more likely is only some minor change in timing was required during versions (maybe the new lcd improvements make the firmware start faster) for flashAir to work again. Does it work reliably now?

rhounsell:
> Did you tried 3.8.0 firmware? Now my FlashAir works fine.

Nope - tried 3.8.0; card isn't recognized.

grzlus:
@leptun I got it from Zaribo (they have it in Europe). I got preassembled, but it's a crap (cold joins), so I re-soldered that. In America you can get it from PrinterSolid. My next update will be OLED, or some custom thing with adjustable brightness or on/off switch for backlight.

Yeah, I was doing this trick with pulling/pushing or restarting in the past. Now it works reliably. Since change I powered my prater 10-15 times and every time it worked.

I can send my old unit as testing unit to Prusa Lab to find root cause 😄 

---

@rhounsell so it must be the LCD with card reader.

leptun:
I've tested the PR and the results are as follows:
Without the PR, the card works in flashair mode, but it is not detected by the firmware as being a flashair. I am able to connect to the card and upload and the printer reacts correctly. The problem is that the printer doesn't detect the fact that the card is flashair.
With the PR nothing changed in my case. Maybe I don't have the timing issues as the others as my card always initializes correctly. It is a W-04 without any firmware update.

rhounsell:
I’ll have to check the FW version of my FlashAir card. It certainly seems like some folks have issues while others have none. On my printer (MK2.5), If I don’t have the patch, the printer doesn’t detect an SD card at all. I haven’t tried connecting via wifi to the card in that situation. I’ll add more info when I get home and have a look...

[edit] It's a Toshiba 64Gb W-04 card - The FlashAir tool 4.0.0 says that the FW is Ver. F15DBW3BW4.00.02.

vintagepc:
> I've tested the PR and the results are as follows:
> Without the PR, the card works in flashair mode, but it is not detected by the firmware as being a flashair. I am able to connect to the card and upload and the printer reacts correctly. The problem is that the printer doesn't detect the fact that the card is flashair.
> With the PR nothing changed in my case. Maybe I don't have the timing issues as the others as my card always initializes correctly. It is a W-04 without any firmware update.

@leptun The cards not offering any flashair features in the firmware is a separate issue; this is purely an initialization fix for those of us where our cards take longer to boot; as a result the printer thinks there is no SD card at all if you have this issue, and some degree of jiggling and manipulation is needed to get it to see the card.  My W-04 FA card has never done any of the stuff like show an IP address etc despite the menu option being set to "FlshAir".  (See #1225)

I have had this patch on my local build for a while and while I used to regularly have issues with card recognition I have not encountered them since applying it.

(I applied the fix quite a while ago, not long after it was first proposed here, so many testing hours/cycles on it).

leptun:
Update: I've started encountering the init error as described in the issue. Resetting the printer or doing the quick reinsert fixes the problem temporarily. I'll test this PR with my card and I'll post the results soon.
Differences compared to before: a LOT more files added to the card (1.1GB total). This is most likely the culprit.

rhounsell:
@leptun FWIW it happens 100% of the time to me, with an almost empty card. Hope the patch fixes the issue for you.

rhounsell:
@leptun Any update on whether this patch solved your issue?

leptun:
@rhounsell I wish I could say I encountered the issue again, but no. Since I can’t actually reproduce the issue again I can’t test. Sorry.

@DRracer what is your opinion on this?

vintagepc:
I had it pretty regularly and have not seen it since applying the patch; if there's anything I can contribute to aid in confirming/testing just let me know.

DRracer:
@vintagepc thanks for confirmation, right now you are one of the few who actually report some improvement with this patch. I'm not against it, we just couldn't verify it yet (or FlashAir cards seem to work with and without it)

DRracer (approved):

rhounsell:
Thank you.

rhounsell:
@michalxfanta FWIW you might consider mentioning this (#2088) pull request in the 3.9 release notes. A number of people have discussed in several issues a problem with their printer properly recognizing a Toshiba FlashAir card over the past year or so, although they seem to be a small percentage of users. Still, it may give some users some hope if they are still struggling with their FlashAir cards.

vintagepc:
^^ Definitely, now that FlashAir is also going to be a supported upload host in Plicer 2.2

PR #1664, opened 2019-03-20
"disable_z()" for Delta PSU

MRprusa3d:

PR #2107, opened 2019-08-14
Fix power panic

mkbel:
This fixes two problems with power panic. Power lost is sometimes not detected. Resumed Z level is sometimes lower.

This reverts "Merge pull request #1664 from MRprusa3d/PFW-811" and following compiler warning fixes. In original commit, there is suspicious sei() command in disable_force_z() without any cli() command.

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -166,17 +166,6 @@ void manage_inactivity(bool ignore_stepper_queue=false);
   #define disable_z() {}
 #endif
 
-#ifdef PSU_Delta
-void init_force_z();
-void check_force_z();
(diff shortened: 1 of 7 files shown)


DRracer (commented):
This is not a fix, but a removal and an especially brutal one - commenting out PSU_Delta define should be sufficient. We now know the problems are related to z-motors being switched to stealth mode. I'd rather prefer having the z-motors fixed.

DRracer:
Since there is already a PR correctly solving the power panic issue with the z-motors stealth switching, can we close this PR?

mkbel:
yes

PR #2127, opened 2019-08-21
Code size reduced by almost 5KB

DRracer:
Extracting common call of:
`plan_buffer_line(current_position[X_AXIS], current_position[Y_AXIS], current_position[Z_AXIS], current_position[3], ...`
saves almost 5KB. The performance penalty is negligible, since these planned lines are usually maintenance moves with the extruder.


XPila (commented):
All OK

XPila (commented):
OK

XPila (changes requested):
OK

XPila (approved):

PR #2134, opened 2019-08-23
Fix repeated power panic restarted print from beginning or jumped at …

mkbel:
…most 65536 B back in file printed from SD card.

As sdpos_atomic was not updated after printer power up and first power panic recovery, it was equal 0. When the first command from SD card was queued its size on SD card was computed as current SD index position minus sdpos_atomic. This was equal to offset from beginning of the file limited to 16 bit storage type. When next power outage occurred earlier then this command was finished and wiped out of queue, this command size (extraordinarily big) was subtracted from sdpos_atomic and saved to EEPROM. This led to up to 65536 B jump back in file printed after next power panic recovery.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -5343,7 +5343,9 @@ if(eSoundMode!=e_SOUND_MODE_SILENT)
     // ----------------------------------
     case 26: 
       if(card.cardOK && code_seen('S')) {
-        card.setIndex(code_value_long());
+        long index = code_value_long();
+        card.setIndex(index);
+        sdpos_atomic = index;
       }
       break;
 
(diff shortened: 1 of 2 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -5343,7 +5343,9 @@ if(eSoundMode!=e_SOUND_MODE_SILENT)
     // ----------------------------------
     case 26: 
       if(card.cardOK && code_seen('S')) {
-        card.setIndex(code_value_long());
+        long index = code_value_long();
+        card.setIndex(index);
+        sdpos_atomic = index;
```
DRracer, line 5353: is the non-atomic assignment ok here?

DRracer (approved):

PR #2141, opened 2019-08-27
Unify preheat menu, add ASA material

mkbel:
Saves 478 B of flash memory.


DRracer (approved):
Not sure if code guarded by TEMP_SENSOR_1_AS_REDUNDANT and WATCH_TEMP_PERIOD ever worked. We are removing it along with the preheat menus refactoring, as it has not ever been tested.

Issue #2089, opened 2019-08-09
Sheet rename cursor does not clear


vintagepc:
Found in 3.8.0-RC1, Mk3S.

The caret below the selected letter does not clear when pushing the knob to move to the next letter.

![IMG_0001](https://user-images.githubusercontent.com/53943260/62814108-a8dae500-badc-11e9-878a-bc1522ce3777.JPG)

Dippyskoodlez:
It felt like this was intended to signify the character had changed.

Issue #2093, opened 2019-08-10
proposal: position of the sheet selection menu


DesC21:
I think it would be nice if you could see which sheet is selected from the main menu before starting a print. When I start a print from the sd card, I have to scroll down to check if the correct sheet is selected. If the menu entry for the sheet selection was higher to the top, for example directly before or after "PRINT FROM SD" (I'd prefer after), the user could immediately see if the print can be started.

vintagepc:
I would advocate for something more visible too; I also nearly did a number on my smooth sheet when I'd accidentally had the printer set to the PC profile. A prompt prior to starting the print could be problematic in the octoprint/remote workflow but perhaps finding a way to display the selected sheet on screen during warmup and pre-heating with the X/Y/Z (X0/Y210/Z50?) in such a position that it's easy to quickly change sheets would be viable. Maybe replace "Bed heating... " with "[Sheet name] heating..."?

jbinfl:
Ohh, I like this suggestion and vintage PCs idea.  that would be very helpful if possible to implement!

TehH4rRy:
This has my vote! I've been lucky to only scuff my PEI by not checking which sheet I have on

vintagepc:
I would be willing to contribute some time to get this implemented and filing a PR; are there any existing leanings from the "official" dev team as to what/how?

Right now I'm contemplating doing something like replacing the %done with the sheet name until the print is underway (since it's obviously going to be 0 until we get up to temp and actually melting plastic), but if there are existing plans for shuffling display elements to accommodate it I would be happy to work within those plans.

vintagepc:
I decided to implement it on the info screen by replacing the percent done when either the printer is heating or there is no percentage (idle). Both regions consume exactly 7 characters, so it works very nicely. PR is filed.

DesC21:
perfect, thanks Prusa Team :-)!

PR #2145, opened 2019-08-29
Fix sheet reset

mkbel:

PR #2146, opened 2019-08-29
PFW-1061: #2093 - Relocate sheet menu, show sheet on info screen

vintagepc:
Implemented #2093 
-Sheet is shown instead of percentage done either if idle, or during
print preheat. (Both consume exactly 7 characters)
-Relocated sheet menu to below SD menu for additional visibility to
current sheet.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -6580,6 +6580,15 @@ static void lcd_rename_sheet_menu()
         lcd_putc(menuData->name[i]);
     }
     lcd_set_cursor(menuData->selected, 1);
+    if (menuData->selected>0)
+    {
+	lcd_set_cursor(menuData->selected-1, 1);    
+	lcd_putc(' ');
+    }
+    else
+    {
+	lcd_set_cursor(menuData->selected, 1);
+    }
     lcd_putc('^');
     if (lcd_clicked())
     {
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -600,7 +600,7 @@ void lcdui_print_percent_done(void)
 	bool num = IS_SD_PRINTING || (PRINTER_ACTIVE && (print_percent_done_normal != PRINT_PERCENT_DONE_INIT));
 	if (!num || heating_status) // either not printing or heating
 	{
-		eeprom_read_block(sheet, EEPROM_Sheets_base->s[selected_sheet].name, 7);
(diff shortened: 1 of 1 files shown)


DRracer:
@vintagepc  Good ideas, thanks for your PR. For now the sheets menu will be like we have it in #2145, but I really like your idea about the status screen showing sheet name instead of percentage in idle. I'll talk about it with our content guys.

vintagepc:
That's fair, looks like we essentially did the same thing for the menu move (though I didn't hide mine if there's only a single sheet in use)

vintagepc:
I'll merge to synchronize with #2145's changes and I realized I can also toss in a fix for #2089 which should be an easy fix.

vintagepc:
Fixed #2089, merged Mk3

Firmware/ultralcd.cpp
```diff
@@ -6542,6 +6552,15 @@ static void lcd_rename_sheet_menu()
         lcd_putc(menuData->name[i]);
     }
     lcd_set_cursor(menuData->selected, 1);
+    if (menuData->selected>0)
```
DRracer: please remove this part from the PR. We know you'd like the cursor to move (and not duplicate) under the edited character, but AFAIK there is another PR for that, isn't it?

DRracer:
@vintagepc I'm marking this as 3.9.x but please note we are still to gain some significant user response about the sheets, so this PR may require some more changes

vintagepc:
Understood, I'm hoping to keep it maintained as new features are added to the firmware.

vintagepc:
I merged 3.8.1-RC1 into my local branch which has this applied and it still functions as expected :+1:

DRracer (changes requested):
besides the cursor '^' change we are ready to merge this, please finish the PR as requested, thank you

Firmware/ultralcd.cpp
vintagepc: Not to my knowledge, I filed it as a separate issue in #2089 but did not make a separate branch or PR for this and a quick search in open requests did not turn up any results for someone else having implemented it.

If you'd still like me to remove it to a separate PR then I'm happy to do so, just say the word.

DRracer: please do so

vintagepc: Done

DRracer (approved):

vintagepc:
I guess I can cross  "Earn 'contributor' badge for Prusa Github" off my bucket list now  😄

PR #2213, opened 2019-09-16
PFW-943: IR sensor gen. II

MRprusa3d:
disconnected PCB detection


leptun:
Can we please get a more details about this new filament sensor? Who gets it? Is there going to be a MK3SS/MK3S2 printer? Are there any schematics/bom files available for the new sensor?

DRracer:
@leptun the new sensor board is equipped with a protection against improper installation (switched wires tolerant) and a resistor, that will allow detection of the sensor board itself.
The resistor may be added to existing sensor boards too.
The printer will then distinguish between non-functional filament sensor and loaded/not loaded filament even at runtime.

boraca:
A lot of people use the mod that adds an LED and a resistor to the filament sensor, to be able to calibrate it. or check if it's working properly without going to the sensor status menu. 
Those are the mods:
https://www.prusaprinters.org/prints/2135
https://www.prusaprinters.org/prints/2139
https://www.thingiverse.com/thing:3565326

It uses a 22K or 560R resistors. Will this change brake anything for those people?

PR #2217, opened 2019-09-17
Mm lay1 cal

mkbel:
Adapt menu and wizard for first layer calibration with any material. PINDA temperature compensation not checked yet.


ojrik:
Calibration of 1 layer does not run / settings / hw setup / steel sheets / sheet 0-7 / first layer cal. /filament 1-5 / type of filament, does not jump to first layer calibration.

DRracer:
@ojrik good catch, that way really doesn't work. Thanks for reporting

DRracer (approved):

PR #2218, opened 2019-09-17
[French translation] Wrong display during mesh bed calibration

Forfelet:
Hello,

During mesh bed calibration, values superiors to 9 are wrongly displayed.
For example, for the point 36, the "6" of "36" is displayed instead of the first position in the first line (instead of the thermometer symbol).

See example below:

The display is a LCD 20x4, but "Calibration de Z : 36" counts 21 caracters.

Best regards,
Cyril

### DIFF
--- a/lang/lang_en_fr.txt
+++ b/lang/lang_en_fr.txt
@@ -136,7 +136,7 @@
 
 #MSG_CALIBRATE_Z_AUTO c=20 r=2
 "Calibrating Z"
-"Calibration de Z"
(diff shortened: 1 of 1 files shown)


DRracer:
@Forfelet thanks for reporting, great timing btw, we are trying to polish the translations right now. Will be included in the next FW.

DRracer:
@Forfelet should be already included in the FW 3.8.1. Thanks for reporting, I'm closing this PR.

Issue #2179, opened 2019-09-08
FW 3.8.0, sheet rename menu. Holding button to adjust Z breaks menu.


BeldrothTheGold:
Hello Team, 

Love the new Sheets menu. Thanks for all of your hard work!

Noticed this bug quite by accident. 

It seems the button stops responding in the rename menu if you hold the button rather than quick push it. Like if you are attempting to adjust Z while in the rename menu. It persists even after the menu times out. I have to reset the printer to fix it.

Hope it's a simple fix. 

V/R
Scott

BeldrothTheGold:
@Panayiotis-git, @DRracer, @leptun 

I have tested the change @Panayiotis-git recommended from https://github.com/prusa3d/Prusa-Firmware/issues/1196#issuecomment-439501656.

This does indeed fix the bug which locks up the UI. I highly suggest this is tested and committed if it doesn't cause any regressions. 

Also, some brief testing of long-press in other menus with the fix in place validates @DRracer's statement that long-press still causes some strange behavior in other settings menus. 

I will work on this some more and submit another pull request.

BeldrothTheGold:
Looks like @DRracer was able to implement an excellent solution.  I will close this issue now.

DRracer:
@BeldrothTheGold thanks ;)

PR #2220, opened 2019-09-19
Fix long-press bug in sheets menu and cancels rename with long-press.

BeldrothTheGold:
This fixes issue #2179 and allows the user to cancel renaming sheets.

### DIFF
--- a/lang/lang_en_cz.txt
+++ b/lang/lang_en_cz.txt
@@ -819,8 +819,8 @@
 "Predehrev k vyjmuti"
 
 #MSG_SELFTEST_PRINT_FAN_SPEED c=18
-"Print fan"
-"Tiskovy vent."
+"Print fan:"
+"Tiskovy vent.:"
 
 #MSG_CARD_MENU
 "Print from SD"
--- a/lang/lang_en_fr.txt
+++ b/lang/lang_en_fr.txt
@@ -62,7 +62,7 @@
 "Auto home"
 "Mise a 0 des axes"
 
-#MSG_AUTOLOAD_FILAMENT c=18
+#MSG_AUTOLOAD_FILAMENT c=17
 "AutoLoad filament"
 "Autocharge du fil."
 
--- a/lang/lang_en_pl.txt
+++ b/lang/lang_en_pl.txt
@@ -1413,6 +1413,7 @@
 #
 "Sheet"
 "Plyta"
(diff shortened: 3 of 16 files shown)


DRracer:
@BeldrothTheGold basically a working approach, however let's generalize it - almost every other dialog suffers from this problem. Are you willing to work on this issues further?

BeldrothTheGold:
> @BeldrothTheGold basically a working approach, however let's generalize it - almost every other dialog suffers from this problem. Are you willing to work on this issues further?

Absolutely interested! What did you have in mind?

Panayiotis-git:
Does the fix referenced at the:

https://github.com/prusa3d/Prusa-Firmware/issues/1196#issuecomment-439501656

fix the problem?

DRracer:
@BeldrothTheGold @Panayiotis-git wouldn't it be better to specify only the allowed scenarios when the long press should work? Looking at the code - it does the opposite.
The long press should only work on the status screen in 2 modes:
1. move z when not printing
2. do live-adj-z for the first few layers when printing
Does it make sense?

leptun:
I think it should also do move-z / live-adjust-z when long-pressing in any of the menus (main menu, settings menu, etc.).

BeldrothTheGold:
@DRracer 

> @BeldrothTheGold @Panayiotis-git wouldn't it be better to specify only the allowed scenarios when the long press should work? Looking at the code - it does the opposite.
>The long press should only work on the status screen in 2 modes:
>move z when not printing
>do live-adj-z for the first few layers when printing
>Does it make sense?

 and @leptun 

> I think it should also do move-z / live-adjust-z when long-pressing in any of the menus (main menu, settings menu, etc.).

These requests seem mutually exclusive.

BeldrothTheGold:
Do we want to retain the current functionality where long-press in any menu will bring up the applicable Z adjustment, and only block this in cases where it causes issues? 

Or only bring up the Z adjustment while in the status menu? If this is the case would long press in menus then return to the previous menu by default?  Do nothing? 

@Panayiotis-git I will test this fix when I get home today.  I'll add it in a build without my commit to see if the bug which causes the UI to stop responding after a long-press in the rename menu is fixed. 

In anycase. I think it's still a good idea to have the long-press in an edit menu cancel the action and return to the previous menu.

PR #2227, opened 2019-09-22
Fix few translation files inconsistencies 

3d-gussner:
All translated files `lang_en_xx.txt` (xx=language) should contain the same format, syntax and text lines, except the translation itself of course.

I ran two scripts `/lang/lang-export.sh` and `/lang/lang-import.sh` scripts to be able to point to it as I will open another issue related to translations as well.

As you can see some aren't consistent and that might generate an issue with `.po` files which are send to translators or even generate issue compiling the language firmware.

You can see already here that the `/lang/po/Firmware.dot` file and other `/lang/po/Firmware_xx.po` files **do not** have the last message from `/lang/lang_en.txt` and `/lang/langen_xx.txt`  except the French translation file is consistent.

### DIFF
--- a/lang/lang_en_cz.txt
+++ b/lang/lang_en_cz.txt
@@ -819,8 +819,8 @@
 "Predehrev k vyjmuti"
 
 #MSG_SELFTEST_PRINT_FAN_SPEED c=18
-"Print fan"
-"Tiskovy vent."
+"Print fan:"
+"Tiskovy vent.:"
 
 #MSG_CARD_MENU
 "Print from SD"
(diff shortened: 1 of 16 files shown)


DRracer:
@3d-gussner good job, thanks for checking the translations, merging now

DRracer (approved):

3d-gussner:
@DRracer Did you see https://github.com/prusa3d/Prusa-Firmware/issues/2228?

3d-gussner:
Looks like this PR is the 1000th closed one ... Hooray

PR #2264, opened 2019-10-11
PFW-1048: fix selftest Z crash. Use stallGuard when testing Z

leptun:
Followup to #1703. The PR is redone in order to comply with recent changes from MK3 branch. As for changes compared to the previous implementation, I am now using the regular home function instead on the old "stepping" find Z function.
As a result, now selftest uses stallGuard when finding Z in order to prevent a nasty bed crash. Also the motion is much smoother since only one big move command is planned instead of many short ones.

@DRracer For testing, I propose that it should be tested on all supported printers: MK2.5(S) and MK3(S). Also, should I close the old PR?

PFW-1048


Firmware/tmc2130.cpp
```diff
@@ -994,7 +994,7 @@ bool tmc2130_home_calibrate(uint8_t axis)
 	uint8_t step[16];
 	uint8_t cnt[16];
 	uint8_t val[16];
-	homeaxis(axis, 16, step);
+	homeaxis(axis, 1, 16, step);
```
DRracer: true instead of 1

DRracer (approved):

PR #2266, opened 2019-10-11
Changed sheet menu to enable easy z height tweaks

Theta-Dev:
I absolutely love the new sheet menu that was introduced with the latest firmware version.

However it is quite tedious to manually adjust the z height of one of the sheets without starting a print or first layer calibration.
You had to select the sheet first, then go back to the settings menu, scroll all the way down to _Live adjust Z_ and then tweak your height.

So I made a small change and added the _Live adjust Z_ menu point to the lcd_sheet_menu, which activates the selected sheet and brings up the Z height selector.

I have already testet it and as far as I can tell it works nice.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -6770,6 +6770,12 @@ static void activate_calibrate_sheet()
     lcd_first_layer_calibration_reset();
 }
 
+static void activate_adjust_sheet()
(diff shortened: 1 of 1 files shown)


3d-gussner:
@Theta-Dev Thanks for the PR, but we can't merge it, sorry. Closing

PR #2268, opened 2019-10-12
Changed sound modes

Theta-Dev:
I removed the button sound from the ONCE sound mode so there is a sound mode for people who want all the prompt and custom sounds but dont like the annoying click sound every time you press the encoder.
So now there is a clear distinction between the 4 sound modes:
SILENT: no sounds except for alerts
ONCE: 1 beep alerts, prompts, custom sounds
LOUD: 3 beep alerts, prompts, custom sounds, click
ASSIST: 3 beep alerts, prompts, custom sounds, click, accessibiliy menu sounds

### DIFF
--- a/Firmware/sound.cpp
+++ b/Firmware/sound.cpp
@@ -6,6 +6,13 @@
 //#include <avr/eeprom.h>
 //#include "eeprom.h"
 
+/*
+SOUND MODES:
+SILENT: 1 beep alerts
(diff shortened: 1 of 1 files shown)


3d-gussner:
@Theta-Dev Thanks for this PR and the idea. At the moment I will close this PR as
- It would need a re-base
- It would differ from the 32-bit firmware.

PR #2271, opened 2019-10-13
Clear "sdprinting" state only when all SD moves are complete

wavexx:
We have a race condition where a filament sensor check and/or a power panic happening during the end of the print are left incomplete. This is due to the "IS_SD_PRINTING" flag being cleared as soon as all commands are queued instead of being cleared only when all commands are effectively performed.

This prevents the internal save_print_* functions to restore the current state of the print.

### DIFF
--- a/Firmware/cmdqueue.cpp
+++ b/Firmware/cmdqueue.cpp
@@ -582,30 +582,8 @@ void get_command()
        ((serial_char == '#' || serial_char == ':') && comment_mode == false) ||
        serial_count >= (MAX_CMD_SIZE - 1) || n==-1)
     {
-      if(card.eof()){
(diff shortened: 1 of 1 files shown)


DRracer:
@wavexx this might be the root cause of issues we've seen on 3.8 RC2 - some specific g-codes ended with the print head standing in the print out.

wavexx:
Unless there's interest in testing some odd corner cases with _just_ this fix, I'd actually close this and keep just #2274 instead

DRracer:
@wavexx OK, let's close this PR and keep on with the #2274

wavexx:
Closing in favor of #2274

PR #2274, opened 2019-10-15
PFW-1028: Fix recovery from filament sensor checks / crash detect / power panic

wavexx:
This PR fixes several issues with mesh_plan_buffer_line which cause undefined/broken behavior when the printer performs a filament check, detects a crash and/or recovers from a power panic.

The first issue is that restore_print_from_ram_and_continue can be called from a nested manage_inactivity() call. We need to properly abort the topmost process_commands() loop in order to avoid smashing the stack and essentially putting the printer in a random state.

A common scenario is a stalled plan_buffer_line which recovers from:

- a filament check
- a paused print
- a stopped print
- a crash detect

We then abort the state correctly also in the mesh_plan_buffer_line call, first by not corrupting the current position of the head, and also by aborting early if we discover we're in an aborted recursive call of process_commands().

Although this "looks" uncommon, a frequent scenario is:

0 process_commands
1 mesh_plan_buffer_line
2 plan_buffer_line
3 filament_check => manage_inactivity
4 process_commands

This PR is based on top on the previous PR #2271 in order to perform proper command buffer flushing with a starved command buffer in order to work at all.

In order to recover properly with an immediate relative move after panic we need to store the amount of relative movement which has been done for the current gcode instruction. This is not possible to perform for a chunked move, such as the one produced by mesh_plan_buffer_line (since the move might still be partially planned *while* the panic occurs). As such, we store the calculated absolute target position for each plan entry instead, which allows us to calculate the relative movement required easily from the current position.


leptun:
@wavexx Do these changes also fix mesh_plan_buffer_line creating stuttering when starting from an empty buffer?

wavexx:
I was hoping it would fix that, but no :(.

There's some extra fixing required on the planner/stepper which conflicts with LA1.5, so I decided to stop there as it would get messy later otherwise.

wavexx:
This became a big melting-pot of several pause/resume state fixes, which was not my original plan, but they're all tightly interconnected.

DRracer (approved):

PR #2307, opened 2019-11-01
Save/restore hotend temperatures above 255C in powerpanic

### LINKED ISSUE #2303: Wrong target temperature for heating up after power failure MK3s+MMU2s
Hi there,
I encountered a firmware problem after experiencing a power failure. My firmware-version is 3.8.0-2684. As the power returned, my i3 MK3s with MMU2s resumed printing and started heating up again. The problem was that the target temperature for the nozzle was wrong. It was set to 14°C.
As a result the nozzle kept cooling down and I had to set the correct nozzle temperature manually. The bed target temperature was correct.
Best regards,
Frans

wavexx:
Save/restore hotend temp using two separate bytes.

Fixes #2303


wavexx:
This will need to be resolved _after_ #2437 is merged.

wavexx:
Rewrote the patch to use 2 consecutive bytes freed by the #2437 so we get some cleaner code.

DRracer (approved):

Issue #2162, opened 2019-09-01
[Request] Extrude 2-3mm on filamanet change


mvasilakis:
I consistently get jams when I insert a filament change (non MMU) when the nozzle cools down and then reheats to extract filament. I've run several tests and this is 100% eliminated when a couple mm of filament are extruded prior to unload. Again this is only an issue when the printer has requested a filament change and the user has not intervened triggering the nozzle to cool.

This is with a MK3s and a properly spec'd ptfe in the hot end. It seems to be less of an issue when the PTFE has opened up after long use but even then I see jams about 50% of the time. I also believe that pulling out a hardened filament tip forces open the diameter of the hole in the hot end PTFE prematurely. Dipping the filament and softening the tip also solves this.

Please consider modifying the code of the change filament action to include a mm or two of extrusion to soften the filament tip and ensure proper extraction. It needs just enough of a dip in the hot end  to soften it up.

gysnst2:
i would like to see the extruder go all the way to the right or left before the filament change starts.  ive had a few failed prints because of the location of the filament change and the purge of the filament

michalxfanta:
Thank you for your idea, we will try it and consider its implementation.

wavexx:
See #1095

PR #2318, opened 2019-11-09
Perform a small extrusion prior to unload

wavexx:
When the filament is left inside the hot zone a larger tip is formed, caused by slow softening of the material. On an MK3, this larger tip will fail to unload as it will jam inside the lower PTFE section, sometimes on the gears, and and sometimes upper PTFE section as well (depending on the viscosity state of the hot tip as it gets ejected).

This PR will perform a small extrusion when unloading manually (though the LCD) using this sequence:

- slow 5mm extrusion
- very slow 0.1mm extrusion to relieve leftover pressure (mostly for flexible materials)
- 2mm retract at max speed
- filament ejection

This sequence essentially extrudes the entire large section of the tip, which avoids jams.
This is *not* done when MMU is operating and/or when a filament runout occurs.

This has been discussed at lengths on issue #1095, and requested on #2162.

In addition, we re-organize the various #defines related to filament change inside the variant files in order to be more meaningful (the actual values are left unchanged).

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -496,7 +496,7 @@ void proc_commands();
 
 void M600_load_filament();
 void M600_load_filament_movements();
-void M600_wait_for_user(float HotendTempBckp);
+bool M600_wait_for_user(float HotendTempBckp);
 void M600_check_state(float nozzle_temp);
 void load_filament_final_feed();
 void marlin_wait_for_click();
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3706,7 +3706,7 @@ static T gcode_M600_filament_change_z_shift()
 #endif
 }	
 
-static void gcode_M600(bool automatic, float x_position, float y_position, float z_shift, float e_shift, float /*e_shift_late*/)
+static void gcode_M600(bool automatic, bool runout, float x_position, float y_position, float z_shift, float e_shift, float /*e_shift_late*/)
 {
     st_synchronize();
     float lastpos[4];
@@ -3748,8 +3748,17 @@ static void gcode_M600(bool automatic, float x_position, float y_position, float
     lcd_change_fil_state = 0;
 
     // Unload filament
-    if (mmu_enabled) extr_unload();	//unload just current filament for multimaterial printers (used also in M702)
-    else unload_filament(true); //unload filament for single material (used also in M702)
+    if (mmu_enabled)
+    {
+        // unload just current filament for multimaterial printers (used also in M702)
+        extr_unload();
+    }
+    else
+    {
+        // unload filament for single material (used also in M702)
+        unload_filament(runout? UnloadType::Runout: UnloadType::Swap);
+    }
+
     //finish moves
     st_synchronize();
 
@@ -8133,7 +8142,8 @@ Sigma_Exit:
 		float e_shift_init = 0;
 		float e_shift_late = 0;
 		bool automatic = false;
-		
+		bool runout = false;
+
         //Retract extruder
         if(code_seen('E'))
         {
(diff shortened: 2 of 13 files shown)


leptun:
@wavexx you know what would be more interesting? It would be performing a ramming sequence just like on the MMU2. It does the small extrusion and it also cools the filament, ensuring a sharp tip.

wavexx:
I did try this early on (the discussion is still in the issue) using the built-in ramming sequence in mmu.cpp, but it didn't work as nicely. I still frequently got jams.

My understanding is that the MMU ramming sequence needs to operate with backpressure of the purge tower in order to work.

wavexx:
Also, the unload jams on the MK3 (w/o MMU) are often caused by a complete cooldown of the hotend. An unload which was performed during a filament change and/or just immediately after a print was always ok. On the other hand, an unload from a cold state would almost always jam.

leptun:
Hmmm. Maybe the small extrusion part can remain as is, however I still think that the cooling move would help also after unload from cooled down hotend. Sometimes I get jams when I attempt to load a badly shaped tip after an unload that didn’t jam.

wavexx:
How does the tip look in such cases?

leptun:
It sometimes has a bigger blob that is usually connected to the rest with a thinner filament (stretched) and is very stringy. I haven’t tested your changes. This is my experience with the MK2 heatbreak. The MK3 heatbreak might behave better.

wavexx:
Indeed, that's the classic result of a normal unload for me.

With this patch, we extrude the blob. The result is a tip which is completely flush though-out.

With PETG/PC I often see the V chamfer on the inside of the nozzle imprinted on the tip, meaning the retract performs a perfectly clean break. With PLA sometimes I still see some stringing, but the tip has no bulge, so the unload is always clean.

I was hoping to eliminate stringing, but given that this is done for manual unloads, it's not a real problem (and yes, I also get stringing without the change anyway).

You're right that doing some cooling moves might improve in this area.
More than ramming, skinnydip aim to solve this better by remelting the string:

https://github.com/prusa3d/PrusaSlicer/pull/2452

matlinder:
Wavexx was this implemented? I am having the same issues that this solution was aimed at fixing. How do I get my mk3s to extrude a little before the unload sequence (without manual intervention).

wavexx:
It hasn't been merged yet, hopefully after 3.9.

Also, as a reminder for myself, I found that 5mm of extrusion prior to unload is sometimes _not_ enough when printing PC at 270-280C. It seems that some heatcreep at these temperatures will soften the PC enough to widen towards the bottom, causing a jam during ejection. PC being PC, it's essentially impossible to extract by just pulling. It seems 2-3mm extra millimiters would do it.

Lacehim:
Any progress on this?  I love printing on my MK3 but it always jams if it cools and I try to unload/swap filament.  My printer is in the garage so I don't hear the beeps.
I just installed the latest 3.9.0 RC3 and nothing has changed with unloading or changing filament, it's just a straight unload.

wavexx:
Btw, with a total of 10mm of extrusion prior to unload I can now always swap any material from any situation (especially from cold-start). No problems with PC any longer.

NiliusJulius:
Any progress on this? It is pretty annoying to have to do a manual extrusion before unloading every single time. Otherwise it jams almost always.

MohanMaker:
This same issue happened to me earlier today! Any progress on this PR @wavexx? Would be great if it was implemented in 3.10. Thanks :).

wavexx:
@leptun should we aim this for 3.12?

leptun:
I think yes, but let's wait for the mmu PR. Maybe we can combine the unload sequences. Afaik, the mmu unload procedure also purges and it even also does a cooling move.

pbotte:
Is there any change to get this into the next release? This has been a known problem for 5+ years. Newbies get a free PLA spool with their printer, but immediately run into jamming problem.   --> Not fair!

wavexx:
Almost grew tired of rebasing it :/

sarusani:
I'm using the MMU style unload for more than a year now and since then never again had a jam. (PLA was mostly OK, but especially PC Blend jammed regularly before the change)
[This](https://github.com/sarusani/Prusa-Firmware/commit/d4f4315cef0aea2551078e72acbfbe9035d8bcf8) is the code I'm using.

wavexx:
I'll try to rebase this once more...

3d-gussner:
Moved to 3.14.0. Add two gcode lines when manual unload or change filament is selected.

PR #2327, opened 2019-11-15
#2089 fix sheet rename cursor

### LINKED ISSUE #2089: Sheet rename cursor does not clear

vintagepc:
Fixes #2089 

No longer leaves a cursor trail: 

image

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -5054,6 +5054,10 @@ static void lcd_rename_sheet_menu()
         lcd_putc(menuData->name[i]);
     }
     lcd_putc_at(menuData->selected, 1, '^');
+    if (menuData->selected > 0)
+    {
+        lcd_putc_at(menuData->selected-1, 1, ' ');
+    }
     if (lcd_clicked())
     {
         if ((menuData->selected + 1u) < sizeof(Sheet::name))


Firmware/ultralcd.cpp
```diff
@@ -6733,7 +6733,15 @@ static void lcd_rename_sheet_menu()
     {
         lcd_putc(menuData->name[i]);
     }
-    lcd_set_cursor(menuData->selected, 1);
+    if (menuData->selected>0)
+    {
+	    lcd_set_cursor(menuData->selected-1, 1);    
+	    lcd_putc(' ');
```
gudnimg: Could be optimised with `lcd_putc_at(menuData->selected-1, 1, ' ');` ;)

vintagepc: Does that put the cursor back in the next character position when it is done?

gudnimg: Yes exactly, it should do the same. Should save a few bytes to call one function instead of two. Not critical but should save some bytes

vintagepc: Done, and further optimized to remove the if/else

gudnimg (approved):
Looks good to me

DRracer (approved):

Issue #2295, opened 2019-10-25
Feature Request: Don't heat the bed when using the autoload/unload filament menu


Josh3D:
The current way of changing filaments is:

- Select autoload/unload filament menu

- The printer will then ask you to select a material that you are either loading or unloading 

- The printer then will preheat the Hotend AND heat bed to those temperatures

I propose the following when selecting the autoload/unload menu, and the temperature selection pops up that it only will heat the Hotend and NOT the heat bed.

Some benefits of this are: 

- Not accidentally burning yourself on the heat bed when you touch it changing filament or taking off the purged plastic that has dropped down on the bed

- After removing the filament you had in your printer previously, loaded in the filament for your next print, the bed would then still be room temperature so you can wipe it down with IPA effectively and then be able to start your next print right away.

JKinch03:
For MMU2 users, you could also 
-Add this feature to the load/unload to nozzle
-Change the "Eject Filament" feature so that it checks if the PINDA probe and the IR sensor are triggered, and if neither are triggered when the eject function is issued, then you wouldn't need to preheat the nozzle or heat bed to eject the filament. If the IR sensor **OR** the PINDA are triggered, then it'll still preheat the nozzle for safety measures.

Prusa-Support:
Duplicate of #422

Issue #422, opened 2018-01-21
Feature Request : Don't preheat bed for filament Unload


SukMeWot:
The current filament unload process should not ask for preheating, instead the menu should ask which fliament you are unloading and set the temperature, do the preheat of the hot end only, then eject, then cool. It should not be incumbent on the operator to select preheat, then unload, then cooldown, and it certainly shouldn't be heating the bed.

Area5142:
The preheat menu could be shown to select filament type (temperature) and only set temperature for the hot end. After unload set temperature to 0.

SukMeWot:
It does but it heats the bed too - MK3

davidhbrown:
This seems poorly implemented, possibly to leverage existing preheat code. I'm unloading filament because the print is done. Heating the bed just wastes energy. (Could/should be labeled with "Firmware 3.6?")

GreenShadowMaker:
Definitely unnecessarily heating the bed with filament unload on with 3.7.0   Not sure if it was happening with 3.5.3. (I don't remember it, or I was unloading before post print cooldown had happened)

GreenShadowMaker:
looks related to  https://github.com/prusa3d/Prusa-Firmware/issues/1688

dstulken:
Any updates on this?

I noticed that the original post (Jan 2018) predates the release of the Prusa Mini (Oct 2019).     One of the best features of the Mini is its small size and portability, and because of this, unloading the filament before moving/transporting the printer is an important part of its usage.

Having the bed heat up for this is completely unnecessary... and having both the hotend and bed stay at their temps after the operation was a surprise.  I didn't even know there was a cooldown option until I found this issue request - so I suspect many others aren't cooling their machines down properly either after doing filament unloads...

SukMeWot:
This has been a bug since I've owned a Prusa - well not a bug but a piece of poor coding that has been present since the beginning.

Prusa-Support:
> [oderbang](https://github.com/oderbang) commented [on Jan 22, 2021](https://github.com/prusa3d/Prusa-Firmware/issues/2985#issue-792172891) • 
> Printer type - MK3
> 
> Is your feature request related to a problem? Please describe.
> When selecting "unload filament" when the bed is cold the printer pre-heats the build plate. this is un necessary if you are just removing filament.
> 
> Describe the solution you'd like
> have a toggle option in the setting to enable/disable build plate pre-heat when using filament unload.

[...]

PR #2335, opened 2019-11-17
Do not heat the bed, on load/unload (or autoload)

Panayiotis-git:
It has been requested a long time ago:

https://github.com/prusa3d/Prusa-Firmware/issues/422
https://github.com/prusa3d/Prusa-Firmware/issues/2295

Since I've already implemented and tested this improvement for months, I thing I can share it.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -1824,12 +1824,42 @@ bool bFilamentPreheatState;
 bool bFilamentAction=false;
 static bool bFilamentWaitingFlag=false;
 
+bool shouldPreheatOnlyNozzle() {
+    uint8_t eeprom_setting = eeprom_read_byte((uint8_t*)EEPROM_HEAT_BED_ON_LOAD_FILAMENT);
+    if (eeprom_setting != 0)
+        return false;
+
+    switch(eFilamentAction) {
+        case FilamentAction::Load:
+        case FilamentAction::AutoLoad:
+        case FilamentAction::UnLoad:
+        case FilamentAction::MmuLoad:
+        case FilamentAction::MmuUnLoad:
+        case FilamentAction::MmuEject:
+        case FilamentAction::MmuCut:
+            return true;
+        default:
+            return false;
+    }
+}
+
+void lcd_print_target_temps_first_line(){
+    lcd_set_cursor(0, 0);
+    lcdui_print_temp(LCD_STR_THERMOMETER[0], (int) degHotend(0), (int) degTargetHotend(0));
+    lcd_set_cursor(10, 0);
+    int targetBedTemp = (int) degTargetBed();
+    if (targetBedTemp) {
+        lcdui_print_temp(LCD_STR_BEDTEMP[0], (int) degBed(), targetBedTemp);
+    } else {
+        lcd_space(10);
+    }
+}
+
 static void mFilamentPrompt()
 {
 uint8_t nLevel;
 
-lcd_set_cursor(0,0);
-lcdui_print_temp(LCD_STR_THERMOMETER[0],(int)degHotend(0),(int)degTargetHotend(0));
+lcd_print_target_temps_first_line();
 lcd_puts_at_P(0,1, _i("Press the knob"));                 ////MSG_PRESS_KNOB c=20
 lcd_set_cursor(0,2);
 switch(eFilamentAction)
@@ -1891,14 +1921,10 @@ switch(eFilamentAction)
 
 void mFilamentItem(uint16_t nTemp, uint16_t nTempBed)
 {
-    static int nTargetOld;
-    static int nTargetBedOld;
     uint8_t nLevel;
 
-    nTargetOld = target_temperature[0];
-    nTargetBedOld = target_temperature_bed;
-    setTargetHotend0((float )nTemp);
-    setTargetBed((float) nTempBed);
+    setTargetHotend0((float)nTemp);
+    if (!shouldPreheatOnlyNozzle()) setTargetBed((float)nTempBed);
 
     {
         const FilamentAction action = eFilamentAction;
@@ -2033,8 +2059,7 @@ void mFilamentItem(uint16_t nTemp, uint16_t nTempBed)
         }
 
         if (bFilamentWaitingFlag) {
-            lcd_set_cursor(0, 0);
-            lcdui_print_temp(LCD_STR_THERMOMETER[0], (int) degHotend(0), (int) degTargetHotend(0));
+            lcd_print_target_temps_first_line();
         }
 
         if (lcd_clicked())
@@ -2043,13 +2068,8 @@ void mFilamentItem(uint16_t nTemp, uint16_t nTempBed)
             if (!bFilamentPreheatState)
             {
                 setTargetHotend0(0.0);
-                setTargetBed(0.0);
+                if (!isPrintPaused) setTargetBed(0.0);
                 menu_back();
-            }
-            else
-            {
-                setTargetHotend0((float )nTargetOld);
-                setTargetBed((float) nTargetBedOld);
             }
             menu_back();
             if (eFilamentAction == FilamentAction::AutoLoad) eFilamentAction = FilamentAction::None; // i.e. non-autoLoad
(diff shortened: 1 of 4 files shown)


wavexx:
I like the behavior. Maybe we can save some space by avoiding the repetition of the entire menu?
Showing a bed temperature of 0 would be fine IMHO.

Panayiotis-git:
@wavexx After the changes, we gain 22 bytes in program space. Every other method I tried resulted to a larger binary.

wavexx:
If I could get a dime for every time I needed this before :+1: 
I essentially always set the hotend temperature manually to avoid heating the bed

Firmware/ultralcd.cpp
```diff
@@ -6326,7 +6333,11 @@ static void mmu_cut_filament_menu()
             bFilamentPreheatState=true;
             mFilamentItem(target_temperature[0],target_temperature_bed);
         }
-        else lcd_generic_preheat_menu();
+        else
+        {
+        	bPreheatOnlyNozzle = false;
```
wavexx: So this one area I have some doubts on. Shouldn't "cut filament" also be a good candidate for bPreheatOnlyNozzle = true?

wavexx (commented):
So I gave this a try, it's nice, but if I already have the bed heated it will set it 0 which is something I do not expect. I also think this could cause problem if preheat is called for unload during the paused state where the bed /should/ be kept on.

When ``bPreheatOnlyNozzle`` is set, instead of setting the bed temperature to zero in all menus, we should just skip setting the bed temperature in ``mFilamentItem`` instead, leaving it as-is.

This avoids all the need for all checks in mFilamentItem_* functions.

Comments?

Firmware/ultralcd.cpp
Panayiotis-git: Since I've never seen this functionality working, I left it untouched. But I think you are right.

Panayiotis-git:
@wavexx indeed the load/unload filament selection during a paused print, set the heated bed to 0 degrees. You have an excellent idea about moving the check into the ```mFilamentItem``` function! The only drawback is that then we should initialize the ```bPreheatOnlyNozzle``` before each call to the ```mFilamentItem``` function.

But then, instead of setting the ```bPreheatOnlyNozzle``` beforehand to true or false, it is enough to implement only the ```bPreheatOnlyNozzle = true```. After the temperatures are set, we can reset the variable back to ```bPreheatOnlyNozzle = false```. The variable should also be reset if the user returns to the previous menu or the menu timeout is expired.

Also, as you proposed, we now set the ```bPreheatOnlyNozzle = true``` to the MMU cutter preheat menu and it is working perfectly.

So, after all these changes, the compiled code size decreased another 108 bytes and now the whole functionality occupies 194 bytes of code.


PS. The load/unload temperatures during a paused print still are not set correctly. For example, if the user cancels the unloading of the filament while waiting for the nozzle to reach the requested temperature, then both the hotend and the heated bed temperatures are set to 0. But this is a preexisting issue, not related to this PR.

wavexx:
> PS. The load/unload temperatures during a paused print still are not set correctly. For example, if the user cancels the unloading of the filament while waiting for the nozzle to reach the requested temperature, then both the hotend and the heated bed temperatures are set to 0. But this is a preexisting issue, not related to this PR.

That sounds bad :/, especially in pause this would detach the print.
What sequence of actions do I actually need to do to get to this state? Print, pause, resume..?

Panayiotis-git:
Steps to reproduce:
1. Start a print.
2. Pause the print (Nozzle temp targets to 0, Heated bed keeps the target temp)
3. Wait for the nozzle to cool down
4. Load/Unload filament, choose one. The hotend starts warming.
5. Press Cancel. This will set both the nozzle's and the bed's target temperatures to 0.

6. At this state, if Resume is selected, then first the hotend will restore its temperature and after it reaches the target temp, then starts the heating of the bed.

wavexx (approved):
I don't know if my review does anything, but I tested the changes and it works perfectly.

wavexx:
I filed a new bug for the unload behavior during a pause to keep track of it. It's a good idea to fix this separately.

wavexx:
After using this again, I'm afraid I've run into the same problem I had into PR #2386. While heating the extruder, ``mFilament`` is called over and over to display the raising temperature. On the first call we clear the ``bPreheatOnlyNozzle``, then on the second the bed is heated anyway :(

Panayiotis-git:
My bad. The last commit broke the main functionality, due to the continues call of the ```mFilament```. And it seems that I've skipped the most obvious tests.
I'll give it another try.

Panayiotis-git:
I returned this PR to a working state. I tried to keep the current logic, but the resulted code/logic was too complex.
So, I reverted to the previous logic, the one that checks at each ```mFilamentItem_xxx``` for the ```bPreheatOnlyNozzle``` value and set the temperatures accordingly. I assigned the temperature value of -1 to the 'keep current set bed temperature'.

Also, I fixed the bug of resetting the bed temperature after a 'Cancel filament loading' during a paused print.

wavexx:
Really like that is also now shows the bed temperature while heating.
Maybe that should be hidden when heating only the nozzle?

Still.. nTarget*Old is technically lost when calling mFilamentItem multiple times, which I think defeats the purpose of the ``bFilamentPreheatState`` flag. I still don't have the complete mental picture of what's the intention of restoring the temperature of a single cycle earlier...

Panayiotis-git:
Since the ```eFilamentAction``` keeps the currently selected ```FilamentAction```, there is no need to define the boolean variable ```bFilamentPreheatState```. The ```mFilamentItem``` depending on the value of the ```eFilamentAction```, decides if it is also required to preheat the bed.

Now the source is as clean and compact as it can get.

Panayiotis-git:
> Still.. nTarget*Old is technically lost when calling mFilamentItem multiple times, which I think defeats the purpose of the `bFilamentPreheatState` flag. I still don't have the complete mental picture of what's the intention of restoring the temperature of a single cycle earlier...

A value of true for the ```bFilamentPreheatState``` indicates that a preheating is already in place. This affects the menu history and if after cancelling the load/unload/... action the preheating should continue or not.

The two variables ```nTargetOld``` and ```nTargetBedOld``` do not have any use at this place. In fact, I removed them and tested all the affected use cases. The system responded as expected: exactly the same.

Since I cannot also find any use for them, I push another commit with these variables removed.

wavexx:
> Since I cannot also find any use for them, I push another commit with these variables removed.

:+1: 

I've been looking at powerpanic issues lately, but I'll get back to this asap :)

wavexx:
Looking back, is #2385 by chance also addressed?

Panayiotis-git:
Yes, this PR also fixes the https://github.com/prusa3d/Prusa-Firmware/issues/2385.

Panayiotis-git:
I've also identified 2 "problems" that have to do with the temperature display, but are not relevant to this PR:
1. The temperature is stored as a float. At some places we round the temperature to the next integer (eg. the status screen) while at other places we just cut the decimal digits. We should follow the same strategy everywhere.
2. There are cases where the custom characters are messed up. The temperature degree symbol gets replaced with the down arrow. We should make sure that whenever we display temperature, we have called the lcd_set_custom_characters_degree beforehand.

@wavexx Do you suggest a new PR?

DRracer:
@Panayiotis-git after some internal (flame-war :) ) discussion, there is another valid opinion on this issue - it is not about technical solution, but if you are unloading and loading a piece of filament, you may very well also want to start printing - therefore the bed heating (it is more likely).
The correct solution to this mutually distinct behavior is either:
- custom preheat profiles (like we have the sheets)
- some setting to allow/disable bed heating while loading/unloading etc.

DRracer:
Marking as FW 3.10 - we shall decide about the generally accepted solution here.

wavexx:
My 2c, I personally really like the behavior since I switch between materials quite often. I also like a lot the cleanup of the preheat flags too.

Panayiotis-git:
I rarely leave the printer with a filament loaded and when I finish a printing I usually  remove the filament and keep it stored into its sealing bag to avoid absorbing humidity.
But most of the times, when I access the printer it is too late and the hot-end has already cooled. So, I have to re-heat the hot-end (and only the hot-end) in order to unload the filament. This was the main reason that I started this implementation almost a year ago and I find it really useful since it solved my problem.

I have also used the filament loading functionality of this PR some times, like to do a cold-pull cleaning to the hot-end.

If I want to start a printing afterwards, I usually go through the preheat menu. Then, after everything reached the target temperatures, I simply select the Load filament function. It is useful if I want to heat uniformly the bed so I preheat it and leave it for a couple of minutes before I start the printing.

In general, I mostly use the Load/unload functionality as it is implemented with this PR rather than the preheat menu.

Regarding the custom preheat profiles, it is indeed a very useful feature especially for the cases when the filament is not included in the printer's list and requires special temperature handling. My opinion is that the custom preheat profiles rather complement than replace this PR. Otherwise, as a user, I would have to re-create the existing list of filaments with custom profiles only to set the bed temperature to 0 degrees.

The solution to use some sort of settings to allow/disable the bed heating while loading/unloading sounds interesting. Although, I would leave it to disabled, I think it gives the user more choices.

Panayiotis-git:
@DRracer , @wavexx I have implemented the solution with the setting to allow/disable the bed heating while loading/unloading the filament, with default value to allow. I will not push it yet since the 3.9.3 release contains some changes to the EEPROM map and there will be some conflicts.

Panayiotis-git:
As mentioned at the [PR#3014](https://github.com/prusa3d/Prusa-Firmware/pull/3014), the current implementation of the button long-press event handling breaks the functionality of the preheat phase of the Load/Unload filament.
A short press of the button selects the Cancel choice and the preheat phase is aborted.
A longer press triggers the long-press event. The behavior on some screens, including the preheat, is to execute a call to the `menu_back()` function. The result is to return to the previous menu and not been anymore able to Cancel the preheating.

The implemented fix is to include the preheat menus to the exclusions of the long-press functionality. Also, modified the redraw loop to always draw the entire screen since the `lcd_quick_feedback()` clears the screen.

DRracer:
@Panayiotis-git is this PR ready for some feature tests? Could you please describe a few test scenarios?

Basically the idea is to allow/disable heatbed heating while doing preheat (which sounds reasonable and thank you for continuing with this work after my last comment). However, are you aware of some edge cases?

Moreover - I can see you are switching texts based on some bool flag. This is perfectly OK at runtime, however are you aware how much more code space do we need for the additional set of preheat texts? Rough guess > 200B more. This is where I'd think about optimization if possible (at reasonable cost of course, no refactoring of half of the FW code base ;) ).

Panayiotis-git:
#### Methods to construct menu items
I have identified the following 3 methods to generate the menu item text. I guess, someone could think many other ways to implement the menu and perhaps more efficient.
1. Create two versions of the menu item text, one for preheating only the nozzle and one for both, and while drawing the menu items decide which text to use.

Preheat only nozzle|Preheat also bed
-|-
`PLA  -  215`|`PLA  -  215/60`
`ABS  -  255`|`ABS  -  255/100`
...|
2. Create only the full version menu item text and then use two String format specifiers to cut the bed temperature if it is not needed.

Initial text|Preheat only nozzle, use `%-11.11S`|Preheat also bed, use `%-18.18S`
-|-|-
`PLA  -  215/60`|`PLA  -  215`|`PLA  -  215/60`
`ABS  -  255/100`|`ABS  -  255`|`ABS  -  255/100`
3. Implement a variation of the `menu_draw_item_puts_P`, that on the runtime it will get the filament name, the nozzle and the bed temperature as parameters and using the correct format specifier it will display the menu text.

The PR implements the 1st method. The 'Preheat only nozzle' version of the menu texts, requires about 90 bytes of extra storage space.
The second method, even though it creates one text per filament, it requires additional code, so it produces similar code size with the 1st method. Also, it makes the assumption that the nozzle temperature will always be 3 digits. Otherwise, the text will be truncated incorrectly.
The 3rd method, quickly becomes inefficient and requires at least 300 additional bytes.

Panayiotis-git:
@DRracer as you've mentioned, this PR provides the user the choice to preheat both the nozzle and the bed in order to load/unload/... the filament or to simply preheat only the nozzle and not wait for the bed that takes considerably more time.
It is used at the Load/AutoLoad/UnLoad/MmuLoad/MmuUnLoad/MmuEject/MmuCut filament actions.

The choice is accessible at the bottom of the Setting menu, as:
`HeatBedOnLoad[Yes/No]`

This PR does a small cleanup of unnecessary code, and also provides fixes for two problems that we identified and are related to this functionality:
1. If a user changes filament during a paused printing and cancels the preheating, then the bed will also cool down
2. The button long-press while preheating will break the correct operation of the printer

I have already tested that the PR works as expected at all the above scenarios and I have it merged to my printers firmware. Of course, I might have missed some corner scenario.

Implemented Code size|Bytes
-|-
Preheat only nozzle|340
Provide the choice as a setting|150
Fix the button long-press problem|324

The last one takes a lot of space because we need to add an exception for each menu item at the `menu_lcd_longpress_func`. If `menu_lcd_longpress_func` is structured differently it would not require that much space. Anyway, we could remove this fix from the PR and handle it separately.

slackista:
Any news on this? Over three years first reported.

tcurdt:
Given the age of this PR I think it's safe to assume this will never be merged (although it is annoying AF).

Maybe worth setting up a GCODE macro for it then.
Should this be enough?

```
M109 S210 ; wait for extruder temp
M412 S0 ; disable filament detection
M83 ; relative mode
G1 E-100 F1000.0 ; unload
```

slackista:
4 years old annoying bug. Please merge.

Panayiotis-git:
The feature implementation including the menu setting, now occupies only 280 bytes.
It is really hard to follow the optimizations that happen during the build. :smiley:

slackista:
How many mk3's has Prusa sold overall? 200k? 500k units? How many billions watt-hours of energy were wasted for bed heating while user wants to only unload the filament after finished printing? How long this bug is well known?: **4.5 years**.

wavexx:
:fire: @DRracer

gudnimg:
@Panayiotis-git Hi 👋 thanks for this PR and sorry this is taking so long to merge. I'm testing your PR at the moment and am seeing it doesn't work as I expected when using the Preheat menu

If I select PLA from Preheat menu, the Bed is heated regardless of what the EEPROM setting is. Is this expected behavior?

If I apply the fix below, then your PR also works for the Preheat menu :) This is super useful for me when testing the MMU. Note I am OK with this not being applied to the Preheat menu, just thought I'd ask about this.

image


I will try to avoid creating conflict with this PR until we have decided whether to merge or not next week. 👍

gudnimg (approved):
Approving the current state of the PR 👍 Works great when using for example "Load to Nozzle" menu.

PR will need one more rebase due to a recent merge to add PA to the filament menu. https://github.com/prusa3d/Prusa-Firmware/commit/3e976e04cada98fc1534d219c0ba0d8cfe87c72b

Panayiotis-git:
@gudnimg, the Preheat action was intentionally left out from this PR. The logic is that from the Preheat menu you can preheat both nozzle and bed for the selected filament. But for the filament handling actions, only the nozzle will be heated.

TY, for your review!

Panayiotis-git:
Travis build failed, with message:
CMake Error at /home/travis/build/prusa3d/Prusa-Firmware/cmake/Check_lang_size.cmake:7 (message):
  Language file MK3S-EINSy10a_fr.bin (13824b) exceeds maximum allowed size of 13568 bytes - Aborting!

Is there something I can do, in the context of this PR, to correct this error?

gudnimg:
@Panayiotis-git no the French translation is currently too large (byte wise). It's failing all builds at the moment, but it should be resolved hopefully very soon.

For this PR, this error should not prevent a merge.

Wallby:
@gudnimg Could the documentation for this be improved?

To find the answer to this I had to browse through GitHub issues of first the firmware for the MK4, which mentions a setting that doesn't exist on the MK3 firmware.
https://github.com/prusa3d/Prusa-Firmware-Buddy/issues/706#issuecomment-2167826286

And the page for the MK3 settings doesn't have any documentation on this: https://help.prusa3d.com/article/lcd-menu-original-prusa-i3_142322
It only has the setting name repeated without any explanation:
> HBed on load - Yes / No

Which to me wasn't clear that that is the setting specifically for filament loading.

PR #2353, opened 2019-12-03
Fix the LCD pause (regression)

wavexx:
This fixes a regression introduced with PR #1899 which would prevent "Pause print" from the LCD to work correctly with the latest mesh_plan_buffer_line fixes.


DRracer (approved):

vintagepc:
Aha, I've been bitten by this twice when running a 12/3 github pull. Good to see it's already been caught/fixed; will pull down and report back whether it solves my occurrence or I've found something new.

PR #2291, opened 2019-10-23
Lang 381 fixes 4

DRracer:
Finalized texts for FW 3.8.1 with one exception - one sentence is intentionally left untranslated, because the whole translations BIN file was larger than 64KB.

PR #2362, opened 2019-12-08
Fix optiboot erase function after address 0xFFFF. Allows language files greater than 64K

leptun:
This PR addresses an issue where if you tried to upload a language file greater than 64K, everything up to address 0xFFFF was fine, but the rest was only garbage data.
The issue might not have manifested to everyone **at first** since the rest of the flash was already erased and you would just write data on the erased portion. The problem starts when you want to change something in the already written area after 0xFFFF. The erase function didn't ever clear anything after 0xFFFF and you would get what appears to be as garbage data, but is in fact just a combination of what was before and what you just uploaded.

This issue has been encountered before by @3d-gussner when he tried packing two more languages into the firmware (resulting in ~80K of lang data), but some of the languages were unusable. It also was an issue with the recent release of fw 3.8 (@DRracer #2291 ), the total lang area was >64K and resulted in one unusable language and a failed upload.

With this PR, the upload process will once again be successful and will allow languages to expand in the near future. It also includes the omitted string from #2291 

PFW-1065

### DIFF
--- a/Firmware/optiboot_w25x20cl.cpp
+++ b/Firmware/optiboot_w25x20cl.cpp
@@ -258,11 +258,11 @@ void optiboot_w25x20cl_enter()
         uint32_t addr = (((uint32_t)rampz) << 16) | address;
         // During a single bootloader run, only erase a 64kB block once.
         // An 8bit bitmask 'pages_erased' covers 512kB of FLASH memory.
-        if (address == 0 && (pages_erased & (1 << addr)) == 0) {
+        if ((address == 0) && (pages_erased & (1 << (addr >> 16))) == 0) {
           w25x20cl_wait_busy();
           w25x20cl_enable_wr();
           w25x20cl_block64_erase(addr);
-          pages_erased |= (1 << addr);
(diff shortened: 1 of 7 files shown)


3d-gussner (approved):
I have tested it with additional languages: Dutch + Turkish and the external flash verification error we got during Firmware Update is gone.
Also the SPI flash dumps are clean now.
To test the maximum of the external flash I created a "20" languages firmware which allocated 223576 bytes. Again the SPI dumps are clean. And choosing these from LCD menu was possible, at least most of them. I guess had some copy paste errors during creating the 20 languages firmware quickly.

3d-gussner:
After fixing my bugs in "lots-of-languanges-firmware" I can approve again that this fix works.
Last test was with 21x 12Kb languages.

I didn't test MK2.5/s as I don't own one. But i don't see any issue here as the MK2.5/s miniRAMbo get's just one additional language in the internal flash.

@wavexx Can you also check this PR and approve it?
@DRracer This PR can be pulled in my opinion any time. There are no dependencies.

DRracer (approved):

PR #2368, opened 2019-12-14
Unretract when resuming a paused print

wavexx:
When pausing from the LCD (or via M601) we retract just before raising Z.
However, when resuming, we never unretract.

``stop_and_save/resume_print_to_ram`` already have the capability to retract/unretract, so we re-use this and save some extra bytes too.


3d-gussner:
Tested before and after with PLA_Whistle SD print. Pause/Resume triggered via LCD menu
- Before the missing line is quite long
- After this fix the missing line is way shorter, but it also looks like the unretract is bit too eraly and ooze out a tiny filament string.

DRracer (approved):

PR #2376, opened 2019-12-18
Fix sheet minimum width on the status screen

leptun:
@DRracer 
Avoids situations like this:
IMG_20191217_173612


vintagepc:
Good catch... all my sheets use the full width so I didn't notice :-)

leptun:
Mine also are the same. Some old beta of 3.8 doesn’t fill the whole space and only put a number as the default name, hence why it was shorter. Minor issue only, but it was detected in time for 3.9

leptun:
Closing in favor of #2402

Issue #2055, opened 2019-07-29
M0/M1 Pause not displaying text/waiting for input


msbeck:
> M1 Hello World Waits until user presses the knob. "Hello World" message is shown as LCD status message.

https://github.com/prusa3d/Prusa-Firmware/wiki/Supported-G-codes

M0 and M1 do not prompt the user with text when pausing. They also do not wait for user input to resume.

```
G1 X10.000 Y200.000 E0
M300 S440 P200
M300 S660 P250
M300 S880 P300
M1 Click to continue 
M108 
M105; return to current temp
```
I would expect the printer to move out of the way, play the beeps, pause, display "Click to Continue" and wait until the LCD button is clicked. Instead, it just pauses, does now show message, and does not resume when button is clicked (standard Prusa menu is up).

The only way to restart is to click Restart in Octoprint, which is not ideal when the pause is one that requires user interaction AT the printer.

This is experienced while printing through OctoPrint on an MK3s. It doesn't work embedded in GCODE nor through the terminal in Octoprint.

Is there another way to go about pausing and waiting for user input?

Thank you

michalxfanta:
We will fix this bug in some of the next releases, thanks for reporting.

3d-gussner:
@msbeck I tried it myself on newest FW and it works with following gcode:
```
G28 W
G80
G1 X10.000 Y200.000 E0
M300 S440 P200
M300 S660 P250
M300 S880 P300
M1 Click to continue
G1 X200.000 Y10.000 E0
;M108 
;M105; return to current temp
```

As `M108` isn't implemented in Prusa firmware see your link above I guess it fails because of the M108 which isn't supported.

Can you please fill in the things below if the problem still exist?

Please, before you create a new bug report, please make sure you searched in open and closed issues and couldn't find anything that matches.

**Printer type** - [e.g. MK3S, MK3, MK2.5S, MK2.5, MK2S, MK2]
**Printer firmware version**-  [e.g. 3.8.1, 3.8.1-RC1, ...]

**MMU Upgrade** - [e.g. MMU2S, MMU2, MMU1]
**MMU upgrade firmware version [e.g. 1.0.6, 1.0.6-RC2, ...]

**Describe the bug**
  A clear and concise description of what the bug is.
  
**To Reproduce**
  Please describe steps to reproduce the behavior.
  
**Expected behavior**
  A clear and concise description of what you expected to happen.
  
**G-code**
[...]

PR #2380, opened 2019-12-21
M600 - Custom message at the Insert filament prompt

Panayiotis-git:
The M600 is a very useful command because it allows the user to change the filament at selected points during the printing.
If the user do not have an automated system (like MMU or Palette) then the M600 is the only solution. But even with the MMU, the user is limited to 5 colors/materials.

Recently, I needed to use 16 different colors per layer on the same print. The current implementation of the M600 command, after unloading the previous filament, it only displays the message "Insert filament", leaving to the user the responsibility to find out which filament is next...

This pull request improves dramatically this functionality, introducing a new parameter that defines a custom message to be displayed after the Insert filament message. The new syntax is:
```
M600 X[pos] Y[pos] Z[relative lift] E[initial retract] L[later retract distance for removal] C[Custom message to show during loading]
```
The size of the custom message is limited to the width of the LCD display.

So, the command:

```
M600 X200 Y0 Z20 CTraffic Red
```
will move the extruder at x=200, y=0, z=z+20, do the unloading and then show the message
```
Insert filament
Traffic Red
```

The C parameter should be the last one and the custom message will include everything after the parameter. Because there are some limitation to the command parser, there are several words that should not be used inside the message because they will trigger other functions. Try for example:

```
M600 G1 X100
```
It will move the extruder at x=100!!

To prevent this, the message should be enclosed between double quotes ("). If double quotes are used, then it is not necessary for the C parameter to be the latest. So, the proposed syntax is:

```
M600 X200 Y0 Z20 C"Traffic Red"
```

Note:
To implement the double quote functionality, the  new method "code_value_string" was introduced at the cmdqueue.h. Also, the two existing "code_seen" methods where improved to exclude everything enclosed between double quotes.

I've been testing and using the improved M600 for a couple of months now and I haven't found any side effects.
The functionality of the M600 command was tested:
- without the MMU2 and with disabled/enabled filament sensor and with enabled the auto-loading
- with MMU2, where no changes were detected in the normal operation of the filament run-out event (spool join on/off).

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -443,7 +443,7 @@ void gcode_M701(float fastLoadLength, uint8_t mmuSlotIndex);
 #define UVLO !(PINE & (1<<4))
 
 
-void M600_load_filament();
+void M600_load_filament(const char* lcd_load_message);
 void M600_load_filament_movements();
 void M600_wait_for_user();
 void M600_check_state();
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3401,7 +3401,7 @@ static void mmu_M600_load_filament(bool automatic) {
     st_synchronize();
 }
 
-static void gcode_M600(const bool automatic, const float x_position, const float y_position, const float z_shift, const float e_shift, const float e_shift_late) {
+static void gcode_M600(const bool automatic, const float x_position, const float y_position, const float z_shift, const float e_shift, const float e_shift_late, const char* lcd_load_message) {
     st_synchronize();
 
     // When using an MMU, save the currently use slot number
@@ -3461,7 +3461,7 @@ static void gcode_M600(const bool automatic, const float x_position, const float
                 st_synchronize();
                 lcd_show_fullscreen_message_and_wait_P(_i("Please open idler and remove filament manually.")); ////MSG_CHECK_IDLER c=20 r=4
             }
-            M600_load_filament();
+            M600_load_filament(lcd_load_message);
         }
         else // MMU is enabled
         {
@@ -7702,7 +7702,7 @@ Sigma_Exit:
     - `L`    - later retract distance for removal, default FILAMENTCHANGE_FINALRETRACT
     - `AUTO` - Automatically (only with MMU)
     */
-    case 600: //Pause for filament change X[pos] Y[pos] Z[relative lift] E[initial retract] L[later retract distance for removal]
+    case 600: //Pause for filament change X[pos] Y[pos] Z[relative lift] E[initial retract] L[later retract distance for removal] C[Custom message to show during loading]
     {
     st_synchronize();
 
@@ -7728,10 +7728,22 @@ Sigma_Exit:
     if (code_seen('X')) x_position = code_value();
     if (code_seen('Y')) y_position = code_value();
 
+    // Add a custom message, during the loading.
+    // It should be the last parameter of the command
+    char lcd_load_message[LCD_WIDTH + 1] = "";
+    if (code_seen('C')) {
+        unquoted_string str = unquoted_string(strchr_pointer);
+        if (str.WasFound()) {
+            const uint8_t len = min(str.GetLength(), LCD_WIDTH);
+            memcpy(lcd_load_message, str.GetUnquotedString(), len);
+            lcd_load_message[len] = '\0';
+        }
+    }
+
     if (MMU2::mmu2.Enabled() && code_seen_P(PSTR("AUTO")))
         automatic = true;
 
-    gcode_M600(automatic, x_position, y_position, z_shift, e_shift_init, e_shift_late);
+    gcode_M600(automatic, x_position, y_position, z_shift, e_shift_init, e_shift_late, lcd_load_message);
 
     // From this point forward, power panic should not use
     // the partial backup in RAM since the extruder is no
@@ -11076,9 +11088,9 @@ void M600_load_filament_movements()
 	st_synchronize();
 }
 
-void M600_load_filament() {
+void M600_load_filament(const char* lcd_load_message) {
 	//load filament for single material and MMU
-	lcd_wait_interact();
+	lcd_wait_interact(lcd_load_message);
 
 	//load_filament_time = _millis();
 	KEEPALIVE_STATE(PAUSED_FOR_USER);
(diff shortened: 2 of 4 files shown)


Panayiotis-git:
The look of the new functionality:
DSC03914_1

and the print that I couldn't complete without this PR:
DSC03928_1

wavexx:
Just some discussion:

- The M601 command inserted by the new "Pause at layer" function in prusa-slicer has the same issue: the custom message is immediately overwritten by the pause prompt
- Same is true from M108 (see #2055)
- M0/M1 is also broken currently (see #1898)
- Generally, M117 is lost very quickly
- Handling quoting as part of code_value() is, IMHO, either overkill if implemented properly or going to fail in subtle ways.

One solution that I was thinking would be the following:

- Always use M117 to set the display LCD
- Record the last millis() along with the message as a time-stamp
- During pause/filament swap/prompt, instead of overwriting the LCD, if ``millis()-lcd_update_stamp`` < 5 seconds (or a similar short time) then *prepend* the prompt instead of overwriting
- Convert M0/M1 to a M117+M601 internally: let's make pause reliable instead of adding another action

So let's assume the gcode:

```
M117 Insert bearings
M601
```

what I'd expect to see on the LCD is "Print paused: Insert bearings".

Panayiotis-git:
@wavexx I'm thinking of a message buffer, where the M117 message is stored and immediately displayed at the status bar. The first following command of the M600, M601, any Mxxx that requires this functionality, will consume this message, print it where it is needed and clear the buffer.

SyntaxNode:
Any status update on this feature request? This would make my life sooo much easier with manual multicolor prints.

Dexter0007:
Hi, I'm also need this functionality... color printing is hard without message

gudnimg:
I've added recently a common function to parse quoted strings, could that be used here to reduce memory usage?

Below is my guess how the code would look like without testing.

```cpp
char lcd_load_message[LCD_WIDTH + 1] = "";
if (code_seen('C')) {
    unquoted_string str = unquoted_string(strchr_pointer);
    if (str.WasFound())
    {
        const uint8_t len = min(str.Length(), LCD_WIDTH);
        memcpy(lcd_load_message, str.GetUnquotedString(), len);
        lcd_load_message[len] = '\0';
    }
}
...
gcode_M600(automatic, x_position, y_position, z_shift, e_shift_init, e_shift_late, lcd_load_message);
```

Panayiotis-git:
Hi @gudnimg, it seems reasonable.

Panayiotis-git:
Found some time and managed to rebase the PR code (forward four years!). @gudnimg, thanks for the suggestion, I've incorporated it in the PR.

Haven't tested the new code though.

3d-gussner:
@Panayiotis-git I tested your PR.
The `lcd_puts_at_P` rows are 0-3 and here my changes
Didn't check it with MMU3

[Fix_PR2380.diff.txt](https://github.com/prusa3d/Prusa-Firmware/files/13459238/Fix_PR2380.diff.txt)

gudnimg:
@Panayiotis-git Have you looked into using M117 like @wavexx describes here: https://github.com/prusa3d/Prusa-Firmware/pull/2380#issuecomment-569430271 I'm not sure if we even need to have a timeout on this message. Should be fairly simple unless I'm missing some edge cases.

It would be better I think since M117 will set the status message buffer and we can just display the message by printing the status line like we already do on the status screen. It will always show the string on the last row of the LCD.

It's unlikely PrusaSlicer will support the M600 C parameter since it would be specific to this firmware only.

Panayiotis-git:
> It's unlikely PrusaSlicer will support the M600 C parameter since it would be specific to this firmware only.

But it already supports it. :)
Well, kind of...

The slicer injects the command `T<tool_number>` whenever a tool change is required.
So, with a little post-processing of the generated gcode file, by replacing the `T` commands with `M600 C"<some name>"` commands, the script is ready!
The attached vbscript file is the one I have used back then to prepare the final gcode files.
[replace_for_M600.vbs](https://github.com/prusa3d/Prusa-Firmware/files/13465434/replace_for_M600.vbs.txt)

### M117 Proposal
In order to use the command `M117`, I assume that at the gcode file it requires something like:
```
M117 Blue Filament
M600
```
This looks good and will similarly do the job. We should check the required changes to the source files and then evaluate which solution is more efficient regarding the Flash memory.

### Integration with slicer
For both the approaches there is a need to somehow integrate with the slicer software. What I can think of is:

1. Use a post-processing script, like the one mentioned above, to replace the `T` commands with the required g-codes.
2. Use macros at the Tool change G-code field of the Printer Settings:
```
{if next_extruder = 0}M117 Red
{elsif next_extruder = 1}M117 Blue
{elsif next_extruder = 2}M117 Green
{elsif next_extruder = 3}M117 Yellow
{endif}
M600
```
I'm not sure though, what will be the effect of the remaining `T` commands in the generated gcode file.

3d-gussner:
@Panayiotis-git Have you been working on this PR to NOT change the `M600` gcode but a combination of `M117 + M600`?
Side note the syntax is `{if next_extruder == 0}` otherwise you get an error in PS2.7

Panayiotis-git:
Hi @3d-gussner,
Just a quick heads-up. I've taken a deep dive into the proposal regarding the use of the M117 code to set messages for the subsequent M600 command, particularly during filament loading. Unfortunately, it seems that the M600 command is clearing out any preset M117 messages right from the filament unload step.

Here's a snippet of the code where the issue seems to be arising:
```
custom_message_type = CustomMsg::FilamentLoading;
lcd_setstatuspgm(_T(MSG_UNLOADING_FILAMENT));
```
So far, I haven't found any workarounds. I'd appreciate your thoughts or any suggestions you might have to tackle this.

3d-gussner:
> Hi @3d-gussner, Just a quick heads-up. I've taken a deep dive into the proposal regarding the use of the M117 code to set messages for the subsequent M600 command, particularly during filament loading. Unfortunately, it seems that the M600 command is clearing out any preset M117 messages right from the filament unload step.
> 
> Here's a snippet of the code where the issue seems to be arising:
> 
> ```
> custom_message_type = CustomMsg::FilamentLoading;
> lcd_setstatuspgm(_T(MSG_UNLOADING_FILAMENT));
> ```
> 
> So far, I haven't found any workarounds. I'd appreciate your thoughts or any suggestions you might have to tackle this.

Maybe we can save the "M117" status message in an additional variable and use it in the `lcd_wait_interact` to be shown.
After it has been shown the new M117_message also should be cleared.

3d-gussner:
@Panayiotis-git Can you close this in favor of #4548

Issue #1164, opened 2018-09-17
Purge Line  Different Lenghts


christizzz:
Hello,

The Purge line are getting different lenghts randomly... the start code is the default one.. sometime it is the short default purge line.. and some times is 160mm long... it seems that the printer has some issue between relative and absolute for this purge line... Relfashing didnt change anything.. it is not so problematic.. but the issue is therr

stahlfabrik:
Yep it has been like that since I got my MK3 last year. Sometimes it is double length

Moorviper:
do you use different materials then it would make sense

christizzz:
Same Gcode...on the Sd card produced 2 different lenghta... shortt lenght: 100mm which is the normal behavior...as in Move in X to 60 then move to 100... The Long purgelines occurs if i have stoped a print before, using the Stop print in the menu.. and starting a new print will generate this long purge line... it is exactly 160...some how it switches to relative system.. it moves 60 then it moves 100.... which makes 160mm purge line... why does this occur after stoping a pprint ? no idea

Panayiotis-git:
IIRC, I had the same problem sometime ago with the MK2.

You can add the g-code G90 to the start g-code of your printer profile, just after mesh bed leveling.

It will switch the printer to absolute positioning mode.

Area5142:
From time to time, I have seen this randomly longer purge line since I got the MK2.
It is also mentioned in this thread on the forum:
https://shop.prusa3d.com/forum/general-discussion-announcements-and-releases-f61/print-keep-stopping-to-unload-and-load-filament-t23245.html#p99172

AbeFM:
Two nights ago mine went so far that it ran off the end of the bed and triggered a stall-and-rehome

GWdd:
I have the had the issue since new, never been a problem.  But, caught a file that does it repeatedly; every time.  The cause was a typo on my part; but the result exactly mimics the random events I've seen. 

In the repeatable case, I shot myself in the foot and used G92 when I wanted to use G90 as part of an anti-ooze strategy. The fact it mimics the hiccups, it might be useful (.1. file).  It seems the printer - for whatever reason, glitches into relative mode during purge.  

That Slic3r sometimes inserts code between the Custom Printer gcode is a concern, since it adds the chance of other processes corrupting the command. As Slic3r did here (.1. file): I tried removing the extra timer lines to no effect.
```
G1 X60.0 E9.0  F1000.0 ; intro line
[...]

PR #2381, opened 2019-12-23
Remove "relative_mode", only use "axis_relative_modes"

wavexx:
``axis_relative_modes`` can cleanly handle G90/G91 and M82/M83 without the need of "relative_mode" entirely.

Change G90/G91 to simply reset all axes according to the requested state, which avoids to check two variables in get_coordinates(), fixing the following incorrect handling:

```
G91   ; all axis relative
M83   ; set extruder absolute
G1 E1 ; should be an absolute move, but still relative
```

This was discovered by @leptun, and should fix the randomly occurring purge-line issues mentioned in #1164

PFW-1075

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -342,10 +342,6 @@ static float next_feedrate;
 // Original feedrate saved during homing moves
 static float saved_feedrate;
 
-// Determines Absolute or Relative Coordinates.
-// Also there is bool axis_relative_modes[] per axis flag.
-static bool relative_mode = false;  
-
 const int sensitive_pins[] = SENSITIVE_PINS; // Sensitive pin list for M42
 
 //static float tt = 0;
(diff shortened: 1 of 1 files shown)


leptun:
@wavexx I’m not sure if this also fixes the purge line length. We might have to also fix the lcd stop command chain (and the random no parking issue as a bonus)

wavexx:
Look at the comment by @Panayiotis-git about G90 in issue #1164  

Regarding the extruder not parking, I really cannot reproduce it. I tried multiple times/combinations today, but no dice. The extruder always parks, although you see the segmented move.

I agree I'd keep this PR pending until we can confirm that this is completely fixed though.

Firmware/Marlin_main.cpp
```diff
@@ -8279,7 +8279,7 @@ void get_coordinates()
   for(int8_t i=0; i < NUM_AXIS; i++) {
     if(code_seen(axis_codes[i]))
     {
-      bool relative = axis_relative_modes[i] || relative_mode;
+      bool relative = axis_relative_modes[i];
```
DRracer, line 8282: wasn't axis_relative_modes array used for remembering the previous state for quick return from relative mode?
It looks that way (judging from the fact, that there is a separate array of axis' relative flags + one global relative flag for all of them).
Or was it just a relict of the past?

DRracer: @leptun explained: Probably some old forgotten array/bool that was used incorrectly. We now do it the right way

DRracer (approved):

PR #2383, opened 2019-12-23
Fix fan_check_error check in usb print resume

wavexx:
Also check for fan_check_error == EFCE_OK in addition to EFCE_FIXED.

This allows resume when printing via USB after an M601.

Reorganize the check in order to fix both #if[n]def FANCHECK cases,
as similarly done in the SDSUPPORT case a few lines below.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -6910,10 +6910,15 @@ static void lcd_main_menu()
   }
 
 
+  if(isPrintPaused && saved_printing_type == PRINTING_TYPE_USB)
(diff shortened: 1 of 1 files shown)


3d-gussner (approved):
I can confirm that the cleared LCD screen effect is gone and everything works fine and better than before.
See https://github.com/prusa3d/Prusa-Firmware/pull/2404#pullrequestreview-339885855

UPDATE: Saw the cleared LCD screen again...but that is "normal"

DRracer (approved):

PR #2386, opened 2019-12-24
Preheat to saved temperature in load/unload during pause

### LINKED ISSUE #2385: [BUG] Load/unload filament in Paused state shouldn't ask for temperature
**Printer type** - MK3/MK3S
**Printer firmware version**- 3.8.0+

**Describe the bug**
When the printer is paused through the LCD or via M601, the hotend is disabled.
In the Tune menu you have the option to Load/Unload the filament to do a filament change.
Since the hotend is disabled, the Preheat menu pops-up, asking for the material.

This shouldn't be the case in the paused state: the saved temperature should always be used to preheat the hotend immediately without asking.

Using the standard preheat menu, among other things, clears the ``bFilamentPreheatState`` flag. If the preheat is cancelled, the bed temperature is reset to 0!

**To Reproduce**
Thanks to @Panayiotis-git for the test case in #2335:

*   Start a print.
*   Pause the print (Nozzle temp targets to 0, Heated bed keeps the target temp)
*   Wait for the nozzle to cool down
*   Load/Unload filament, choose one. The hotend starts warming.
*   Press Cancel. This will set both the nozzle's and the bed's target temperatures to 0.
*   At this state, if Resume is selected, then first the hotend will restore its temperature and after it reaches the target temp, then starts the heating of the bed.
  
**Expected behavior**
Using load/unload when the printer is paused should preheat the hotend with the saved extruder temperature.

wavexx:
When the printer is paused, use the saved_extruder_temperature directly
in preheat_or_continue, which is used by load/unload filament in the
menu. The material temperature is known is this case.

Also ensure bFilamentPreheatState is set. If the preheat is cancelled,
the temperature returns to the paused state instead of being cleared
(which would cool down the bed, detaching the print!)

Fixes #2385


wavexx:
Filament autoloading needs to have the same treatment, so consider this as a WIP until I also handle this case.

Panayiotis-git:
With the fix applied, the heated bed is not allowed to cool down.
However, if the Cancel is pressed, the nozzle's target temperature remains at the saved_extruder_temperature. Perhaps, it should be set back to 0.0:

```setTargetHotend0(isPrintPaused ? 0.0 : (float )nTargetOld);```

wavexx:
You're right, indeed. I'd like to make nTargetOld to be zero when mFilamentItem is entered so that no extra check is necessary..

wavexx:
Wow.. the entire menu implementation is.. bad :(

Using ``bFilamentPreheatState`` is abused to go back 1 or 2 levels so if the temperature is already in range no feedback is shown. I'm not sure if this fully intended...!?

``nTargetOld`` is also reset at every lcd refresh, so whatever "old" was before entering the preheat menu it's immediately lost as soon as preheating is started, making it useless.

My guess is that ``bFilamentFirstRun`` was the flag intended to keep the initial value, but was simply forgotten. It's almost never used, and it's set to true exactly before showing the "load/unload filament" entries. It's currently cleared too soon to be useful in ``preheat_or_continue``.

We could use that the same way you used ``bPreheatOnlyNozzle``: use it to store the initial value of the temperature, then clear it immediately in ``mFilamentItem``. Instead of checking for ``bFilamentPreheatState`` at the end, I'd always restore from nTargetOld, but there's some reshuffling to do, and we need to set it carefully before each menu entry that calls preheat_or_continue.

Not nice...

@Panayiotis-git do you want to work on this? I'd close this PR, since the changes would conflict...

wavexx:
Closing this until all the kinks are ironed out

Issue #2161, opened 2019-08-31
Feature Request: Rehome X/Y in tune menu when paused


vintagepc:
I'd like to request a "tune" menu item to re-home X and Y. I've found myself needing/wanting to do this in a number of cases, either when having paused a print to address something like warp/lifting or filament change with an unintended bump of the X/Y causing a layer shift. 

I can think of any number of other cases where it might be nice to re-home X and Y on demand mid print after addressing an issue, which could save a several hour print from becoming a spaghetti blob.

Not a feature that will be used particularly often, but a lifesaver when it is actually needed, since connecting to USB resets the printer.

michalxfanta:
Now, when printing is paused, the extruder moves to the left-rear position, how much is it different from the left-front position, which you need?

vintagepc:
That doesn't re-home the printer's zero position, does it? So if there's a minor layer shift it will still persist. But if you can rehome X/Y when paused it can be easily corrected.

3d-gussner:
@michalprusa I think it is not the pause parking position is an issue. But WE clumsy users may unintentionally hit the bed or the extruder and then a Rehome xy funtcion in Tune menu would prevent a failed print due to US clumsy users forced a layer shift.

vintagepc:
^^ Yeah, or there have been a few times where I've been sitting next to the printer and heard it thump and skip a step from hitting a print booger; I can pause it right away, clip the booger, and (ideally) rehome and continue on as normal; if I'd left it alone it would have introduced a layer shift; now I'd only have a small imperfection that can be more easily trimmed or filled.

michalxfanta:
OK, I finally understand that.

3d-gussner:
> ^^ Yeah, or there have been a few times where I've been sitting next to the printer and heard it thump and skip a step from hitting a print booger; I can pause it right away, clip the booger, and (ideally) rehome and continue on as normal; if I'd left it alone it would have introduced a layer shift; now I'd only have a small imperfection that can be more easily trimmed or filled.

Another good example where a Rehome xy in Tune menu could save our prints .... Last Night a User saved my Print...lalala...He cut the bugger and Rehomed xy ... lalala ... Last Night a User saved my Print 

Glad I am not a song writer :-)

michalxfanta:
@vintagepc please contact me at michal.fanta@prusa3d.cz thx

vintagepc:
[...]

PR #2390, opened 2019-12-28
Fix filament change behavior when issued in paused state

### LINKED ISSUE #2387: [BUG] "Tune" menu should be available during Pause
**Printer type** - MK3/MK3S
**Printer firmware version**- 3.8.1

**Describe the bug**
When the printer is paused, "Preheat" is available, which doesn't make sense.
"Tune" should be kept.

"Tune" also includes the ability to do a filament swap, which is often something you'd might need during a pause (instead of going through the unload+load route), with the added benefit of the extruder moving to the front.

In the last case though, "Change filament" also needs to restore the hotend temperature before attempting to load, which is something that is currently missing (loading will happily happen with 0 degrees, with the cold-extrusion prevention saving the day).

**To Reproduce**
* Pause print via the LCD or M601
  
**Expected behavior**
* "Preheat" shouldn't be available during a pause
* "Tune" should be though.

wavexx:
Keep the "Tune" menu alive when the print is paused, disabling "Preheat" instead which doesn't make much sense.

All the actions which can be performed in Tune make sense during a pause, but this change is *especially* useful for the "Change filament" action, which moves the extruder to the front-left.

Most changes are related to gcode_M600 tweaks to handle a filament change correctly when issued from within the paused state via the LCD:
- re-heats the hotend to the saved temperature prior to unloading and cools it down again after reload
- keeps the filament in a retracted state, to be consistent with the paused state.

Fixes #2387

PFW-1097

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3718,7 +3718,7 @@ static void gcode_M600(bool automatic, float x_position, float y_position, float
 
     //First backup current position and settings
     int feedmultiplyBckp = feedmultiply;
-    float HotendTempBckp = degTargetHotend(active_extruder);
+    float HotendTempBckp;
     int fanSpeedBckp = fanSpeed;
 
     lastpos[X_AXIS] = current_position[X_AXIS];
@@ -3726,10 +3726,23 @@ static void gcode_M600(bool automatic, float x_position, float y_position, float
     lastpos[Z_AXIS] = current_position[Z_AXIS];
     lastpos[E_AXIS] = current_position[E_AXIS];
 
-    //Retract E
-    current_position[E_AXIS] += e_shift;
-    plan_buffer_line_curposXYZE(FILAMENTCHANGE_RFEED);
-    st_synchronize();
+    if (automatic || !isPrintPaused)
+    {
+        HotendTempBckp = degTargetHotend(active_extruder);
+
+        // Retract E
+        current_position[E_AXIS] += e_shift;
+        plan_buffer_line_curposXYZE(FILAMENTCHANGE_RFEED);
+        st_synchronize();
+    }
+    else
+    {
+        HotendTempBckp = saved_extruder_temperature;
+
+        // Reheat the extruder
+        setTargetHotend(saved_extruder_temperature, active_extruder);
+        mmu_wait_for_heater_blocking();
+    }
 
     //Lift Z
     current_position[Z_AXIS] += z_shift;
@@ -3795,11 +3808,23 @@ static void gcode_M600(bool automatic, float x_position, float y_position, float
     //Not let's go back to print
     fanSpeed = fanSpeedBckp;
 
-    //Feed a little of filament to stabilize pressure
     if (!automatic)
     {
-        current_position[E_AXIS] += FILAMENTCHANGE_RECFEED;
-        plan_buffer_line_curposXYZE(FILAMENTCHANGE_EXFEED);
+        if (isPrintPaused)
+        {
+            // Return to retracted state during a pause
+            current_position[E_AXIS] -= default_retraction;
+            plan_buffer_line_curposXYZE(FILAMENTCHANGE_RFEED);
+
+            // Cooldown the extruder again
+            setTargetHotend(0, active_extruder);
+        }
+        else
+        {
+            // Feed a little of filament to stabilize pressure
+            current_position[E_AXIS] += FILAMENTCHANGE_RECFEED;
+            plan_buffer_line_curposXYZE(FILAMENTCHANGE_EXFEED);
+        }
     }
 
     //Move XY back
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3853,7 +3853,10 @@ static void gcode_M600(bool automatic, float x_position, float y_position, float
 	fsensor_check_autoload();
 #endif //IR_SENSOR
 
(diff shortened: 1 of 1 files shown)


vintagepc:
Spiffy, this will make my work for #2161 a lot easier since I no longer need to do the legwork to expose tune, just add the menu entry!

wavexx:
Yup, I was thinking about that in recent changes. Just doing a G29 though might not work out of the box though - looking at the code earlier on, I spotted a couple of cases where the MBL offset might not be reapplied correctly after homing. Keep an eye on that

vintagepc:
I assume you're referring to within the actual G28 X Y (not 29? that's ABL...) code? Yeah, that's pretty messy... Without MBL the resume/recovery might not be perfect ("good enough" unless your bed is pretty bad...) It probably makes more sense to do a custom function and invoke homeaxis() directly then

wavexx:
Yes, G28 sorry. I would expect G28 X Y to "work" with MBL on though, because that's certainly what I would expect ;)

vintagepc:
Aye, until you dive into it and realize it blows away  Z regardless...
I'll poke at it and see if we can just do a clean direct home instead to avoid altering stored MBL. Seems reasonably safe since pause print already raises Z and avoids concerns about actually hitting the part. I eagerly await the PR dev program to see what that entails... things would be easier/faster if I didn't have to share my sole printer for both printing and tinkering XD... then I could probably even bust out my ICE/ISP for a massive QoL improvement in debugging.

DRracer:
@wavexx @vintagepc can you think of potential edge cases what the Tune menu may bring in Pause? Are there any items which may break the print or damage the machine?

vintagepc:
Not that I can think of, it all seems safe. I never did get around to adding the ability to rehome X/Y in there... :smile:

wavexx:
The only catch here is that "filament change" needs to consider that the paused state handles retract in a different way compared to the regular printing. This is already handled in the PR though, since being able to swap filament in the paused state was my main priority. Similarly, having Tune during pause is something I also considered for #3006 

Thinking out loudly here, one thing I don't remember how the changed settings carry over after the print is resumed (since in that moment the saved settings are recalled). If you change temperature or speed in the paused state, it makes sense that you'd like to carry those over to the resumed print..

Panayiotis-git:
@wavexx The resume of a paused print restores the hotend but not the bed temperature. One thing to consider is that there is a bug at the load/unload filament functionality and when during a paused print the user aborts the filament loading/unloading by selecting Cancel, then both hotend and bed set values are reseted to 0. This issue is handled at the https://github.com/prusa3d/Prusa-Firmware/pull/2335

wavexx:
I'm using your PR :). However in this case I was wondering what happens if you pause the print, immediately change the feedrate in the tune menu, then resume. Is the new feedrate kept, or is it forgotten? Same for temperatures. This is because we save the old temperature in a temporary, but the tune menu applies the value to the current setting. While resuming we overwrite the current settings with the saved ones..

wavexx:
After looking into this more, I think the behavior is still OK as-is:

- changing the bed temperature from the Temperature menu persists, because we don't save/change the bed temp during pause.
- updating the extruder temp of the saved print feels "broken", since we automatically cool-off the extruder. The user might want to change the extruder temp for other reasons during the paused state (change filament manually), so restoring the _original_ print temp is ok.

The only exceptions I think make sense are:

- Flow
- Feedrate

Flow is saved during pause, but it could make sense to persist over to the saved print.

Feedrate should instead change the feedrate of the _saved_ print, not for the feedrate of the paused state. This is unique to this PR though: the encoder handling in any menu needs to be updated, so it's technically "already broken" and I'd handle this separately.

wavexx:
Never mind, Flow sets extrudemultiply, which is _not_ saved/restored during pause, so it's ok too. Feedrate is really the only exception I see.

wavexx:
I've logged this as issue #3122 to keep track of this issue independently

wavexx:
Well, it turns out "Tune" was included during pause already in commit 59c2b7e79543f81089c32ca8db3aff0bb86b16bd.

I dropped the relevant commit. Looks like fixing the M600 behavior included this PR is still applicable.

Firmware/Marlin_main.cpp
```diff
@@ -3828,7 +3853,10 @@ static void gcode_M600(bool automatic, float x_position, float y_position, float
 	fsensor_check_autoload();
 #endif //IR_SENSOR
 
-    lcd_setstatuspgm(MSG_WELCOME);
+    if (isPrintPaused)
+        lcd_setstatuspgm(_i("Print paused"));////MSG_PRINT_PAUSED c=20 r=1
```
gudnimg, line 3857: Doesn't this overwrite the M117 message?

gudnimg: I see now my comment doesn't really make sense in the context of this PR.

When a M600 is issued during a paused print, it doesn't really matter if the M117 was called prior.

gudnimg:
Testing this on 3.13.2 and can confirm this PR is still very much relevant.

When I trigger a filament change when a print is paused (at least with MMU) it will attempt to change the filament while the target temperature is 0°C!

Target temperature is also 0°C when unloading.

gudnimg:
Tested the changes on top of latest MK3 branch (3.13.2 RC1). I created a separate PR with some minor changes on top https://github.com/prusa3d/Prusa-Firmware/pull/4408

3d-gussner:
@wavexx Closing in favor of #4408

PR #2394, opened 2019-12-31
Avoid more recursive behavior in fsensor_update

wavexx:
In the MK3 (with the optical/laser sensor) ``fsensor_update`` enqueues commands to the main loop to check for the filament status. In some circumstances (mesh_plan_buffer_line->sync->mesh_plan_buffer_line->filament_update->process_commands->runout) the recursive call to ``process_commands()`` will cause a stack overflow. The stack at this point is so large that frequently the stack guard doesn't even trigger.

We remove the recursive call to process_commands and replace it instead by calling ``plan_buffer_line`` directly. We also move some filament status checks earlier up in the call chain to shrink the stack further while the nested check is being performed.

As an added bonus, while we check for the filament status, we briefly lift the extruder to avoid scorching the print during the 2-second pause.

PFW-1086

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -8606,7 +8606,8 @@ if(0)
 #ifdef PAT9125
 				fsensor_autoload_check_stop();
 #endif //PAT9125
-				fsensor_update();
+                if (fsensor_enabled && !saved_printing)
+                    fsensor_update();
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


leptun:
@wavexx I don't understand why they keep pushing commands to the cmdqueue instead of calling plan_buffer_line. Is it maybe to reduce stack size? I don't think so since they run process_commands right after they enqueue them.

DRracer (approved):

PR #2395, opened 2020-01-02
Improve raise_z_above to always raise when at Z_MIN

wavexx:
When check_z_endstop is set, Z_MIN_PIN is checkend regardless of the
moving direction to support Z calibration. This prevents the ability
to use _just_ SG when moving upwards.

But since we know the extruder is at Z_MIN, it's always safe to raise
irregardless, so we can dodge the issue.

PFW-1074


3d-gussner (approved):
Works fine.
During testing an issue with z 210 been displayed on screen after reset and preheat this fix solved it.

DRracer (approved):

PR #2402, opened 2020-01-08
Do not show sheet on status screen if less than 2 are initialized

leptun:
Just as the title says, if less than 2 sheets are initialized, the current sheet will not be shown on the status screen while idle.

PFW-1077

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -630,7 +630,7 @@ void lcdui_print_percent_done(void)
 	{
 		eeprom_read_block(sheet, EEPROM_Sheets_base->s[eeprom_read_byte(&(EEPROM_Sheets_base->active_sheet))].name, 7);
 		sheet[7] = '\0';
-		lcd_printf_P(PSTR("%s"),sheet);
+		lcd_printf_P(PSTR("%-7s"),sheet);
 	}
 	else
 	{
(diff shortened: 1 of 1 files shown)


Firmware/ultralcd.cpp
```diff
@@ -622,21 +622,24 @@ void lcdui_print_feedrate(void)
...
-		sprintf_P(per, num?_N("%3hhd"):_N("---"), calc_percent_done());
-		lcd_printf_P(_N("%3S%3s%%"), src, per);
+		const int8_t sheetNR = eeprom_read_byte(&(EEPROM_Sheets_base->active_sheet));
+		const int8_t nextSheet = eeprom_next_initialized_sheet(sheetNR);
+		if ((nextSheet >= 0) && (sheetNR != nextSheet))
+		{
+			char sheet[8];
+			eeprom_read_block(sheet, EEPROM_Sheets_base->s[sheetNR].name, 7);
+			sheet[7] = '\0';
+			lcd_printf_P(PSTR("%s"),sheet);
```
DRracer: please improve the formatting string so that shorter names do not screw the status screen (what is written next to it)
something like "%-7.7s"

leptun: I used "%-7s" as that is enough

DRracer (approved):

PR #2404, opened 2020-01-08
New lcd_print_stop function

leptun:
This PR addresses the countless issues with the stop print function. It is based on and requires #2382. 
It does NOT retain MBL active as that would cause shifting of the Z axis position because of the two planner_abort_hard() calls.
@wavexx Does the MBL really need to stay enabled after stop?

PFW-1075


3d-gussner (commented):
Test results leptuns Fix pause->stop->PrintFromSD scenario firmware

- 1st SD print: nice purge line, stopped and extruder moved to park position without hick-up
- 2nd SD print: nice purge line, paused with move to park position, resumed, stopped and extruder moved to park position without hick-up
- 3rd SD print: nice purge line, stopped and extruder moved to park position without hick-up

looking good
FW 3.8.1-2869 git hash 9dfd2491

I will test Octoprint as well

3d-gussner (commented):
est results leptuns Fix pause->stop->PrintFromSD scenario firmware #2 with Ocotprint
1st SD print started via Octoprint:
nice purge line
- "ENCLOS->1" on display line 4 which is wrong it should be "PLA_WH~1"
- stopped via LCD and extruder moved to park position without hick-up
- Hotend and Bed target temp to 0

2nd SD print started via Octoprint:
- nice purge line
- "ENCLOS->1" on display line 4 which is wrong it should be "PLA_WH~1"
- paused via LCD with move to park position, hotend target temp to 0, bed stayed up
- resumed via LCD STRANGE the LCD cleared and i thought printer reset but it came back
- paused via Octoprint with move to park position, hotend target temp to 0
- resumed via Octoprint STRANGE the LCD cleared and i thought printer reset but it came back
- stopped via LCD and extruder moved to park position without hick-up

3rd SD print started via Octoprint:
- nice purge line
- "ENCLOS->1" on display line 4 which is wrong it should be "PLA_WH~1"
- paused via Octoprint (M400) with move to park position, hotend target temp to 0
- resumed via Octoprint (M24) STRANGE the LCD cleared and i thought printer reset but it came back
- Communication lost and Octoprint hang up on me REBOOT of Octoprint

4th SD print started via Octoprint:
- nice purge line
- "ENCLOS->1" on display line 4 which is wrong it should be "PLA_WH~1"
- paused via Octoprint (M400) with move to park position, hotend target temp to 0
- resumed via Octoprint (M24) STRANGE the LCD cleared and i thought printer reset but it came back
- stopped via Octoprint (M108) with move to park position, hotend + bed target temp to 0

5th Octoprint print started via Octoprint
- nice purge line
- Display shows USB 0% not filename in line 4
- The menu during printing shows "Live adjust Z", "Tune", "Resume print", "Stop print", "Fail stats" - and "Support"
- Why is there the "Resume print" and not "Pause Print" ?
- Octoprint pause: Hotend and bed stay at print temp. Extruder doesn't move
- Octoprint resume: continuous print
- LCD Resume print ... prints something else (previous print PLA_WH~1) Display changed to SD 3%
This is a BUG
- stopped the print via LCD menu and extruder moved just a bit up NOT in park position. Hotend and bed target temp are 0

Why does Questions:
Why does it show the wrong filename on display?
Why does the screen clears with resume?
Why does the Printing menu shows "Resume print" when an Octoprint is running? Selecting it also prints some other gcode.
FW 3.8.1-2869 git hash 9dfd2491

Could not reproduce the printing menue BUG during Octoprint print started from Octopring showing "resume print " and "stop print"

3d-gussner (approved):
Approved

DRracer (approved):

wavexx:
> @wavexx Does the MBL really need to stay enabled after stop?

Only for M999. We can store the state previous of mbl and re-enable it on M999 though.

When I read the documentation about M999 on marlin though, it feels like the intended usage is different. In our case we don't allow restart after a heater error and I cannot really think of a useful case for M999.

I didn't test this PR yet (yeah I know it's merged), but there's a catch. Calling planner_abort_hard will free the planner, but you still need to let the the process_command function (which likely is being frozen somewhere) to complete by returning.

You might have a stopped process_command being frozen after parsing M140 but prior to setting the actual heater to on. Stop is called, it parks the heat, but then process_commands finishes running re-setting the heater to on.

That was the idea behind the lcd_commands_type: set some stuff to be done "later" (when the process_command call is complete). It was badly done, but it made sense.

leptun:
@wavexx so what do you propose? Moving all stop commands in a single step and still use lcd_commands_type implementation?

wavexx:
Yes, pretty much like before, but without having all those stop subcommands.
I'm not sure why the lcd_commands_type had many steps where one could do..  (ideas?)

leptun:
They would queue commands and also set lcd messages, hence why things had to be split. Now that we plan lines the issue is no more.

leptun:
@wavexx I might change it so all commands are done in one lcd_commands_type step avoiding having to call planner_abort_hard two times. Anyway, it's probably not that bad right now and we'll have the chance to do so for 3.9-RC1. I'd still disable MBL though. planner_abort_hard() might mess up if we don't disable MBL since it also compensates for current MBL offset in current_position

wavexx:
I'm fine with disabling MBL. I actually would be in favor of even deprecating M999...

leptun:
Me too. That will come in the next planned release, though. There's no time now to change anything and we still have a few weeks to finish this area.

PR #2405, opened 2020-01-08
⚡️New SD menu

leptun:
This PR is a compilation of several changes I've made for some time regarding the "Print from SD" menu and file sorting in general.

For a starter, let me tell about the **sdPrint menu**.
The main problem that I see with it is that it was a custom implementation of the menu. This is not that bad by itself, but combined with the fact that it also had to do the scrolling of text, it soon became obsolete in the way it was designed and it just became very inefficient compared to the rest of the menus. There's also the fact that while the text was scrolling, it used some really weird delays and it was a blocking function overall.
With my new sdPrint menu implementation I also use a custom implementation, but I try to make it as similar to the other custom menus as possible.
Now, the idea behind this menu is to be as non-blocking as possible. I was able to achieve this by some submenu trickery. ~~We now consider the scrolling text an actual submenu and the code seamlessly transitions between the big menu and the small menu which only deals with scrolling text on the current line. This part of the implementation could be improved as the code is not entirely straightforward, but I think it is a good start.~~ The menu is now still one submenu, but it has three main states in which it can be: Uninitialized, standard and scrolling.
- Uninitialized retrieves some variables which are slow to load, are constantly needed and don't change under normal circumstances.
- Standard is just the regular rendering which deals with the menu structure
- Scrolling deals only with one line at a time scrolls the text on that one.

Another very important change I did to the sdPrint menu is **caching the file positions**. Before this change, the code had to search all files by filename on the SD card which is really slow since it had to load each and every file for it to compare them and determine that it has found the file. What I did to combat this is caching the actual position of the file on the sd card just like in the sorting algorithm so that I can just call getfilename_simple(uint32_t) instead of getfilename(index) which has to load all files in the work directory till it gets to that one that is requested. The advantages to this method are obvious: The menu is blazing fast now even if you have 10 or 100 files in that directory.
There are some disadvantages though, but I don't think they outweigh the advantages:
- The benefits only work for the first 100 files (just like with sorting). The rest use the old(slow) method.
- It uses more ram, ~400B more for the cache. There are ways to reduce that to ~300B, but that is the least you could ever get. At the very least it is static, so no dynamic allocation issues should happen with this.

And now for the one you've probably been waiting for the most, **faster sorting**.
Sorting with many files is really slow and you might wait more than 30 seconds sometimes. This is the fault of the bubble sort algorithm which is rather inefficient for this use case. The main problem with it is that it has to compare many times until everything is sorted and the worst part is that it has to load at least one file for every comparison. As a better approach, I chose to use the shellSort algorithm which should be much better suited for this kind of application than bubble sort. It is a bit more complex, but it is still in-place and it has to do much fewer comparisons, hence why it is faster.
The only problem I had with it is that normally it wouldn't sort folders, but I solved this by running the algorithm two times: Once to sort all files and to move the folders at the beginning, the second one to sort only the folders.

Bonus feature: When flashair mode is enabled, you get an additional button for reloading the files. Should make resetting the firmware for the uploaded file to appear unnecessary.

I'd be glad if anyone could provide feedback on this.

PFW-1146


Panayiotis-git:
Nice work! 
First impression is that it is fast. The scrolling through the file list is quick and responsive.

I've noticed that the long press to switch to the move Z axis, is no longer available when in the sd card menu. At the original implementation, when the long press happens during a long filename scrolling, the menu change is postponed till the end of the scrolling (missing conditions to break the scrolling). Also afterwards, when returning from the move Z menu, it first scrolls the entire filename and then prints the rest of the file list.

Perhaps, during the folder sorting (second pass) the message should mention 'folders' instead of 'files'

Panayiotis-git:
When entering a folder, the first refresh shows short file names. It requires a one line scroll to refresh with the long filenames.

leptun:
Folder issue: I am aware of that issue. I’ll try to fix it later. Right now this PR is not really a priority for me. More important is releasing 3.9 (soon) and ironing out the bugs that might rise from it.

Thanks for testing it. All the feedback will be used later when I come back to it.

leptun:
@Panayiotis-git I'm back at this PR. I've just pushed a fix for lcd_subdir and lcd_updir. I'm still making some improvements, so stay tuned.

leptun:
@Panayiotis-git I also tried to fix the longpress behavior. Now you can enter the menu also when it's scrolling. There's a small render issue when you return from the move_z menu, but that is much better than before. Can you please provide feedback on the latest SD menu?

leptun:
I've just done a lot of reorganization. The code is much cleaner now. I've also fixed the long-press behavior completely.

mkbel:
If the RAM consumption is perceived as a problem, wouldn't it be possible to use RAM allocated for screen lifetime only? There is already mechanism for that.

leptun:
@mkbel I need the uint32_t array of file positions to be generated during sorting and then used in the menu.
I think these are my only options:
- use a static array. As it is now
- dynamically allocate the array to the needed size based on the file count sorted. Max 100 files/dirs
- dump all that data in some unused buffer. Worst option in my opinion.
I also would like to avoid generating the array when entering the menu and dumping it on exit as that can generate a few seconds of delays in some situations.

The only such mechanism you talk about that I know of is menu_data[], but that is 32B long and I need 400B (or a more efficient way to store those file positions).

mkbel:
Yes, I meant menu_data[]. Yes I meant to increase menu_data[] to 400 bytes. I thought that menu_data[] is already bigger, but at least this approach saves 32B of RAM, it is better than nothing.

mkbel:
But I understand that you don't want the drawback of index lost on exit of SD menu.

leptun:
That would be ideal. I’d like to avoid showing the user a “Loading files” screen every time they try to open the menu.

mkbel:
What about saving index to EEPROM? As you already said index generating is slow, so it wouldn't be big penalty to save it to EEPROM which is slow to write. But you don't need to cache EEPROM read, EEPROM is fast to read, so you can read index directly from EEPROM.

mkbel:
And there would be no EEPROM write overhead if index doesn't change, as normaly EEPROM update function is used, which doesn't erase and write cell if content is already the same.

mkbel:
So if content of SD is not changed, EEPROM is not updated.

leptun:
That is an interesting idea. I’ll try it and see how it goes. What worries me is the write endurance. Going into a directory will completely overwrite the cache.

mkbel:
Oh, a thought that whole content of SD is indexed at once.

mkbel:
This wouldn't be great idea to write EEPROM when navigation subdirectories.

mkbel:
EEPROM cell erase/write cycle endurance is 100 000.

mkbel:
So it is not good to write something more then 50 times a day to the same cell.

Panayiotis-git:
It came to my mind so let's throw it on the table.
What if the results of the sorting are stored in a file at the current directory of the SD card? While entering to a folder, do the sorting only if the file does not exist or if changes are detected at the content of the folder till the creation of the file. Count of files/folders and timestamps could be used to detect changes.

leptun:
@mkbel how is the SD menu and sorting handled on the Mini?

mkbel:
I don't know the answer. @DRracer is in charge of answering you.

DRracer:
@leptun this is the engine, which is now (upcoming FW 4.1) responsible for keeping track of files and directories displayed to the user:
https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/master/src/gui/lazyfilelist.h

gogglespisano:
> do the sorting only if the file does not exist or if changes are detected at the content of the folder till the creation of the file. Count of files/folders and timestamps could be used to detect changes.

Agreed

Firmware/Marlin_main.cpp
```diff
@@ -750,23 +750,32 @@ static void factory_reset(char level)
...
-			softReset();
-
+#ifdef FILAMENT_SENSOR
+		fsensor_enable();
+		fsensor_autoload_set(true);
+#endif //FILAMENT_SENSOR
+		break;
+	case 3:
+		menu_progressbar_init(EEPROM_TOP, PSTR("ERASING all data"));
+		Sound_MakeCustom(100,0,false);
```
DRracer: the sound call has been extracted to the beginning of this function, please remove this `Sound_MakeCustom(100,0,false);` line from here

Firmware/menu.cpp
```diff
@@ -551,4 +553,33 @@ uint8_t menu_item_edit_P(const char* str, T pval, int16_t min_val, int16_t max_v
 template uint8_t menu_item_edit_P<int16_t*>(const char* str, int16_t *pval, int16_t min_val, int16_t max_val);
 template uint8_t menu_item_edit_P<uint8_t*>(const char* str, uint8_t *pval, int16_t min_val, int16_t max_val);
 
-#undef _menu_data
+static uint8_t progressbar_block_count = 0;
```
DRracer, line 556: Actually this solution is not that bad - I can't imagine what situation may require having 2 independent progress bars on the LCD. The only syntactic sugar (+a bit of safety of use) may be wrapping the progress bar into a class + a gobal static instance of it (ideally a singleton, but gcc 4.9 on AVR creates default 16B locks around singletons, which is not something we want).

Firmware/messages.c
```diff
@@ -79,7 +79,7 @@ const char MSG_PRESS_TO_UNLOAD[] PROGMEM_I1 = ISTR("Please press the knob to unl
 const char MSG_PRINT_ABORTED[] PROGMEM_I1 = ISTR("Print aborted"); ////c=20
 const char MSG_PULL_OUT_FILAMENT[] PROGMEM_I1 = ISTR("Please pull out filament immediately"); ////c=20 r=4
 const char MSG_RECOVER_PRINT[] PROGMEM_I1 = ISTR("Blackout occurred. Recover print?"); ////c=20 r=2
-const char MSG_REFRESH[] PROGMEM_I1 = ISTR("\xF8" "Refresh"); ////
+const char MSG_REFRESH[] PROGMEM_I1 = ISTR("\x04" "Refresh"); ////
```
DRracer, line 82: is this change related to the update for the refresh icon/custom character?

leptun: Yes

Firmware/Marlin_main.cpp
leptun: If you need it right now for testing, please remove it.

Firmware/cardreader.cpp
```diff
@@ -821,177 +825,160 @@ void CardReader::presort() {
...
 		// By default re-read the names from SD for every compare
 		// retaining only two filenames at a time. This is very
 		// slow but is safest and uses minimal RAM.
-		char name1[LONG_FILENAME_LENGTH + 1];
+		char name1[LONG_FILENAME_LENGTH];
 		uint16_t crmod_time_bckp;
 		uint16_t crmod_date_bckp;
 
+		#if HAS_FOLDER_SORTING
+		uint16_t dirCnt = 0;
```
DRracer, line 837: a question: since we don't support sorting > 255 files this may get reduced to uin8_t along with other related variables as well. However, I can't see all the potential edge cases of such a change

leptun: It will probably work with it reduced to uint8_t. Not sure if it will generate warnings tho.

DRracer: just something to think about (makes code shorter and faster ;) )

leptun: I'll leave it like this for now since FAT32 uses 16bit dir entries, that meaning we can fit 65535 files in one directory. Because I don't want to generate unpredictable behaviour, I'll leave everything file count related to uint16_t, even if uint8_t might suffice in this case

DRracer (approved):
After long-term tests and several fixes this looks good to go. Even with the bubble sort algorithm (as before) it is still slightly faster than the previous SD menu implementation.

Important edge cases have been fixed/handled, so this new code is much safer.

Future improvements are possible, but the originally proposed shell sort alg. is not mature enough yet (thus not enabled by default).

leptun:
Due to some issues with the shellSort algorithm, it was temporarily reverted to bubbleSort. Until these issues are ironed out or we implement a better algorithm, it will stay like this. The shellSort code is still there, just disabled.

PR #2408, opened 2020-01-09
Clear the USB printing counter when cancelling a print

smurfix:
Small patch to show the "real" menus immediately after printing is cancelled (i.e. steppers are off) when printing serially / via USB. Having to wait for the timeout doesn't make sense.


3d-gussner:
@smurfix Thanks for the PR, but we will not change M84. Closing

smurfix:
? What do you mean, "change M84"? This patch doesn't change what M84 does, which is (essentially) to tell the printer that printing is finished. Thus it makes sense to show the menus that usually get displayed when the printer is idle *now*, instead of waiting for the timeout.

PR #2420, opened 2020-01-16
Version changed (3.9.0 build 3175)

DRracer:


XPila (approved):

3d-gussner:
Shouldn't it be `#define FW_VERSION "3.9.0-BETA"`  and `#define FW_DEV_VERSION FW_VERSION_BETA`

At least for the beta tag? This way user has to confirm on LCD that it is a BETA version if they compile it their self.

Issue #2415, opened 2020-01-14
[FEATURE REQUEST] LCD option to unload new filament on M600


Chompworks:
Problem:
When using M600, after using the LCD to remove filament, the user is prompted to load new filament. When that filament triggers the sensor, a small length is extruded and the LCD prompts for extruding more, resuming the print and a few other options. At the moment there is no way to remove the filament again in the current LCD menu flow. 

Proposed/Ideal solution:
Add a "unload filament" option (for non-MMU firmwares) AFTER the loading of new filament. In the below scenario, the user could then select this, and the firmware would invoke the unload procedure again, and then request the user to insert new filament again.

Why this is useful:
There is no easy way to remove a piece of filament if it was fed in with an unsuitable plug at the end. In that situation the filament plug tends to get stuck in the small bowden tube directly below the bondtech gears, but there is no LCD menu option currently to retract the filament. The only other option of forcing further extrusion is undesirable as it may grind the filament, and even break the plug off from the filament feed, necessitating opening up the extruder and removing it manually. Often that means there is no good way to save the print in progress.

If there was a way in the LCD menus to allow a user to choose to retract the filament, this would allow the user to reinsert the filament after trimming a problematic plug. The above scenario would be avoided, without loss of the print in progress.

wavexx:
I know I had this issues a few times (I opened the extruder to disengage the gears and pulled manually).

thew1se:
I was somehow able to get to the move axis menu and retract z-axis. And then resume print.  Can't remember how though...it's been a while.

Prusa-Support:
For the time being, the recommended method to get around these accidental occurrences is:
- open the idler door,
- guide the filament with needle pliers,
- purge more after the intervention.

Michele Moramarco
Prusa Research

Ro3Deee:
> This issue has been closed due to lack of recent activity.

seems it's planned https://github.com/prusa3d/Prusa-Firmware/pull/2423#event-10009222632

Prusa-Support:
Yes 🙂 by now it is merged too, hence ready for the next FW release.

Michele Moramarco
Prusa Research

PR #2423, opened 2020-01-17
LCD option to unload new filament on M600 (issue #2415)

Panayiotis-git:
I have met sometimes the situation described at the issue #2415 and I think that it would be useful to add the option.

Since the screen was already filled with the header and the three choices, I assigned the "Yes" choice to the header. So the new screen is:
```
>Changed correctly
 Filament not loaded
 Color not correct
 Unload filament
```

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -446,7 +446,7 @@ void gcode_M701(float fastLoadLength, uint8_t mmuSlotIndex);
 void M600_load_filament();
 void M600_load_filament_movements();
 void M600_wait_for_user();
-void M600_check_state();
+bool M600_check_state_and_repeat();
 void load_filament_final_feed();
 void marlin_wait_for_click();
 float raise_z(float delta);
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3473,7 +3473,6 @@ static void gcode_M600(const bool automatic, const float x_position, const float
         }
         if (!automatic)
             repeat = M600_check_state_and_repeat();
-        
     }
     while (repeat);
 
@@ -11107,7 +11106,7 @@ void M600_load_filament() {
 	}
 	KEEPALIVE_STATE(IN_HANDLER);
 
-    M600_load_filament_movements();
+	M600_load_filament_movements();
 
 	Sound_MakeCustom(50,1000,false);
 }
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3435,17 +3435,19 @@ static void gcode_M600(const bool automatic, const float x_position, const float
     plan_buffer_line_curposXYZE(FILAMENTCHANGE_XYFEED);
     st_synchronize();
 
-    // Unload filament
-    if (MMU2::mmu2.Enabled()) {
-        eject_slot = MMU2::mmu2.get_current_tool();
-        mmu_M600_unload_filament();
-    } else {
-        // Beep, manage nozzle heater and wait for user to start unload filament
-        M600_wait_for_user();
-        unload_filament(e_shift_late);
-    }
-    st_synchronize();          // finish moves
-    {
+    bool repeat = false;
+    do {
+        // Unload filament
+        if (MMU2::mmu2.Enabled()) {
+            eject_slot = MMU2::mmu2.get_current_tool();
+            mmu_M600_unload_filament();
+        } else {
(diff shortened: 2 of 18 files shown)


Panayiotis-git:
The translations were done using the Google translate.

Panayiotis-git:
Completed the non-MMU functionality testing, with enabled/disabled autoloading.

Panayiotis-git:
I've rebased the code to the latest MK3 branch.

This PR also addresses two minor issues:

1. If the answer to the question "Was filament unload successful?" is "No", then the printhead moves to the left in order to make accessible the filament port. During the head movement the screen is supposed to show the message MSG_PLEASE_WAIT. But instead, it returns to the status screen.

2. Also, after unloading the filament and selecting to start the next filament loading, only for the first time, instead of showing the message "loading filament" the status screen is shown and the progress bar at the bottom.

I tested the PR with plain MK3S. No side-effects are expected to the MMU functionality, because of the code structure.

Firmware/Marlin_main.cpp
```diff
@@ -3435,28 +3435,31 @@ static void gcode_M600(const bool automatic, const float x_position, const float
...
         {
             KEEPALIVE_STATE(PAUSED_FOR_USER);
             uint8_t choice =
                 lcd_show_fullscreen_message_yes_no_and_wait_P(_i("Was filament unload successful?"), false, LCD_LEFT_BUTTON_CHOICE); ////MSG_UNLOAD_SUCCESSFUL c=20 r=3
+            lcd_update_enable(false);
             if (choice == LCD_MIDDLE_BUTTON_CHOICE) {
                 lcd_clear();
                 lcd_puts_at_P(0, 2, _T(MSG_PLEASE_WAIT));
-                current_position[X_AXIS] -= 100;
+                current_position[X_AXIS] = 100;
```
3d-gussner, line 3462: Why did you change that?

Panayiotis-git: If I repeatedly choose the No answer at the question "Was filament unload successful?", the print head moves to the left again and again, till it crashes to the left edge.

Another solution could be to check if the print head is at the far right and the filament port is not accessible, only then move 100 units to the left.

But then why not always move the head to a fixed position to the left?

3d-gussner:
I have been testing your PR and there are few changes I would like to address.

It works well and I would like to approve it with the changes.

I can make a Pr later in the afternoon or here the diff file if you want to do it yourself. The diff is based on MK3 branch with hash https://github.com/prusa3d/Prusa-Firmware/commit/02a5228f5e3034f5b9dcf9b201eb73c1341b9ee0 so a `git apply` may not work correctly.
Updated the message lengths, po files and removed some white spaces 
- `MSG_CORRECTLY c=19`
- `MSG_UNLOAD_FILAMENT c=18` as it is also used in a SUBMENU which are limited to 18 chars
   
[MK3_M600_Issue_2415_few_changes.diff.txt](https://github.com/prusa3d/Prusa-Firmware/files/13523227/MK3_M600_Issue_2415_few_changes.diff.txt)


Thanks again for a very well prepared PR.

Panayiotis-git:
Hi @3d-gussner, thank you for your comments.
`git apply` worked perfectly!

3d-gussner (approved):
Tested on MK404
image

Firmware/Marlin.h
```diff
@@ -446,7 +446,7 @@ void gcode_M701(float fastLoadLength, uint8_t mmuSlotIndex);
 void M600_load_filament();
 void M600_load_filament_movements();
 void M600_wait_for_user();
-void M600_check_state();
+boolean M600_check_state_and_repeat();
```
DRracer: please use C++ standard `bool`

Firmware/Marlin_main.cpp
```diff
@@ -3435,28 +3435,31 @@ static void gcode_M600(const bool automatic, const float x_position, const float
...
-        eject_slot = MMU2::mmu2.get_current_tool();
-        mmu_M600_unload_filament();
-    } else {
-        // Beep, manage nozzle heater and wait for user to start unload filament
-        M600_wait_for_user();
-        unload_filament(e_shift_late);
-    }
-    st_synchronize();          // finish moves
-    {
+    boolean repeat = false;
```
DRracer: `bool`

Firmware/Marlin.h
Panayiotis-git: Strong Java influence...

PR #2427, opened 2020-01-21
Properly restore the saved E position in PP even if relative

leptun:
This PR fixes the PowerPanic problems in 3.9.0-beta that would result in spewing a large amount of filament at best and overwritten memory at worst.
The reason for the issue was that the E position was not restored if the E axis was relative. This is of course not ideal and for a long time it didn't cause any problems. The issue (most likely) started to manifest itself after @wavexx improved the pause/resume functions to compensate for the incomplete segment extrusion.
Video where you can see the issue (at 4:08): https://photos.app.goo.gl/3Wb6duTHssrvTXQA6
@DRracer


leptun:
Closed in favor of #2430

PR #2431, opened 2020-01-22
selfTest workflow optimalization

MRprusa3d:
PFW-1081


3d-gussner (approved):
Tested several times on MK3s with "old" IR sensor and had no issues with this change compared to previous version.

DRracer (approved):

PR #2433, opened 2020-01-22
🐛Even more fixes to lcd_print_stop

leptun:
I've been told that there are some issues sometimes with the lcd_print_stop where it would get stuck in the middle of the stop sequence inside a st_synchronize call. Weird issues would start to happen if you would play with the lcd, even to the point where the print randomly resumed. The probable cause is that the memory would get filled in certain edge cases and would result in this madness. We now unroll the stack and abort any ongoing functions before running the stop sequence. This should free up all the used ram and avoid the issue mentioned above.
Thanks @wavexx 
@DRracer 

PFW-1098


leptun:
Needs to be checked again

PR #2434, opened 2020-01-22
Do not sync before "PRUSA uvlo" during EEPROM recovery

wavexx:
While the extruder is unparking after a PowerPanic recovery you can stop
the print via the LCD. This stops the print during "G4", before "PRUSA
uvlo" is scheduled. Without this, the following powerup will try to
resume a print which was aborted.

Let the PP resume parse "PRUSA uvlo" immediately, clearing the uvlo flag
as soon as stop is available. There's no need to sync anyway.


wavexx:
Closing in favor of #2437 which handles it in a slightly different way

PR #2436, opened 2020-01-25
Do not clear axis known position when Z is set to silent

leptun:
Also removed unused (forgotten) `z_microsteps` variable as we no longer unshift the Z axis to the next full step when we switch to silent mode.
@DRracer 

PFW-1083


wavexx:
I would straight-out remove the comments entirely. The axis is not cleared when doing disable\_\*, so neither should disable_force\_\*.

leptun:
@wavexx Done

DRracer (approved):

PR #2430, opened 2020-01-21
Powerpanic E axis fixes

wavexx:
The E axis position wasn't recovered properly after a power panic which could result in wild overextrusion during recovery and/or a crash.

In this PR we reworking the recovery procedure to set the E axis position and unretractions correctly.

PFW-1079

PR #2437, opened 2020-01-26
Accurate PowerPanic E/Z recovery

wavexx:
There are multiple PowerPanic issues in the current code which lead to an inconsistent layer height just after a simple recovery (with the nozzle potentially dipping into the build plate), completely inaccurate Z height after a second panic, wild overextrusion and/or a crash during recovery due to a uninitialized E position.

In this PR we overhaul how the uvlo procedure is handled by correctly storing and recovering the physical Z position across a power failure. We also always move the Z axis towards a true fullstep during the procedure, so that no Z change is likely to be introduced while the power is off.

The E axis position and unretractions are now done correctly.

Because there's still frequently enough power left, we also re-enable the final extruder move in the first power failure to avoid the extruder getting stuck to a thin string of filament while cooling (which could lead to a detached print during recovery).

Z height is now kept consistent across any number of power failures.

This PR supersedes both #2434 and #2430 (both included here), as most of the uvlo handling code has been rewritten and/or modified.

PFW-1079


leptun:
Could we make a variable that defines the speed of all these movements? There’s pause, resume, powerPanic recover, M600, crash detection recovery and probably more. Wouldn’t it be better if we used a single define everywhere for speed?

wavexx:
I'd love to do this. It's full of magic speeds and hard-coded values...

DRracer:
@leptun let's think about it. We were talking about it with @wavexx a couple minutes ago - the current solution with hard coded variables scattered around the code is not the best practice of course.

DRracer:
@leptun @wavexx let's merge this one and check the power panic. We may improve the constants later in a separate PR

DRracer (approved):

Issue #1787, opened 2019-04-30
Power Panic input initialize


GWdd:
It appears that a short in the Power Panic switch or wiring will not be detected at power on, leaving the circuit to turn off heaters, but not be detected by the CPU; so everything else appears to function normally, including starting a print.

This condition would happen if the firmware reads and stores the state of the power panic pin, and accepts a low as normal. Subsequent tests will detect no change and not react to the failed condition.

wavexx:
I don't have the reference of the rambo at hand, but if that can be done I guess a better place to put this would be during self-test

wavexx:
Related: #2446

The FW *should* definitely check for this to avoid losing state during short repeated interruptions. We might need to do something on top of that to help the user detect that the wiring is wrong.
Skipping the test when the printed is uncalibrated and delay it to the first self-test is an option.

PR #2446, opened 2020-01-29
Refuse to start a print if power is lost before arming uvlo

wavexx:
If power has been lost during startup already a falling edge would be
skipped, causing the print to continue and lose its state without
being able to save again.

Check for a low line after arming the interrupt and simply wait
for reset.

PFW-1079

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -10480,6 +10480,14 @@ void serialecho_temperatures() {
 }
 
 #ifdef UVLO_SUPPORT
+void uvlo_drain_reset()
+{
+    // burn all that residual power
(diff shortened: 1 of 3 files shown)


leptun:
Hmmm... checking for the low pin might not be ideal. There was another issue where they asked for checking for shorted PowerPanic pin during selftest. This would make doing that impossible without some changes. What if we do that only if EEPROM_UVLO is different than 0? This way it would also allow me to use manual PowerPanic on my MK2.5SS that doesn’t have the hardware for this function.

leptun:
@wavexx Similar as you wrote in #1787, except we also check for EEPROM_UVLO.

DRracer (approved):

Issue #2280, opened 2019-10-18
Frequent error 'Some problem encountered, Z-leveling enforced' - Rel 3.8.0 - Mk2.5S


joris57:
When mesh calibration runs, I frequently get an 'Some problem encountered, Z-leveling enforced' error. Moving back to rel. 3.7.1 the problem goes away. Printer prints fine.

Maybe the limits for this check are set too tight in rel 3.8.0?

leptun:
A fix is on it's way. If everything goes according to plan, it should be fixed in 3.8.2.
Duplicate of #1651

joris57:
Thanks for the info, but it is worth noting that #1651 is about an issue in 3.7.0-RC1, where the issue I encounter does not manifest itself (in my set-up) when I migrate back to rel. 3.7.1

Kachidoki2807:
@joris57, do you mean that your issue doesn't disappear when probing with heatbed off?
The #1651 thread mixup some different issues, but it solves at least one. 
Could you be so kind to try out the firmware posted by @leptun [here](https://github.com/prusa3d/Prusa-Firmware/issues/1651#issuecomment-538714471)?
This one solved my 'Z-leveling enforced' issue, related to the heatbed PWM.

If it doesn't work better for you, then it will confirm that there are at least two issues giving this error.

joris57:
Sorry for he delay in replying. I tried the firmware referred to by @Kachidoki2807 and I am still experiencing the same issue.

madchiller:
Any updates on this... since updating to 3.8.1 this issue occurs frequently, and it is starting to get me..    Any RC of 3.8.2?  I’m down to test...  thanks guys!  Or has this made it to the 3.9 beta?

leptun:
@madchiller the latest MK3 branch has the mention linked above merged. It should greatly improve the number of rechecks.
You need to compile the fw by yourself. It is not present in any releases as of now. It will be present in the upcoming 3.9.0-RC1.

Deneteus:
@leptun Is this really only an issue because the Pinda V2 is not shielded? Mine started doing it after upgrading to the last beta (prusa3d_fw_MK3_3_9_0-beta_3175.hex) but it wasn't doing it before that. It's only failing near the back left side where the power comes into the heat bed.

![image](https://user-images.githubusercontent.com/6380390/74416641-0c86f780-4e0b-11ea-973d-5a02b4575f9a.png)

[...]

Issue #1651, opened 2019-03-19
[3.7.0-RC1] Repeated enforced Z calibration during 7x7 bed leveling


sgoll:
Since 3.7.0-RC1 I am using the new 7x7 mesh bed leveling (with 5 measurements per position). Now I have the problem that intermittently the bed leveling done before each print detects some problem, runs the enforced Z calibration and then re-runs the bed leveling.

Sometimes, this Z calibration is enforced a second or third time before finally the bed leveling succeeds.

This also happens when nothing at all was changed: the printer was switched on all the time (so no skip in the Z steppers occurred) and the steel sheet was not replaced (so it should still be aligned just like in the first run of the mesh bed leveling).

When is the Z calibration enforced? Is there some setting to tune down the sensitivity?

L0rdS474n:
Have this too

JohnnyPhantom27:
Me too!

JohnnyPhantom27:
Working on finding a cause. My buddy also has the MK3 printer with the new 3.7.0-RC1, his works fine. I decided to try the calibration wizard. It told me my motor belt was loose. I made it tighter and ran the test again, everything went great until the bed leveling. I did the 4 calibration points with paper, then put on the steel sheet as instructed, it got to the 2nd point in the test (front center of the bed) and failed. Same behavior as when I just tried printing. Got a failed: check nozzle and bed for debris. I cleaned the steel sheet, the bed, and the nozzle again. I also gave the screws holding the bed in place a snug turn with the hex key. re-trying...

JohnnyPhantom27:
Cleaned the bed, nozzle, leveling sensor, and gave the bed screws a snugging. I get past the wizard now, but during the first layer calibration, it fails during the bed leveling, front center each time. I hope my trial and errors helps.

PavelSindler:
Thank you all for your reports.

JohnnyPhantom27:
Figured maybe a factory reset might help, since I did have some manual leveling adjustments in the printer before going to 3.7.0-RC1, it did not :(

L0rdS474n:
> Figured maybe a factory reset might help, since I did have some manual leveling adjustments in the printer before going to 3.7.0-RC1, it did not :(

I also did a factory restore with no solution on my MK3S

geeksville:
[...]

PR #2447, opened 2020-01-29
Turn off bed while probing during MBL

leptun:
Many people have been experiencing retries in the probing during Mesh Bed Leveling after the 7x7 version was introduced. It has been determined that the bed heating was causing interference to the Pinda probe and it would cause false triggers. This PR disables the heating right when the probing happens so that no switch of state (on->off or off->on). This greatly reduces the retries count.
Overall, the probing accuracy is not increased, it's just the retries that are mitigated.
Issue describing the issue further:
Closing #2280 
Closing #1651 
@DRracer 

PFW-1019


Firmware/heatbed_pwm.cpp
```diff
@@ -137,12 +140,17 @@ ISR(TIMER0_OVF_vect)          // timer compare interrupt service routine
 	case States::ONE:             // state ONE - we'll either stay in ONE or change to FALL
 		OCR0B = 255;
 		slowCounter += slowInc;   // this does software timer_clk/256 or less
-		if( slowCounter < pwm ){
-			return;
+		if (!bedPWMDisabled){ //disable heating as soon as possible
```
DRracer: is it really correct to have the slowCounter outside of the "if" statement? It basically continues to tick the SW (aka slow) PWM. Wouldn't it be better to just keep the States::ONE when bed PWM is to be disabled?

Firmware/heatbed_pwm.cpp
```diff
@@ -137,12 +140,17 @@ ISR(TIMER0_OVF_vect)          // timer compare interrupt service routine
...
+			}
+			if( (soft_pwm_bed << 1) >= (255 - slowInc - 1) ){  //@@TODO simplify & explain
+				// if slowInc==2, soft_pwm == 251 will be the first to do short drops to zero. 252 will keep full heating
+				return;           // want full duty for the next ONE cycle again - so keep on heating and just wait for the next timer ovf
+			}
 		}
-		if( (soft_pwm_bed << 1) >= (255 - slowInc - 1) ){  //@@TODO simplify & explain
-			// if slowInc==2, soft_pwm == 251 will be the first to do short drops to zero. 252 will keep full heating
-			return;           // want full duty for the next ONE cycle again - so keep on heating and just wait for the next timer ovf
+		else if (pwm > 200){ //if duty cycle is high and BED PWM is disabled keep heater on. Prevents overcooling
```
DRracer: is this "else" really necessary (with the above comment in mind)?

Firmware/Marlin_main.cpp
```diff
@@ -4973,11 +4973,7 @@ if(eSoundMode!=e_SOUND_MODE_SILENT)
 					z_offset_u = eeprom_read_word((uint16_t*)(EEPROM_BED_CALIBRATION_Z_JITTER + 2 * (ix + iy * 3 - 1)));
 				}
 				z0 = mbl.z_values[0][0] + *reinterpret_cast<int16_t*>(&z_offset_u) * 0.01;
-				#ifdef SUPPORT_VERBOSITY
-				if (verbosity_level >= 1) {
-					printf_P(PSTR("Bed leveling, point: %d, calibration Z stored in eeprom: %d, calibration z: %f \n"), mesh_point, z_offset_u, z0);
-				}
-				#endif // SUPPORT_VERBOSITY
+				// printf_P(PSTR("Bed leveling, point: %d, calibration Z stored in eeprom: %d, calibration z: %f \n"), mesh_point, z_offset_u, z0);
```
DRracer: probably we can remove this change, it is not related to the features in this PR

leptun: Reverted the testing change

Firmware/heatbed_pwm.cpp
```diff
@@ -45,6 +45,12 @@
 // If there are any change requirements in the future, the signal must be checked with an osciloscope again,
 // ad-hoc changes may completely screw things up!
 
+// 2020-01-29 update: we are introducing a new option to the automaton that will allow us to force the output state
+// to either full ON or OFF. This is so that interference during the MBL probing is minimal.
+// To accomplish this goal we use bedPWMDisabled. It is only supposed to be used for brief periods of time as to
+// not make the bed temperature too unstable. Also, careful consideration should be used when using this
+// option as leaving this enabled will also keep the bed output in the state it stopped in.
+
```
DRracer, line 53: excellent explanation

DRracer (approved):

GilesBathgate:
Hi @leptun I wanted to try this out on MK2 firmware. I know it was to solve an issue with @DRracer's fast PWM (i.e `heatbed_pwm.cpp`) which only exists in MK3.

However, in my tests with a bed temp heated to only 36°C, I've noticed a considerable difference in probe values if the bed is ON vs OFF and wanted to test this patch.

I've attempted to backport your change here: https://github.com/GilesBathgate/Prusa-Firmware/commit/c70d00aef091025f85508f92e3ff380e316470b5

I wondered if you can take a look and see if I've understood this correctly.

leptun:
@GilesBathgate Hello. Yes, this is what I had in mind when I implemented it here. One thing to note, though: With the heatbed_pwm.cpp the idea was to avoid the fast PWM pulses when switching the state. The way you implemented it follows exactly what we did, but it might not be that useful overall. You can try and see if you get improfements with it as it is now, but you might want to try something like this:
Put the `if (!bedPWMDisabled) {...}` so that it only prevents a WRITE(1) to the bed. This way it will always stay off during the probing. If you use it as it is now the bed will stay either on or off (no switching)

GilesBathgate:
@leptun So leave the state changing code in place and just disable the writes? 

i.e `if (!bedPWMDisabled) WRITE(HEATER_BED_PIN, 1);`
and `if (!bedPWMDisabled) WRITE(HEATER_BED_PIN, 0);`

GilesBathgate:
@leptun So you suggest only disabling the write that turns the bed on i.e:

`if (!bedPWMDisabled) WRITE(HEATER_BED_PIN, 1);`

but leave all the other code to execute.

leptun:
Yes, that should work. So that the bed will disable itself once the PWM cycle finishes, but it won't get enabled again. Not sure what happens when the bed is at full power, but otherwise it should be fine

GilesBathgate:
>Not sure what happens when the bed is at full power.

It gets hot! :joy:

GilesBathgate:
@leptun So to test whether this was working I added the following gcode commands:

```c++
    case 88: //just for test
        bedPWMDisabled=1;
        break;
    case 89:
        bedPWMDisabled=0;
        break;
```

I then use the application EMF Detector for my mobile phone to do basic detection of the magnetic flux caused by the PWM. I just place my phone directly on the heated bed.

image

Then I turn the heat bed up to 40°C

image

Unfortunately, I am not able to get the EMF to go back down by issuing my test gcodes.
If I set the temp back to 0°C (I.e OFF) the EMF goes down immediately.

GilesBathgate:
@leptun I figured it out, there was an additional place where `WRITE(HEATED_BED_PIN,1)` was being called: See https://github.com/GilesBathgate/Prusa-Firmware/commit/ad98d7c58fbc72fe824e88894c597fc4001f33db#diff-719af9fc6a59d745b8e9c0164f8dcebfR1532

Works as expected now but I will share some results from actual MBL probing tests next.

PR #2449, opened 2020-01-30
Fix nozzle size menu

koenkooi:
The switch statement is currently a bit confused, selecting 0.25mm will save 0.40, but selecting 0.40 will save 0.60 and selecting 0.60 will save 0.25.

Signed-off-by: Koen Kooi <koen@dominion.thruhere.net>


koenkooi:
Nevermind, I'm an idiot :) This is how it selects the next entry, not where it saves the size to EEPROM.

leptun:
No problem 😉

PR #2453, opened 2020-02-02
save stats with M84

bladel:
Save stats after succesfull print / M84 command for printing over serial port


3d-gussner:
Closing it as we we are working on similar solution for host systems.

PR #2454, opened 2020-02-02
Fix compiler warnings

smurfix:
This patch removes all compiler warnings.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -4947,7 +4947,6 @@ if(eSoundMode!=e_SOUND_MODE_SILENT)
 		}
 		#endif // SUPPORT_VERBOSITY
 		int l_feedmultiply = setup_for_endstop_move(false); //save feedrate and feedmultiply, sets feedmultiply to 100
-		const char *kill_message = NULL;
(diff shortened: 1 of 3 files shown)


leptun:
@smurfix Thanks for submitting the PR. We are planning to fix these compiler warnings.
For example, the SD related warnings will be fixed in a future version by #2405 
ONE_TO_FALL: That is ok
kill_message: I think that was left there intentionally as a reminder to improve/fix it later
`int chars = lcd_printf_P(_N(" F0  "));`: same as above. The code is currently disabled, but it should be fixed later.
lcd_check_gcode_set: Planned to be enabled soon
lcd_belttest: Ok. The implementation of the screen might change a bit as it is not very efficient
bAction: I need to ask the author of the code what was the reason for it, but otherwise looks ok

mkbel:
We applied some of your fixes:
https://github.com/prusa3d/Prusa-Firmware/pull/2725
https://github.com/prusa3d/Prusa-Firmware/pull/2727
https://github.com/prusa3d/Prusa-Firmware/pull/2728

Remaining compiler warnings will be fixed different way. Thank you for your contribution.

3d-gussner:
@smurfix Thanks for your PR, we have fixed multiple compiler warnings. Closing

PR #2092, opened 2019-08-10
PFW-1062: Mk3 Toggle and Translations

leptun:
Hi everyone,

In this PR I've implemented a way to make the Toggle menus much more efficient.

As it is now in the MK3 branch, all toggle menu options have a translated counterpart:
```
#MSG_FSENSOR_ON
"Fil. sensor  [on]"
"Fil. senzor [zap]"

#MSG_FSENSOR_NA
"Fil. sensor [N/A]"
"Fil. senzor [N/A]"

#MSG_FSENSOR_OFF
"Fil. sensor [off]"
"Fil. senzor [vyp]"
```
This is very wasteful and with this PR I intend to turn that into:
```
#MSG_FSENSOR
"Fil. sensor"
"Fil. senzor"
```

and common for all toggles:
```
#
"Off"
"Vyp"

#
"On"
"Zap"

#
"N/A"
"\x00"
```

@DRracer @mkbel In theory this will result in significantly reduced flash usage. I've made the necessary changes for this feature to work and changed the "Fans check" message to use this stile of toggle. If you think this is worth developing further, please let me know and I'll start changing the rest of the messages to this. I've also added a "toggle" character at the end. I think it looks nice this way. It can be changed easily to blank like before. One thing I noticed is that some translated toggles don't fit or barely fit. Wouldn't it be better if the toggle section (the one between []) was moved 1 or 2 characters to the right? This way everything fits better, especially with translations.
image

@michalxfanta I've noticed that translations are somewhat neglected, especially french and italian. What do you think of sending an email out like you did for beta testers where you ask for help on existing translations and maybe also new languages?

PR #2455, opened 2020-02-02
Add more nozzles

smurfix:
There are more commonly used nozzle diameters than 0.25/0.4/0.6mm.

This patch adds some of them. A better fix would probably be to freely change the current diameter in 0.05mm increments, but this way works for me.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -5465,7 +5465,19 @@ uint16_t nDiameter;
 
 switch(oNozzleDiameter)
      {
+     case ClNozzleDiameter::_Diameter_150:
+          oNozzleDiameter=ClNozzleDiameter::_Diameter_200;
+          nDiameter=200;
+          break;
+     case ClNozzleDiameter::_Diameter_200:
+          oNozzleDiameter=ClNozzleDiameter::_Diameter_250;
+          nDiameter=250;
+          break;
      case ClNozzleDiameter::_Diameter_250:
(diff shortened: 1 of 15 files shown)


leptun:
Imagine having to do this prior to #2092. It would have been a nightmare ;)

leptun:
The po files are not important to the build process and are only used by external translators. They are automatically generated using one of the scripts in the repo.

3d-gussner:
@smurfix Please review your PR as there are conflicts with latest firmware release.
I really like that you also provided the lang/po/ files.🥇
After you fixed these conflicts we need to update the `lang_en.txt` and `lang_en_xx.txt` files.
I can do it if you like or read the [translations.md](https://github.com/prusa3d/Prusa-Firmware/blob/MK3/lang/translations.md) how to do that.
[Update] please remove the `po`files from your PR.

3d-gussner (commented):
Please remove the `po` files as you haven't added a new translations.

JKinch03:
It would be nice to get ones like 0.8mm and 1mm, and maybe 0.2mm too.

smurfix:
Well, IMHO we should use the wheel to do 0.05mm increments from 0.15 to 1.25. This long list of possible nozzle diameters is not a long term solution.

JKinch03:
True. The knob selection would make it better and easier than a long list

3d-gussner:
@smurfix I will close this PR as we will not implement it this way.

PR #2459, opened 2020-02-03
Improved lcd_menu_statistics() implementation

leptun:
@DRracer 
Fix formatting for both menu states.
Disable timeout to status while in this menu.
Do not clean the screen on every refresh. It no longer blinks.
No longer use deprecated method of generating live-updating menus.

PFW-418

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -2830,8 +2830,8 @@ void lcd_menu_statistics()
 			"%S:\n"
 			"%10ldh %02hhdm %02hhds"
 		    ),
-            _i("Filament used"), _met,  ////c=18 r=1
-            _i("Print time"), _h, _m, _s);  ////c=18 r=1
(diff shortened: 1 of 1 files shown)


Firmware/ultralcd.cpp
```diff
@@ -2840,29 +2841,20 @@ void lcd_menu_statistics()
...
-			manage_inactivity(true);
-			_delay(100);
-		}
-		KEEPALIVE_STATE(NOT_BUSY);
-		lcd_quick_feedback();
-		menu_back();
+			"%10ldd %02hhdh %02hhdm"
+            ),
+            _i("Total filament"), _filament_m,
+            _i("Total print time"), _days, _hours, _minutes);
```
DRracer: we can also add translation hints (c=18 r=1) to both texts here

leptun: Don't we already have these in the menu "render" above the function?

DRracer (approved):

PR #2473, opened 2020-02-07
#2472 #2356 #2484 belt test improvements

### LINKED ISSUE #968: Belt tension status check - addition to menu please
Given  there is code written to test it ...  could it be added to as a menu item, (eg in calibration menu) as currently I believe the only way it can be measured is to run the complete “new machine” wizard which takes ages, if you’re just working on getting the belt tension sorted. This is a common thing new people with their mk3s (and presumably  other models) do once the machine is working as part of their initial “dialing stuff in” activities. To have to wait for what feels like 15/20 minutes for the machine to run the whole wizard is very frustrating whilst iterating just on belt tension - it would be great to be able to test and display the belt tension(s) alone.

### LINKED ISSUE #2356: [3.9.x] Belt test feature is broken
This feature does not work on my printer. If I try to use it, Z moves up ~20mm, and I get a "Selftest error" on screen instructing me to check X axis length. X never attempts to move, and the updated status value is 0. Regular self-test functions as expected.

Mk3S/MMU2 with 12/3 github pull & 1.0.6 firmware. 

Video: 

https://vimeo.com/377413263

### LINKED ISSUE #2472: [ENHANCEMENT] Warn that stealth mode can't be used for belt calibration.
Enter what type of printer or upgrade the enhancement applies to.
**Printer type** - MK3s

I was trying to "calibrate" my belts via the belt status option inside calibration menu and every time, a printer told me that there was an error and that I should "check axis length x". But all I needed to do was to disable stealth mode.

P.S.: The calibration was kinda useless anyway, as even after some serious tightening, the belt status went up instead of down.
  
Maybe add a warning that it can't be calibrated in stealth mode or automatically disable the stealth mode before the "calibration".

### LINKED ISSUE #2484: [ENHANCEMENT] Better "Belt test" implementation
As of now, the function always raises the Z a certain amount. `raise_z_above()` should be used instead.
Also the LCD rendering could be improved.

Note for later.

### LINKED ISSUE #2547: [BUG] After Belt test the X,Y steppers are locked.
**Printer type** - MK3S
**Printer firmware version**-  3.9.1-RC1

**Describe the bug**
When printer belt test routine is finished, the X, Y steppers are locked.
  
**To Reproduce**
Run Belt test and when finished try to move bed or extruder by hand.
  
**Expected behavior**
After Belt test the X and Y steppers should be released immediately,.
  
**G-code**
N/A

**Video**
N/A

vintagepc:
Forces high power mode for belt test.

- Fixes #2472, Fixes #2356, resolves #2484 

Edit: Also fixes #968 , closes #2547 

PFW-1104

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -7446,30 +7446,28 @@ void lcd_belttest()
     uint16_t   X = eeprom_read_word((uint16_t*)(EEPROM_BELTSTATUS_X));
     uint16_t   Y = eeprom_read_word((uint16_t*)(EEPROM_BELTSTATUS_Y));
 	lcd_puts_at_P(0,0,_i("Checking X axis  ")); // share message with selftest
-	lcd_set_cursor(0,1), lcd_printf_P(PSTR("X: %d "),X);
+	lcd_set_cursor(0,1), lcd_printf_P(PSTR("X: %d -> ..."),X);
     KEEPALIVE_STATE(IN_HANDLER);
     
 	// N.B: it doesn't make sense to handle !lcd_selfcheck...() because selftest_sg throws its own error screen
 	// that clobbers ours, with more info than we could provide. So on fail we just fall through to take us back to status.
     if (lcd_selfcheck_axis_sg(X_AXIS)){
 		X = eeprom_read_word((uint16_t*)(EEPROM_BELTSTATUS_X));
-		lcd_printf_P(PSTR("-> %d"),X); // Show new X value next to old one.
+		lcd_set_cursor(9,1), lcd_printf_P(PSTR("%d"),X); // Show new X value next to old one.
         lcd_puts_at_P(0,2,_i("Checking Y axis  "));
-		lcd_set_cursor(0,3), lcd_printf_P(PSTR("Y: %d "),Y);
+		lcd_set_cursor(0,3), lcd_printf_P(PSTR("Y: %d -> ..."),Y);
 		if (lcd_selfcheck_axis_sg(Y_AXIS))
 		{
 			Y = eeprom_read_word((uint16_t*)(EEPROM_BELTSTATUS_Y));
-			lcd_printf_P(PSTR("-> %d"),Y);
-			lcd_set_custom_characters_nextpage();
+			lcd_set_cursor(9,3),lcd_printf_P(PSTR("%d"),Y);
 			lcd_set_cursor(19, 3);
-			lcd_print(char(2));
+			lcd_print(LCD_STR_UPLEVEL);
 			lcd_wait_for_click_delay(10);
 		}
     }
 	
 	FORCE_HIGH_POWER_END;
     KEEPALIVE_STATE(NOT_BUSY);
-	lcd_set_custom_characters(); // restore status screen chars.
 }
 #endif //TMC2130
 
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -7832,7 +7832,7 @@ bool lcd_selftest()
 
 static void reset_crash_det(unsigned char axis) {
 	current_position[axis] += 10;
-	plan_buffer_line_curposXYZE(manual_feedrate[0] / 60, active_extruder);
+	plan_buffer_line_curposXYZE(homing_feedrate[axis] / 60, active_extruder);
 	st_synchronize();
 	if (eeprom_read_byte((uint8_t*)EEPROM_CRASH_DET)) tmc2130_sg_stop_on_crash = true;
 }
@@ -7853,17 +7853,15 @@ static bool lcd_selfcheck_axis_sg(unsigned char axis) {
 	tmc2130_home_exit();
 	enable_endstops(true);
 
-	if (axis == X_AXIS) { //there is collision between cables and PSU cover in X axis if Z coordinate is too low
-		raise_z_above(17,true);
-		tmc2130_home_enter(Z_AXIS_MASK);
-		st_synchronize();
-		tmc2130_home_exit();
-	}
+
+	raise_z_above(MESH_HOME_Z_SEARCH);
+	st_synchronize();
+	tmc2130_home_enter(1 << axis);
 
 // first axis length measurement begin	
 	
 	current_position[axis] -= (axis_length + margin);
-	plan_buffer_line_curposXYZE(manual_feedrate[0] / 60, active_extruder);
+	plan_buffer_line_curposXYZE(homing_feedrate[axis] / 60, active_extruder);
 
 	
 	st_synchronize();
@@ -7873,11 +7871,11 @@ static bool lcd_selfcheck_axis_sg(unsigned char axis) {
 	current_position_init = st_get_position_mm(axis);
 
 	current_position[axis] += 2 * margin;
-	plan_buffer_line_curposXYZE(manual_feedrate[0] / 60, active_extruder);
+	plan_buffer_line_curposXYZE(homing_feedrate[axis]  / 60, active_extruder);
 	st_synchronize();
 
 	current_position[axis] += axis_length;
-	plan_buffer_line_curposXYZE(manual_feedrate[0] / 60, active_extruder);
+	plan_buffer_line_curposXYZE(homing_feedrate[axis]  / 60, active_extruder);
 
 	st_synchronize();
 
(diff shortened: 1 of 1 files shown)


leptun:
@vintagepc You must also disable TMC high power when returning with an error. Anyway, the function is due to some restructuring and improvements, so I’d wait a bit with this PR. (Probably 3.9.0 release).

vintagepc:
wow, I really derped there...
Anyway, tentative reorg pushed up and removed an unused variable. Will confirm later today it works when I finish actually printing.

vintagepc:
:+1: Just compiled and ran updated version. Still functions as expected regardless of stealth or normal mode.

vintagepc:
fixes #2484 
- Display is now built incrementally as the test progresses
- Display is more useful to see old and new numbers
- Reused existing strings for translation
- Removed unnecessary defines and if-elses on errors
- Fixed continuous raising of Z each time the test is run.

Also, thanks to you guys for putting up with my attempts to learn the innards :+1: 

(Ignore my test numbers, I have 0.9 degree steppers which makes them a lot lower than normal)
IMG_0006

leptun:
Looking good. Does it show in any way on the screen when a value is not yet calculated?
I might simplify the checkmark symbol in the future. Having to swap custom character banks can be avoided pretty easily now, especially for this screen.

vintagepc:
> Looking good. Does it show in any way on the screen when a value is not yet calculated?

 Not explicitly, but only one value is shown while it is in progress. e,g:
"Checking X axis"
"[old value]"

The " -> [new value]" appears after the measurement is complete.

I'd originally considered doing "Old: [value] New:[value]" to be a bit clearer, but realized pretty quick that different languages could lead to unpleasant sizing problems.

leptun:
Hmmm. I think there still are ways to improve the formatting. We’ll look into that aspect after the 3.9.0-RC1 is released.
Can you please split the PR into two: render changes and everything else? The others will probably get merged faster, but the rendering will take a bit longer since translations also need to be created for these.

vintagepc:
I deliberately re-used strings from the selftest so we should already have translations on file for those. 
Do I need to update any of the translation files in this situation or does the setup automatically detect identical phrases?

vintagepc:
Additionally, it would be pretty easy to do something like 
`215 -> ...` or `215 -> ???`
while the new value is being measured. Neither adds any translation overhead or language dependence.

vintagepc:
Tweaked the layout a little more to make it clear the measurement is in progress, also ditched the custom character switch

IMG_0016

IMG_0017

vintagepc:
I see we're getting really close to release, judging by the number of remaining 3.9 PRs. Should I quickly address #2547 in this PR, or are we done with changes and in final QA now?

leptun:
The 3.9 final won’t be released soon unfortunately. Because of that LA1.5 bug, we have to release another RC.

I’ve been thinking of another issue with belttest that we could fix. Currently, the motor current is not reduced like during homing. This makes the axis harder to trigger. I was thinking we should also “home_enter()” each axis so that the currents would be lower. Also, it would be better if we also used the homing speed during the belttest, not the manual movement speed.

I’ve already started making these changes myself, but never got to finishing them. I’ll post a link to my changes once I push them later this day.

vintagepc:
That's fair - I know we do need to address the LA bug.
Feel free to file a PR against my branch, I'll merge it in and then add a trailing M84 if you haven't already.

leptun:
@vintagepc https://github.com/vintagepc/Prusa-Firmware/pull/5/
The PR is created. Please look carefully through the code as there might be errors.
I'm pretty sure that the steppers are disabled at the end. Maybe they are left on after an error?
I also (somehow) fixed the wrong behavior where the axis would stay at MIN after measuring the length. This would result in a thump as the axis disengaged at the end. Now they move ~10mm (don't remember) away from the endstop after measuring the axis.
We also now use homing speed instead of manual speed and we use the lower driver currents just like in G28

vintagepc:
Thanks, I'll get to this after work tomorrow and we can get it buttoned up.

Firmware/ultralcd.cpp
```diff
@@ -7436,54 +7436,38 @@ static void lcd_belttest_v()
...
-    if (!_result){
-        lcd_belttest_print(_i("Error"), X, Y);
-        lcd_clear();
-        return;
+    
+	// N.B: it doesn't make sense to handle !lcd_selfcheck...() because selftest_sg throws its own error screen
+	// that clobbers ours, with more info than we could provide. So on fail we just fall through to take us back to status.
+    if (lcd_selfcheck_axis_sg(X_AXIS)){
+		X = eeprom_read_word((uint16_t*)(EEPROM_BELTSTATUS_X));
+		lcd_set_cursor(9,1), lcd_printf_P(PSTR("%d"),X); // Show new X value next to old one.
```
DRracer: %u for unsigned ints ... hopefully the stats won't get negative, so no harm here...

Firmware/ultralcd.cpp
```diff
@@ -7436,54 +7436,38 @@ static void lcd_belttest_v()
...
+    lcd_clear();
+	// Belttest requires high power mode. Enable it.
+	FORCE_HIGH_POWER_START;
+    
     uint16_t   X = eeprom_read_word((uint16_t*)(EEPROM_BELTSTATUS_X));
     uint16_t   Y = eeprom_read_word((uint16_t*)(EEPROM_BELTSTATUS_Y));
-    lcd_belttest_print(_i("Checking X..."), X, Y);
-
-    _delay(2000);
+	lcd_puts_at_P(0,0,_i("Checking X axis  ")); // share message with selftest
```
DRracer: this can be narrowed down to a simple lcd_printf without the explicit cursor position
... lcd_clear should set the cursor to [0,0]

vintagepc:
Thanks for the review. Will address in my cleanup tonight.

leptun:
@vintagepc Small issue: The ... that is displayed on the new value sticks around even after the new value is printed and results in a '.' after the value is displayed.

leptun:
Correction: The text is printed a bit too much to the left. One character to the right for the second numbers and the formatting will be perfect.

vintagepc:
Looks like I forgot to push a fix for that - there was a bug in the initial version where the alignment was off, but as you can see in the pictures a few posts up it overlays perfectly.

leptun:
Also, the steppers are always disabled at the end, so that issue can be closed or further clarification is needed.

vintagepc:
👍 I'll test it myself and if I also can't repro that I'll mark this PR as closing that too. My memory is fuzzy  but I think it may have been possible in the initial implementation in the RC.

Firmware/ultralcd.cpp
vintagepc: Fixed!

Firmware/ultralcd.cpp
vintagepc: Fixed!

vintagepc:
Pulled in @leptun's changes. For some reason the PR wants to merge 25 changed files, so I did a manual compare and edited the relevant functions. 

Also fixed the printf() calls.

Motors are disabled afterwards, even in error cases.

I suspect it is to be expected that the lower feed rates/currents will result in different numbers.My belt test values drop by about 60 after these changes. 

That may have implications for the online manual, which has suggestions for expected ranges that may need to be updated...?

leptun:
I really don’t know about the online manual. Maybe @drracer can help here.
All those changes are probably the fact that I merged the local MK3 branch instead of merging remote MK3.

It’s actually moving faster, not slower. I was thinking that since we use homing motor currents, why not also use the homing speed?

vintagepc:
I stand corrected on the speed, I just assumed it was lower due to lower currents. :-)

If we are changing the manual I'd rather suggest we move away from the notion these numbers are 1:1 belt tension; there are many factors that play into the motor load (bearings, lubrication, alignments, weight, what have you). Ideally, all other things being equal, they likely could be a proportional measure of belt status but there are just too many variables, especially with kit assembly, for that to be definitive.

My recommendation would be to refer to them as axis loading measurements and the acceptable range as such - and possible causes for if you are outside the range, depending on too high/low. There is a lot of confusion in the forums around these numbers as it stands and now seems like a good time to clear that up. People get too caught up in the "tension" aspect and the numbers and don't investigate the bigger picture of issues with their motion systems.

Firmware/ultralcd.cpp
```diff
@@ -7870,17 +7853,15 @@ static bool lcd_selfcheck_axis_sg(unsigned char axis) {
...
+
+	raise_z_above(MESH_HOME_Z_SEARCH);
+	st_synchronize();
+	tmc2130_home_enter(1 << axis);
 
 // first axis length measurement begin	
 	
 	current_position[axis] -= (axis_length + margin);
-	plan_buffer_line_curposXYZE(manual_feedrate[0] / 60, active_extruder);
+	plan_buffer_line_curposXYZE(homing_feedrate[axis] / 60, active_extruder);
```
DRracer: @vintagepc  homing_feedrate vs. manual_feedrate = 3000 vs. 2700 - since this code is being used in normal self test, which has been around for quite a while (and since the selftest is probably failing more with the 3000 feedrate), please revert the homing_feedrate back to manual_feedrate in this whole PR. Otherwise it looks reasonably well

vintagepc: Done!

vintagepc:
requested changes have been made. :+1:

DRracer (approved):

DRracer:
@vintagepc thank you, approved, merging

PR #2474, opened 2020-02-08
Add X-Y home to tune menu if print is paused

### LINKED ISSUE #2161: Feature Request: Rehome X/Y in tune menu when paused

vintagepc:
Fixes #2161. 

Note: This is intended to be merged after #2390 since it depends on the Tune menu being available while the print is paused. 

There are a number of reasons one might need this feature, e.g. accidental slip while dealing with filament change/jammed filament, catching a layer shift quickly enough, etc. 

I tested it on my printer (apologies in advance, I'm sure the forced X shift will make some people cringe...)

https://vimeo.com/390095133

and it homed and resumed flawlessly with only a tiny zit to show for it, about the same as you'd get anyway. 

I did also verify the menu item is not present if the print is not paused... inserting a G28 mid print stream seems like a bad idea.... :-)

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -5441,7 +5441,7 @@ static void lcd_tune_menu()
     if (!farm_mode)
         MENU_ITEM_FUNCTION_P(_T(MSG_FILAMENTCHANGE), lcd_colorprint_change);
 #endif
-    if (isPrintPaused) // Don't allow rehome if actively printing. Maaaaybe it could work to insert on the fly, seems too risky.
+    if (isPrintPaused) {// Don't allow rehome if actively printing. Maaaaybe it could work to insert on the fly, seems too risky.
         MENU_ITEM_GCODE_P(_T(MSG_AUTO_HOME), PSTR("G28 X Y"));
     }
 #ifdef FILAMENT_SENSOR
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -5441,7 +5441,9 @@ static void lcd_tune_menu()
     if (!farm_mode)
         MENU_ITEM_FUNCTION_P(_T(MSG_FILAMENTCHANGE), lcd_colorprint_change);
 #endif
-
+    if (isPrintPaused) // Don't allow rehome if actively printing. Maaaaybe it could work to insert on the fly, seems too risky.
+        MENU_ITEM_GCODE_P(_T(MSG_AUTO_HOME), PSTR("G28 X Y"));
+    }
 #ifdef FILAMENT_SENSOR
     MENU_ITEM_SUBMENU_P(_T(MSG_FSENSOR), lcd_fsensor_settings_menu);
 #endif //FILAMENT_SENSOR
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -241,6 +241,8 @@ static void lcd_sheet_menu();
 static void menu_action_sdfile(const char* filename);
 static void menu_action_sddirectory(const char* filename);
 
+static void lcd_rehome_xy();
+
 #define ENCODER_FEEDRATE_DEADZONE 10
 
 #define STATE_NA 255
@@ -5335,6 +5337,15 @@ void stepper_timer_overflow() {
 }
 #endif /* DEBUG_STEPPER_TIMER_MISSED */
 
+static void lcd_rehome_xy() {
+	// Do home directly, G28 X Y resets MBL, which could be bad.
+	homeaxis(X_AXIS);
+	homeaxis(Y_AXIS);
+	lcd_setstatuspgm(_T(MSG_AUTO_HOME));
+	lcd_return_to_status();
+	lcd_draw_update = 3;
+}
+
 
 static void lcd_colorprint_change() {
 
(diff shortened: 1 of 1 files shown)


wavexx:
> I tested it on my printer (apologies in advance, I'm sure the forced X shift will make some people cringe...)

This is no different than the extruder re-homing after a crash (on the print), which IMHO a "regular" scenario. Of course, it's better the printer doesn't require rehoming since it's not necessarily as precise, but the option is very nice to have.

> https://vimeo.com/390095133

Uuh, interesting extruder/shroud combo! What is this?

> and it homed and resumed flawlessly with only a tiny zit to show for it, about the same as you'd get anyway.

I suppose homeaxis(X/Y) does account for the MBL correctly right after rehoming?

My plan was to put some ~0.5mm plastic under a corner of the sheet put a print right on top of it.
You really need something bigger to notice.

vintagepc:
> This is no different than the extruder re-homing after a crash (on the print), which IMHO a "regular" scenario. Of course, it's better the printer doesn't require rehoming since it's not necessarily as precise, but the option is very nice to have.

Yep, not needed 99% of the time but pretty damn handy to have if there's an incident during a filament change/jam or so, not necessarily a print collision. 

> Uuh, interesting extruder/shroud combo! What is this?

It's a skelestruder (https://www.thingiverse.com/thing:2845416) with a pancake stepper. ~200g lighter than stock, and geared. It's a standard fan shroud but I dangled an LED off the front of it because I was tired of not being able to see the nozzle when ambient light is dimmer.

> I suppose homeaxis(X/Y) does account for the MBL correctly right after rehoming?

Yes, I read through the code and it doesn't throw away the MBL like "G28 X Y" would.

wavexx:
> > I suppose homeaxis(X/Y) does account for the MBL correctly right after rehoming?
> 
> Yes, I read through the code and it doesn't throw away the MBL like "G28 X Y" would.

It shouldn't throw it away of course, but it should definitely recalculate the logical MBL-adjusted Z at the new position, since a normal move to the homing location *would* shift Z, so moving back need to undo that shift.

vintagepc:
> > > I suppose homeaxis(X/Y) does account for the MBL correctly right after rehoming?
> > 
> > 
> > Yes, I read through the code and it doesn't throw away the MBL like "G28 X Y" would.
> 
> It shouldn't throw it away of course, but it should definitely recalculate the logical MBL-adjusted Z at the new position, since a normal move to the homing location _would_ shift Z, so moving back need to undo that shift.

Ah, gotcha. Though I think that's implicit because the Z raising/resume is done by the pause/unpause as opposed to this option.

vintagepc:
I think I'm going to revert the change to explicit Homeaxis() and restore G28 X Y. 
I re-read the G28 gcode more closely and it looks like it does handle the MBL correctly, disabling it for the home and restoring it afterwards when only X/Y are specified. Looks like there may have been some confusion surrounding the previous implementation that used this and the risks in regards to MBL.

wavexx:
I fixed several issues that caused problems with powerpanic, but I'm not sure I got them all since I didn't read all the cases supported by G28.

But in essence, homing needs to account for the MBL shift between last known position at the final position after homing. A simple Z travel might not move the the right spot otherwise, since this delta needs to be subtracted correctly. Imagine that the current Z offset is 1mm and moving to 0,0 would move the extruder by -0.5. You home, the extruder stays at 1mm. But when you move back to the original spot, the extruder will be moved +0.5.

Another complication that I've seen in G28 is that instead of working with relative distances, sometimes the Z position is recalculated from the absolute current_position. This might not give the same results depending on the order of the instructions.

vintagepc:
Thanks. I'll review this again. Sounds like plan C is better, just do it explicitly in this function rather than rely on what may or may not be happening during homeaxis or G28.

Firmware/ultralcd.h
```diff
@@ -48,6 +48,7 @@ void prusa_statistics(int _message, uint8_t _col_nr = 0);
 void lcd_confirm_print();
 unsigned char lcd_choose_color();
 void lcd_load_filament_color_check();
+void lcd_rehome_xy();
```
3d-gussner: Please fix the build error after a rebase

```
//<user>/Prusa-Firmware/MK3/Firmware/ultralcd.cpp: In function 'void lcd_rehome_xy()':
//<user>/Prusa-Firmware/MK3/Firmware/ultralcd.cpp:5340:13: error: 'void lcd_rehome_xy()' was declared 'extern' and later 'static' [-fpermissive]
 static void lcd_rehome_xy() {
             ^~~~~~~~~~~~~
In file included from //<user>/Prusa-Firmware/MK3/Firmware/ultralcd.cpp:7:0:
//<user>/Prusa-Firmware/MK3/Firmware/ultralcd.h:64:6: note: previous declaration of 'void lcd_rehome_xy()'
 void lcd_rehome_xy();
      ^~~~~~~~~~~~~
```

3d-gussner (changes requested):
After rebase and fixing I will approve this PR

3d-gussner (approved):
Tested on MK404

PR #2494, opened 2020-02-14
Use combined creation/modification file time stamps for sorting

DRracer:
There are scenarios when simple modification time is not enough (on MS Windows) - for example - extract an old g-code from an archive onto the SD card.
In such case the file's creation time is the current time (which is correct), but the modification time stays the same - i.e. old. 
Therefore let's pick the most recent timestamp from both creation and modification timestamps.
PFW-1093

### DIFF
--- a/Firmware/cardreader.cpp
+++ b/Firmware/cardreader.cpp
@@ -136,8 +136,17 @@ void CardReader::lsDive(const char *prepend, SdFile parent, const char * const m
 						SERIAL_ECHOPGM("Access date: ");
 						MYSERIAL.println(p.lastAccessDate);
 						SERIAL_ECHOLNPGM("");*/
-						modificationDate = p.lastWriteDate;
-						modificationTime = p.lastWriteTime;
+						crmodDate = p.lastWriteDate;
+						crmodTime = p.lastWriteTime;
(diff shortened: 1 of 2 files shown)


leptun (approved):
Tested: Ok

Firmware/cardreader.cpp
```diff
@@ -136,8 +136,17 @@ void CardReader::lsDive(const char *prepend, SdFile parent, const char * const m
...
 						SERIAL_ECHOLNPGM("");*/
-						modificationDate = p.lastWriteDate;
-						modificationTime = p.lastWriteTime;
+						crmodDate = p.lastWriteDate;
+						crmodTime = p.lastWriteTime;
+						// There are scenarios when simple modification time is not enough (on MS Windows)
+						// For example - extract an old g-code from an archive onto the SD card.
+						// In such case the creation time is current time (which is correct), but the modification time
+						// stays the same - i.e. old.
+						// Therefore let's pick the most recent timestamp from both creation and modification timestamps
```
mkbel, line 145: I would consider moving this documentation to where the rest of this function documentation is. (Line 66 of this file.)

mkbel (approved):

PR #2508, opened 2020-02-22
#2477 Improved M-code control for stealth mode/normal mode

### LINKED ISSUE #2477: [ENHANCEMENT]Stealth-mode/Normal-mode  M-code
**Printer type** - Mk3S
**MMU Upgrade** - MMU2S

As noted in #1676 , stealth/power mode is more than just motor currents now. MBL is painfully slow when in stealth mode (esp. 7x7), and very quick in high power mode. This is partly because of all the time spent switching modes before Z-probing.
  
I can live with the added noise for a few minutes, so I'd be thrilled if it was possible to use an M-code to toggle between normal and silent modes. That way, one could change to high power before MBL, and back to stealth for the print. 

If you already have an M-code number in mind or a suggestion for a good one, I'd be happy to implement this myself.

vintagepc:
This closes #2477 and allows dynamic gcode mode changes, for example to get crash detect in risky print areas, or faster MBL:
```
M914
G28 W
G80
M915
```
Features & changes:
- Enable M914/915 for TMC2130 to allow setting stealth and normal mode via g-code.
- Optional "P" argument to write the change to EEPROM when using M code
- Optional 'R' argument to revert to EEPROM value.
- Default behaviour is temporary, i.e. any change is lost either on M84 or reboot.
- Fix possible logic hole causing FORCE_HIGH_POWER_MODE_START to not shift modes if PSU_Delta is defined and there's a stepper timeout before MBL (e.g. slow heating or high target temp)
- Updated stealth mode menus in Tune and Settings menu to indicate if an override is in place. (Menu shows "M915" or "M914" as appropriate. Chosen because no translation is required)
- Shared code for M914/5, partially shared with force_high_power_mode()

Picture, during MBL in stealth mode with an override:
IMG_0018


vintagepc:
Hm, something's not quite right with MBL. Investigating.

vintagepc:
I think the issue has been resolved, it now plays nicely with MBL again.

vintagepc:
Closing - deprecated by #4053

Issue #2511, opened 2020-02-26
[FEATURE REQUEST] MK3 Firmware to handle increased MMU filament count


cjbaar:
**Printer type** - MK3S
**MMU Upgrade** -MMU2S (modified)

**Is your feature request related to a problem? Please describe.**
I have been building a mod to the MMU to utilize more than 5 filaments (currently testing with 12). I am able to get a working prototype using the parts and MCU from the original MMU2 kit. I am also able to create a slice models successfully with Prusa Slicer, and I can see tools up to T11 (as expected) in the output gcode. However, the MK3 firmware seems to be hard-coded to ignore any tool changes greater than T4 (shown on-screen as F5). I have done debugging on the MMU side, and confirmed that when the gcode for a higher tool is reached by the printer, no tool change command is forwarded to the MMU, and the printer does not attempt an unload.
  
**Describe the solution you'd like**
Ideally, the firmware should send the gcode data as-is, without putting restrictions on the tool number. If necessary, the printer could query the MMU for the value of EXTRUDERS when it queries for firmware versions.

Panayiotis-git:
I'm really interested in your solution!

I recently did some prints with 16 different filaments and I did the changes by hand. An improvement was required to the firmware to keep filament changes under control:
https://github.com/prusa3d/Prusa-Firmware/pull/2380

I also have in mind to modify the MMU to something expandable (to multiples of 5?) utilizing the same controller.

Have you shared any of your designs?

vintagepc:
This is likely because internally we need to allocate memory for tracking filament depletion in things like spooljoin - currently fixed at 5 slots, of course. 
It would likely not be too hard to abstract those count checks into a #define in the variants. The main ones I'm aware of are in Autodeplete.h and the T-code handling. There are likely many more - LCD menus, Load codes, cut codes...

cjbaar:
I've started to look through the code, and see a lot of hard-coding around this. As you mention, in AutoDelete, there is a definition for number of filaments:
    `static const uint8_t filamentCount = 5;`
But, then, that variable does not seem to be used in other places that reference the MMU. For example, the T commands are parsed into an ENUM of all MMU commands, which I don't quite get, instead of using just "T" and a variable.

I am not super concerned with accessing all of the slots from the menu (ie, "load to nozzle"), just letting the gcode be understood and not truncated.

cjbaar:
[...]

PR #2515, opened 2020-03-01
#2511 Increase maximum MMU filament count

cjbaar:
These changes are intended to allow the MK3S to understand and communicate filaments values greater than the default 5 (T0-T4). In the interest of maximizing the number, while still keeping the AutoDeplete routines intact around a single integer, I have gone with a maximum of 32. The value of the filament count is stored in variable, in the hopes that it can later be dynamically read and updated from the MMU itself.

The highlights of the changes include the following:
1) Create variable mmu_filament_count (default 5)
2) Allow gcode characters outside of 0-4 (updated to 0-9)
3) Extend manual ENUM commands for T,E,K,L from 0-4 to 0-31
4) Dynamically create "load" and "eject" filament menus
5) Change auto deplete storage from uint8_t to uint32_t
6) Correct some LCD display issues with the double-digit filaments

I have tested with a range of filament->filament changes successfully, using a prototype model with up to 12 slots. I have not tested the auto-deplete, as I have never actually used it, but the functions there remain unmodified. I have also not tested the LCD output with other languages, but have attempted to abbreviate strings that I think are too long when using double digits.


cjbaar:
I am not sure I have permission to link PR to issues. This is related to https://github.com/prusa3d/Prusa-Firmware/issues/2511

PR #2526, opened 2020-03-05
Mk3 decouple XYZE relative modes

### LINKED ISSUE #2514: [BUG][3.9.x] Significant E-axis behaviour difference (breaks prints) from 3.8.1
**Printer type** - MK3S
**Printer firmware version**-  3.9.0 RC / 3.8.1

**MMU Upgrade** - MMU2S

**Describe the bug**
There appear to be print-breaking E-axis differences with certain gcode combinations that were not present in 3.8.1.
I kept having issues with the attached g-code while running 3.9.0-RC (+ my various PRs merged in). 
Figuring it was something I did, I flashed back to stock RC. No dice.

I went even further back to 3.8.1, at which point the attached g-code prints fine.

In 3.9.0, The perimeter prints fine, but any retraction over-retracts and the rest of the print continues without extruding until the over-retraction gets to the point it trips the filament sensor.

There's something about the way this gcode does retractions that highlights the issue - regular sliced code from prusaslicer prints fine.

( for reference, the attached g-code is a first-layer speed and height calibration generated from 3dOptimizer.com  - it appears to be functional for many people on many makes of printers.)
  
**To Reproduce**
Print the attached g-code on 3.9.0-RC
  
**Expected behavior**
Gcode should print as expected, e.g. as it does in 3.8.1
  
**G-code**
[First_layer.gcode.zip](https://github.com/prusa3d/Prusa-Firmware/files/4269742/First_layer.gcode.zip)

**Video**
~Happy to capture one if necessary; I just don't have the time at the moment to set up a camera/flash versions/etc.~

update: plans changed, captured video. Pay attention to the orange rotating gear during retractions. Exact same g-code, both printers:
3.8.1 behaviour: https://vimeo.com/394576663
3.9.0-RC behaviour: (will upload tomorrow, I hit my weekly vimeo upload limit)

### LINKED ISSUE #2516: [BUG] 3.9.0-RC1 Multi-material print stops extruding after prime line, works fine in 3.8.1
MK3S fw 3.9.0-RC1 (w/ non-stock geared extruder)

MMU2S fw 1.0.6

PrusaSlicer 2.1.0 release or 2.2.0-beta

**Describe the bug**
This problem started after flashing to 3.9.0-RC1. Flashing back to 3.8.1 resolved the issue, so I'm pretty sure it's firmware related.

Note that I'm using a non-stock geared extruder with 490 E-steps and 16 microsteps on the E axis.  I don't know if this is related to the issue or not; I haven't tried it on a stock extruder.

After printing the initial prime line, the printer moves to start printing the wipe tower. Other than a small blob, the printer stops extruding altogether for the rest of the print. The printer goes through all the motions of printing the rest of the print (moving in X/Y/Z axes), but the E axis only turns back and forth very slightly, instead of trying to extrude. It isn't a clog or jam, as it happens every time at exactly the same place (start of the wipe tower), and stopping & restarting will print a successful prime line every time. Single color prints work fine with no issues. 

This happens with files sliced by both PrusaSlicer 2.1.0 final, and 2.2.0 beta. It happens regardless of whether I have "Prime all printing extruders" enabled or disabled. If enabled, all 5 filaments will successfully prime, but extrusion stops as soon as it moves to start printing the wipe tower. 

I tried changing many different slicer settings, including the M900 K value in the filament gcode (tried K30 or K0.05) but there was no difference.

Wild guess: Some problem with the new LA1.5 code with a geared extruder?

G-code: https://www.dropbox.com/s/co8zsxaywitjstk/ps220beta%20Worf%20badge%20part%201_0.2mm_PLA_MK3SMMU2S_1d17h36m.zip?dl=0

3MF: https://www.dropbox.com/s/ks3unyqcbznwkb9/worf%20and%20geordi.3mf?dl=0

Video: https://www.dropbox.com/s/drd7xngjc3nwap1/IMG_4689.mov?dl=0
  
**To Reproduce**
Print the gcode file using fw 3.9.0-RC1. If the issue doesn't happen on stock, you may need to set E-steps to 490 and E microstepping to 16 to reproduce it.
  
**Expected behavior**
Print should succeed just like when using fw 3.8.1.

leptun:
This PR splits G90/G91 from M92/M93
This fixes the change that came in the 3.9 prerelease versions which unfortunately caused incompatibility with some gcodes.
G90/G91 no longer affects the extruder relativity. This is in contrast to marlin 2.0, which handles these commands just like we did prior to this PR.
Gcodes that were tested and were confirmed to work:

- File sliced with the latest PrusaSlicer MK3 profile.
- Gcode that comes with the printer on the SD card.
- https://github.com/prusa3d/Prusa-Firmware/issues/2514

Fixes #2514 
Fixes #2516 
PFW-1103

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -5365,7 +5365,7 @@ if(eSoundMode!=e_SOUND_MODE_SILENT)
 
     /*!
 	### G90 - Switch off relative mode <a href="https://reprap.org/wiki/G-code#G90:_Set_to_Absolute_Positioning">G90: Set to Absolute Positioning</a>
-	All coordinates from now on are absolute relative to the origin of the machine.
+	All coordinates from now on are absolute relative to the origin of the machine. E axis is left intact.
     */
     case 90: {
 		axis_relative_modes &= ~(X_AXIS_MASK | Y_AXIS_MASK | Z_AXIS_MASK);
@@ -5374,7 +5374,7 @@ if(eSoundMode!=e_SOUND_MODE_SILENT)
 
     /*!
 	### G91 - Switch on relative mode <a href="https://reprap.org/wiki/G-code#G91:_Set_to_Relative_Positioning">G91: Set to Relative Positioning</a>
-    All coordinates from now on are relative to the last position.
+    All coordinates from now on are relative to the last position. E axis is left intact.
 	*/
     case 91: {
 		axis_relative_modes |= X_AXIS_MASK | Y_AXIS_MASK | Z_AXIS_MASK;
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -178,7 +178,7 @@ float default_retraction = DEFAULT_RETRACTION;
 
 
 float homing_feedrate[] = HOMING_FEEDRATE;
-uint8_t axis_relative_modes;
+uint8_t axis_relative_modes = 0;
 int feedmultiply=100; //100->1 200->2
 int extrudemultiply=100; //100->1 200->2
 int extruder_multiply[EXTRUDERS] = {100
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -181,6 +181,8 @@ float homing_feedrate[] = HOMING_FEEDRATE;
 
 //Although this flag and many others like this could be represented with a struct/bitfield for each axis (more readable and efficient code), the implementation
 //would not be standard across all platforms. That being said, the code will continue to use bitmasks for independent axis.
+//Moreover, according to C/C++ standard, the ordering of bits is platform/compiler dependent and the compiler is allowed to align the bits arbitrarily,
+//thus bit operations like shifting and masking may stop working and will be very hard to fix.
 uint8_t axis_relative_modes = 0;
 
 int feedmultiply=100; //100->1 200->2
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -178,9 +178,7 @@ float default_retraction = DEFAULT_RETRACTION;
 
 
 float homing_feedrate[] = HOMING_FEEDRATE;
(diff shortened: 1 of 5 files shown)


vintagepc:
I'll merge and tinker with this tonight and run through some more of the code from the tool that gave me issues in #2514 
Thanks!

vintagepc:
Running a print now, all seems good, E axis behaving as expected again.

Firmware/Marlin_main.cpp
```diff
@@ -5367,21 +5365,19 @@ if(eSoundMode!=e_SOUND_MODE_SILENT)
 
     /*!
 	### G90 - Switch off relative mode <a href="https://reprap.org/wiki/G-code#G90:_Set_to_Absolute_Positioning">G90: Set to Absolute Positioning</a>
-	All coordinates from now on are absolute relative to the origin of the machine. E axis is also switched to absolute mode.
+	All coordinates from now on are absolute relative to the origin of the machine.
```
DRracer: E axis is left intact

Firmware/Marlin_main.cpp
```diff
@@ -178,9 +178,11 @@ float default_retraction = DEFAULT_RETRACTION;
 
 
 float homing_feedrate[] = HOMING_FEEDRATE;
-// Currently only the extruder axis may be switched to a relative mode.
-// Other axes are always absolute or relative based on the common relative_mode flag.
-bool axis_relative_modes[] = AXIS_RELATIVE_MODES;
+
+//Although this flag and many others like this could be represented with a struct/bitfield for each axis (more readable and efficient code), the implementation
+//would not be standard across all platforms. That being said, the code will continue to use bitmasks for independent axis.
+uint8_t axis_relative_modes = 0;
```
DRracer, line 186: Moreover, according to C/C++ standard, the ordering of bits is platform/compiler dependent and the compiler is allowed to align the bits arbitrarily, thus bit operations like shifting and masking may stop working and will be very hard to fix.

leptun: Updated

DRracer (approved):

PR #2397, opened 2020-01-04
Show the FlashAir IP address

odaki:
Displays the FlashAir IP address in the "Support" menu if the card is available.
If no IP address has been assigned yet, "0.0.0.0" will be displayed.

PR #2088 works fine with my FlashAir W-03 32GB.
However, even if a FlashAir IP address is assigned, the IP address is not displayed.
So, instead of the old commented out code, I created a small patch that shows the IP address.

The patch tested with a FlashAir W-03 32GB.

Please merge if you like.

PR #2541, opened 2020-03-15
Fix unexpected timeout for accessing SD card

odaki:
Prusa-Firmware has a 32-bit millisecond counter for various timeout checks.
However, when accessing the SD card via SPI, the millisecond timer value is treated as a 16-bit timer (the maximum is 65.5 seconds).
So, if the lower 16 bits of the system's millisecond timer were close to 0xffff, the timeout was too early (in a bad case, 1 millisecond instead of 300 milliseconds).

Therefore, the counter variable was changed from 16 bits to 32 bits.

---
For example, if the timing of inserting the SD card happens to be the value of 0x00001ffff millis of the system timer, this condition will occur.


odaki:
If possible, please fix it together with #2397 to stabilize the process of accessing FlashAir.

Thank you.

odaki:
Excuse me. I've withdrawn this PR because I misunderstood the processing.

PR #2549, opened 2020-03-23
EEPROM documentation 

3d-gussner:
This is an attempt to document the EEPROM via doxygen.

- All D-codes in EEPROM Table have been tested
- Some values or functions are unknown to me and marked with `???` ... some `???` are just place holder
- During test I found few EEPROM space allocations to be too big
  - `EEPROM_FARM_NUMBER` allocated 3 bytes but only uses 2 bytes for the numeric farm number 000-999
  - `EEPROM_CRASH_DET` allocated 5 bytes but only uses 1 byte to store the status
  - Therefore I added `EEPROM_FREE_NR1 - EEPROM_FREE_NR5` to keep the actual space allocation intact.
  - Tried to find out via github `blame` if one of these has been changed lately but it doesn't look like that.

To verify the default/factory rest values the printer has been reset to delete `All Data` and `Shipping prep` via factory reset several times.
Also searched the source code for valid values.
Tested LCD menu settings to confirm with the `D3 Axyyyy Cz` D-code that the correct values are in the documentation.

Few EEPROM values aren't initialized after factory reset and only a power off/on set the correct default values, here the known to me list:
```
EEPROM_PRINTER_TYPE
EEPROM_BOARD_TYPE
EEPROM_SOUND_MODE
EEPROM_BACKLIGHT_LEVEL_HIGH
EEPROM_BACKLIGHT_LEVEL_LOW
EEPROM_BACKLIGHT_MODE
EEPROM_BACKLIGHT_TIMEOUT
```
This should be addressed in another PR

The `EEPROM_MMU_FAIL_TOT` and `EEPROM_MMU_LOAD_FAIL_TOT` are uint16 but in `ultralcd.cpp` I found a bug that these are used as uint8. I will prepare another PR for this.

Also the D-codes are case sensitive and I will prepare another PR and update the RepRap Wiki as well after it has been pulled.

Please let me know if something is wrong or should be changed.

PFW-985

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -9124,8 +9124,8 @@ Sigma_Exit:
 #### End of D-Codes
 */
 
-  /** @defgroup GCodes G-Code List 
-  */
+/** @defgroup GCodes G-Code List 
+*/
 
 // ---------------------------------------------------
 
--- a/Firmware/eeprom.h
+++ b/Firmware/eeprom.h
@@ -48,7 +48,8 @@ static_assert(sizeof(Sheets) == EEPROM_SHEETS_SIZEOF, "Sizeof(Sheets) is not EEP
   - Octoprint does support D-codes
   - _Pronterface_ does <b>not</b> support D-codes
 
-
+  ### !!! D-codes are case sensitive so please don't use upper case A,C or X in the address you want to read !!!
+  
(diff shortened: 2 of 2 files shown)


3d-gussner:
If PR is approved can you update the Version from 0.9.1 to 1.0.?!

DRracer (approved):

PR #2554, opened 2020-03-25
Mk3 dcodes case sensitive

3d-gussner:
Read/write to EEPROM / RAM / Flash are case sensitive as A,C and X are used by the code.
The count has to decimal.


DRracer (approved):

DRracer (approved):

3d-gussner:
Updated RepRap Wiki

PR #2571, opened 2020-04-01
🐛 Fix few LCD issues

3d-gussner:
#### Issue
Some EEPROM values are uint16 but have been read/displayed in LCD Fail Statistics as uint8 or limited to 000-999
This could cause strange results on display.
- EEPROM_MMU_FAIL_TOT is uint16 but it was read as uint8
- EEPROM_MMU_LOAD_FAIL_TOT is uint16 but it was read as uint8

Other values have been read correctly BUT limited to 000-999 on display.

#### Changes
Due to the limited LCD space I had to shorten few messages.

Also updated the `c=xx r=yy` information in the `lang_en.txt / lang_en_??.txt` files so translators know what the message limit is.

By using an existing `MSG_FSENSOR` instead of several `Fil. sensor` I could save 28bytes of program storage space.
It doesn't sound a lot, but by removing these "duplicate" messages from the `lang_en.txt/lang_en_??.txt` also the needed "2nd language" space got smaller and could save additional space on the internal flash.

So by re-using the MSG_FESNSOR we kind of saved ~ 28*2 bytes

#### Test
Tested the LCD changes with and without the MMU2 connected and used `D3` Dcode to manipulate the output on the display.

#### To-do
Maybe translators need to review the `lang_en_??.txt`files to ensure that the translation isn't too long.
With the `c=xx r=yy` they should now know the limit.

#### Known issue
Some Statistics are really at the 20x4 limit and may be hard to read, but I tried my best to keep them similar and readable.


3d-gussner:
Close PR for the moment as documentation and manuals may need to be updated.
We will discuss and decide that later what will happen.

Issue #2570, opened 2020-04-01
[BUG]Division by zero with "M900 K15"


smurfix:
**Printer type** - MK3S
**Printer firmware version**-  current MK3 master

**Describe the bug**

This code results in `new_K` =0 when `k<=15`. This later results in a division-by-zero error and a printer reboot.

```
  // Approximate a LA10 value to a LA15 equivalent.
static float la10c_convert(float k)
{
    float new_K = k * 0.004 - 0.06;
    return (new_K < 0? 0: new_K);
}
```

**To Reproduce**
Print with `M900 K=16`. Works. Now print with `K=15`. Crash.

wavexx:
Did you find this by chance during a test of because you happened to use a K value below 15 with some material?

wavexx:
I'm looking at this along with #2566. Do you have a short gcode that makes the print crash/reboot? I'd love to add this as a test case.

smurfix:
I am printing with a 0.8mm nozzle and noticed splodged bottom infill at the edges even when the overlap is set to zero. The gcode that comes with PrusaSlicer reduces K to 18 when the nozzle is set to 0.6 so I tried with 15, which then crashed before emitting a single blob of filament. Editing the gcode to K=16 rescued the print but still has a bit too much infill.

Thus I assume that basically any print will die when you set K=15.

wavexx:
> Thus I assume that basically any print will die when you set K=15.

Not always. Note that the entire range down to K=0 is actually a legitimate value for LA. There is a problem later on that can cause either a zero division or (in case of the other issue) excessive loss of precision, but it also depends on the segment at hand.

I can reproduce nicely the issue in #2566 but couldn't get a crash right away

smurfix:
OK. I'll try to recreate the offending file.

vintagepc:
Just to clarify here... you are setting `M900 K15` right? Not `M900 K**=**15`? I ask because of the quotes in the title. `K=15` is not valid gcode.

smurfix:
Yes, M900 K15, sorry.

UltiCa:
I have the same problem with M900 K15 on my Prusa. I do not know, if there is a division by zero problem, but if I set M900 K15 (M900 K30 is running quite well), then next command cashes the Prusa (MK3S with MMU2S and 3.8.1). He does nothing, the display is frozen 1-2-3 seconds later the Prusa reboots.

[...]

Issue #2566, opened 2020-03-30
[BUG]SUPER slow speed with new firmware


forestqc69:
MK3 - 3.9.0-RC1-3272

I've been printing the same piece for 1 week, after upgrading to the current RC firmware, print speed is SUUUPER slow, the time remaining countdown didn't move for 5 minutes or more.  Non-print Movements are as fast as usual I would say.. And yes, print speed is 100%.

[3xtubeV2_Mod1_0.6n_0.2mm_PLA_MK3S_5h37m.gcode.zip](https://github.com/prusa3d/Prusa-Firmware/files/4403523/3xtubeV2_Mod1_0.6n_0.2mm_PLA_MK3S_5h37m.gcode.zip)

leptun:
How does it manifest? Really low acceleration? I can't test the gcode at the moment sine all my printers are printing face shields.

forestqc69:
I flashed back to 3.8.1 and regular speed resumed, I'd say it like 10-15 time slower on print movements, non-print movement seemed at regular speed.  It's so slow I can't even see accelerations, it just all printing movement that are so slow.. I did put the speed to 120% and it wasn't even close to going at the regular speed..

DRracer:
@forestqc69 what nozzle are you using?

DRracer:
@forestqc69 
original gcode - really slow
E jerk set to 4.5 in gcode - no change
E jerk 10 in gcode - a little less slow
change from K15 to K30 - normal speed
change from K15 to K0.03 - normal speed
It looks like K15 is being converted to some bad value, we'll look at it, thanks for reporting ;)

forestqc69:
0.6mmLe 31 mars 2020 11:12 a.m., DRracer <notifications@github.com> a écrit :
@forestqc69 what nozzle are you using?

wavexx:
Related  #2570

sn4k3:
yeah that also happen to me with my ABS profile when i had K=15, had to return to 18 to be able to print. Bad conversions...

Issue #2543, opened 2020-03-16
[BUG] Firmware 3.9.0-RC1 print quality degradation compared to 3.8.1


nikkolade:
**Printer type** - MK3S
**Printer firmware version**- 3.9.0-RC1
**MMU Upgrade** - MMU2S
**MMU upgrade firmware version 1.0.6

**Describe the bug**

Print source: https://www.thingiverse.com/thing:3494496
and explicitly: https://www.thingiverse.com/download:6386081

While printing the above STL sliced with PrusaSlicer 2.20-RC2 and using PETG material works just fine when using the MK3S firmware 3.8.1 (build result on the left), the very same gcode results an awful quality after the printer was upgraded to 3.9.0-RC1 (end result on the right).
Please, see the attached picture.

  
**To Reproduce**
Upgrade firmware MK3S firmware to 3.9.0-RC1
and print the above mentioned STL file.
  
**Expected behavior**
The print quality should not degrade, compared to previous firmware version.

 

![IMG_20200316_163306](https://user-images.githubusercontent.com/3157534/76774208-4f424380-67ac-11ea-9702-df1111194ab4.jpg)

wavexx:
Could you post the 3mf file saved by PrusaSlicer using the same settings you used to print it? Also posting the gcode itself would help so we can have a look.

nikkolade:
Sure, here is the 3mf file.


[3mf.zip](https://github.com/prusa3d/Prusa-Firmware/files/4339149/3mf.zip)

nikkolade:
And the gcode
[gcode.zip](https://github.com/prusa3d/Prusa-Firmware/files/4339156/gcode.zip)

vintagepc:
Hm... have you tried re-calibrating your linear advance K-value with 3.9.0? I'm not saying the default shouldn't work fine, but I'm curious whether calibrating for the 1.5 K-value makes the issue go away.

nikkolade:
No, I haven't. What is the proper way to run the linear advance calibration ?
My printer is currently busy for the next 12 hours, but I can certainly try this out after it's idle again.

leptun:
You can find instructions on what Linear advance is and how to calibrate it here: https://marlinfw.org/docs/features/lin_advance.html
You are looking for LA 1.5.

vintagepc:
[...]

PR #2591, opened 2020-04-09
Fix chained wipes in Linear Advance 1.5

wavexx:
LA assumes all the nozzle pressure is released at the end of each
extrusion, which makes calculating the required pressure advance during
travels and retracts not normally necessary.

This is not always true in our planner, since the E axis is explicitly
ignored when not in use, but also due to E-jerk allowing a non-linear
jump in speed. And since the compression factor is currently tied by XYZ
axes and not independently calculated, this can result in a wrong
estimation of final pressure in several conditions.

In this PR we change the underlying assumptions about pressure:

1) Pressure is no longer lost when LA is disabled: if a retract is
followed by an unretract of the same length, the pressure will be likely
maintained entirely. This also holds true during travels, as long as the
retract length can overcome all the backpressure (which is the case in
all but the most noodly materials)

2) Pressure is released as soon as possible during travels: we now
enable LA also during travels, but under the sole condition of undoing
excess pressure.

If a wipe (or more generally, any travel) has been chained to an extrusion
move, the cruising phase will now also release any excess pressure so
that the subsequent extrusion re-starts from an ideal zero as previously
assumed in 1). This will reduce the amount of ooze during the wipe, but
most importantly avoids a small-but-constant build-up of material after
the travel if the wipe couldn't effectively lose the pressure in the first place.

This hopefully addresses #2543 

We also address the related issues #2570 and (partially) #2566. We now check
for zero e/D ratio as a special flag to handle wipes, which always avoids a 0
at the denominator. We also perform the acceleration limiting one step
earlier to avoid 32bit overflow, which did incorrectly limit low-compression
segments.

PFW-1105

### DIFF
--- a/Firmware/planner.cpp
+++ b/Firmware/planner.cpp
@@ -1065,10 +1065,12 @@ Having the real displacement of the head, we can calculate the total movement le
      * delta_mm[E_AXIS] >= 0    : Extruding or traveling, but _not_ retracting.
      * |delta_mm[Z_AXIS]| < 0.5 : Z is only moved for leveling (_not_ for priming)
      */
-    block->use_advance_lead = extruder_advance_K
+    block->use_advance_lead = extruder_advance_K > 0
                               && delta_mm[E_AXIS] >= 0
                               && abs(delta_mm[Z_AXIS]) < 0.5;
     if (block->use_advance_lead) {
+        // all extrusion moves with LA require a compression which is proportional to the
+        // extrusion_length to distance ratio (e/D)
         e_D_ratio = (e - position_float[E_AXIS]) /
                     sqrt(sq(x - position_float[X_AXIS])
                          + sq(y - position_float[Y_AXIS])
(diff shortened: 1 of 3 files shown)


Firmware/planner.cpp
```diff
@@ -1082,10 +1082,10 @@ Having the real displacement of the head, we can calculate the total movement le
...
         if (e_D_ratio > 3.0)
             block->use_advance_lead = false;
-        else {
-            const uint32_t max_accel_steps_per_s2 = cs.max_jerk[E_AXIS] / (extruder_advance_K * e_D_ratio) * steps_per_mm;
-            if (block->acceleration_st > max_accel_steps_per_s2) {
-                block->acceleration_st = max_accel_steps_per_s2;
+        else if (e_D_ratio > 0) {
+            const float max_accel_per_s2 = cs.max_jerk[E_AXIS] / (extruder_advance_K * e_D_ratio);
+            if (cs.acceleration > max_accel_per_s2) {
+                block->acceleration_st = ceil(max_accel_per_s2 * steps_per_mm);
```
DRracer, line 1088: Is this really ceil() - aka the smallest integral value that is not less than x (that's something else than rounding).

wavexx: It is. We do that when calculating most acceleration values (see a few lines above where the non-limited value of acceleration_st is computed).

DRracer (approved):

PR #2623, opened 2020-04-20
⚡️LA1.0 compatibility improvements

leptun:
PFW-1113


wavexx:
I like the la10_reset on stop, however I'm thinking of the consequences of resetting the mode with K0.

My original wish was, of course, to lock a print to either LA10 or LA15 so that any value seen during a tool change or a fill type that temporarily disables LA cannot change mode as well.

This will restore the un-adjusted e-jerk, which is not great if you just want to disable LA for infill (one example).

Maybe I'm over-cautious here.

leptun:
Closed in favor of another PR

PR #2632, opened 2020-04-25
IR fsensor fixes

DRracer:
The IR sensor v0.3 or older may have got false-detected as v0.4 and then get disabled, which would have been extremely annoying. Sadly, I found the issue right after tagging the RC2 :( (Murphy's law works...).

PFW-1111


dragomirecky (approved):

3d-gussner:
Just a short summery of testing this PR, believe me you don't want to read 1000+ lines what has been created during tests.

Tested the initial PR and the results were promising.

Just few hours after RC2 has been tagged we found a random issue:
It was possible to trigger the 0.3 IR sensor in a way that it was detected as new 0.4 IR sensor.
Within few hours we could test an updated version.
Tested again, approved bug fixed.

Today we tried to think of and generate "rare/potential" issue and there is still some room to improve the user experience in odd/wired/random scenarios.

These will be addressed in next PR and will be hopefully part of FW 3.9.0 final.

PR #2627, opened 2020-04-22
⚡Add Linear Advanced 1.5 Max value

3d-gussner:
### Purpose of this PR
To determine if `M900 Kx` is a LA15 or LA10 value we had a "hard-coded" value of 10.
During testing and a lot of discussions we couldn't agree on one fixed value as we didn't have all Linear Advanced 1.5 Filament K values on hand.

### Solution
We decided to add one variable `LIN_ADVANCE_15_K_MAX` so this can be updated if needed just in one place.

### ToDo
- The LA15 K values for all "know"/supported filaments have to be defined by PrusaSlicer-settings team and the community.
- Update the `LIN_ADVANCE_15_K_MAX` if needed.

## Dependencies
No known dependencies to other PRs.
However, I tested it with https://github.com/prusa3d/Prusa-Firmware/pull/2623 as this PR also makes sense at this moment.

### Tests
- This firmware has been compiled and tested:
   - LA10 K values above `LIN_ADVANCE_15_K_MAX`  has been converted correctly.
   - LA15 K values below/equal to `LIN_ADVANCE_15_K_MAX` have been used correctly.
- As there is an option to disable the LA10 -> LA15 conversion I also compiled and tested this firmware:
   - LA10 K values above `LIN_ADVANCE_15_K_MAX` have been ignored and the LA15 K value was set to `K0.00`.
   - LA15 K values below/equal to `LIN_ADVANCE_15_K_MAX` have been used correctly.
   - LA15 K values above `LIN_ADVANCE_15_K_MAX` have been ignored and the LA15 K value was set to `K0.00`.

#### Test log:
- Firmware based on MK3 (today) + https://github.com/prusa3d/Prusa-Firmware/pull/2623 + `#define LIN_ADVANCE_15_K_MAX 9.99`.

```
Recv: FIRMWARE_NAME:Prusa-Firmware 3.9.0-RC2 based on Marlin FIRMWARE_URL:https://github.com/prusa3d/Prusa-Firmware PROTOCOL_VERSION:1.0 MACHINE_TYPE:Prusa i3 MK3S EXTRUDER_COUNT:1 UUID:00000000-0000-0000-0000-000000000000
[...]
Send: M900 K45
Recv: LA10C: Linear Advance mode: 1.0
Recv: LA10C: Adjusted E-Jerk: 4.50
Recv: echo:Advance K=0.13
Recv: ok
[...]
Send: M900 K0
Recv: LA10C: Linear Advance mode: UNKNOWN
[...]

PR #2639, opened 2020-04-28
Adjustable Linear Advance limits and thresholds

wavexx:
Based on PR #2627, we rename and add some constants to avoid hard-coding limits in several parts of the code. This PR also decouples the LA1.5 maximum value from the detection threshold, although it's now simply defined in terms of the first, matching existing behavior.

We also clamp the result of `la10c_convert()` to respect `LA_K_MAX` which would otherwise could be exceeded by a large LA10 value.

The mode detection is also reset when stopping a print, which is a perfectly safe thing to do (taking parts of PR #2623).

PFW-1113

### DIFF
--- a/Firmware/Configuration_adv.h
+++ b/Firmware/Configuration_adv.h
@@ -285,11 +285,13 @@
 #define LIN_ADVANCE
 
 #ifdef LIN_ADVANCE
-  #define LIN_ADVANCE_K 0  // Unit: mm compression per 1mm/s extruder speed
-  //#define LA_NOCOMPAT    // Disable Linear Advance 1.0 compatibility
-  //#define LA_LIVE_K      // Allow adjusting K in the Tune menu
-  //#define LA_DEBUG       // If enabled, this will generate debug information output over USB.
-  //#define LA_DEBUG_LOGIC // @wavexx: setup logic channels for isr debugging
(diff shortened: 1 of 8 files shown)


wavexx:
I now incorporated the mode reset also during K0, effectively doing the same as PR #2623.

By moving `la10c_reset()` from G80 into `finishAndDisableSteppers()` we handle reset both at the end of a print (via M84 - which is additionally _checked_ when printing via SD) and also via `lcd_print_stop()` as previously done.

I've moved out finishAndDisableSteppers out of stepper.cpp, as it clearly doesn't belong there (it includes a call to functions in both planner and Marlin_main).

3d-gussner (approved):
Here some short tests and results.
- MMU gcode with two LA10 materials. Works as expected.✔
- Single material gcode with "new" PS profile having LA15 K value 1st and directly after that the LA10 K value. Works as expected.✔
- Print tiny cube. Works as expected.✔
  - It detects the LA15 K value 1st and ignores the following LA10 K value ✔
  - At the end with M84 it resets the LA10C: Linear Advance mode: UNKNOWN ✔
- Canceling a print in Octoprint also resets the LA to Unknown.  Works as expected.✔
- Testing LA K values close to 10 shows that the #PR is working as expected. ✔

DRracer (approved):

PR #2640, opened 2020-04-28
Fixes delayed display of choices for multiscreen messages for short messages

aeoa:
If a short message is displayed using the function lcd_show_multiscreen_message_two_choices_and_wait_P and this message requires only a single page, then the two choices are currently only displayed after waiting 5 seconds.

This pull request moves the code responsible for displaying the choices in front of the code that does the waiting and other stuff so the choices are displayed immediately. As far as I can tell this shouldn't have any side effects.


3d-gussner:
@gudnimg I think this is obsolete, please check.

gudnimg:
Yup this one is obsolete. We are calling `lcd_show_choices_prompt_P()` to show the choices.

Closing PR as resolved

Issue #2606, opened 2020-04-14
  [ENHANCEMENT] Adjust MM_PER_ARC_SEGMENT default value


FormerLurker:
**Printer type** - All Prusa Printers
**MMU Upgrade** - NA (this is purely a printer firmware request)

**Is your enhancement related to a problem? Please describe.**
I have been experimenting with a new algorithm I created that converts G0-G1 commands into G2-G3 commands to eliminate stutter introduced by serial streaming and to compress the GCode files by approx 40-60%.  However,the conservative value of 1 for MM_PER_ARC_SEGMENT creates very noticeable flat edges that should clearly be curves.  This results in a noticable decrease of print quality around tight curves (around 1mm or less) when using the arc commands, even though the tool paths are practically identical (+- 0.025mm).  Longer gentler arcs turn out fantastic, and are indistinguishable from those printed using G1 commands.  Please let me know if you need to see comparisons.  

**Describe the solution you'd like**
I would recommend a value of at most 0.1mm for quality that is comparable to the output of PrusaSlicer when using G2/G3, as long as that won't affect printer performance in your opinion.  If it's possible to go even lower, down to the resolution of the stepper, that would be ideal if it doesn't negatively affect printer performance.  In any case, I believe 1mm is too conservative.  Since most people are not using G2/G3 in everyday printing, I believe this has been overlooked.

Here is the relevant setting in the Configuration_adv.h file:
https://github.com/prusa3d/Prusa-Firmware/blob/66ffb8489ac34cd62605ea744e262c429f7c479f/Firmware/Configuration_adv.h#L296

I would be more than willing to extensively test a build that contained the altered setting.  I'm running an MK2.5 w MMU2.0.  I am planning to compile a version with that setting modified as soon as possible, but wanted to reach out first.

Thank you very much for taking the time to read this, and for all of your efforts!

FormerLurker:
I've been experimenting with the ```MM_PER_ARC_SEGMENT ``` and ```N_ARC_CORRECTION``` defines. I increased the ```N_ARC_CORRECTION``` setting inversely as I decreased the value of ```MM_PER_ARC_SEGMENT ``` so that the same number of corrections per mm are being applied.

[...]

PR #2657, opened 2020-05-07
Add arc interpolation features (G2/G3) and M214 command for controlling settings

FormerLurker:
I don't expect this pull request to be immediately accepted, and would greatly appreciate your thoughts and feelings about the changes I have made.  This is my first attempt to modify any firmware, so I expect a lot of critical feedback.

I recently created a program to convert G0/G1 commands into G2/G3 commands where possible, and noticed that some arcs were coming out as flat edges.  These edges were clearly rounded in the original gcode, and I tracked the issue down to the MM_PER_ARC_SEGMENT setting.  I did not have much luck at all decreasing this setting.  A value of 0.1 (similar to the minimum distance between segments in my original file) yielded terrible stuttering that degraded the print quality severely.  A value of 0.75 yielded flat edges in some places.  I couldn't seem to find a good value that produced rounded small curves that didn't result in stuttering that impacted quality, so I started exploring the Marlin 2.0 code, and came up with a few ideas of my own that I've added to this pull request.

I've added the following features from Marlin 2.0 to enhance arc interpolation:

**MIN_ARC_SEGMENTS** - Minimum number of segments in a full circle of any radius (0 = disabled)
**ARC_SEGMENTS_PER_SEC** - Number of segments per second to be generated based on current F (0 = disabled)

Note that when the two settings above conflict, the smallest arc length calculated from the above two settings is used.  Also MM_PER_ARC_SEGMENT is the max segment length allowed.

I've added a third setting to prevent extremely small segments from being generated:

**MIN_MM_PER_ARC_SEGMENT** - The minimum length of any segment (0 = disabled)

I've also added a new command, M214 (is that and appropriate gcode?) with the following definition that can be used to control the above settings:

    ### M214 - Set Arc configuration values (Use M500 to store in eeprom)

    #### Usage

        M214 [N] [S] [R] [F]

    #### Parameters
    - `N` - A float representing the max and default millimeters per arc segment.  Must be greater than 0.
    - `S` - A float representing the minimum allowable millimeters per arc segment.  Set to 0 to disable
    - `R` - An int representing the minimum number of segments per arcs of any radius,
            except when the results in segment lengths greater than or less than the minimum
            and maximum segment length.  Set to 0 to disable.
    - 'F' - An int representing the number of segments per second, unless this results in segment lengths
            greater than or less than the minimum and maximum segment length.  Set to 0 to disable.

Please let me know if you feel the parameters I used (N, S, R and F) seem appropriate to you.

I also added M214 settings to the M503 configuration display command, which produces output similar to this:
```
Recv: echo:Arc Settings: N=Arc segment length max (mm) S=Arc segment length Min (mm), R=Min arc segments, F=Arc segments per second.
Recv: echo:  M214 N1.00 S0.50 R16 F0
```
I have tested M500 to ensure it works as expected for storing these settings to EEPROM.  You will find these new settings within the ConfigurationStore.h file:

```
    // Arc Interpolation Settings, configurable via M214
    float mm_per_arc_segment;
    float min_mm_per_arc_segment;
    int min_arc_segments; // If less than or equal to zero, this is disabled
    int arc_segments_per_sec; // If less than or equal to zero, this is disabled
```

I've added default settings for the arc interpolation to each .h file in the /variants folder and have removed the settings from configuration_adv.h.

I have tested installing the latest firmware release (3.8.1) and then upgrading to my variant, and the default settings seemed to load correctly.

The thing I am NOT sure about is the initialization of these new settings.  Though I'm sure you will review this completely, and some changes will be necessary, I am not at all sure that I did the initialization properly within ConfigurationStore.cpp.  Any help or advice there would be greatly appreciated.

I also removed the N_ARC_CORRECTION setting and functionality completely.  I was seeing bad interpolations using a wide variety of settings (endpoints quite far from the arc itself) due to the small angle approximation of COS and SIN.  I removed the small angle approximation completely, so there is one COS and one SIN calculation that happens at the beginning of the interpolation loop as long as there is more than one segment.  This change immediately resulted in arcs whose endpoints fit nearly exactly along the arc itself for all sizes I tested.  The correction factor seemed completely unnecessary after using accurate SIN/COS values.  I realize there is a trade-off here, since using those functions results in the first arc segment being inserted into the planner a bit later, but in my testing not only did I not notice any performance difference, but the resulting arcs fit better, especially in arcs of a small radius or where MM_PER_ARC_SEGMENT is large.  I did experiment with using a cutoff angle for deciding when to use true SIN and COS, but I did not notice any performance gains at all, so I dropped that in favor of simpler code.  Please let me know what you think of this.  I wrote a small program that generates the gcode equivalent of the interpolated arcs the firmware is producing if you are wondering how I discovered this.

Additionally, I think the default settings may need to be tweaked somewhat, though they seem to work well for me.  There is still some slowdown (but no stuttering) with the default settings when printing arcs with G2/G3 of a very small radius at high speeds (50+mm/sec), and I think that could be eliminated.  Now that I've added M214, testing should be MUCH easier.  I was hoping to find a way to print multiple copies of the same object and inject the M214 command between each object, but I have not yet found a way to do that automatically.  I'm sure there is a way, I just haven't found it yet.

If you need some gcode that includes a lot of arcs to test these changes, please let me know.  I have code for the MK2.5 w/mmu 2.0, and could generate gcode for any model you want.  I created a 1 layer file containing 6 circles using archimedean chord fill where each circle is printed at a different speed (from 10-60MM/s) to test stuttering, which was quite useful and illuminating.  

Thank you very much for your time and consideration!


Firmware/ConfigurationStore.cpp
```diff
@@ -268,6 +276,11 @@ bool Config_RetrieveSettings()
                 memcpy_P(&cs.max_acceleration_units_per_sq_second_silent[i],&default_conf.max_acceleration_units_per_sq_second_silent[i],sizeof(cs.max_acceleration_units_per_sq_second_silent[i]));
             }
         }
+        // Initialize arc interpolation settings if they are not already (Not sure about this bit, please review)
+        if (0xff == cs.mm_per_arc_segment) cs.mm_per_arc_segment = DEFAULT_MM_PER_ARC_SEGMENT;
```
leptun: The default for float values is not handled correctly. Look above at how it was done for `cs.max_feedrate_silent`. You can't compare the invalid float value to 0xFF. You would never get a valid match unless the float was 255.f

Wait. I might have don goofed. Please wait a second. max_acceleration_units_per_sq_second_silent is uint32, not float

FormerLurker: I was pretty sure I botched that bit.  Are the int settings handled properly?

I will examine ```cs.max_feedrate_silent``` and make the appropriate changes.

leptun:
```
for(uint8_t j = 0; j < sizeof(float); ++j)
            {
                if(0xff != reinterpret_cast<uint8_t*>(&(cs.max_feedrate_silent[i]))[j]) initialized = true;
            }
            if (!initialized) memcpy_P(&cs.max_feedrate_silent[i],&default_conf.max_feedrate_silent[i], sizeof(cs.max_feedrate_silent[i]));
```

It is done correctly for this float.

leptun: As for the int values, it would be better if you would cast the cs.int value to uint16_t and comparing to 0xFFFF.
So it would be something like this:
`if (((uint16_t)cs.min_arc_segments) == 0xFFFF) cs.min_arc_segments = DEFAULT_MIN_ARC_SEGMENTS;`
I'm not entirely sure that the code I just wrote works, but the idea is that you compare to the correct data type. What you are doing is comparing a int16_t value to 0xFF. It should not work correctly.
If cs.min_arc_segments is uninitialized, then it's value would be 0xFFFF or -1. This depends on whether you compare it as a signed value or as an unsigned value.

I'm not sure if what I just said makes much sense. If you have questions, please tell me ;)

leptun: One more thing. Both arc_segments_per_sec and min_arc_segments don't really need to be int. They can be unsigned int or uint16_t and nothing will change. It might be easier for you with initialization if they were unsigned

Firmware/Marlin_main.cpp
```diff
@@ -7238,6 +7238,87 @@ SERIAL_PROTOCOLPGM("\n\n");
 
     }break;
     #endif // FWRETRACT
+    /*!
```
leptun, line 7508: @FormerLurker It's nice that you documented the gcode. This should be standard in Prusa-Firmware as it keeps everything as tidy as possible. If I'm not mistaken, there's a section that is only for titles and links to all gcodes. It should be added there as well. https://github.com/prusa3d/Prusa-Firmware/blob/43c22e7fe3f0673a93defc2a6a3c25161400be23/Firmware/Marlin_main.cpp#L3511
@3d-gussner Is the one that wrote the gcode documentation in the firmware. He should be able to tell you more about this and make sure that it is documented properly everywhere

Firmware/Marlin_main.cpp
```diff
@@ -7238,6 +7238,87 @@ SERIAL_PROTOCOLPGM("\n\n");
...
+    */
+    case 214: //!@n M214 - Set Arc Parameters (Use M500 to store in eeprom) N<MM_PER_ARC_SEGMENT> S<MIN_MM_PER_ARC_SEGMENT> R<MIN_ARC_SEGMENTS> F<ARC_SEGMENTS_PER_SEC>
+    {
+        // Extract N
+        float n = cs.mm_per_arc_segment;
+        float s = cs.min_mm_per_arc_segment;
+        int r = cs.min_arc_segments;
+        int f = cs.arc_segments_per_sec;
+        // Extract N
+        if (code_seen('N'))
```
leptun: Please avoid using N. It has special meaning when printing from USB. It might not generate a conflict, but let's stay on the safe side :)

Firmware/Marlin_main.cpp
```diff
@@ -7238,6 +7238,87 @@ SERIAL_PROTOCOLPGM("\n\n");
...
+        {
+            n = code_value();
+            if (n <= 0 || (s != 0 && n <= s))
+            {
```
leptun: No need to write a message so complex to the user. It's enough if you just break from the command processing switch early. At most tell the user that a wrong argument was provided.

Firmware/Marlin_main.cpp
FormerLurker: Whoops.  I had that in my test project, but I copied the changes into a fresh fork to keep things clean.  I will add it back.  Here is what I had:

```
//!@n M214 - Set Arc Parameters (Use M500 to store in eeprom) N<MM_PER_ARC_SEGMENT> S<MIN_MM_PER_ARC_SEGMENT> R<MIN_ARC_SEGMENTS> F<ARC_SEGMENTS_PER_SEC>
```
Good catch.  @3d-gussner, what do you think?

Firmware/Marlin_main.cpp
```diff
@@ -7238,6 +7238,87 @@ SERIAL_PROTOCOLPGM("\n\n");
...
+                SERIAL_ECHORPGM(MSG_UNKNOWN_COMMAND);
+                SERIAL_ECHO(CMDBUFFER_CURRENT_STRING);
+                SERIAL_ECHOLNPGM("\"(1)");
+                break;
+            }
+        }
+        // Extract S
+        if (code_seen('S'))
+        {
+            s = code_value();
```
leptun: Use code_value_float() for floats

Firmware/Marlin_main.cpp
FormerLurker: Absolutely fine with me.  Any suggestions as to what I should use instead of N?

leptun: You can use P, W or any other letter that doesn't have a special meaning somewhere else. I recommend that you use P

Firmware/Marlin_main.cpp
FormerLurker: Perhaps something like this:

messages.c - Add invalid argument value message
```
const char MSG_INVALID_ARGUMENT_VALUE[] PROGMEM_N1 = "Invalid argument value: "; ////
```

Marlin_main.cpp
```
                SERIAL_ECHO_START;
                SERIAL_ECHORPGM(MSG_INVALID_ARGUMENT_VALUE);
                SERIAL_ECHO("S");
                SERIAL_ECHOLNPGM("");
```

I couldn't find an existing invalid argument value message.  I could easily just break as you mentioned if this is not acceptable.
??

Firmware/motion_control.cpp
```diff
@@ -25,121 +26,140 @@
...
-      r_axis0 = -offset[axis_0]*cos_Ti + offset[axis_1]*sin_Ti;
-      r_axis1 = -offset[axis_0]*sin_Ti - offset[axis_1]*cos_Ti;
-      count = 0;
+void mc_arc(float* position, float* target, float* offset, float feed_rate, float radius, uint8_t isclockwise, uint8_t extruder)
+{
+    // Extract the position to reduce indexing at the cost of a few bytes of mem
+    float p_x = position[X_AXIS];
+    float p_y = position[Y_AXIS];
+    float p_z = position[Z_AXIS];
+    float p_e = position[E_AXIS];
```
leptun: We are quite tight on SRAM. Is it that much of a performance hit if you use `position` and `target` directly in the code? I'm pretty sure the compiler has some optimization enabled for indexing with a const value.

Firmware/Marlin_main.cpp
leptun: For one it's repeating code. Also, no other gcode command does this and I don't think it's really necessary. Breaking directly is probably good enough in this case.

Firmware/motion_control.cpp
FormerLurker: I can switch them back quite easily.  It was noticeable, but not huge.  As an added benefit, clamp to endstops could be added for each segment.  Do you think I should add that too (see the commented out code right before the call to plan_buffer_line) after I convert back to using target and position arrays?

Firmware/motion_control.cpp
```diff
@@ -25,121 +26,140 @@
...
+
+        for (i = 1; i < segments; i++) { // Increment (segments-1)
+            r_axisi = r_axis_x * sin_T + r_axis_y * cos_T;
+            r_axis_x = r_axis_x * cos_T - r_axis_y * sin_T;
+            r_axis_y = r_axisi;
+
+            // Update arc_target location
+            p_x = center_axis_x + r_axis_x;
+            p_y = center_axis_y + r_axis_y;
+            p_z += linear_per_segment;
```
leptun: There's a really big problem with this that was already present with the previous arc implementation. The Z motion doesn't account for MBL. Am I mistaken? It linearly interpolates the current->target position without compensating with the MBL map

Firmware/motion_control.cpp
leptun: It would be best if we clamped to software endstops.  Btw, how well does Crash Detection work during arcs?

Firmware/motion_control.cpp
FormerLurker: I have to be honest in that I have never tried to use the Z parameter with G2/G3.  However, is MBL not happening here:

https://github.com/prusa3d/Prusa-Firmware/blob/43c22e7fe3f0673a93defc2a6a3c25161400be23/Firmware/planner.cpp#L801-L806

??

leptun: Thanks for pointing this out. For some reason I thought that you needed to use `mesh_plan_buffer_line()` to have MBL compensation, but now I realize that all that function does is split a long G0/1 segment into shorter ones and `plan_buffer_line` does all the MBL work, which is great.

leptun:
I'll give this PR a try tomorrow if I have time. Can you please post the gcode you mentioned in the initial comment? I'd like one for MK3(S) without MMU if that is possible.

I like the idea of this PR quite a lot. It would be very interesting if arc commands were more streamlined into slicers... Oh well, guess we're stuck with the crappy STL format for now. It would be so nice if STEP files could be used in Slicers so that actual arcs in the model files could be used for generating the gcode...

FormerLurker:
Yes sir, and thank you for your review!  I will have all the changes posted to the repository by then.

Firmware/ConfigurationStore.cpp
FormerLurker: I'm still a bit confused about the floats initialization, but it looks like the code should loop through each byte of the float and see if some part of it is not equal to 0xff.  I'm assuming this needs to be done because we don't know the size of a float until the firmware is compiled?   

Anyway, things were looking a bit ugly, so I added these two functions definitions to ConfigurationStore.h:

```
bool is_setting_initialized(float setting_value);
bool is_setting_initialized(uint16_t setting_value);
```

with the following implementations.
```
// Returns true if the setting_value has been initialized, false if not.
bool is_setting_initialized(float setting_value)
{
    // Look at each byte of the arc float settings and see if any byte is initialized
    for (uint8_t j = 0; j < sizeof(float); ++j)
    {
        // If any part of the current byte is not == 0xff, the setting_value has been
        // initialized
        if (0xff != reinterpret_cast<uint8_t*>(&setting_value)[j]) return true;
    }
    // Every byte of the setting_value == 0xff, not initialized
    return false;
}

// Returns true if the setting_value has been initialized, false if not.
bool is_setting_initialized(uint16_t setting_value)
{
    return (0xFFFF != setting_value);
}
```
and I have the following in Config_RetrieveSettings():

```
        // Initialize arc interpolation settings if they are not already
        if (!is_setting_initialized(cs.mm_per_arc_segment)) cs.mm_per_arc_segment = DEFAULT_MM_PER_ARC_SEGMENT;
        if (!is_setting_initialized(cs.min_mm_per_arc_segment)) cs.min_mm_per_arc_segment = DEFAULT_MIN_MM_PER_ARC_SEGMENT;
        if (!is_setting_initialized(cs.min_arc_segments)) cs.min_arc_segments = DEFAULT_MIN_ARC_SEGMENTS;
        if (!is_setting_initialized(cs.arc_segments_per_sec)) cs.arc_segments_per_sec = DEFAULT_ARC_SEGMENTS_PER_SEC;
```

Any obvious problems with that?

Firmware/motion_control.cpp
FormerLurker: I thought the EXACT SAME THING, and in fact added mesh_plan_buffer_line before I realized the same thing you did.  The segments from mc_arc will be plenty short :)

Firmware/motion_control.cpp
FormerLurker: Alas, I only have an MK2.5, so I'm not sure (I don't think this feature is supported until MK3, but I could be wrong).  It would be particularly difficult to recover in the middle of an arc.  I believe we would have to save the arc interpolation settings and the segment number that failed in order to guarantee printing starts at the exact same place.  It is definitely doable though, at least if I understand Crash Detection properly (which I may not).

Firmware/ConfigurationStore.cpp
leptun: I don’t see any obvious flaw with the code above.
We cycle through all 4 bytes that make the float value and check if they are different than 0xFF. If all are 0xFF, then it is uninitialized.

The reason that we can’t just compare the float to some value is because, unlike an integer, a float that is all 0xFFFFFFFF is not a valid number (NaN). You can’t just compare it to 0xFFFFFFFF as that is an uint32 integer and has a different float representation.

Firmware/motion_control.cpp
leptun: The save and resume code is pretty complex. I’m quite sure that it is not handled correctly with G2/3 (and it never was, so it’s not your fault).

Firmware/ConfigurationStore.cpp
FormerLurker: Fantastic.  I just tried downgrading to the current release firmware, doing an M502 + M500 to clear settings, then installed my fork, but my old M214 settings were retained.  I guess the old values are just sort of stuck in the EEPROM.  I'm not sure how to test the initialization code now that I have values set in the EEPROM.  Thoughts?

It may be easier if you try it out tomorrow when you give this code a go.  You can just do an M503 and see what values are shown for the M214 command.  If it doesn't look like this:

```
M214 P1.00 S0.50 R20 F0
```

that means my init code is flawed.

wavexx:
Related #2606

FormerLurker:
Here is a zip file containing gcode you can use to test the PR sliced for the MK3S:  

[PRFiles.zip](https://github.com/prusa3d/Prusa-Firmware/files/4596064/PRFiles.zip)

Description of Contents:

* 6_cylinder_project/6_cylinders_6_speeds.3mf - Contains a 3mf file for the 6 cylinder test, which prints six single layer cylinders with Archemedian infill, each with a speed ranging from 10mm/s to 60mm/s.
* G2_G3_Test_Gcode folder - Contains three files that utilize G2/G3.  One is the 6 cylinder test, the other two are Benchys sliced at 0.2mm layer height.  One uses the quality profile, the other uses the speed profile.
* Original files folder - Contains the original gcode used to produce the files in the G2_G3_Test_Gcode folder.
* 6_cylinders_showing_print_speeds.jpg - An image captured from Simplify3D's gcode analysis feature showing the print speeds.  The 10mm/s cylinder will be towards the back left and the 60mm/s cylinder is towards the front right.
* ArcConversionInfo.jpg - Some interesting statistics regarding the conversion from the original files to those using G2/G3.  I thought you might like to see this.

If you would like to try other models, or if you want me to convert other gocde files, please let me know.  The full source for the Arc Welder conversion library, as well as a console application that can be used to produce the G2/G3 code from the command line are available in the [ArcWelderLib](https://github.com/FormerLurker/ArcWelderLib) repository that you can find in my github pages.  See the ArcWelderConsole project for the command line version.  There is also an InverseProcessor project in there that can be used to output gcode that shows the actual line segments that will be put into the planner (though it now needs some slight modification due to the recent modifications I made for this pull request).  The project was created using Visual Studio, and some alterations will need to be made if you want to compile it from source.  I could send windows binaries if you want. 
 
You might also like to try the [ArcWelderPlugin](https://github.com/FormerLurker/ArcWelderPlugin) for OctoPrint.  It is in pre-release status ATM, but you can install from the OctoPrint plugin manager by clicking *Get More* and pasting this url into the *From url...* box:  ```https://github.com/FormerLurker/ArcWelderPlugin/archive/0.1.0rc1.dev2.zip```

Please let me know if I can do anything else to help you test, and thank you for spending your valuable time on this PR!

FormerLurker:
First, congratulations on getting the new release out!

Second, I wanted to mention that I will be implementing some enhancements from [this pull request](https://github.com/MarlinFirmware/Marlin/pull/18144) that may bring the small angle approximation back into the picture.  It turns out adding another term to the approximation makes a HUGE difference and has a relatively small cost.

ochm:
Hi all, do you plan add this feature to firmware, please?

arne182:
I am also in support of this PR

aguschanchu:
Please add this feature to the firmware!

marchetb:
Add this please 🙏  👍

XRyu:
Thumbs up for this! Please bring this to the firmware 👍

FormerLurker:
Working on some changes to bring this more in-line with Marlin 2.  Re-adding N_ARC_CORRECTION and the small angle approximation with an additional term since it seems to be plenty accurate after that.  Should speed up getting the first segment into the planner.  Will also update M214 and M500 to reflect this.  FYI, I'm also going to try to get M214 into marlin 2 where it could make it's way eventually (hopefully) into the Buddy firmware.

frango1963:
Please add this feature to the firmware!

FormerLurker:
Fyi, this has been slow going, but it should be ready in the next month.

Cjkeenan:
I hope this goes through in the near future as I have been having issues with my Pi Zero W plugged into the back of my mk3s+ and small circular details in which the nozzle seems to just smudge the entire detail leaving a mess, ex. very small half spheres looking more like smudged cylinders.

It only happens when printing from OctoPrint even when the exact same GCode file is used from the SD card. One thing to note, the smudging seems to be a little less noticeable when using OctoPrint safe-mode but still not as good as the local SD card. 

My best guess now is that there is micro-stuttering happening due to the large quantity of G-codes associated with curves and the less than optimal specs of the Zero W. I was hoping this PR along with  your ArcWelder plugin may help to alleviate that additional overhead and this would be a good addition to go along with the official Prusa Zero W guides as a sort of optimization.

FormerLurker:
I've committed some recent and substantial changes, but have noticed one issue at the last minute.  I will fix this, test, and amend this pull request ASAP, so no need to review this yet.  Thank you.

FormerLurker:
So, I added one line variable to the ConfigurationStore, a uint8_t, and now the compilation is failing.  I noticed in the build details it says it is using 100% of the available memory, but I'm getting 93% using the .h file indicated in the error log.  Any ideas as to what is going on?

Edit:  The .h file that failed was 1_75mm_MK3-EINSy10a-E3Dv6full.h.

FormerLurker:
Here is what sketch claims to use for me when building for the MK3.h file:

image

leptun:
You are building without multi language support. In reality, the code size is dangerously close to 100%

3d-gussner:
@FormerLurker The firmware source code is for MK3, MK3S/+, MK2.5 and MK2.5S/+ (MK2, MK2S), so you have to pay attention to your code/PR that it will compile correctly for all variants.

The Travis CI test compiles all variants and checks if they succeed:
- The MK3S build passes see https://travis-ci.com/github/prusa3d/Prusa-Firmware/builds/212922781#L512 and https://travis-ci.com/github/prusa3d/Prusa-Firmware/builds/212922781#L576-L577
- The **MK3** build **fails** as it is using for resources than available see https://travis-ci.com/github/prusa3d/Prusa-Firmware/builds/212922781#L654 and https://travis-ci.com/github/prusa3d/Prusa-Firmware/builds/212922781#L702-L707
- The MK2.5S build passes see https://travis-ci.com/github/prusa3d/Prusa-Firmware/builds/212922781#L711 and https://travis-ci.com/github/prusa3d/Prusa-Firmware/builds/212922781#L755-L756
- The MK2.5 build passes see https://travis-ci.com/github/prusa3d/Prusa-Firmware/builds/212922781#L955 and https://travis-ci.com/github/prusa3d/Prusa-Firmware/builds/212922781#L1002-L1003

So rule of thumb is to compile the firmware for all variants upfront before pushing a PR.

The MK3 build contains the PAT9125 filament sensor which is more complex and so needs more resources compared to MK3S/+.
Usually the MK2.5S+, MK2.5, MK2 and MK2S are way smaller as these don't have the TMC2130 drivers and use less resources compared to the MK3 and MK3S+.

Variants by recourse usage (high to low):
- MK3 TMC2130 + PAT9125
- MK3S/+ TMC2130
- MK2.5 PAT9125
- MK2.5S

FormerLurker:
Thank you @leptun and @3d-gussner, I will work to reduce the LOC.  It's right on the edge.  I think the optimizer was omitting some code earlier due to an unreachable code block (code error due to a last minute change).  I can def shave off a bit.  Will report back.

FormerLurker:
OK, got the memory in line, just barely, but I've found a bug that exists still in the latest version of Marlin 2.  The problem is that there is a remainder when calculating the z and e travel per segment.  It is assumed that Z and E each travel the same amount per segment, but if there is a remainder, that needs to be subtracted from the calculations at the start of the interpolation loop.  I will work on this, and will hopefully be able to keep the memory in check.  I do have some ideas for how to achieve this, the simplest being to spread the difference out across all the segments.

FormerLurker:
Ok, I did extensive testing and learned that there was actually no problem with the final segment.  My flow rate calculations should have been ignoring extremely small movements where either the linear distance or extrusion amount was very close to zero.

Anyway, I've cleaned the comments up a bit and made one very minor modification.  I believe this is now ready for review.

sslupsky:
It's been suggested earlier in this thread (@leptun), perhaps this feature should be integrated into slicer?  (Actually, I am quite surprised to learn this is not already done within slicer).

FormerLurker:
@sslupsky, regarding this comment:

> It's been suggested earlier in this thread (@leptun), perhaps this feature should be integrated into slicer? (Actually, I am quite surprised to learn this is not already done within slicer).

Adding arc support to the slicer is somewhat different than this pull request, though.  The point of this PR is to improve the printer's ability to print arcs (the current version does not produce accurate geometry, and will stall terribly if the mm_per_arc_segment is reduced to compensate), and to control the arc generation settings via GCode, which can be extremely useful, say by increasing the resolution around exterior perimeters where the speed is lower, but keeping it high for gyroid infill and the like.

Recently I've been running lots of firmware tests on both the release version, a version with MeatPacker support, and my fork, but when I print code with G2/G3, I really must switch to my own fork to achieve the quality I would expect.  Adding G2/G3 to the slicer will have to come after compatibility with the printer's firmware is straightened out, IMHO.  If you'd like to try out ArcWelder with PrusaSlicer and help test the firmware modifications, please comment within the [ArcWelderLib repository discussions](https://github.com/FormerLurker/ArcWelderLib/discussions) to keep this PR clean, thanks!

FormerLurker:
I just noticed there are new conflicts.  I will attempt to resolve these ASAP.

Cjkeenan:
> a version with MeatPacker support

Is this firmware version public for testing? I want to see if it helps with print issues using a ZeroW.

mmirate:
I took a stab at fixing the merge conflicts; the result is [a new PR pointing to this PR's branch](https://github.com/FormerLurker/Prusa-Firmware/pull/1).

mmirate:
So this now seems mergeable again; modulo some automerge commits which y'all will probably want squashed-together (after using them for any necessary merge troubleshooting); and modulo the inclusion of contents from the `MK3_3.10.0` branch (which, given the EEPROM changes that have apparently been happening during the RCs, might be a good idea? or not? idk).

DRracer:
@FormerLurker @mmirate we are considering this nice PR for evaluation in the next FW release (3.11)

mmirate:
More merge conflicts means [more fixes for the merge conflicts](https://github.com/FormerLurker/Prusa-Firmware/issues/2).

FormerLurker:
Thanks @mmirate!!!  I appreciate the PR.  Merged.

Have-Purple:
I'd like to see this in production.

mmirate:
> I'd like to see this in production.

That certainly cannot happen until these latest merge conflicts are resolved.

And, I'm sorry to spam the main thread again, but the easiest way for that is for @FormerLurker to merge my months-old PR to their branch: https://github.com/FormerLurker/Prusa-Firmware/pull/3.

mmirate:
Okay, after a minor little force-push to fix an accidental squash-merge, this PR is once more free of merge-conflicts.

mtaumike:
So is this coming or is it still sitting out there?

stephenjamieson:
Would love to see this in the next release.

Katherine1:
Same, I'm really hoping this gets approved before it needs to be rebased again.

joshuacant:
I'd love to see all the hard work in here get merged.

veenified:
@leptun A lot of folks really looking forward to this feature. Any hopes of finding time to review this PR before it needs to be rebased again?

GhostlyCrowd:
+1 I've been running these changes as well as meatpack serial compression changes together for months now, working great.

mtaumike:
Is there a status update from Prusa folks on this?

stephenjamieson:
Unfortunate this isn't getting the attention it deserves.

sigboe:
I hope this gets merged soon

3d-gussner:
@FormerLurker Sorry it took so long to get this into the official firmware. As we are developing and merging for FW3.12.0 I would like to ask you to re-base a final time  so we can review and merge it soon. We got some additional free resources due to change of build environment and improved code, so the chances are very high that this PR will make it into FW3.12.0

leptun:
Since this PR is old (sorry for taking this long), there seem to be a lot of commits which shouldn't be there. It might be necessary to rebase the changes manually onto the latest MK3 branch.

mmirate:
FYI @FormerLurker I am more than happy to do the rebase again, but among other things I'm at a conference this weekend, so this will happen next Wednesday at the earliest.

FormerLurker:
@leptun, this is great news, thank you!  No problem regarding the timeframe.  I understand how these things go.

@mmirate, that is very kind of you to offer, and I'll take you up on it.  I've been totally buried here recently, and probably wouldn't get around to this for a couple weeks, so anything you can do to speed the process along would be very very much appreciated.

FormerLurker:
@mmirate, I will have free time to deal with this in the next week or so.  Let me know if you're still interested in rebasing the fork?  If not, I'll try to do it manually.  Thanks!

mmirate:
@FormerLurker I'm absolutely interested, but I've been way busier than expected recently. ☹️ I'm terribly sorry. Please go ahead with your attempt.

FormerLurker:
@mmirate,

> I'm terribly sorry.

You've contributed so much to this you have NOTHING to be sorry about!  I hope things cool down for you soon!

I'll work on this during the next week.

FormerLurker:
I have finished the rebase locally, but had to make a few changes to the configuration store and data types to bring it in line with changes to the origin.  It will take at least another week for me to thoroughly test the changes because I'm going out of town.  You can expect this PR to be updated by the week of the 24th of April at the earliest, and the weekend after that at the latest.  Thanks!

FormerLurker:
I tried to bring my code changes within ConfigurationStore.cpp in line with the existing style and to use the new is_uninitialized function.  Please take a look at that to ensure the defaults are properly initialized.  I did test this, but there were likely already values in my EEPROM, so it's difficult to tell if it's working properly.

I tested M214, fetching default values, storing new values into EEPROM, and several prints using arc interpolation (including using the Z parameter).

Thanks!

Firmware/ConfigurationStore.cpp
```diff
@@ -273,65 +282,72 @@ bool Config_RetrieveSettings()
                 memcpy_P(&cs.max_acceleration_units_per_sq_second_silent[i],&default_conf.max_acceleration_units_per_sq_second_silent[i],sizeof(cs.max_acceleration_units_per_sq_second_silent[i]));
             }
         }
+        // Initialize arc interpolation settings if they are not already (Not sure about this bit, please review)
```
leptun: I tested this PR on a printer with a fresh eeprom (with values already loaded from a previous stock fw version). The values loaded correctly (checked via M503)

Firmware/Marlin_main.cpp
```diff
@@ -7501,9 +7503,46 @@ SERIAL_PROTOCOLPGM("\n\n");
       }
 
     }break;
-#endif // FWRETRACT
+    #endif // FWRETRACT
+    /*!
+    ### M214 - Set Arc configuration values (Use M500 to store in eeprom)
```
leptun, line 7509: todo:
@3d-gussner We need to update the reprap gcode wiki with this gcode

Firmware/Marlin_main.cpp
```diff
@@ -7501,9 +7503,46 @@ SERIAL_PROTOCOLPGM("\n\n");
...
+        float p = code_seen('P') ? code_value_float() : cs.mm_per_arc_segment;
+        float s = code_seen('S') ? code_value_float() : cs.min_mm_per_arc_segment;
+        unsigned char n = code_seen('N') ? code_value() : cs.n_arc_correction;
+        unsigned short r = code_seen('R') ? code_value() : cs.min_arc_segments;
+        unsigned short f = code_seen('F') ? code_value() : cs.arc_segments_per_sec;
+
+        // Ensure mm_per_arc_segment is greater than 0, and that min_mm_per_arc_segment is sero or greater than or equal to mm_per_arc_segment
+        if (p <=0 || s < 0 || p < s)
+        {
+            // Should we display some error here?
```
leptun, line 7537: We could add an error message, but in any case, the user can check afterwards with M503 that the values loaded successfully. The other M20x commands don't have any checks at all, so I think this is fine as it is now.

leptun (approved):
I am aproving the PR.
Everything M214 related seems to work properly. I also played a bit with G2 and it worked correctly

Firmware/Marlin_main.cpp
FormerLurker: Agreed.  I would have added one, but there was a memory crunch at the time.

Firmware/ConfigurationStore.cpp
FormerLurker: Thanks, I appreciate it!  It seemed to be working for me, but better safe than sorry.

leptun:
@FormerLurker I found a but with the G2/3 commands. Turns out hard stopping an arc move is not handled properly and the print head starts doing funny movements in the process and even losing the current position. See the following script:
[arcs.zip](https://github.com/prusa3d/Prusa-Firmware/files/8592800/arcs.zip)
https://photos.app.goo.gl/duduGsUVGoHSdT7k8

Now, I have a fix for the movements continuing after pause and stop, but there is another problem. The last instruction is executed again on resume. Because G2/3 uses I and J as relative offsets, the offsets are applied to the resume position, not to the start position of the arc. Do you know any way to work around this issue? Just thinking out loud, if R was specified instead of I and J, the problem would be gone.

Also, can I rebse the branch again on the latest MK3 branch and add the first fix I mentioned?

FormerLurker:
Was this issue introduced by this PR, or did it exist previously?  I suspect this has been an issue all along.

Second, regarding the repeated command, I'm assuming the current position (from the point of failure) is being sent to prepare_arc_move (Marlin_main.cpp line 9694) along with the destination position, right?  I actually thought about recovery at some point, since the printer could be in the middle of an interpolation of a single G2/G3 when paused/during recovery.  

I can think of the following two solutions:

1. For pausing, we could just let the current command complete, I suppose.  This will not help if power is lost, unfortunately.
2. A better, but more challenging solution would be to send and store the index of the current interpolation as well as the starting point of the arc (the previous value for current_position) somewhere, and add an index parameter to mc_arc.  Index = 0 means we would not skip any motions.  As long as we can store the original starting position of the command, I think it could be made to resume appropriately by not calling plan_buffer_line for the *Index* iterations (including the final move to the destination).  During a power failure, I'm not sure if there would be enough time to save these two additional values, but maybe I'm wrong?  It would be an int and 3 floats.

I haven't tried your code yet, but I will do that here shortly.  It may be that I don't understand the problem 100% yet.

Feel free to add whatever you need to the code, or I could add it to my branch if that's more convenient for you.  Please let me know.

leptun:
@FormerLurker I'll add that quick pause/stop glitch fix to this PR. We can fix the resume relative positions in a follow-up PR. I also think this was broken all along (it's not like anyone was using G2/3 before arcwelder). I plan to merge this PR on Monday, but the pause/resume relative bug really needs to be fixed. Otherwise a wrong move will be queued on resume if the pause event is timed badly. The pause events can come from:
- lcd
- gcode
- tmc drivers
- filament sensor
- power panic
- fan error
So in theory all of the above are broken during a G2/3 at the moment, which means the bug definitely needs to be fixed. Using the R argument is just a workaround since the radius can be applied anywhere while on an arc.

I think for this bug fix I'll wait for @wavexx to be available again since he is the one who fixed relative motion in G1 commands. He probably knows best how to approach this issue.

FormerLurker:
I appreciate that.  I need to work through the code you mentioned so I understand how exactly pause/resume is working.  I agree this is worth fixing, and will try to assist as much as possible.

wavexx:
When a pause, crash or power panic is triggered, we save the current position in the gcode stream, the current position of the printer (at the time of the crash), and the destination for the move (``saved_target``).

When we recover, the head is moved back to the position where it has been stopped, and the old gcode is re-issued. In this case, ``destination`` is replaced by ``saved_target`` (since "destination" cannot be easily reconstructed from the current instructions, it can be a relative move). The current position and the destination include all axes (extruder included), so for a linear move it's actually trivial: we just replan a brand new segment, which in turn may be segmented again.

Why this is done this way? It's because actually pretty conservative in terms of buffers required. G1 similarly splits the line internally for MBL. We need to be able to stop anywhere in any of the segments generated, and recover at any point. Also important, we need to be able to interrupt and recover a move which has also _previously_ been interrupted (let's say a G1 command paused, then resumed, then paused again). And we need to consider that the interruption is done asynchronously from the main loop: the global "destination" will be referring to a different segment, so we need to save such values directly in the block buffer, which is the largest buffer we have in memory.

This is a little bit trickier for the arc move. Saving extra parameters is not an issue, as long as they are not in the block buffer where they blow up easily.

A better approach would be to change how we save the current state. Instead of trying to resume the current gcode instruction at a higher level, we can save the position index of the segment generated by the move - so one layer below. It does require many more changes to the pause/resume machinery though, which would be nice to avoid (at least at this stage if we want to get this into 3.12).

wavexx:
To continue, a good question would be what would be the absolute minimum required to be able to resume an interrupted arc move. Assuming we have moved from the original starting position, should we record the X/Y offsets (appropriately shifted at the time of the pause)? Can we avoid this entirely somehow and or save something else internally that allows us to continue the previous arc using less memory?

FormerLurker:
One option would be to store the original starting position at the beginning of g2/g3, and the segment index, which i can save to some global during interpolation.  Then we can continue the arc, skip the appropriate number of segments, and pick right back up where it left off, without redrawing any of the segments.  This assumes the original gcode will be reissued, adding the index as a parmeter.

This would require an extra G1 to be sent during resume to move to either the original starting position, or to the current position when pause was issued.  The latter is preferred, since the Z position may change throughout the G2/G3.  The former would require motion, but no extrusion, until the appropriate segment index is hit.

What do you think?  Am I missing anything critical?

wavexx:
The problem with storing the starting point is just memory at the moment. This requires 4 floats for each block + the index (resulting in 272 bytes required).  

That would be fine if we redo also G1 so that we don't need the end position anymore.

I was hoping for something more compact. Can we calculate the radius from the supplied XY offset instead? That would remain constant during recovery. With this we could probably reduce the requirement to 2 floats per block: radius + extrusion remaining.

Firmware/Configuration_adv.h
```diff
@@ -289,9 +289,7 @@
   //#define LA_DEBUG_LOGIC     // @wavexx: setup logic channels for isr debugging
 #endif
 
-// Arc interpretation settings:
-#define MM_PER_ARC_SEGMENT 1
-#define N_ARC_CORRECTION 25
+// Arc interpretation settings : Moded to printer default settings (Configuration_prusa.h)
```
wavexx: I would remove this comment entirely (there's also a typo: "moded" -> "moved")

FormerLurker:
Makes sense.  Let me think about it a bit more.  We do have the advantage of being able to do some more complicated math for the recovery than during a normal command.

FormerLurker:
Actually storing the radius and extrusion remaining may just do the trick.  I don't think we can calculate the radius from the offset since we don't know the starting point.

Firmware/motion_control.cpp
```diff
@@ -25,121 +26,137 @@
...
+
+            // Update Position
+            position[X_AXIS] = center_axis_x + r_axis_x;
+            position[Y_AXIS] = center_axis_y + r_axis_y;
+            position[Z_AXIS] += linear_per_segment;
+            position[E_AXIS] += segment_extruder_travel;
+            // Clamp to the calculated position.
+            clamp_to_software_endstops(position);
+            // Insert the segment into the buffer
+            plan_buffer_line(position[X_AXIS], position[Y_AXIS], position[Z_AXIS], position[E_AXIS], feed_rate, extruder, position);
```
wavexx, line 152: Depending on what we plan to do during recovery, the last argument to plan_buffer_line here the 4 floats are up for "abuse". In G1 this would be "target", since the final coordinate is the invariant during the move.

What we do for for arcs can be different.

wavexx (approved):

wavexx:
I approved this, I think we can work on it as we go :)

Firmware/motion_control.cpp
leptun: It can be different, but we still need to store XYZE since all of these could be relative. I just find it annoying that I and J is always relative as well. The only purpose of gcode_target is to fix issues with relative gcode commands

wavexx:
The radius can be calculated only when the first (non interrupted) move is scheduled. We always know when we're working on a recovered move, so we can just re-use the old radius in that case.

leptun:
Well, considering we have 2 approvals, I'm going to merge this now and work from there.
As a side note, this PR also greatly improves the quality of prints which use G2/3 commands:
image
image
(your fw being the current state of the branch with aw gcode)

Firmware/motion_control.cpp
wavexx: Depending on how we see this, 4 extra floats will always be required somehow for either the (absolute) start or the end of the move because of this :/

leptun:
@wavexx 
> The radius can be calculated only when the first (non interrupted) move is scheduled. We always know when we're working on a recovered move, so we can just re-use the old radius in that case.

We don't store the radius globally. Only the offsets from the start position. Also, these get overridden by the next G2/3 command, so I'd be careful with those. The XYZE cached values also get overridden by any G1 command since the global cache for XYZE is common between G0/1/2/3

wavexx:
> We don't store the radius globally. Only the offsets from the start position. Also, these get overridden by the next G2/3 command, so I'd be careful with those. The XYZE cached values also get overridden by any G1 command since the global cache for XYZE is common between G0/1/2/3

By this I mean store the radius just for the move. This would need to be stored for each block for this reason.
When stopping, extract the radius from the block, and supply that radius again as the first G2 during recovery.

Yes, it's ugly :]

leptun:
Using just the radius will create inconsistencies in the print. Consider the following scenario: If the move is stopped at the middle of a G2 segment, the radius should be actually smaller than if it was at a segment end (exactly the radius). This is probably good enough for high precision segments (max 1mm), but the lower resolution you go (>1mm), the worse the offset will get after resume. Also, the restarted move will not aligh with the segment split of the previous move. I think this will most likely result in some kind of visual artifact.

Firmware/motion_control.cpp
leptun: Can you explain how it's done differently in marlin 2.0 that could be used here (in order to save on resources)?

Firmware/Configuration_adv.h
leptun: I'll wait for this to be resolved before merging

FormerLurker:
> If the move is stopped at the middle of a G2 segment, the radius should be actually smaller than if it was at a segment end (exactly the radius)

I'm not sure this is true.  Typically you would draw an arc on paper by calculating the midpoint of the circle (move from starting point by the relative offset), and drawing from the current position to the target position.  However, If you flip your compass around (pointy end on the current position instead of the center of the circle), and draw a circle of a fixed radius, it will always intersect in the same two places as long as the pointy end is on the arc.  The correct intersection depends on the direction.  Here's a crude sketch I made with a screwdriver, a pen, and a piece of foam, so it's not great, but accurate enough to get the idea:

PXL_20220430_182122528

From that, you can reconstruct the offset by finding the intersections of two circles, selecting the correct intersection (not sure how to do that just yet), and just figuring the X and Y coordinates of the vector from the starting point to the intersection.  This assumes we will know the current absolute position, and that it is on the arc somewhere.

I'll try to create a program that will, given a G2/G3 command and a point along the arc, will attempt to create a new G2/G3 command to complete the arc with only the radius, the extrusion remaining, and the target coordinates (do we know the final destination of the G2/G3, or just the final destination of the interrupted segment?)

leptun:
I'm not sure I understand the explanation. I'll try to explain my part a bit more:
image

Consider the following sequence:
```
G1 X0 Y0 ; start at the origin
G2 X30 Y30 I30 J0
```
In the picture above, the resolution of the arc was intentionally dropped to highlight the issue.
The problem I see is that the hard stop event can happen at any point during the arc. In the picture you have the perfect arc (dashed line) and the actual interpolated arc by G2 (normal line). Since the stop event can happen at any moment, the "current_position" at the hard stop position can be anywhere along the interpolated segments, not on the perfect arc. This can cause problems since the distance to the center of the arc isn't technically constant (it's the lowest at the midpoint of every interpolated segment). On resume, the arc is restarted, but the starting position will most likely be wrong because the arc would actually start a bit closer to the center in the midpoint case. If the pause instead happens exactly at the beginning/end of an interpolated segment, then the arc after resume will perfectly match the one before the resume. I'm not sure how this can be avoided besides not doing a hard stop in the middle of a segment.

If we use start position, end position and radius for the second arc, this is what it would look like (blue line):
image

leptun:
Also, something I didn't consider at all is the handling of broken arc commands. What I mean by this is when the center offsets create invalid geometry. These arcs can be identified by the fact they draw an arc, and then do a straight last move to the target position. I'm 100% sure we can't replicate their behaviour on resume (not that we would want to do that, but still something to consider).

DRracer (approved):
Approving as well - but the pause/resume feature must be resolved in the future to make this reliably for all scenarios.

FormerLurker:
First, THANK YOU SO MUCH for merging this in!  I'm so thankful for all of your time and effort!

Next, I'm not sure where to carry on the discussion about pause/resume.  Should I create an issue where we can work on a solution?

leptun:
Yes, please create an issue about this where we can dump all the info we have so far (and ideas)

3d-gussner:
@FormerLurker Thanks again for this PR. Finally it got merged. :partying_face: 
Are you familiar with https://reprap.org/wiki/G-code and editing it?
As you have introduced the `M214` g-code we should also add this to the wiki.
Please let me know if you are doing this or if I should do it.

FormerLurker:
@3d-gussner, thank all of YOU for your time and support!  I'm familiar with the wiki (I have used it EXTENSIVELY over the years), but I've never edited anything.  I will take a look as soon as I have some time, and will let you know how it goes.

arekm:
Usage question.

PrusaSlicer doesn't do G2/G3 for arcs, right?

Based on https://github.com/prusa3d/PrusaSlicer/issues/4352 the only way to play with this is ArcWelder plugin, right? (and no solution to wire that into prusaslicer)

Edit: never mind, this explains everything
https://github.com/FormerLurker/ArcWelderLib#running-arcwelder-from-slic3r-slic3rpe-prusaslicer-and-superslicer

wavexx:
AFAIK it hasn't been integrated in prusaslicer yet because post-processing the gcode as arc-welder does can result in non-coherent segments across layers. This can result in different printing behavior from multiple fronts that the slicer cannot control, so it makes reported slicing issues much harder to reason about.

It's ironic that it was merged into a 32bit FW, where I would consider it's primary feature (the resulting g-code space savings) not so important.

FormerLurker:
@wavexx,

>AFAIK it hasn't been integrated in prusaslicer yet because post-processing the gcode as arc-welder does can result in non-coherent segments across layers.

FYI, I have heard this before from several people, but have never seen an example of this with any reasonable resolution setting.  Also, the resolution could be set by the slicer to prevent this.  Every time this 'issue' has been reported to me, it's been a visualization error within some slicer or gcode viewer.  I can think of many reasons why this feature isn't added to PrusaSlicer, but I don't think the above is a good one.

> It's ironic that it was merged into a 32bit FW

The MK3 firmware is 32 bit?  What about the MK2?  I ran this fork on my MK2 a bunch for testing.

> the resulting g-code space savings

For me, the main feature is reduced serial transfer so I can stop hearing about 'zits' being created when using OctoPrint :)  It was a rare issue, but annoying, and created a lot of github issues for Octolapse.

leptun:
@FormerLurker 
Clarification from @wavexx (since I was also confused):
> @leptun: What was merged into the 32 bit fw?
> @wavexx: Not in the 32bit but for the 32bit. I was referring to the fact that the bambulab slicer (rebadged prusaslicer) includes arc-welder, but the printer doesn't really need it (I actually think it's counter-productive if you don't need the serial speed)

The MK3 is still the 8 bit firmware. Only Mini and newer are the 32 bit firmware (buddy firmware). The MK2 is a separate fork of the 8 bit firmware, but it is not being actively maintained.

johnnynoone:
@3d-gussner
**M214** is still not described on [RepRap wiki](https://reprap.org/wiki/G-code). If you can, please add it. Thank you in advance. :)

3d-gussner:
@johnnynoone Thanks for the reminder. Done https://reprap.org/wiki/G-code#M214:_Set_Arc_configuration_values @FormerLurker Is that okay so?

PR #2665, opened 2020-05-14
Prevent successful compiling when lang_en_xx.txt aren't correct

3d-gussner:
Prevent successful compiling when the message count and size aren't correct.

Users or scripts may create/modify the `lang_en_??.txt` files in a way that the LCD output isn't correct.

Added some extra checks:
- The message count of the `lang_en.txt` will be stored temporally in `lang_en.cnt`
  - All other translations compare their message count against the English message count.
    - If the xx count is equal to EN count the script displays the information and continues to generate the lang_xx files needed for multi language support.
	- If the xx count is NOT equal to EN count the script displays the information and stops
- The maxsize for 2nd language will be extracted from `../Firmware/config.h` and will be stored temporally in `lang_en.max`
  - All other translations compare their size against the maxsize.
    - If the xx size is less than the maxsize the script displays the information and continues to generate the lang_xx files needed for multi language support.
	- If the xx size is greater or equal than the maxsize the script displays the information and stops

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,8 @@ Firmware/Doc
 /lang/progmem1.var
 /lang/text.sym
 /lang/textaddr.txt
+/lang/lang_en*.cnt
+/lang/lang_en*.max
 /build-env/
 /Firmware/Firmware.vcxproj
 /Firmware/Configuration_prusa_bckp.h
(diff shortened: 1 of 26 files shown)


3d-gussner:
Examples:
- successful build: https://travis-ci.com/github/prusa3d/Prusa-Firmware/builds/168383924
- failed build: https://travis-ci.com/github/prusa3d/Prusa-Firmware/builds/165986696#L619-L623

3d-gussner:
Close as included in https://github.com/prusa3d/Prusa-Firmware/pull/3212

Issue #2238, opened 2019-09-26
first layer calibration for different nozzle sizes


ligniant:
I do have the following feature request:

Now it is possible to set the nozzle size in the firmware and you are implementing first layer calibration for different materials. It would be great if the built in first layer calibration could be valid for different nozzle diameters as well. Until now it is only usable with the 0.4 nozzle. Now it could be possible, that the printer is performing individual gcodes depending on the configured nozzle.

Kachidoki2807:
Duplicate of #2177

midnightsmith89:
OK, so this has been "scheduled" since 2019, whats up?  Had to search this tonight when trying to set up my 0.6 nozzle that is pre setup in the slicer! C'mon prusa!

Issue #2177, opened 2019-09-07
New firmware update Nozzle diameter setup


hutje:
> Nozzle diameter setup
This an initial release of this function. We need to finish the coding on the PrusaSlicer side in order to fully enable it. By default, all Original Prusa i3 printers are shipped with the 0.4mm nozzle. In case you change the nozzle, you can set the correct diameter in the printer's menu (see Settings -> HW setup -> Nozzle d.). The printer will then compare the stored value of the nozzle diameter with parametres in the G-code. If there is a mismatch, you will be informed to prevent a print failure.

Will this feature also correct the first layer test, where the amount of filament is being calculated when using 0.2mm, 0.6mm or 0.8mm?

3d-gussner:
@hutje As far i can see it will not compensate the extra/less extrusion. If you activate it it will warn you or even prevent the print depending on your settings.

I think the best is to make your own 1st Layer calibration gcode and store it on the SD card if you need another filament type or nozzle size.

Update: 07 sep 2019 20:00 GMT+1
Just add G87 at the end of the gcode to let the printer know that 1st layer cal have been done.

leptun:
This could be generated on the fly though. Might be worth looking into in the future.

3d-gussner:
@leptun As this 1st layer gcode uses some space, I am not a big fan of it. And making the code even bigger to support different nozzle size and filament types doesn't make lot of sense to me.

leptun:
I think it is already generated on the fly with formulas. Pretty sure it can be easily implemented to support a different variable for nozzle diameter. If @DRracer approves this idea, I’ll look into it.

DRracer:
This feature doesn't perform any corrections in the G-code - it is here to warn the user, that the G-code was sliced for a different nozzle diameter.
Implementing 1st layer calibration for other diameters of nozzles is an interesting idea. Of course it is only about changing material flow through the nozzle...

Kachidoki2807:
To avoid opening a new issue, since the title is generic enough, I will expose my issue here.
On one of my machines, I use a 0.80mm nozzle.
So this feature is useless to me and I have to disable the warning.
I would like to address two points:

**First:**
As far as I read in the code, the selected value is stored in EEPROM as an integer (250, 400 or 600). And it is even stored once more (but why?) as an enum (25, 40, 60).
[...]

PR #2682, opened 2020-05-24
Nozzle diameter compensation in first layer calibration

GilesBathgate:
This is a potential fix for  #2177 (and #2238 ) Completely untested, and not sure if integer math is the correct approach, but perhaps this serves as a conversation starter.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -1131,6 +1131,7 @@ void lcd_commands()
 		float length = 20 - width;
 		float extr = count_e(0.2, width, length);
 		float extr_short_segment = count_e(0.2, width, width);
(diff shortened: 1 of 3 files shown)


GilesBathgate:
@DRracer Is anyone willing to discuss this suggestion. Perhaps an approach simply of setting the extrusion multiplier would be preferable.

digitalentropy:
This is a really important feature for handling different (especially larger) nozzle diameters. I am really surprised it's been ignored so long.

GilesBathgate:
@mkbel It was definitely mentioned in the PRUSA LIVE #35 live chat. Perhaps someone else is working on this in which case it can be closed, otherwise, should I clean up the PR?

PR #2688, opened 2020-05-25
[WIP] Avoid strings in RAM if possible

DRracer:
- remove RAM located strings from reset_menu
- optimize calls printing one character
Saves 45B of RAM (!) and 60B of FLASH.

PFW-1123


DRracer:
closing in favour of a more recent PR #2982

PR #2690, opened 2020-05-25
Update fastio

GilesBathgate:
This just updates fastio.h from the latest marlin firmware. I did this primarily to get the `SET_INPUT_PULLUP` macro, but also reduces divergence.


wavexx:
Do you know if there have been significant changes and/or improvements to look out for?
I'm also interested in this.

GilesBathgate:
Aside from SET_INPUT_PULLUP, which I thought was useful, the main reason for doing this was to pull some of the codebase up to date with the marlin firmware. I think that should be possible and a good thing to do for many other areas. This was just an initial experiment, as to whether that is at all possible.

DRracer:
@GilesBathgate Thank you for your PR. Indeed, it is possible to update the fastio and macros from the latest Marlin2.

However, we don't see any significant improvement of doing so for the quite matured 8bit Prusa-Firmware for the Original Prusa MK3/S. Moreover, without going through these changes line-by-line and testing each of them, we can pull in some really-hard-to-debug errors, especially with the macro definitions whose error and side effect may not be caught at compile time.

We'd prefer picking only the absolutely necessary stuff if needed for some further work.

Similar experiments in the Prusa-Firmware-Buddy repo so far always ended up with bugs which we had to chase for several weeks, which is something to be avoided.

GilesBathgate:
@DRracer Shall I change the PR to only include `SET_INPUT_PULLUP` which is quite useful?

DRracer:
@GilesBathgate possibly, please get in touch with @leptun , who would like to use some of the macros as well for his work. Ideally, make one new macros.h only with the necessary stuff.

leptun:
@GilesBathgate I want to remake this PR: https://github.com/prusa3d/Prusa-Firmware/pull/2786
I was told that I shouldn't port all of marlin's macros.h and only copy what is essential for me.
For my PR I need `TERN_`, `ENABLED` and `DISABLED`. If others will be needed, they can be added later on.
I tried only implementing what I needed, but these macros just branch way too much for me to understand and I ended up not being able to compile. As a compromise I tried pulling in all of macros.h and also taking some other scattered macros from our firmware and adding them to macros.h, but that was still too much even though it worked.

I also found an additional scattered macro here: You can include it in your new macros.h as well.
Let me know if I can do anything for you that might make this process simpler. I'll remove my macros.h from my PR so that there aren't any merge conflicts.

wavexx:
Looking back at this, it seems that the new definition of _WRITE_C is using the same trick used by _TOGGLE. In our source we have a CRITICAL_SECTION there.

I checked the history of the fastio source we currently have but I couldn't pin-point the origin of the critical section. We need to ensure the code doesn't happen to rely on that somewhere.

Overall moving the various avr variants into each file is also much nicer. We would need to verify that macros for fastio_1280 haven't been changed for some reason, but otherwise this would be nice to include.

PR #2692, opened 2020-05-27
ALTFAN implementation

leptun:
PFW-1109


Firmware/temperature.h
```diff
@@ -270,10 +270,12 @@ void check_fans();
...
-#endif
+#ifdef EXTRUDER_ALTFAN_DETECT
+  extern bool extruder_altfan_detect();
+#endif //EXTRUDER_ALTFAN_DETECT
 
 extern unsigned long extruder_autofan_last_check;
 extern uint8_t fanSpeedBckp;
 extern bool fan_measuring;
 
+#endif
```
DRracer: no newline at the end of file

Firmware/tone04.c
```diff
@@ -0,0 +1,110 @@
...
+	#define CRITICAL_SECTION_END    SREG = _sreg;
+#endif //CRITICAL_SECTION_START
+
+
+#include "fastio.h"
+
+void timer4_init(void)
+{
+	//save sreg
+	uint8_t _sreg = SREG;
```
DRracer: you have the CRITICAL_SECTION_START just a few lines above ;)

Firmware/tone04.c
```diff
@@ -0,0 +1,110 @@
...
+	//disable interrupts for sure
+	cli();
+
+	SET_OUTPUT(BEEPER);
+	WRITE(BEEPER, LOW);
+	
+	SET_OUTPUT(EXTRUDER_0_AUTO_FAN_PIN);
+	
+	TCCR4A = (1 << WGM40);
```
DRracer, line 36: it would be good to explain the timer setup in comments - especially why is something set

Firmware/tone04.c
```diff
@@ -0,0 +1,110 @@
...
+	{
+		TCCR4A &= ~(1 << COM4C1);
+		OCR4C = 0;
+		WRITE(EXTRUDER_0_AUTO_FAN_PIN, duty);
+	}
+	else
+	{
+		TCCR4A |= (1 << COM4C1);
+		// OCR4C = duty;
+		OCR4C = (((uint32_t)duty) * ((uint32_t)((TIMSK4 & (1 << OCIE4A))?OCR4A:255))) / ((uint32_t)255);
```
DRracer, line 64: uint16_t would be sufficient, don't you think? Duty can be 0-255, the second part of the statement can be up to 255 max.

Firmware/tone04.c
```diff
@@ -0,0 +1,110 @@
...
+{
+	WRITE(BEEPER, 1);
+}
+
+ISR(TIMER4_OVF_vect)
+{
+	WRITE(BEEPER, 0);
+}
+
+void tone4(__attribute__((unused)) uint8_t _pin, unsigned int frequency)
```
DRracer: a more precise type would be nice (`uint16_t frequency`)

Firmware/tone04.c
```diff
@@ -0,0 +1,110 @@
...
+	uint8_t prescalarbits = 0b001;
+	uint32_t ocr = F_CPU / frequency / 2 - 1;
+	
+	if (ocr > 0xffff)
+	{
+		ocr = F_CPU / frequency / 2 / 64 - 1;
+		prescalarbits = 0b011;
+	}
+	
+	uint8_t _sreg = SREG;
```
DRracer: critical section again ;)

Firmware/tone04.c
```diff
@@ -0,0 +1,110 @@
...
+	cli();
+	TCCR4B = (TCCR4B & 0b11111000) | prescalarbits;
+#ifdef EXTRUDER_0_AUTO_FAN_PIN
+	OCR4C = (((uint32_t)OCR4C) * ocr) / (uint32_t)((TIMSK4 & (1 << OCIE4A))?OCR4A:255);
```
DRracer, line 99: uint16_t again?

Firmware/tone04.c
```diff
@@ -0,0 +1,110 @@
...
+#endif //EXTRUDER_0_AUTO_FAN_PIN
+	OCR4A = ocr;
+	TIMSK4 |= (1 << OCIE4A) | (1 << TOIE4);
+	SREG = _sreg;
+}
+
+void noTone4(__attribute__((unused)) uint8_t _pin)
+{
+	uint8_t _sreg = SREG;
```
DRracer: critical section again

Firmware/tone04.c
```diff
@@ -0,0 +1,110 @@
...
+	cli();
+	TCCR4B = (TCCR4B & 0b11111000) | (1 << CS42) | (1 << CS40);
+#ifdef EXTRUDER_0_AUTO_FAN_PIN
+	OCR4C = (((uint32_t)OCR4C) * (uint32_t)255) / (uint32_t)((TIMSK4 & (1 << OCIE4A))?OCR4A:255);
```
DRracer, line 115: uint16_t again

DRracer (changes requested):
please address my comments in the code

Firmware/tone04.c
leptun: Forgot about that. Probably got reverted during the many changes I made to this function

Firmware/tone04.c
leptun: @DRracer I'm also trying to handle the scenario where a fan speed duty cycle change happens during a tone or tone4() is called while another tone is happening. I think it is safer if we stick with the larger data type just like Arduino did with their Tone()

DRracer: ok

DRracer (approved):

lord-carlos:
Is there any documentation on how to control or disable this?
My sunon hotend fan is spinning kinda slow.

3d-gussner:
@lord-carlos You can change the speed of the ALTFAN in your variant file https://github.com/prusa3d/Prusa-Firmware/blob/MK3/Firmware/variants/1_75mm_MK3S-EINSy10a-E3Dv6full.h#L337

vertigo235:
Why was it decided to cut Sunon fans to half speed?

DRracer:
@vertigo235 this is a generic solution for multiple fan types - if we detect speeds above 4k RPM (which is the default of the NOCTUA side fan) at the start, then we try to lower the RPM. The original idea was to keep the alternative fan's noise lower.

lord-carlos:
> The original idea was to keep the alternative fan's noise lower.

Hello @DRracer, thank you for your quick replay.
I assume what my esteemed colleague vertigo235 wanted to know is **why** you want to lower the noise on alternative fans?
What is the usecase?

* User has mk3s
* User is not satisfied with stock fan
* Installs alternative fan
* RPM halfed
* ???

If the user wants to have a low noise fan, I assume he will keep the Noctua? Or if it's defective buys a replacement noctua.

gregsaun:
Could you make this feature adjustable via the LCD menu? The community would really appreciate to have the choice between silence and performance/reliability.

TinyKatz:
> If the user wants to have a low noise fan, I assume he will keep the Noctua? Or if it's defective buys a replacement noctua.

I concur here.  I bought a non-Noctua fan because I WANT more air passing through the hotend.  If I was worried about noise, I would have kept the Noctua.  This particular idea should be reconsidered, or redesigned for menu controllable fan limitations.

threeclaws:
I use a sunon fan because the noctua gave me issues, cutting the RPM's in half on the sunon will just put me back in the same position as before.

cwkowalski:
I agree. Performance should not be risked for noise by default, reducing the max fan speed should be an optional setting and not the default. Like stealth mode.

thisiskeithb:
Please make this adjustable via LCD.

vertigo235:
> @vertigo235 this is a generic solution for multiple fan types - if we detect speeds above 4k RPM (which is the default of the NOCTUA side fan) at the start, then we try to lower the RPM. The original idea was to keep the alternative fan's noise lower.

Since alternative fans are installed by the user for reasons that care to them, why are you trying to keep the alternative fans noise lower?  The most likely reason that someone has installed an alternative fan is to improve the performance of the fan.

broodro0ster:
Please make this adjustable via the LCD screen so we don’t run into heatcreep issues again due to too low fan speeds.

josefprusa:
Noctua is very bad with deliveries during the pandemic so we need to have a solution if we need to use alternate fans and of course we need to stick to the promised sound levels the printer makes, this solution doesn’t require any user intervention and works. Another menu option which would confuse 150k MK3S users is not a good solution. Maybe disabling this by g-code might be an option.

vertigo235:
GCODE is an option yes, please consider that this affects many users who have upgraded their Noctua fan.  Upgrading the Noctua fan is the #1 recommended upgrade to improve the printer, especially in warm/hot environments.

vertigo235:
I think the GCODE option is a great idea @josefprusa , allowing the user to set the fan speed for their own setup would be GREAT they could adjust it as needed for sound.  Please do this, it would be very useful and positive for the community, thank you for listening and responding to this.

threeclaws:
> Noctua is very bad with deliveries during the pandemic so we need to have a solution if we need to use alternate fans and of course we need to stick to the promised sound levels the printer makes, this solution doesn’t require any user intervention and works. Another menu option which would confuse 150k MK3S users is not a good solution. Maybe disabling this by g-code might be an option.

What about building it into the code for silent mode? The toggle is already there and it would be more elegant than gcode but any solution that allows me to stay on stock firmware is a good one IMO.

broodro0ster:
A Gcode option would be fine as well. But compiling your own firmware is a no go imo.

gregsaun:
@josefprusa why don't you create a "low noise" menu from the LCD menu. This would activate low fan noise and stealth mode. I don't understand why we would have to choose stealth mode from the LCD menu and the fan via g-code, this sounds to be a disorganized solution.

thisiskeithb:
> Maybe disabling this by g-code might be an option.

That’d be a good compromise. Normal users won’t have to look at an option that’ll never apply to them and Advanced users can handle sending gcode commands since they are likely already doing that for other upgrades like a BMG extruder.

dryja123:
Here’s an idea, why not create a prompt that will prompt the user if the fan speed is detected higher than what a Noctua fan could deliver?  

For example: “Alternative fan detected, would you like to enable silent fan mode?”

Also create a menu option to allow the user to disable the feature if the performance is unsatisfactory. 

There are already RPM prompts built in. Not sure why this wouldn’t be a good option. 

Edit: if you went the GCODE route to disable the feature, would I have to go back and edit all of my production GCODE?  That would be annoying.

Panayiotis-git:
Noctua provides a resistor adapter to plug in series with the fan if you want to reduce speed/noise

gregsaun:
@Panayiotis-git very nice point!

gregsaun:
@Panayiotis-git Very nice point!

@josefprusa As mentioned by @Panayiotis-git just add a resistor to your Delta fan to make it running slower. This will save place for your firmware, reduce testing, does not add complexity for the end user, and the community is free to modify the fan without having to modify firmware or use of a start g-code.

sdooweloc:
Please make this an option on the screen instead of making us compile our own firmware. I grabbed an alternative fan because I have had others have issues with heat creep on the stock fan. Halving the rpm of my replacement fan would defeat the replacement. If I wanted it to be as quiet as the noctua i would just use the noctua. I wanted the extra airflow and am willing to sacrifice the sound levels for that.

LupoLoopy:
I couldn't give two hoots about noise level. If my priority was noise level, I'd have kept the stock fan. My priority is performance/stopping heat-creep, so I replaced my fan.

This really strikes me as a solution for non-problem, akin to 'we decided to speed limit the new Ferrari for consumers who cared about the MPG'. Who in the Ferrari audience cares about MPG? If you want economy, you buy a Fiat. Or more analogy accurate, keep your existing Fiat.

If you insist onto integrating such a thing into the code, essentially gutting non-stock hardware, at least let me kill it without having to play the compilation game. Menu or G-Code, I'm fine with, as long as it's documented. Both, even better, I guess?

{edit: Totally appreciate you have Noctura supply issues and need a solution for your manufacturing pipeline. It would just be nice if the solution didn't geld a widely deployed and recognised airflow-upgrade for those who already own the printers)

draeath:
Perhaps a setting to control the RPM that this logic operates on? Default to the RPM of the Noctua to get the effects you desire out-of-the-box, but folks who have replaced their fans can set things up as they wish.

This also covers them if they *do* want to limit RPM for noise, and is also implementable via gcode if that's the route decided.

I replaced mine with a Delta ASB0405HHA-AF00 (7k RPM) and if this change was in place, my resulting speed would be 3.5k, *lower than the original Noctua!* Not acceptable, I need a way to control this without having to customize my firmware. (not that I can't do that, I just don't believe it should be necessary).

CodeMonkeyX:
> @josefprusa As mentioned by @Panayiotis-git just add a resistor to your Delta fan to make it running slower. This will save place for your firmware, reduce testing, does not add complexity for the end user, and the community is free to modify the fan without having to modify firmware or use of a start g-code.

Nothing is that simple though. He would either have to buy harnesses (like the low noise plug Noctua makes), or have an employee sit there and add resisters to each fan. Even if you get harnesses they add another connector, and bulk into the fan cable, then he would have to redesign the extruder to give space for the additional connector.

Either way that's another employee you are paying to mod a fan that 90% of users do not care about. For power users I think the GCode option is fine. If you are a power user it's not a big deal to add a custom gcode to disable fan speed limits for your printer.

mmalluck:
I also replaced my fan to address a heatcreep issue. Please don't nerf my device. The last thing I need is a firmware that brings back hotend jams.

Menu option is preferred. Gcode is acceptable, but less optimal as I'd have to modify all of my presliced files.

gregsaun:
@CodeMonkeyX they talk to Delta directly, they can have this made by them "in the wires". And Noctua fans are probably much more expensive than Delta so not necessarily more expensive. But yes this is a point to verify on their side.

Note that MK2.5(S) has already a connector for the Noctua to use 5V from somewhere else as the Rambo Mini is 12V on the fan.

zafuquazar:
Since this speed reduction is triggered by higher than Noctua fan speeds sensed at startup - am I correct in assuming that turning off “fan sense” will prevent the speed reduction from happening?

ethsanders:
Please make this a toggle in settings instead of forced.

TheMrCaveman:
This is a terrible idea, people change the fan because the noctua fan is simply not powerfull enaugh for some aplications and people change the fan for that reason. Now we have to compile firmware ourselfs to get it back to to work...........

alomex21:
As @josefprusa said, adding an option to the gcode commands would be the best since is the least invasive, non-complex, and probably more user-friendly (Obligatory Documentation ofc) option there is. As far I can see, the only argument here is the heat creep and the high ambient temperature of the room, and if I remember right, the Prusa farm is already quite hot and pretty sure they haven't got any problem at all, the worst scenario would two or three would fail.
IMO this is an interesting change, but pretty sure it doesn't affect like 90% of Prusa MK3 users.
Cheers!

ogland:
Limiting the speed of maglev fans with PWM destroys them. So in other words, most users with the sunon fan mod will have their hardware destroyed by this PR.

Check note 7 in this document:

https://www.google.com/url?sa=t&source=web&rct=j&url=http://www.tme.eu/Document/3cb1207277218a2fbbe696db35c9eace/A1259XBL-TC%2528A25000350G-00%2529-0.pdf&ved=2ahUKEwi-4uLs9NTqAhVPfZoKHcxJAC0QFjAJegQIARAB&usg=AOvVaw0EC9qIblrCkSXr3RSIg8V-

Evoplatypus:
Why not link it to "silent" mode? If someone wants the printer quiet they would run it like that anyway.

vertigo235:
The Menu option should say "Hotend Fan Mode: [Poor / Normal]"

PaulGD03:
I use silent mode a lot, and that doesn't necessarily mean I want to run my hotend fan slower, it doesn't matter to me anyway, as the blower fan is louder. Is it that hard to just add an entry to HW Setup where you can override the standard fan speed? You can of course keep it at the "SILENT" speed when stock, but I see no reason we shouldn't have an easy way to change it, we can change the blower speed too afterall, and that thing is louder than the hotend fan anyway.

vertigo235:
> I use silent mode a lot, and that doesn't necessarily mean I want to run my hotend fan slower, it doesn't matter to me anyway, as the blower fan is louder. Is it that hard to just add an entry to HW Setup where you can override the standard fan speed? You can of course keep it at the "SILENT" speed when stock, but I see no reason we shouldn't have an easy way to change it, we can change the blower speed too afterall, and that thing is louder than the hotend fan anyway.

Careful, the next update will be to limit all blower fans to 50% :D

Evoplatypus:
@vertigo I get it. I did the same thing and changed the fan and here I am giving my 2 cents. Don't beat a dead horse.

@paulgamedev your comments is what I thought of too after I posted. I never use it because I want performance, but I can see your scenario.

vertigo235:
Lighten up guys, I like to tell jokes, I think this will be addressed, I love the idea of allowing user settable GCODE. We are all friends here.

SigmaRelief:
Taking a step back, is there a way that this could be used to better harmonize printers?  Other g code parameters are unitized so 1mm is the same everywhere, but fan speed is arbitrary.  A simple 100% = 3500RPM to properly set full range could be used as a scale and a separate Max fan speed = XXXX RPM much like firmware speed limits on the axis motors would serve as a high end trim.  If done elegantly this would be just another useful feature to fine tune performance.  Elegance may be difficult however and there would still be many debates regarding g codes, menu options, and defaults for new firmware.

The bigger thing is new customers will be getting a more powerful fan stock which is nothing but good news for the community provided they can easily enable this additional performance.  A short prompt in the initial calibration wizzard would ensure everyone gets to pick their preference.

fheatherz:
Please make this a menu selectable or gcode option. I print in my office, and in the summer I will get heat creep unless I blast the AC for the whole print. I changed to a Sunon fan, problems gone. It's actually not very noisy, and even if it's louder than a noctua I am fine with that.

josefprusa:
So after brainstorming here with the @DRracer the current plan is to make a hidden menu (we still have to agree on how to call it but imagine triple click in support menu or similar for example) for options like this. I really don't want to make the normal menus more crowded with features which are unrelated/confusing for users with no mods. Contrary to the feeling from the online communities, less than 3% of users ever did any mod based on our customer survey. With 150k+ MK3s in the wild, anything confusing kills the support and regular issues don't get resolved swiftly. BTW, I don't like making this decision under pressure of flamewar quickly when we would have more time to think about this in the near future as the release would get closer ...

ogland:
@josefprusa So will the sunon maglev fan destroyer feature be activated by default or do the users have to activate this manually?

And for what it's worth, I think your proposed solution is very good and it's nice that you are hearing the voices of the community like this. I'm just concerned about the ones having a fan mod not realizing that they need to deactivate a feature in order to not have their hardware destroyed. There needs to be a super clear disclaimer I think.

lord-carlos:
> @ josefprusa So will the sunon maglev fan destroyer feature be activated by default or do the users have to activate this manually?

Low rpm alt fan has to be active by default, no? Otherwise new users who get a printer with an delta fan would have to active it themself. Would be a weird request for brand new users.
On the other end, users who switched out stock fan for something else know what they do.

ogland:
@lord-carlos I can assure you that plenty of users who change the fan don't know what they do. No fan limiting by PWM should definitely be activated by default. I think a silent mode would be better because:
1) A delta at high speed is better for print quality of low TG filaments
2) Sunon maglev fan users can have their fans destroyed

josefprusa:
@DRracer we might be able to detect the current users of modded fans if the printer boots up with the new firmware without the flag to run wizard on first boot, detect the rpm and unlock it. This wouldn't change the behavior for current user? Just in theory for now.

gregsaun:
@josefprusa would that means if we have a Sunon, reset the config and boot then we might destroy the fan, no?

josefprusa:
@gregsaun that is, why there are release notes. Feel free to propose another solution while maintaining our parameters of not bothering new users.

gregsaun:
@josefprusa why not adding a resistor on the fan like adapters of Noctua fans?

vertigo235:
I still like the GCODE idea the best but all I care is that this is being addressed and community voices are being heard. 

Thank you Prusa team.

fheatherz:
As someone with an upgraded fan, I'm ok with anything that doesn't involve compiling my own firmware. Gcode, secret menu, autodetect, whatever. Thanks for listening to us! :)

draeath:
> less than 3% of users ever did any mod based on our customer survey

@josefprusa just a caution there - that's of the people who completed the survey, or installed such modifications after doing so. I don't recall seeing this survey ever myself, for example. Gotta be careful of hidden biases in selection for such surveys. I see PhD candidates running into that trap frequently!!

Thank you for taking the time to listen to us, and consider our viewpoints on this. Personally, I'm happy so long as we have a way to adjust this behavior without having to flash an unofficial build.

MartinMajewski:
> Contrary to the feeling from the online communities, less than 3% of users ever did any mod based on our customer survey. 

@josefprusa Surveys are essential, but to be taken with a big grain of salt. I have two highly modified MK2.5S and one MK3S in the making, and no survey has ever reached my mailbox. I completed the Mini survey last week, telling you my Mini is not modded. But this is only because I have waited for that thing for months, and so did others. There are any worth mentioning mods out there yet. But between the survey and now, I have printed a back cover for the LCD screen for the Mini. Is this enough of a mod to make the survey obsolete?

I think that if you guys want to go with Delta fans, you should make it in a way that the hardware you change incorporates the mods, not stuff that is already out in the wild. Incorporate a resistor, an adapter harness, or whatnot for those, but leave the firmware alone. Next year you find another fan supplier, and what then?

After you changed the extruder fan size from 30x30 to 40x40, the extruder fan is the smallest noise issue IMHO.

PR #2698, opened 2020-05-29
MK3s IR sensor improvement

3d-gussner:
Both IR sensor versions are "automatically" detected:
After Factory reset / Fw upgrade from <FW3.9.0 / Shipping AND after switching IR sensor from 0.4 to 0.3 (on the fly) and 0.3 to 0.4 after restart:

- when filament is loaded and later unloaded/status change ( need to wait for FS_CHECK_COUNT )
- when filament is *un*loaded and later loaded/status change ( need to wait for FS_CHECK_COUNT )

Also Support -> Fil. sensor v. reads the EEPROM value and shows the correct status

In case the voltage is in "bad range 1.5-3.0V" the Settings -> Fil. sensor goes in [N/A] and outputs a message on serial.

Resources used [MK3 FW3.9.0](https://travis-ci.com/github/prusa3d/Prusa-Firmware/builds/166731977) vs this [PR](https://travis-ci.com/github/prusa3d/Prusa-Firmware/builds/168712148):
| MK3S | 248810 | 6486 | 249218 | 6494 | +408 | +8 |
| MK3 | 250396 | 6515 | 250396 | 6515 | +0| +0 |
| MK25S 13a | 222264 | 6310 | 222264 | 6310 | +0| +0 |
| MK25S 10a | 222144 | 6310 | 222144 | 6310 | +0| +0 |
| MK25 13a | 226274 | 6358 | 226274 | 6358 | +0| +0 |
| MK25 10a | 226004 | 6358 | 226004 | 6358 | +0| +0 |

PFW-1116


Firmware/Marlin_main.cpp
```diff
@@ -9474,44 +9474,65 @@ static uint16_t nFSCheckCount=0;
...
+						}
+					} else {
+						nFSCheckCount=0;
+					}
+				} 
+				//! If and only if minVolt is in range <0.0, 0.3> and maxVolt is in range  <4.6, 5.0V>, I'm considering a situation with the old fsensor
+				else if( minVolt < IRsensor_Ldiode_TRESHOLD 
+				 && maxVolt > IRsensor_Hopen_TRESHOLD && maxVolt <= IRsensor_VMax_TRESHOLD
+				){
+					bool bTemp = (!CHECK_ALL_HEATERS);
```
DRracer: Can we refactor this whole if()else if()... better? It is a copy of the if(..) check above but only with a few changed variables.

Firmware/fsensor.cpp
```diff
@@ -731,26 +731,39 @@ void fsensor_update(void)
 }
 
 #ifdef IR_SENSOR_ANALOG
-/// This is called only upon start of the printer or when switching the fsensor ON in the menu
-/// We cannot do temporal window checks here (aka the voltage has been in some range for a period of time)
+//! This is called only upon start of the printer or when switching the fsensor ON in the menu
```
DRracer: it is not necessary to change the doxygen comment syntax

Firmware/ultralcd.cpp
```diff
@@ -2169,14 +2169,16 @@ static void lcd_support_menu()
...
        case ClFsensorPCB::_Old:
-            MENU_ITEM_BACK_P(_T(MSG_03_OR_OLDER));
+            MENU_ITEM_BACK_P(_T(MSG_IR_03_OR_OLDER));
             break;
        case ClFsensorPCB::_Rev04:
-            MENU_ITEM_BACK_P(_T(MSG_04_OR_NEWER));
+            MENU_ITEM_BACK_P(_T(MSG_IR_04_OR_NEWER));
             break;
        case ClFsensorPCB::_Undef:
+            MENU_ITEM_BACK_P(_T(MSG_IR_UNKNOWN));
```
DRracer: why did you split the _Undef and default branch when they are identical?

Firmware/ultralcd.cpp
```diff
@@ -5725,7 +5727,9 @@ void lcd_hw_setup_menu(void)                      // can not be "static"
 
 #ifdef IR_SENSOR_ANALOG
     FSENSOR_ACTION_NA;
-    MENU_ITEM_FUNCTION_P(PSTR("Fsensor Detection"), lcd_detect_IRsensor);
+	//! Fsensor Detection isn't ready for mmu yet it is temporarily disabled.
```
DRracer: fix indentation

Firmware/ultralcd.cpp
```diff
@@ -7542,14 +7546,19 @@ static bool lcd_selftest_IRsensor(bool bStandalone)
 
 static void lcd_detect_IRsensor(){
     bool bAction;
-
+	bool loaded;
```
DRracer: fix indentation in the whole function

DRracer (changes requested):
please address my comments

Firmware/fsensor.cpp
```diff
@@ -207,7 +207,7 @@ void fsensor_init(void)
 	}
 	printf_P(PSTR("FSensor %S"), (fsensor_enabled?PSTR("ENABLED"):PSTR("DISABLED")));
 #ifdef IR_SENSOR_ANALOG
-     printf_P(PSTR(" (sensor board revision:%S)\n"), (oFsensorPCB==ClFsensorPCB::_Rev04) ? _T(MSG_04_OR_NEWER) : _T(MSG_03_OR_OLDER));
+     printf_P(PSTR(" (sensor board revision:%S)\n"), ((oFsensorPCB==ClFsensorPCB::_Undef) ? _T(MSG_IR_UNKNOWN): ((oFsensorPCB==ClFsensorPCB::_Rev04) ? _T(MSG_IR_04_OR_NEWER) : _T(MSG_IR_03_OR_OLDER))));
```
DRracer: Can we do this relatively complex version text selection at one place? There are similar cases in ultralcd.cpp.
something like:
`const char *FsensorIRVersionText(){
switch(oFsensorPCB){
case oFsensorPCB==ClFsensorPCB::_Undef: return _T(MSG_IR_UNKNOWN);
....
}...`

Firmware/ultralcd.cpp
```diff
@@ -7542,14 +7546,19 @@ static bool lcd_selftest_IRsensor(bool bStandalone)
...
     bMenuFSDetect = true;                               // inhibits some code inside "manage_inactivity()"
-    bAction = lcd_show_fullscreen_message_yes_no_and_wait_P(_i("Is filament loaded?"), false, false);
-    if(bAction){
-        lcd_show_fullscreen_message_and_wait_P(_i("Please unload the filament first, then repeat this action."));////c=20 r=4
+	//! Check if filament is loaded. If it is loaded stop detection.
+	//! @todo Add autodetection with MMU2s
+	loaded = (digitalRead(IR_SENSOR_PIN) == 0);
```
DRracer: are you sure this `loaded` will work even if the fsensor is unconnected, damaged etc.?

Firmware/ultralcd.cpp
3d-gussner: Should we use spaces or tabs?

Firmware/ultralcd.cpp
```diff
@@ -7783,8 +7800,7 @@ bool lcd_selftest()
 				_progress = lcd_selftest_screen(TestScreen::FsensorOk, _progress, 3, true, 2000); //fil sensor OK
 			}
 #endif //PAT9125
-//#ifdef IR_SENSOR_ANALOG
-#if (0)
+#ifdef IR_SENSOR_ANALOG
```
DRracer: The main trouble with this was that the user would have been forced to remove the loaded filament on an assembled printer - just to check the two voltages.
That's why there was the #if 0
Did you find a way around it?

Firmware/ultralcd.cpp
3d-gussner: I get what you mean as I don't read the raw voltage value but just the digital value.
- Unconnected should result as digital 1 
- A broken one we don't know for sure what it would be.
But if you like we can keep the Fsensor detection asking for the status.

Firmware/ultralcd.cpp
3d-gussner: That is something I didn't know that assembled printers have filament loaded...

have to think about that

Firmware/fsensor.cpp
```diff
@@ -611,45 +611,44 @@ void fsensor_enque_M600(){
...
-			err |= (fsensor_oq_yd_sum < FSENSOR_OQ_MIN_YD); // total measured distance is below limit
-
-            fsensor_restore_print_and_continue();
-            fsensor_autoload_enabled = autoload_enabled_tmp;
-            fsensor_oq_meassure_enabled = oq_meassure_enabled_tmp;
-
-            unsigned long now = _millis();
-            if (!err && (now - fsensor_softfail_last) > FSENSOR_SOFTERR_DELTA)
-                fsensor_softfail_ccnt = 0;
+if (fsensor_watch_runout && (fsensor_err_cnt > FSENSOR_ERR_MAX))
```
DRracer: shouldn't this `if` be offset by 4 spaces too?

Firmware/ultralcd.cpp
3d-gussner: This section is in the selftest and enables the check if the sensor is working correctly:
- If filament is loaded the user needs to unload and redo the selftest
- If filament is unloaded the user will need to load in the sensor (but not in extruder"

Firmware/fsensor.cpp
3d-gussner: You are right sorry ... fixing it now

Firmware/fsensor.cpp
3d-gussner: Fixed by @leptun

Firmware/fsensor.cpp
3d-gussner: fixed https://github.com/prusa3d/Prusa-Firmware/pull/2698/commits/6d07ea72664cc2d79f62078d696af6ccacbc47f3

Firmware/fsensor.cpp
```diff
@@ -629,129 +644,146 @@ void fsensor_update(void)
...
+                        // sequence for direct data reading from AD converter
+                        DISABLE_TEMPERATURE_INTERRUPT();
+                        nMUX1=ADMUX;        // ADMUX saving
+                        nMUX1=ADMUX;        // ADMUX saving
+                        nMUX2=ADCSRB;
+                        adc_setmux(VOLT_IR_PIN);
+                        ADCSRA|=(1<<ADSC);  // first conversion after ADMUX change discarded (preventively)
+                        while(ADCSRA&(1<<ADSC))
+                            ;
+                            ADCSRA|=(1<<ADSC);  // second conversion used
```
DRracer: this indentation is incorrect - this looks like the `ADCSRA|=(1<<ADSC);` is a part of the while cycle above which isn't true.

Firmware/ultralcd.cpp
DRracer: yes, but that's why we moved the detection to runtime - not to force the user to remove and insert the filament

Firmware/fsensor.cpp
```diff
@@ -629,129 +644,146 @@ void fsensor_update(void)
...
-
-								   // if we are printing and FS action is set to "Pause", force pause the print
-                                   if(oFsensorActionNA==ClFsensorActionNA::_Pause)
+                        uint8_t nMUX1,nMUX2;
+                        uint16_t nADC;
+                        bIRsensorStateFlag=false;
+                        // sequence for direct data reading from AD converter
+                        DISABLE_TEMPERATURE_INTERRUPT();
+                        nMUX1=ADMUX;        // ADMUX saving
+                        nMUX1=ADMUX;        // ADMUX saving
```
DRracer: why is this duplicated?

DRracer (approved):

PR #2709, opened 2020-06-03
M-Code to get silent mode status

hergtoler:
I'm really not sure what conventions are for adding codes, but in the spirit of sharing, here's what I did.  I'm using this to allow an Octoprint plugin to display embedded time estimates properly based on whether running in silent or normal mode.  A link to the plugin changes:
[https://github.com/kanocz/octopi_eta_override/pull/6](url)

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -8417,6 +8417,15 @@ Sigma_Exit:
 
 #endif //TMC2130_SERVICE_CODES_M910_M918
 
+    */
+	case 919:
+    {
+#ifdef TMC2130
+		printf_P(_N("SILENT MODE: %S\n"), (tmc2130_mode==TMC2130_MODE_SILENT)?_N("ENABLED"):_N("DISABLED"));
+#endif //TMC2130
+    }
+    break;
+
     /*!
 	### M350 - Set microstepping mode <a href="https://reprap.org/wiki/G-code#M350:_Set_microstepping_mode">M350: Set microstepping mode</a>
     Printers with TMC2130 drivers have `X`, `Y`, `Z` and `E` as options. The steps-per-unit value is updated accordingly. Not all resolutions are valid!
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -8417,6 +8417,9 @@ Sigma_Exit:
 
 #endif //TMC2130_SERVICE_CODES_M910_M918
 
+    /*!
+	### M919 - Get silent mode status <a href="https://reprap.org/wiki/G-code#M919:_Report_silent_mode_status">M919: Report silent mode status</a>
+	
     */
 	case 919:
     {
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -8417,17 +8417,17 @@ Sigma_Exit:
 
 #endif //TMC2130_SERVICE_CODES_M910_M918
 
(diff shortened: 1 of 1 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -8417,6 +8417,18 @@ SERIAL_PROTOCOLPGM("\n\n");
 
 #endif //TMC2130_SERVICE_CODES_M910_M918
 
+    /*!
+	### M919 - Get silent mode status <a href="https://reprap.org/wiki/G-code#M919:_Report_silent_mode_status">M919: Report silent mode status</a>
+	
+    */
+	case 919:
+    {
+#ifdef TMC2130
```
leptun: It would be better if you moved the `#ifdef TMC2130` to encompass the entire case, so that on a printer without TMC2130 you will get "unknown gcode" or something like that automatically. There's no reason to handle something like this with the MK2.5/S since there the Mode doesn't change the max speed/acceleration

leptun (changes requested):
@hergtoler Change this, but otherwise it looks ok to me. Maybe @3d-gussner has something to say about the documentation section.

awenelo:
Would a command to send the mode be better? (instead of if silent mode is enabled) That way, if another mode were to be added(such as super-fast mode), we could append a condition for that to the existing output.

leptun:
@awenelo That is also another thing, But here he is only using tmc2130_mode which has 2 states. Unless you implement TMC2209 drivers in the firmware (which I did) or feel brave enough to trust the HYBRID_THRESHOLD, then I don't see why you would ever implement more than two modes

3d-gussner:
See https://github.com/prusa3d/Prusa-Firmware/pull/2508
And M914 https://reprap.org/wiki/G-code#M914:_Set_TMC2130_normal_mode is already defined in RepRap G-code wiki.

3d-gussner:
I think we could use the idea from @vintagepc and add another parameter that "reads" the status and is default to read if `M914` is sent without any parameter.

vintagepc:
I don't see any reason I couldn't add a serial output indicating the current status to that PR

vintagepc:
M914/5 are flagged as "set" operations so I don't think we should change the default behaviour to only query the status. But we could add a Q argument to query status without changing, and this would return the same thing as shown in the menu (normal/stealth/M91x override)

I am happy to revise that PR to accommodate this after work tonight, just let me know and I will do that.

3d-gussner:
Sure a change has to be discussed but I would prefer to minimise the special tmc m codes as we already have too many with the same name but different functions depending on the firmware.

hergtoler:
Adding a query to M914/M915 would certainly get the job done, though it would be nice if it was included in the default build, not require special #define TMC2130_SERVICE_CODES_M910_M918 and rebuild.

Edit:
An alternative that could work for my use case would be to modify the M73 code to only print status for the active mode rather than both "NORMAL" and "SILENT" on every occurrence.

hergtoler:
> An alternative that could work for my use case would be to modify the M73 code to only print status for the active mode rather than both "NORMAL" and "SILENT" on every occurrence.

Like this: https://github.com/prusa3d/Prusa-Firmware/pull/2735

3d-gussner:
@hergtoler Thanks for the PR, we may use your idea and use a different way.

3d-gussner:
@hergtoler Thanks for the idea and PR! Please close this PR if you think that https://github.com/prusa3d/Prusa-Firmware/pull/4388 is okay for you.

PR #2721, opened 2020-06-06
Fix high speed deceleration

leptun:
If you run the steppers at high speed (350mm/s+) or use higher microstepping (x32+), then you would soon begin to notice that high speed motion accelerates just fine, but once it starts decelerating the axes jerk twice and the speed behaves very weird.
Here is the issue in action: https://photos.app.goo.gl/JYH4MrEAuLz2v3W5A
To get that effect I simply home the axis, `M350 Y32` for higher microstepping and then do a move at 200mm/s. To see the issue easier I recommend lowering the acceleration with `M204 S350`. It really becomes visible at that acceleration.
With this commit the issue disappears entirely.

### DIFF
--- a/Firmware/planner.h
+++ b/Firmware/planner.h
@@ -77,8 +77,8 @@ typedef struct {
   unsigned char direction_bits;             // The direction bit set for this block (refers to *_DIRECTION_BIT in config.h)
   unsigned char active_extruder;            // Selects the active extruder
   // accelerate_until and decelerate_after are set by calculate_trapezoid_for_block() and they need to be synchronized with the stepper interrupt controller.
-  long accelerate_until;                    // The index of the step event on which to stop acceleration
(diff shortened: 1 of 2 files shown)


vintagepc:
I wonder if I was observing a similar effect with 0.9 degree steppers at 16 microsteps, it would always choke on deceleration during quick travel moves... Issue goes away if X/Y microsteps is set to 8.

leptun:
Yes, this PR should fix that issue

vintagepc:
Nice, I merged it locally and can confirm I'm no longer getting horrible noises/lost steps during fast deceleration. (0.9 steppers, 16 microsteps, 200 steps/mm). Will keep an eye on it when running more prints.

leptun:
Closing in favor of #2792

PR #2724, opened 2020-06-08
Implement MIN/MAX AMBIENT safety checks

wavexx:
Take advantage of the NTC thermistor found on the Einsy as an additional
safety measure, following the steps of the other MIN/MAXTEMP errors.

Introduce two configurable params AMBIENT_MINTEMP and AMBIENT_MAXTEMP
in the variant defines and set them for the MK3/MK3S to -30/+100
respectively.

AMBIENT_MINTEMP is primarily intended to catch a defective board
thermistor (to ensure MAXTEMP would be properly triggered) and thus the
trigger temperature is set just above the sensing limit and well below
the operating range.

AMBIENT_MAXTEMP is set at 100C, which is instead 20C above the maximum
recommended operating temperature of the Einsy. The NTC thermistor is
located just above the main power connector on the bottom of the board,
and could also help in detecting a faulty connection which can result in
rapid overheating of the contacts.

As for MAXTEMP, we cut power to the heaters, print fan and motors to
reduce power draw. Resume is not possible except by resetting the
printer, since the user is highly advised to inspect the board for
problems before attempting to continue.

PFW-1114

### DIFF
--- a/Firmware/temperature.cpp
+++ b/Firmware/temperature.cpp
@@ -639,6 +639,7 @@ void manage_heater()
     return; 
 // more precisely - this condition partially stabilizes time interval for regulation values evaluation (@ ~ 230ms)
 
+  // ADC values need to be converted before checking: converted values are later used in MINTEMP
   updateTemperaturesFromRawValues();
 
   check_max_temp();
--- a/Firmware/temperature.cpp
+++ b/Firmware/temperature.cpp
@@ -180,6 +180,12 @@ static int bed_minttemp_raw = HEATER_BED_RAW_LO_TEMP;
 #ifdef BED_MAXTEMP
 static int bed_maxttemp_raw = HEATER_BED_RAW_HI_TEMP;
 #endif
+#ifdef AMBIENT_MINTEMP
+static int ambient_minttemp_raw = AMBIENT_RAW_LO_TEMP;
+#endif
+#ifdef AMBIENT_MAXTEMP
+static int ambient_maxttemp_raw = AMBIENT_RAW_HI_TEMP;
+#endif
 
 static void *heater_ttbl_map[EXTRUDERS] = ARRAY_BY_EXTRUDERS( (void *)HEATER_0_TEMPTABLE, (void *)HEATER_1_TEMPTABLE, (void *)HEATER_2_TEMPTABLE );
 static uint8_t heater_ttbllen_map[EXTRUDERS] = ARRAY_BY_EXTRUDERS( HEATER_0_TEMPTABLE_LEN, HEATER_1_TEMPTABLE_LEN, HEATER_2_TEMPTABLE_LEN );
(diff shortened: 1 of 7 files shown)


Firmware/temperature.cpp
```diff
@@ -1484,32 +1504,24 @@ void min_temp_error(uint8_t e) {
 }
 
 void bed_max_temp_error(void) {
-#if HEATER_BED_PIN > -1
-  //WRITE(HEATER_BED_PIN, 0);
-#endif
   if(IsStopped() == false) {
     SERIAL_ERROR_START;
-    SERIAL_ERRORLNPGM("Temperature heated bed switched off. MAXTEMP triggered !");
+    SERIAL_ERRORLNPGM("Heaters switched off. MAXTEMP BED triggered !");
```
DRracer: Can we build these strings from parameters? There are multiple instances of them and only MAXTEMP/ MINTEMP BED/AMBIENT actually change.

Firmware/temperature.cpp
```diff
@@ -1428,19 +1452,16 @@ enum { LCDALERT_NONE = 0, LCDALERT_HEATERMINTEMP, LCDALERT_BEDMINTEMP, LCDALERT_
 uint8_t last_alert_sent_to_lcd = LCDALERT_NONE;
 
 void max_temp_error(uint8_t e) {
-  disable_heater();
```
DRracer, line 1431: I'd rather keep the disable_heater() call here - just to be safe

Firmware/temperature.cpp
```diff
@@ -1459,12 +1480,11 @@ void min_temp_error(uint8_t e) {
 	return;
 #endif
 //if (current_temperature_ambient < MINTEMP_MINAMBIENT) return;
-  disable_heater();
```
DRracer, line 1462: also keep the disable_heater() here

DRracer (changes requested):

Firmware/temperature.cpp
wavexx: Stop() does that of course as the first call. Maybe we should move Stop() as the first action instead?

wavexx:
Instead of adding a duplicate call, what about moving Stop() first, which makes more sense anyway?
Stop() will disable heaters in the first step too.

wavexx:
Rebased on current master. I re-introduced the disabled_heaters calls, as well as building the error strings out of parameters. Despite two new small lcd functions, still saves 188 bytes.

There's probably a bit more to save during the initialization of the minmax/temp[raw] tables IMHO.

Firmware/temperature.cpp
```diff
@@ -1427,20 +1451,47 @@ enum { LCDALERT_NONE = 0, LCDALERT_HEATERMINTEMP, LCDALERT_BEDMINTEMP, LCDALERT_
...
 uint8_t last_alert_sent_to_lcd = LCDALERT_NONE;
 
+
+//! update the current temperature error message
+//! @param type short error abbreviation (PROGMEM)
+//! @param func optional lcd update function (lcd_setalertstatus when first setting the error)
+void temp_update_messagepgm(const char* PROGMEM type, void (*func)(const char*) = lcd_updatestatus)
+{
+    char msg[LCD_WIDTH];
+    strcpy_P(msg, PSTR("Err: "));
```
DRracer, line 1461: interestingly, filling the msg this way results in 30B shorter code than in `char msg[] = "Err: ";`

Firmware/temperature.cpp
```diff
@@ -1484,44 +1532,61 @@ void min_temp_error(uint8_t e) {
...
 #ifdef DEBUG_DISABLE_MINTEMP
 	return;
 #endif
-//if (current_temperature_ambient < MINTEMP_MINAMBIENT) return;
-#if HEATER_BED_PIN > -1
-    //WRITE(HEATER_BED_PIN, 0);
-#endif
-	static const char err[] PROGMEM = "Err: MINTEMP BED";
+    disable_heater();
+	static const char err[] PROGMEM = "MINTEMP BED";
```
DRracer: indentation inconsistent with lines above

DRracer (changes requested):
Please fix the indentation inconsistency at one spot, then we can merge this.

Firmware/temperature.cpp
wavexx: Tried a few different approaches, I was surprised too. It's a bit of a gimmick to pass the update function, but I didn't want to duplicate the str* function twice.

When I was looking to unify some code I think we could actually also remove the "Err: " prefix in a few other places by adding a proper lcd function to set error messages. But it had a bit too many changes for this PR. Something to think about for the future though.

Firmware/temperature.cpp
wavexx: I think it's just the usual tab/spaces expanding differently yet again, they're aligned in the code. Expanding it now.

DRracer (approved):

PR #2725, opened 2020-06-08
Fix compiler warning enumeration value 'ONE_TO_FALL' not handled in s…

mkbel:
…witch.

leptun:
Yes, I noticed this as well. That enum state is never actually used as it is not needed. All we should do is comment out ONE_TO_FALL in the States class. If I'm not mistaken there is already a PR that fixes this issue and some other warnings, but it also includes some stuff that shouldn't be changed.


3d-gussner:
https://github.com/prusa3d/Prusa-Firmware/pull/2454

DRracer (approved):

PR #2727, opened 2020-06-08
Unused function warning fix.

mkbel:
(cherry picked from commit 54e2b6a829a221cc3abbff3a39ca7d3cafed3a09)
Pick only unused function warning fix.
#2454

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -5602,6 +5602,7 @@ do\
 }\
 while (0)
 
+#if 0 // temporarily unused
 static void lcd_check_gcode_set(void)
 {
 switch(oCheckGcode)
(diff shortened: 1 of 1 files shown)


Firmware/ultralcd.cpp
```diff
@@ -5602,6 +5602,7 @@ do\
 }\
 while (0)
 
+#if 0 // temorarily unused
```
DRracer: tem*P*orarily ;)

DRracer (changes requested):
typo ;)

Firmware/ultralcd.cpp
mkbel: There, I fixed it.

DRracer (approved):

PR #2729, opened 2020-06-08
Update farm mode status screen rendering

leptun:
Also fixes warning.


mkbel:
Can I close this PR? It looks like it is printing farm number. But as far as I remember we decided together and with farm operator to drop farm number and print "FRM" string in farm mode instead.

leptun:
Yes, you can close this one. I’ll submit a PR that removes it soon.

PR #2669, opened 2020-05-17
Report Long Filename to Octoprint

hergtoler:
Uses the long filename when outputting the SD open filename so that Octoprint can display the long filename during monitoring.

PR #2731, opened 2020-06-08
Report Long Filename to Octoprint

GilesBathgate:
Uses the long filename when outputting the SD open filename so that Octoprint can display the long filename during monitoring.

### DIFF
--- a/Firmware/cardreader.cpp
+++ b/Firmware/cardreader.cpp
@@ -415,14 +415,14 @@ void CardReader::openFile(const char* name,bool read, bool replace_current/*=tru
     if (file.open(curDir, fname, O_READ)) 
     {
       filesize = file.fileSize();
+      getfilename(0, fname);
       SERIAL_PROTOCOLRPGM(_N("File opened: "));////MSG_SD_FILE_OPENED
-      SERIAL_PROTOCOL(fname);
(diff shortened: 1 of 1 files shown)


BackSlasher:
@leptun I see you approved #2669 which contains the same commit.
Could I ask you to please review this as well?

I'm an interested party :)
Thank you for your time!

leptun (approved):
As per the first comment, I'm reapproving this PR.

BackSlasher:
image
@leptun do we need someone else to approve?

neslekkim:
Won't spaces in filenames break the current behaviour on M20?, an parameter on M20 that returns a richer format instead?

GilesBathgate:
@neslekkim So the change involves code where it used to write to the serial:

```
File opened: {shortname}
Size: {filesize}
```

it now writes:
```
File opened: {longname}
Size: {filesize}
```

Since `File opened` already has a space, I don't see how spaces could be a problem.

GilesBathgate:
@neslekkim I think your comment is for #2934 This PR doesn't change the output of M20

DRracer:
Could you please have a look at PR #3113 ? It should solve the long file names reporting and a bit more.
If you agree, can we close this PR?

PR #2735, opened 2020-06-10
Print progress serial message, only send for active mode

hergtoler:
Not sure what desired operation is here, but this lets you know which mode is active and reduces serial activity, while still providing the progress for the active mode.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -6201,8 +6201,18 @@ Sigma_Exit:
 
 		{
 			const char* _msg_mode_done_remain = _N("%S MODE: Percent done: %d; print time remaining in mins: %d\n");
-			printf_P(_msg_mode_done_remain, _N("NORMAL"), int(print_percent_done_normal), print_time_remaining_normal);
-			printf_P(_msg_mode_done_remain, _N("SILENT"), int(print_percent_done_silent), print_time_remaining_silent);
+#ifdef TMC2130
+			if(tmc2130_mode==TMC2130_MODE_NORMAL) 
+			{
+#endif
+				printf_P(_msg_mode_done_remain, _N("NORMAL"), int(print_percent_done_normal), print_time_remaining_normal);
+#ifdef TMC2130
+			}
+			if(tmc2130_mode==TMC2130_MODE_SILENT) 
+			{
+				printf_P(_msg_mode_done_remain, _N("SILENT"), int(print_percent_done_silent), print_time_remaining_silent);
+			}
+#endif
 		}
 		break;
 


wavexx:
This makes a lot of sense to me, we don't need both !?

arekm:
What's up with this? Makes sense but not merged. Without this we have no idea which estimate is the right one to use.

wavexx:
@leptun what do you think? IMHO this can go into 3.12

leptun:
@wavexx We need to make sure this doesn't break Prusa Link. Not sure if they need both messages

arekm:
prusa/link/printer_adapter/structures/regular_expressions.py:

````
PRINT_INFO_REGEX = re.compile(
    r"^SILENT MODE: Percent done: (?P<progress>-?\d+); "
    r"print time remaining in mins: (?P<time>-?\d+) ?.*$")
````

and only that - not looking for NORMAL mode, so looks like PrusaLink always uses silent mode line.

Fortunately that regex allows adding new things at the end.

"%S MODE: Percent done: %d; print time remaining in mins: %d; active: %d\n"

so leave both lines, add "active: 1" for active, "active: 0" for inactive and prusalink should be still happy while we would know which estimate is a good one.

Other approach - just one new line in output with current mode.

**Comment by wavexx** [comment]
Looks like it would make sense to fix prusalink to match either.

**Comment by leptun** [comment]
What about octoprint support? Are there any popular plugins that make use of this (or maybe something more in octoprint itself)? If yes, then those need to be checked as well to make sure everything is still working.

**Comment by arekm** [comment]
Octoprint itself doesn't use this and doesn't issue M73 g-code.

There is popular https://plugins.octoprint.org/plugins/prusaetaoverride/  plugin which where this issue started. Author of this pull request wanted to improve that plugin, so proper time is used (currently it's hardcoded to look for "NORMAL MODE").

**Comment by arekm** [comment]
Plugin update that adds support for firmware which emit only one MODE line (and also both mode lines) got released.
https://github.com/kanocz/octopi_eta_override/releases/tag/0.2.0

prusalink client code - repository is most likely private, so someone with access needs to fix it

**Comment by arekm** [comment]
Any idea if prusalink code was fixed, so this could be merged into MK3?

**Comment by leptun** [comment]
@arekm I'm trying to push this idea to the Prusa Link devs:
> so leave both lines, add "active: 1" for active, "active: 0" for inactive and prusalink should be still happy while we would know which estimate is a good one.

If I get confirmation that it is ok, I'll modify the PR in that way and also solve the conflict with Marlin_main.cpp

**Comment by leptun** [comment]
@arekm How does this look like for a colorchange print?
```
NORMAL MODE: Percent done: 0; print time remaining in mins: 49; Change in mins: -1; active: 0
NORMAL MODE: Percent done: 0; print time remaining in mins: 49; Change in mins: 22; active: 0
SILENT MODE: Percent done: 0; print time remaining in mins: 49; Change in mins: -1; active: 1
SILENT MODE: Percent done: 0; print time remaining in mins: 49; Change in mins: 22; active: 1
```
This was generated by:
```
M73 P0 R49
M73 C22
M73 Q0 S49
M73 D22
```

With my code, if this wasn't split in 4 commands, you'd get only 2 messages, but unfortunately it's split everywhere for some reason, so there's that. With the MK3 branch, I think you get 8 lines, so this is definitely still an improvement. The way I did it is so it prints the NORMAL message if any normal parameter was provided (P/R/C) and SILENT if any silent parameter was provided (Q/S/D).
Can I push these rebased changes to your PR?

BTW, my changes take up 74B extra of flash compared to the current MK3 branch.

arekm:
That way is also fine. I'll update [prusaetaoverride/](https://plugins.octoprint.org/plugins/prusaetaoverride/) plugin for this format.

(btw this PR is by @hergtoler  but I'm sure any solution that solves the problem will be welcome)

leptun:
@arekm Ah. Didn't see this wasn't your PR :)
In that case, @hergtoler may I push my changes to this PR?

leptun:
I any cases, here are the changes for the time being: https://github.com/prusa3d/Prusa-Firmware/compare/MK3...leptun:Prusa-Firmware:M73_fixes?expand=1

DRracer:
Do I understand correctly, that the serial msg NORMAL/STEALTH will be reported only once based on the current TMC mode?

That means changing the mode on the printer requires an M73 command to occur to be able to report changed time to Octoprint/PrusaLink and will not happen instantly like now?

Do we want that?

3d-gussner:
@hergtoler Thanks for the PR.

I will close it, as some host systems may expect the existing report and I want to avoid issues there.

PR #2736, opened 2020-06-11
Update README.md

mkbel:
Mention gawk dependency for multi language.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,14 +26,28 @@ The firmware for the Original Prusa i3 printers is proudly based on [Marlin 1.0.
 
 1. Clone this repository and checkout the correct branch for your desired release version.
 
-2. Set your printer model. 
+1. Set your printer model. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,14 +26,28 @@ The firmware for the Original Prusa i3 printers is proudly based on [Marlin 1.0.
 
 1. Clone this repository and checkout the correct branch for your desired release version.
 
-2. Set your printer model. 
+1. Set your printer model.
```
DRracer, line 29: numbering got broken

DRracer (changes requested):
Numbering of steps got broken

README.md
mkbel: I removed numbering on purpose, so it is numbered automatically, and doesn't need to be renumbered once another point is added.

DRracer (approved):

PR #2774, opened 2020-07-08
Added nozzle cooldown by enabling part cooling fan

stefan-golinschi:
This speeds up the cooldown process by enabling the cooling fan.


leptun:
But doing it like this will leave the fan on even after everything cools.
Maybe it would be better if instead we used the “wait for cooldown” function that is used by calibrations. That one also disables the fan when it finishes. If this is to be implemented here as well, a quick abort option should exist as well by clicking the knob.

stefan-golinschi:
I agree. `lcd_wait_for_cool_down()` seems to be a better approach.

PR #2776, opened 2020-07-09
Remove numbers from level 1 headers in readme

### LINKED ISSUE #2750: [QUESTION] Why are there numbers in front of some headers?
In `README.md`, there are numbers in front of the `Automated tests`, `Documentation` and `FAQ` headers, but not `Build` and `Table of contents`. Why are there numbers in front of some, but not all level 1 headers? Also, why do these numbers start at 3, instead of 1?

awenelo:
The headers `Automated tests`, `documentation` and `FAQ` all have numbers like they are part of a list. However, These numbers start at 3 and don't appear to correlate to anything else.

Fixes #2750

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,9 +15,9 @@ The firmware for the Original Prusa i3 printers is proudly based on [Marlin 1.0.
      * [Using Arduino](#using-arduino)
      * [Using Linux subsystem](#using-linux-subsystem-under-windows-10-64-bit)
      * [Using Git-bash](#using-git-bash-under-windows-10-64-bit)
-   * [Automated tests](#3-automated-tests)
-   * [Documentation](#4-documentation)
-   * [FAQ](#5-faq)
+   * [Automated tests](#automated-tests)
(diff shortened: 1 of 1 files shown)


DRracer:
@awenelo in fact, the numbers do correspond to something. The level of headers is not kept consistent though:
1. Development environment preparation
2. Source code compilation

awenelo:
The Linux build has its own numbering from 1-5, and in between the `Windows > Using Arduino` section of readme and the `Automated Tests` section is `Using Linux subsystem under Windows 10 64-bit` and `Using Git-bash under Windows 10 64-bit`. I feel like the numbering system is really confusing right now, and not having a numbered list spread out across the page with multiple unnumbered headers in between is the simplest way to help with that.

DRracer:
Marking as 3.10 - we shall improve the whole doc page

3d-gussner:
@awenelo Thanks for the PR, but we have updated the README. Closing

PR #2784, opened 2020-07-29
altfan killswitch

leptun:
Related to #2692 
PFW-1134

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -8570,7 +8570,15 @@ Sigma_Exit:
 	break;
 
     /*!
-	### M999 - Restart after being stopped <a href="https://reprap.org/wiki/G-code#M999:_Restart_after_being_stopped_by_error">M999: Restart after being stopped by error</a>
+    ### M666 - Enter experimental menu
+    Only used by Prusa
+    */
+    case 666:
+        menu_submenu(lcd_experimental_menu);
+    break;
+
+    /*!
+    ### M999 - Restart after being stopped <a href="https://reprap.org/wiki/G-code#M999:_Restart_after_being_stopped_by_error">M999: Restart after being stopped by error</a>
     @todo Usually doesn't work. Should be fixed or removed. Most of the time, if `Stopped` it set, the print fails and is unrecoverable.
     */
     case 999:
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -8570,14 +8570,6 @@ Sigma_Exit:
 	break;
 
     /*!
-    ### M666 - Enter experimental menu
-    Only used by Prusa
-    */
-    case 666:
-        menu_submenu(lcd_experimental_menu);
-    break;
-
-    /*!
     ### M999 - Restart after being stopped <a href="https://reprap.org/wiki/G-code#M999:_Restart_after_being_stopped_by_error">M999: Restart after being stopped by error</a>
     @todo Usually doesn't work. Should be fixed or removed. Most of the time, if `Stopped` it set, the print fails and is unrecoverable.
     */
--- a/Firmware/eeprom.h
+++ b/Firmware/eeprom.h
@@ -359,6 +359,8 @@ static_assert(sizeof(Sheets) == EEPROM_SHEETS_SIZEOF, "Sizeof(Sheets) is not EEP
 | ^					| ^			| ^										| 00h 0			| ^						| LCD backlight mode: __Dim__						| ^				| ^
 | 0x0D30 3376		| uint16	| EEPROM_BACKLIGHT_TIMEOUT				| 01 00 - ff ff | 0a 00h 65535			| LCD backlight timeout: __10__ seconds				| LCD menu		| D3 Ax0d30 C2
 | 0x0D2C 3372		| float		| EEPROM_UVLO_LA_K						| ???			| ff ff ff ffh			| Power panic saved Linear Advanced K value			| ???			| D3 Ax0d2c C4
+| 0x0D2B 3371		| uint8		| EEPROM_ALTFAN_OVERRIDE				| 0-1			| 00h					| ALTFAN override									| LCD menu		| D3 Ax0d2b C1
+| 0x0D2A 3370		| uint8		| EEPROM_EXPERIMENTAL_VISIBILITY		| 0-1			| 00h					| Experimental menu visibility						| LCD menu		| D3 Ax0d2a C1
 
   
 | Address begin		| Bit/Type 	| Name 									| Valid values	| Default/FactoryReset	| Description 										| Gcode/Function| Debug code
@@ -562,9 +564,10 @@ static Sheets * const EEPROM_Sheets_base = (Sheets*)(EEPROM_SHEETS_BASE);
 #define EEPROM_UVLO_LA_K (EEPROM_BACKLIGHT_TIMEOUT-4) // float
 
 #define EEPROM_ALTFAN_OVERRIDE (EEPROM_UVLO_LA_K-1) //uint8
+#define EEPROM_EXPERIMENTAL_VISIBILITY (EEPROM_ALTFAN_OVERRIDE-1) //uint8
 
 //This is supposed to point to last item to allow EEPROM overrun check. Please update when adding new items.
-#define EEPROM_LAST_ITEM EEPROM_ALTFAN_OVERRIDE
+#define EEPROM_LAST_ITEM EEPROM_EXPERIMENTAL_VISIBILITY
 // !!!!!
 // !!!!! this is end of EEPROM section ... all updates MUST BE inserted before this mark !!!!!
 // !!!!!
--- a/Firmware/eeprom.h
+++ b/Firmware/eeprom.h
@@ -561,8 +561,10 @@ static Sheets * const EEPROM_Sheets_base = (Sheets*)(EEPROM_SHEETS_BASE);
 
 #define EEPROM_UVLO_LA_K (EEPROM_BACKLIGHT_TIMEOUT-4) // float
 
+#define EEPROM_ALTFAN_OVERRIDE (EEPROM_UVLO_LA_K-1) //uint8
+
 //This is supposed to point to last item to allow EEPROM overrun check. Please update when adding new items.
-#define EEPROM_LAST_ITEM EEPROM_UVLO_LA_K
+#define EEPROM_LAST_ITEM EEPROM_ALTFAN_OVERRIDE
 // !!!!!
 // !!!!! this is end of EEPROM section ... all updates MUST BE inserted before this mark !!!!!
 // !!!!!
--- a/Firmware/temperature.cpp
+++ b/Firmware/temperature.cpp
@@ -153,6 +153,7 @@ uint8_t fanSpeedBckp = 255;
   uint8_t fanState = 0;
 #ifdef EXTRUDER_ALTFAN_DETECT
   bool extruderFanIsAltfan = false; //set to Noctua
+  uint8_t altfanOverride = 0;
 #endif //EXTRUDER_ALTFAN_DETECT
 #endif
 
@@ -224,6 +225,14 @@ bool extruder_altfan_detect()
 	setExtruderAutoFanState(3);
 
 	SET_INPUT(TACH_0);
+
+	altfanOverride = eeprom_read_byte((uint8_t *)EEPROM_ALTFAN_OVERRIDE);
+	if (altfanOverride == EEPROM_EMPTY_VALUE)
+	{
+		altfanOverride = 0;
+		eeprom_update_byte((uint8_t *)EEPROM_ALTFAN_OVERRIDE, altfanOverride);
+	}
+
 	CRITICAL_SECTION_START;
 	EICRB &= ~(1 << ISC61);
 	EICRB |= (1 << ISC60);
@@ -241,6 +250,18 @@ bool extruder_altfan_detect()
 	setExtruderAutoFanState(1);
 	return extruderFanIsAltfan;
 }
+
+void altfanOverride_toggle()
+{
+    altfanOverride = !altfanOverride;
+    eeprom_update_byte((uint8_t *)EEPROM_ALTFAN_OVERRIDE, altfanOverride);
+}
+
+bool altfanOverride_get()
+{
+    return altfanOverride;
+}
+
 #endif //EXTRUDER_ALTFAN_DETECT
 
 // return "false", if all extruder-heaters are 'off' (ie. "true", if any heater is 'on')
@@ -494,7 +515,7 @@ void setExtruderAutoFanState(uint8_t state)
 	if (fanState & 0x01)
 	{
 #ifdef EXTRUDER_ALTFAN_DETECT
-		if (extruderFanIsAltfan) newFanSpeed = EXTRUDER_ALTFAN_SPEED_SILENT;
+		if (extruderFanIsAltfan && !altfanOverride) newFanSpeed = EXTRUDER_ALTFAN_SPEED_SILENT;
 		else newFanSpeed = EXTRUDER_AUTO_FAN_SPEED;
 #else //EXTRUDER_ALTFAN_DETECT
 		newFanSpeed = EXTRUDER_AUTO_FAN_SPEED;
@@ -1356,7 +1377,7 @@ void temp_runaway_stop(bool isPreheat, bool isBed)
 		SERIAL_ERROR_START;
 		isBed ? SERIAL_ERRORLNPGM(" THERMAL RUNAWAY ( PREHEAT HEATBED)") : SERIAL_ERRORLNPGM(" THERMAL RUNAWAY ( PREHEAT HOTEND)");
 #ifdef EXTRUDER_ALTFAN_DETECT
-		extruderFanIsAltfan = false; //full speed
+		altfanOverride = true; //full speed
 #endif //EXTRUDER_ALTFAN_DETECT
 		setExtruderAutoFanState(3);
 		SET_OUTPUT(FAN_PIN);
@@ -1446,7 +1467,7 @@ void max_temp_error(uint8_t e) {
     WRITE(FAN_PIN, 1);
     WRITE(BEEPER, 1);
 #ifdef EXTRUDER_ALTFAN_DETECT
-    extruderFanIsAltfan = false; //full speed
+    altfanOverride = true; //full speed
 #endif //EXTRUDER_ALTFAN_DETECT
     setExtruderAutoFanState(3);
     // fanSpeed will consumed by the check_axes_activity() routine.
--- a/Firmware/temperature.cpp
+++ b/Firmware/temperature.cpp
@@ -1377,7 +1377,7 @@ void temp_runaway_stop(bool isPreheat, bool isBed)
 		SERIAL_ERROR_START;
 		isBed ? SERIAL_ERRORLNPGM(" THERMAL RUNAWAY ( PREHEAT HEATBED)") : SERIAL_ERRORLNPGM(" THERMAL RUNAWAY ( PREHEAT HOTEND)");
 #ifdef EXTRUDER_ALTFAN_DETECT
-		altfanOverride = true; //full speed
+		altfanOverride = 1; //full speed
 #endif //EXTRUDER_ALTFAN_DETECT
 		setExtruderAutoFanState(3);
 		SET_OUTPUT(FAN_PIN);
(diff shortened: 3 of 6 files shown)


3d-gussner:
Resources previous pulled [PR](https://travis-ci.com/github/prusa3d/Prusa-Firmware/builds/171499649) vs this [PR](https://travis-ci.com/github/prusa3d/Prusa-Firmware/builds/177660451):
| MK3S | 245948 | 6441 | 246266| 6442 | +318 | +1 |
| MK3 | 247148 | 6471 | 247316 | 6471 | +168| +0 |
| MK25S 13a | 219568 | 6264 | 219736 | 6264 | +168| +0 |
| MK25S 10a | 218910 | 6266 | 219088 | 6266 | +178| +0 |
| MK25 13a | 223482 | 6314 | 223644 | 6314 | +0| +0 |
| MK25 10a | 222944 | 6314 | 223006 | 6314 | +0| +0 |

Firmware/eeprom.h
```diff
@@ -359,6 +359,8 @@ static_assert(sizeof(Sheets) == EEPROM_SHEETS_SIZEOF, "Sizeof(Sheets) is not EEP
 | ^					| ^			| ^										| 00h 0			| ^						| LCD backlight mode: __Dim__						| ^				| ^
 | 0x0D30 3376		| uint16	| EEPROM_BACKLIGHT_TIMEOUT				| 01 00 - ff ff | 0a 00h 65535			| LCD backlight timeout: __10__ seconds				| LCD menu		| D3 Ax0d30 C2
 | 0x0D2C 3372		| float		| EEPROM_UVLO_LA_K						| ???			| ff ff ff ffh			| Power panic saved Linear Advanced K value			| ???			| D3 Ax0d2c C4
+| 0x0D2B 3371		| uint8		| EEPROM_ALTFAN_OVERRIDE				| 0-1			| 00h					| ALTFAN override									| LCD menu		| D3 Ax0d2b C1
```
3d-gussner, line 362: Default value if `00h 0`

Firmware/eeprom.h
```diff
@@ -359,6 +359,8 @@ static_assert(sizeof(Sheets) == EEPROM_SHEETS_SIZEOF, "Sizeof(Sheets) is not EEP
...
+| 0x0D2A 3370		| uint8		| EEPROM_EXPERIMENTAL_VISIBILITY		| 0-1			| 00h					| Experimental menu visibility						| LCD menu		| D3 Ax0d2a C1
```
3d-gussner, line 363: see above

Firmware/Marlin_main.cpp
```diff
@@ -8570,7 +8570,7 @@ SERIAL_PROTOCOLPGM("\n\n");
 	break;
 
     /*!
-	### M999 - Restart after being stopped <a href="https://reprap.org/wiki/G-code#M999:_Restart_after_being_stopped_by_error">M999: Restart after being stopped by error</a>
+    ### M999 - Restart after being stopped <a href="https://reprap.org/wiki/G-code#M999:_Restart_after_being_stopped_by_error">M999: Restart after being stopped by error</a>
```
3d-gussner, line 8573: What is wrong with the spacing?

Firmware/eeprom.h
3d-gussner: Did we want to move the experimental to another EEPROM space?

Firmware/eeprom.h
```diff
@@ -561,8 +563,11 @@ static Sheets * const EEPROM_Sheets_base = (Sheets*)(EEPROM_SHEETS_BASE);
 
 #define EEPROM_UVLO_LA_K (EEPROM_BACKLIGHT_TIMEOUT-4) // float
 
+#define EEPROM_ALTFAN_OVERRIDE (EEPROM_UVLO_LA_K-1) //uint8
```
3d-gussner, line 566: see above

Firmware/ultralcd.cpp
```diff
@@ -2096,6 +2096,9 @@ static void lcd_preheat_menu()
 //! 	@code{.unparsed}
 //! 	| Voltages           |	MSG_MENU_VOLTAGES
 //! 	@endcode
+//!
+//!
+//! 	| Experimental       |	c=18 r=1
```
3d-gussner, line 2101: no need for `r=1` as it is only one row

Firmware/ultralcd.cpp
```diff
@@ -2210,6 +2222,12 @@ static void lcd_support_menu()
   MENU_ITEM_SUBMENU_P(_i("Voltages"), lcd_menu_voltages);////MSG_MENU_VOLTAGES c=18 r=1
 #endif //defined VOLT_BED_PIN || defined VOLT_PWR_PIN
 
+  if (_md->experimental_menu_visibility)
+  {
+    MENU_ITEM_SUBMENU_P(PSTR("Experimental"), lcd_experimental_menu);
```
3d-gussner, line 2227: missing `//MSG_MENU_EXPERIMENTAL c=18` or so

Firmware/ultralcd.cpp
```diff
@@ -9160,3 +9185,25 @@ void lcd_crash_detect_disable()
...
+    eeprom_update_byte((uint8_t *)EEPROM_EXPERIMENTAL_VISIBILITY, oldVal);
+}
+
+void lcd_experimental_menu()
+{
+    MENU_BEGIN();
+    MENU_ITEM_BACK_P(_T(MSG_BACK));
+
+#ifdef EXTRUDER_ALTFAN_DETECT
+    MENU_ITEM_TOGGLE_P(_N("ALTFAN det."), altfanOverride_get()?_T(MSG_OFF):_T(MSG_ON), altfanOverride_toggle);
```
3d-gussner, line 9205: missing `//MSG_MENU_ALTFAN c=18` or so

Firmware/temperature.cpp
```diff
@@ -1356,7 +1377,7 @@ void temp_runaway_stop(bool isPreheat, bool isBed)
 		SERIAL_ERROR_START;
 		isBed ? SERIAL_ERRORLNPGM(" THERMAL RUNAWAY ( PREHEAT HEATBED)") : SERIAL_ERRORLNPGM(" THERMAL RUNAWAY ( PREHEAT HOTEND)");
 #ifdef EXTRUDER_ALTFAN_DETECT
-		extruderFanIsAltfan = false; //full speed
+		altfanOverride = true; //full speed
```
DRracer: altfanOverride is an uint8_t ... prefer not assigning true/false into non-bool variables

Firmware/temperature.cpp
```diff
@@ -153,6 +153,7 @@ uint8_t fanSpeedBckp = 255;
   uint8_t fanState = 0;
 #ifdef EXTRUDER_ALTFAN_DETECT
   bool extruderFanIsAltfan = false; //set to Noctua
+  uint8_t altfanOverride = 0;
```
DRracer: theoretically, we could merge extruderFanIsAltfan and altfanOverride into one var to save 1B of RAM, but at the cost of a slightly more complex code elsewhere.
However, checking if we have an altfan and have an override may actually get shorter:
`if (extruderFanIsAltfan && !altfanOverride) ` into something like `if (altfan == 1)` in case:
#define extruderFanIsAltfanMask 1
#define altfanOverrideMask 2
... if you know what I mean...

Firmware/ultralcd.cpp
```diff
@@ -2108,11 +2111,12 @@ static void lcd_preheat_menu()
 static void lcd_support_menu()
 {
 	typedef struct
-	{	// 22bytes total
+	{	// 23bytes total
 		int8_t status;                 // 1byte
 		bool is_flash_air;             // 1byte
 		uint8_t ip[4];                 // 4bytes
 		char ip_str[3*4+3+1];          // 16bytes
+        uint8_t experimental_menu_visibility; //1byte
```
DRracer: indentation ... there are tabs in the lines above ;)

DRracer (changes requested):

DRracer (approved):

3d-gussner (commented):
Stupid me forgot to submit comments

Firmware/Marlin_main.cpp
leptun: It was indented using tabs. The surrounding code is with spaces. Looking again in that area of code, it is extremely inconsistent.

Firmware/ultralcd.cpp
leptun: I just copied how the Debug menu was documented. Changing now.

Firmware/ultralcd.cpp
leptun: It is not translated, so I assumed it doesn't need it.

Firmware/ultralcd.cpp
leptun: same as above

Firmware/ultralcd.cpp
3d-gussner: Still good practice to document it in case it will be translated or merged to `messages.*`

Firmware/ultralcd.cpp
3d-gussner: Same as above

vertigo235:
How do you enable the experimental menu? where is it located

leptun:
@vertigo235 The experimental menu can be enabled by going into the "HW Setup" menu. While in the menu you have to long-press in order to toggle the Experimental menu visibility. The menu will appear at the end of the Support menu. The visibility and the ALTFAN override are persistent and can only be changed easily via the menu. If you really want to do the changes via gcode, you can use the EEPROM read/write Dcode.

vertigo235:
I see, yes, that's what I deduced but there is no indication in the hw menu that it was enabled.  

Tried it a few times and it is there now.  Perhaps the longpress should be in the support menu not in the hw menu?

vertigo235:
or maybe the experimental menu should be part of HW Setup

wavexx:
Can we have a self-destruct option in the Experimental menu?
Every machine needs to have this option ;)

vertigo235:
Gotta have a 'DO NOT USE' toggle in there for sure.

vertigo235:
I think that up, up, down, down, left, right, left, right, B, A , Start would have been most appropriate to enable the powerup menu

wavexx:
I kid I kid.. but daamn, these would be 5 star easter eggs :)

gregsaun:
Is the reduction speed activated after we flash the firmware on our current Prusa printers?

leptun:
Yes. You have to manually disable the altfan detection from the menu in order to have it at 100% speed

gregsaun:
So despite the concerns we have reported and other alternatives we have proposed to you, Prusa has decided to deliberately destroy some hotend fans used by the community. Wonderful...

ogland:
This is outrageous and evil, since you at Prusa are aware of what it will do and still chose this. The message you are sending again and again is extremely hostile towards modifications.

Panayiotis-git:
Just an idea: What if, when the altfan is detected, a message is shown to the user informing about the speed change and where to navigate to revert the change? Perhaps a choice to directly go to that page?

(Sorry, I cannot view the sources so I cannot check if this is even doable with the current implementation.)

CacikeOtoao:
This has to be a joke, there is no way that you would implement a piece of code that will instantly damage a third party piece of equipment that was installed properly. If indeed that is the way this works you might be exposing yourself to possible consumer law infractions in some places. You HAVE to make sure there is a warning in the LCD for the people that won’t be aware of this upon flashing.

3d-gussner:
I know people will hate me for this, but here my 2ct:
- Prusa provides a firmware that is tailored around their printers to perform in their specs. So if they decide that their chosen ALTFAN at 50% speed is sufficient to cool the hotend and quiet enough than it is up to them to tailor their firmware.

- Everyone that modifies the hardware/electronics/design should be prepared to also to modify the firmware.
   - I have only one Prusa MK3s left "stock" all others are modded and some need modified firmware. 
   - I can't expect from any company to suite all wishes and if I change something I have to deal with it.

- Some may argue that they changed the Noctua fan as it performs (in their opinion / including me in some cases) poorly and now flashing the "stock" firmware they run only at 50%. I changed my hotend fan at some Prusa printers and so I have to do make sure they operate like I expect them to:
  - I can comment out `#define EXTRUDER_ALTFAN_DETECT` in the variation file.
  - I can change `#define EXTRUDER_ALTFAN_SPEED_SILENT 128` to `255` to have full speed again.
  - I can use this "hidden" menu to change the behavior.
- I am OUT OF SPECS with my machine mods, so **I** have to do something about it.


> So despite the concerns we have reported and other alternatives we have proposed to you, Prusa has decided to deliberately destroy some hotend fans used by the community. Wonderful...

@gregsaun The user has chosen to got OUT OF SPECS with their mod. Prusa added an option, maybe not the best but at least they listened to the community and added it to their "stock" firmware. And why should this change destroy fans?



> This is outrageous and evil, since you at Prusa are aware of what it will do and still chose this. The message you are sending again and again is extremely hostile towards modifications.

@ogland What is outrageous or evil about that Prusa adds an option the switch the behavior of an ALTFAN without need to compile you own custom firmware.


> This has to be a joke, there is no way that you would implement a piece of code that will instantly damage a third party piece of equipment that was installed properly. If indeed that is the way this works you might be exposing yourself to possible consumer law infractions in some places. You HAVE to make sure there is a warning in the LCD for the people that won’t be aware of this upon flashing.

@CacikeOtoao
1. The code doesn't instantly damage instantly anything! Their choice of ALTFAN at 50% has been tested and approved to operate as well as the previous one.
2. Glad that I am not living in a place that this kind of things have a chance to get to court. You modify a printer fan so you are out of specs using a not correct/tailored to NEW specs firmware.


Things could be handled differently BUT these kind of comments aren't helping at all.

Please make constructive comments and provide ideas like Panayiotis-git did.
Don't expect Prusa to solve your selfmade issues.

If you need custom firmware checkout vertigo235/zaribo/my github and other places.

3d-gussner:
I am big fan of Mods BUT these need sometimes custom firmware:
- Love the Bear but needs a modified firmware as the MK3 Extruder R4++ design is different than what Bear/Zaribo/Bondtech using (based on R3 diamensions)
- Bondtech users need a custom firmware as the dimensions are different.
- Love the faster Sunon and LDO part cooling fans as they are way better performing BUT they are defiantly louder. I am OK with louder but Prusa MK3s users expect a "silent" printer.
- Love the work and effort vertigo235 and others do to make MODs accessible to a wide range of people that aren't that comfortable to compile their firmware.

Love this community and it make me really SAD and disappointed when these kind of flares and negative non-constructive messages pop-up.

gregsaun:
> @gregsaun The user has chosen to got OUT OF SPECS with their mod. Prusa added an option, maybe not the best but at least they listened to the community and added it to their "stock" firmware. And why should this change destroy fans?

I 100% agree with the out of specs, no problem with this. However, there are simple solution to avoid this that took much less time to implement than all those talks have generated...

CacikeOtoao:
@3d-gussner so for clarification purposes. If I have a MK3S with a Sunon fan and I flash 3.9.1 with the current implementation been cited here, the Sunon will not suffer any damage. It will only spin at 4,000RPM until I enable the AltFan function from the secret menu?

wavexx:
@3d-gussner I'll add my (serious) view of this: you don't need to go "out of spec" to run into this issue.

- Noctua breaks down, order a replacement
- Slap on the first compatible 40mm fan you have while waiting to avoid downtime

This is IMHO 100% legitimate and done with the best intentions (matching power/cfm requirements).
I know I did it on other printers a number of times.

gregsaun:
@josefprusa you said outside github: _it autodetects is for users who have it installed while updating and you will be able to switch this on/off in the menu._ -> Do you mean the that when we flash the firmware it autodetects we are using a Sunon and ask you if you want to disable it?

I am sorry but I understand this differently than the answer from @leptun I had here: https://github.com/prusa3d/Prusa-Firmware/pull/2784#issuecomment-670665473. To me this means that current users will have to manual change this setting in a hidden menu otherwise it will BY DEFAULT damage the fan.

If you warn the current user and give him the choice to NOT deliberately damage the fan then I don't see the problem.

Could we have a precision on how this works then?

vertigo235:
@gregsaun That's not how it worked when I flashed it, the FW certainly does detect it and changes your fan speed, but there is no warning or notification to the user as far as I could tell. 

You would have to know to activate the secret menu, and then how, enable to power your machine up again. 

It's really easy to not know this even if you are an experienced user, I had built the FW at least 3 times and never noticed the slower fan speed until another community member told me about it.

gregsaun:
@vertigo235 So why @josefprusa is angry at me then?

I have to warn my community that they might damaged permanently their fan on the next firmware upgrade. I should have kept this silent and let my customers damage their fan because they didn't see this little hidden menu and the reduction of speed?

3d-gussner:
Easier can be seen differently by people.

Prusa has their point to not make this standard 100% fan speed as it fulfilled their specs for the fan of choice at 50%. Also make it non visible to the majority of their users is a good point. 

Modders would like to have it done differently and everything customizable as possible.

So what is the correct way to do this? There is no ONE way to satisfy everyone.

Where I am very disappointed in is the reaction of the some people of community ranting about a compromise Prusa tried to fit both needs.

Standard user want to use realible hardware and software and not be overwhelmed by the choices they have to make.

Modders should be glad that they can deactivate the slow 50% fan to 100% in stock firmware even it is bit tricky to do that. And it has to be done ONCE.

I personally was surprised that Prusa considered this option and spent developing time on something that is out of their specs.

I live in a country where people first complain and think later. 

The 3D printing community is better than this.

gregsaun:
Yes so what is the problem if I inform my community that they MIGHT DAMAGE THE FAN?

vertigo235:
FWIW I'm OK with the solution, but all the builds that I make will have ALTFAN set to 255 as god intended, so the ALTFAN killswitch will effectively do nothing anyhow.

vertigo235:
@gregsaun As long as you communicate that they may damage the fan unless they disable the ALTFAN detection, then I can't see how there is any problem with that.  The ability to disable the "feature" in the stock FW has been addressed. 

Thank you Prusa team for listening to the community.

gregsaun:
@vertigo235 I did obviously provide an explanation on how to activate the hidden menu to not damage the fan. I think it is important to also warn the customers before the firmware is available to be sure everyone gets the info in time.

vertigo235:
As it is an unreleased feature, things can change based on feedback (as it already has), so we should give the Prusa team time to take take the feedback and perhaps adjust before the final release.  

To that note, maybe it was premature to sound a general alarm since the feature set is not complete yet for the next release, at least I think?

3d-gussner:
@gregsaun maybe you should advise your users to use the @vertigo235 Bear firmware as it is well maintained and very fast updated.
I will also add the 100% altfan to the Zaribo firmware as most printers will have sunon hotend Fans.

gregsaun:
@vertigo235 I agree but we have discuss this since several weeks now, we have a merge here, a confirmation from the developer how it will work and how it will be done in the menu... I think it was the moment to inform my community on how the firmware evolve. I personally know that they don't read release notes so I did prefer to be safer than sorry.

gregsaun:
> @gregsaun maybe you should advise your users to use the @vertigo235 Bear firmware as it is well maintained and very fast updated.
> I will also add the 100% altfan to the Zaribo firmware as most printers will have sunon hotend Fans.

That was obviously part of my post, explaining the future of the Bear calibration firmware and that we are going to change how things are made. I was waiting on the developer program to make this smoother for the community but this is not available today.

josefprusa:
@gregsaun Autodetection is what we agreed on with @DRracer who is the FW team leader at the time when we were first planning how to implement it and that is how it is going to be. I am sorry that external devs don’t know everything, next time confront @DRracer. Stop your witch hunt until the beta of the FW is out.

gregsaun:
@josefprusa then it is on your side Jo, I've asked here before positing to be sure it was the correct behavior. I have just taken the decision to be safer than sorry regarding my customers.

vertigo235:
I'm a little confused now , does that mean this feature is incomplete? Is it not behaving as intended?

gregsaun:
@DRracer sorry you are in this position, I really appreciate all your quality work and I am deeply convinced that Prusa is very lucky to have you.

vertigo235:
If your external developers don't know what is going on then you certainly can't expect the general community to know either.

vertigo235:
Was it a mistake in assuming this is the proper forum for customers to provide feedback and testing results from merged code in the Prusa FW?  

If we do this we are accused of being witch hunters?

gregsaun:
If a customer has a hotend fan running at 5000RPM, will it go down to 2500RPM?

gregsaun:
If we reset the firmware, in what mode we will be then?

vertigo235:
Nobody resets their firmware after updating so resetting EEPROM upon upgrading is a poor assumption that your users will do that. 

Is that part of the recommendation as you update the FW? I could be out of process.

vertigo235:
I would not think that I would be expected to re calibrate all my sheets etc after updating my FW

gregsaun:
@vertigo235 no but you might have a customer resetting the firmware after seeing a weird behavior.

Issue #2760, opened 2020-06-25
[BUG]External perimmiters with 0.6mm nozzle


2Kasi2:
**Printer type** - MK3S
**Printer firmware version**-  3.9.0-3421

**SD card or USB/Octoprint**
  I used SD card

**Describe the bug**
  Well there are 2 issues, but they are both with external perimiters on my prints. 
1st issue is when i use .3mm layer height with .6mm nozzle the "seam" is terrible. It looks like the nozzle starts to extrude a bit too late and therefore there is a gap(see picture of skull). Also the seam function doesnt properly put the seam where it is supposed to. I set it to Rear, but as you can see some of it was in the front!
Second issue is ugly external perimiter on 1 side of the print, and they might be related because it kinda widens as the Zaxis rises. I found that each new layer on top of previous one was a bit worse, both with the bad seam and the ugly finish.
Also with the secong print you can see that the cilinder which is in one corner kinda sticks out of the print(see pics)

**To Reproduce**
 The bad seam I used both .2 and .3mm layer heights with .6mm nozzle. It happend both while using Prusa PLA white, and PETG prusament Gold. The PLA was much better as the PETG. 
The second issue which might 
  
**G-code**
First Gcode is of the Skull, printed in PETG, it had the wide and wrongly placed seam.
Second is a pot and the seam here seems to look better, as the "seams" (or whatever you call them) are on the corner. But one side of the print has this ugly finish. 
(cpuldnt upload Gcode directly so i converted it to .txt)

** Actual gcodes**
[second_issue.txt](https://github.com/prusa3d/Prusa-Firmware/files/4832419/second_issue.txt)
[issue_one.txt](https://github.com/prusa3d/Prusa-Firmware/files/4832427/issue_one.txt)

**pictures of the Skull (issue one)**
![IMG_20200625_175338](https://user-images.githubusercontent.com/67429462/85756145-d89fd500-b70e-11ea-8f8e-838e2744f13f.jpg)
![IMG_20200625_175352](https://user-images.githubusercontent.com/67429462/85756155-d9d10200-b70e-11ea-8651-aeaccf47d905.jpg)
![IMG_20200625_175326](https://user-images.githubusercontent.com/67429462/85756157-da699880-b70e-11ea-90bb-88623be8375c.jpg)
![IMG_20200625_175342](https://user-images.githubusercontent.com/67429462/85756163-db022f00-b70e-11ea-8a78-954699a46be9.jpg)

[...]

Issue #2757, opened 2020-06-23
[BUG] FW 3.9.0 perimeter quality issue


wavexx:
**Printer firmware version**-  3.9.0

**Describe the bug**
As reported by @rcalis, FW 3.9.0 seem to cause issues in perimeters:

![](https://user-images.githubusercontent.com/19313838/83067027-d0268280-a01a-11ea-8112-c6ef8f9def9c.JPG)

In this issue I'd like to investigate _this_ specific pattern, which seems to be caused by higher E step/mm counts (geared extruders such as the BMG).

The Prusa-Slicer 3mf file as well as the generated gcode is important. If using the BMG, also please report the detailed machine settings which might be set in the FW by using M503

haarp:
On a Bondtech BMG, the default E usteps of 32 and 830 steps-per-mm produced a very obvious such pattern. This was alleviated somewhat by reducing it to 16/415. No further improvements were visible by lowering it again to 8/207.5.

Sorry, I don't have any comparison pictures.

wavexx:
Very useful to know. Do you have a model where I can try this? I also have a BMG

haarp:
> Very useful to know. Do you have a model where I can try this? I also have a BMG

Yes. It's either left or right handle of [this model](https://www.thingiverse.com/thing:3750319/files) Only the first 5mm or so are relevant. My print settings were PLA, 0.28mm layers, rear seams. Firmware 3.8 prints it fine, 3.9 shows problems at the chamfer on the bottom, which are very obvious with 32 e-usteps.

I'm also attaching the gcode I've used, but it's truncated to 5mm or so. The full print settings are at the bottom of the file.
[handle+clip_0.4n_0.28mm_PLA_MK3.gcode.zip](https://github.com/prusa3d/Prusa-Firmware/files/4820155/handle%2Bclip_0.4n_0.28mm_PLA_MK3.gcode.zip)

wavexx:
So I had an initial guess about this issue being the result of dual/quad stepping within LA, although at least with the model from @rcalis I have I couldn't really trigger this even with the BMG.

It's still a bug though, so I fixed this in my branch nonetheless. Would you be willing to try this in parallel? Do you know how to build the FW youself or need some help?

haarp:
> Would you be willing to try this in parallel? Do you know how to build the FW youself or need some help?

[...]

Issue #2751, opened 2020-06-20
[BUG] Updated to 3.9.0 Firmware (Mk3s), and caused bad bridging and seam issue?


john5634:
Tried to run old gcode for previous firmware, but it just makes the prusa restart?

So i re sliced the files using the 3mf that originally generated the gcode, and the print now looks terrible.

Any suggestions? The issue is in two places (edge going up seems to be a seam setting issue, even though I didn't change the 3mf settings, and the bridging is now terrible in the top center hole)

The one of the left is after the new firmware.

MK3S
3.9.0

No Upgrades or changes

SD card

[older-GCODE-that-makes-the-printer-restart_0.25mm_PLA_MK3S_3h41m.txt](https://github.com/prusa3d/Prusa-Firmware/files/4807329/older-GCODE-that-makes-the-printer-restart_0.25mm_PLA_MK3S_3h41m.txt)

Link to 3mf that was used for perfect prints on old firmware, that no longer makes perfect prints

https://drive.google.com/file/d/1SJ_oDQvqbeupTQVIlz9YH-Ol__7fLIA7/view?usp=sharing

IMAGE SHOWING THE ISSUES:
![2020-06-19](https://user-images.githubusercontent.com/12575407/85190666-6c455180-b288-11ea-9f06-ae7a0371f85a.jpg)

3d-gussner:
Thanks for detailed bug report. I think it is kind of duplicate of #2693

Can you check out https://github.com/prusa3d/Prusa3D-Test-Objects/issues/13
And report back your results?
Would be great if you also could have a look at https://github.com/prusa3d/Prusa3D-Test-Objects/issues/14 as the included firmware has both potential fixes.

DRracer:
All this should be fixed in FW 3.9.1 final, can we close this issue @john5634 ?

john5634:
Yeah! Of course! Thank you so much.

Issue #2674, opened 2020-05-21
[BUG]3.9.0 Final, retraction problems.


Glasofruix:
**Printer type** -MK3S
**Printer firmware version**-  3.9.0

**Describe the bug**
Very stringy prints with identical gcode used on 3.8.1 and 3.9.0 RC3 to the point of parts fusing together. No problem whatsoever before the upgrade.
  
**To Reproduce**
Just print something
  
**Expected behavior**
Little to no wispy strings
  
Here's a pic:

https://i.imgur.com/h4UY3NZ.jpg

awenelo:
Can you please attach your Gcode? I suspect this has something to do with the upgrade to LA1.5, resulting in different K values.

Related to #2543?

DRracer:
@Glasofruix looking at the photo you posted - the filament is very wet and the temperature is too high - the skirt is full of tiny bubbles and the whole print is not shiny.

Glasofruix:
> 
> 
> @Glasofruix looking at the photo you posted - the filament is very wet and the temperature is too high - the skirt is full of tiny bubbles and the whole print is not shiny.

It's not supposed to be shiny it's a matte filament, the spool is barely a month old and was properly stored with dessicant, the skirt always looked like that. My prints were fine before i upgraded to 3.9.0. I'll try to dry it though and continue testing with different filaments. It's would be funny if it's just a coincidence.

DRracer:
@Glasofruix ok, thanks for additional info. Now please post your gcode so that we can try to reproduce your issue. Also the brand and filament type would be helpful

mullcom:
Did you solve your problem. for me i get some blobs on my print with petG and  just open it. it seams that some times it push out to much filiment. try to change values. i get better result of higher speed but still blob coming.

DRracer:
All this should be fixed in FW 3.9.1 final, can we close this issue @Glasofruix ?

Prusa-Support:
Thanks for reporting.

This issue has been awaiting for response until now, moreover, it was supposedly fixed. Therefore, the issue will be closed now.

Michele Moramarco
Prusa Research

PR #2785, opened 2020-07-29
Linear Advance 1.5 Fixes

wavexx:
This PR fixes multiple issues with LA1.5 for both stock and geared extruders, as discovered and reported by multiple users in #2543, #2674, #2693, #2751, #2757, #2760.

Major issues covered:

- A bad timer check prevented fast LA ticks to be scheduled, effectively limiting LA as speed increased or with shorter intervals due to geared extruders (173aa2d)
- A bad sign check would cause over-retraction as extrusion speed/nozzle size increased (c08f37d, fb5f09d)
- Tighter accounting for advance ticks (since #2591) lead to the discovery of two more subtle issues:
  - Bad compression factors calculated during segments without cruising would now propagate forward instead of being ignored, generally leading to bad corners and a final over-retract. Fixed by correctly calculating the compression factor at the junction (1554895, 753e651, 7c140bc, a36efcb)
  - Numerical errors in short segments (<0.1mm) would cause LA to continuously re-adjust the pressure in high-res models, starving the planner and further reducing the quality of the print in a vicious loop.
    Since the error can be large and difficult to overcome without further performance penalties, this was fixed by artificially limiting LA *back* to the acceleration/deceleration phase so that short segments can necessarily only have a limited influence. The effective compression will still converge to an average  result which is more stable than what is being calculated currently (c08f37d, c54474f). This still preserves the ability to handle chained wipes, but now is doing so only when wiping (no extrusion).

The code scheduling advance ticks has been simplified and speed-up (50a0982, feafc5e). Other commits only perform minor cleanups.

Notably, There is now also an optional automatic LA adjustment for flow control (9b8f642) which can change the behavior of M221 or the FLOW settings in the LCD. Since this behavior is highly dependent on the intended behavior of the adjustment, we believe that keeping this disabled is actually the intended behavior. See a08ca19 for a discussion.

PFW-1125

### DIFF
--- a/Firmware/Configuration_adv.h
+++ b/Firmware/Configuration_adv.h
@@ -288,6 +288,7 @@
   #define LA_K_DEF    0        // Default K factor (Unit: mm compression per 1mm/s extruder speed)
   #define LA_K_MAX    10       // Maximum acceptable K factor (exclusive, see notes in planner.cpp:plan_buffer_line)
   #define LA_LA10_MIN LA_K_MAX // Lin. Advance 1.0 threshold value (inclusive)
+  //#define LA_FLOWADJ         // Adjust LA along with flow/M221 for uniform width
   //#define LA_NOCOMPAT        // Disable Linear Advance 1.0 compatibility
   //#define LA_LIVE_K          // Allow adjusting K in the Tune menu
   //#define LA_DEBUG           // If enabled, this will generate debug information output over USB.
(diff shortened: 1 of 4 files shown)


3d-gussner:
Resources previous pulled [PR](https://travis-ci.com/github/prusa3d/Prusa-Firmware/builds/171499649) vs this [PR](https://travis-ci.com/github/prusa3d/Prusa-Firmware/builds/177709952):
| MK3S | 245948 | 6441 | 246418 | 6443 | +470| +2 |
| MK3 | 247148 | 6471 | 247486 | 6473 | +338| +2 |
| MK25S 13a | 219568 | 6264 | 219952 | 6266 | +384| +2 |
| MK25S 10a | 218910 | 6266 | 219406 | 6268 | +496| +2 |
| MK25 13a | 223482 | 6314 | 223860 | 6316 | +378| +2 |
| MK25 10a | 222944 | 6314 | 223454 | 6316 | +510| +2 |

DRracer (approved):
approving for FW 3.9.1 based on the numerous tests performed by us and the community

PR #2789, opened 2020-08-02
Use/enable TMC2130 DEDGE support in MK3/MK3S

wavexx:
The TMC2130 support Double Edge stepping aka ``DEDGE`` where both the rising and lowering edge of the stepping pin count as a step instead of requiring a full cycle.

DEDGE is advantageous as it removes one extra toggle for each stepper pulsed (plus a delay, in some circumstances), saving a few µs here and there.

We introduce 2 macros to step: STEP_NC_HI and STEP_NC_LO.

In normal mode, both perform as their name implies. In DEDGE mode instead STEP_NC_HI inverts the line, while STEP_NC_LO becomes a no-op. Both modes are still fully supported. The macros are specialized for each axis, and still respect INVERT_?_AXIS as defined in the variant file.

We also introduce a few constant definitions the delays required by the TMC2130, and use them when appropriate instead of hard-coding delays:

- step pauses (2µs minimum)
- direction changes (100µs minimum)

Existing code that was manipulating the pins directly has been changed to use either the macros or the helper functions defined in tmc2130.h. We also fix the behavior of TMC2130_INTPOL_* which was previously ignored in the variant files.

PFW-1192


DRracer:
@wavexx @vintagepc is this ready to get merged into MK3?

Firmware/tmc2130.cpp
```diff
@@ -678,25 +696,32 @@ static uint8_t tmc2130_rx(uint8_t axis, uint8_t addr, uint32_t* rval)
...
-#define _DO_STEP_Y      { WRITE(Y_STEP_PIN, !INVERT_Y_STEP_PIN); asm("nop"); WRITE(Y_STEP_PIN, INVERT_Y_STEP_PIN); asm("nop"); }
-#define _DO_STEP_Z      { WRITE(Z_STEP_PIN, !INVERT_Z_STEP_PIN); asm("nop"); WRITE(Z_STEP_PIN, INVERT_Z_STEP_PIN); asm("nop"); }
-#define _DO_STEP_E      { WRITE(E0_STEP_PIN, !INVERT_E_STEP_PIN); asm("nop"); WRITE(E0_STEP_PIN, INVERT_E_STEP_PIN); asm("nop"); }
+#ifdef TMC2130_DEDGE_STEPPING
+#define _DO_STEP_X      TOGGLE(X_STEP_PIN)
+#define _DO_STEP_Y      TOGGLE(Y_STEP_PIN)
+#define _DO_STEP_Z      TOGGLE(Z_STEP_PIN)
+#define _DO_STEP_E      TOGGLE(E0_STEP_PIN)
+#else
+#define _DO_STEP_X      { WRITE(X_STEP_PIN, !INVERT_X_STEP_PIN); delayMicroseconds(TMC2130_MINIMUM_PULSE); WRITE(X_STEP_PIN, INVERT_X_STEP_PIN); }
```
DRracer: calling/inlining delayMicroseconds will definitely take more time than the original "nop" instruction, why was this changed?

vintagepc: Looks like it's to support variable minimum pulse times. Could probably special-case it in a #define so that if the pulse time is 0 it just `nop`s but otherwise adds a delay call

leptun: Do we really need that nop? It’s not like it makes any real difference

wavexx: I can make an exception for TMC2130 to use the nop as before, much nicer.

_DO_STEP* is not used by the stepping code, this is essentially just for XYZCAL. But removing the the call is still nice ;)

vintagepc: > Do we really need that nop? It’s not like it makes any real difference

Preventing Compiler optimizations that could eliminate the actual change, perhaps?

wavexx: Note that some delay is necessary by the spec. A nop seems to be always sufficient on the Einsy, I wouldn't remove it completely. On an externally wired 2130 the TMC2130_MINIMUM_PULSE is fully required.

leptun: It is not optimized away. I've done 8MHz toggles before. I'm pretty sure the registers are defined as volatile.
As for external drivers, never had issues with missed step pulses, even with 20cm wires and horrible board design

wavexx: Ok, I'll remove the nop then. I actually ran into missed step though in my case

wavexx: (on an external 2130)

wavexx: Ok, I removed all delays entirely. @leptun could you try a build _without_ DEDGE and see if everything is still a-ok?

There are still a couple of "nops" in sm4.c for XYZCAL.
We could go the extra mile and remove those as well.

leptun: I wanted do say that there should be no NOP when DEDGE is used. With regular stepping it is unfortunately needed as you've said. I thought you meant to keep the delay with DEDGE, which didn't make sense to me. There should only be delays between the HIGH->LOW transition of the nonDEDGE mode

wavexx: I'll refine this further in a few minutes

wavexx: Ok, the regular variant now restores the nop instruction, but still allows to use delayMicroseconds when not using TMC2130. DEDGE removes all details (as it was the case before).

wavexx:
Hopefully I've ironed out all concerns. The regular (non-DEDGE) variant now uses a single nop instruction as it used to, but still allows variable delays if needed.

Tweaked sm4.cpp to use the same efficient toggle mechanism as used by stepper.cpp.

I've tested both variants merged on the current master and both work as expected, including XYZcal.

Firmware/tmc2130.h
```diff
@@ -29,6 +29,18 @@ extern uint8_t tmc2130_sg_homing_axes_mask;
 #define TMC2130_WAVE_FAC1000_MAX 200
 #define TMC2130_WAVE_FAC1000_STP   1
 
+#define TMC2130_MINIMUM_PULSE 0   // minimum pulse width in uS
+#define TMC2130_SET_DIR_DELAY 20  // minimum delay after setting direction in uS
+#define TMC2130_SET_PWR_DELAY 0   // minimum delay after changing pwr mode in uS
+
+#ifdef TMC2130_DEDGE_STEPPING
+#define TMC2130_MINIMUM_DELAY //NOP
```
DRracer, line 37: Personally I prefer `/*NOP*/`, because that terminates the comment when placed into the source code by the preprocessor

DRracer (approved):

Firmware/tmc2130.h
wavexx: Isn't the comment already eaten during parsing? (the expression should expand to nothing)

wavexx:
```
#define A b //NOP
A C
```
```
% gcc -E test.c
# 1 "test.c"
# 1 "<built-in>"
# 1 "<command-line>"
# 31 "<command-line>"
# 1 "/usr/include/stdc-predef.h" 1 3 4
# 32 "<command-line>" 2
# 1 "test.c"

b C
```

Looks like (I was going from memory)

DRracer: ok

PR #2791, opened 2020-08-05
Fix stack smashing in temperature/fsensor ISR

wavexx:
The temperature and fsensor ISR re-enable interrupts while executing.

However, we still need to protect the epilogue of the ISR so that
the saved return address is not altered while returning.

We hoist the body of the function out of the isr in both cases for
clarity (and to avoid a stray return bypassing the lock/cli), so that
the re-entrant portion is clearly indicated.

This should fix the "STATIC MEMORY OVERWRITTEN" error messages randomly
happening when stepping at high frequency (where either isr is
preempted more frequently).

PFW-1135

### DIFF
--- a/Firmware/fsensor.cpp
+++ b/Firmware/fsensor.cpp
@@ -478,22 +478,8 @@ bool fsensor_oq_result(void)
 }
 #endif //FSENSOR_QUALITY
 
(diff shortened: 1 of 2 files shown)


DRracer (approved):
an ugly workaround fixing a nasty bug

PR #2792, opened 2020-08-05
Mk3 fix high speed deceleration

wavexx:
These are further tweaks to https://github.com/prusa3d/Prusa-Firmware/pull/2721 to fix high-speed deceleration on the MK3 when using high step counts.

We fix the type of a block_t::acceleration_rate (which is also positive), as well re-introduce the un-optimized C reference implementation of Mult* functions.

We also check again for step_rate underflow.

Combined with #2791 this allows reliable deceleration at higher microstep counts (M350 X32 Y32)


DRracer (approved):
to be tested thoroughly

PR #2809, opened 2020-08-18
Pfw 1134

leptun:
Altfan detection is overriden by default after updating firmware.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -2096,9 +2096,6 @@ static void lcd_preheat_menu()
 //! 	@code{.unparsed}
 //! 	| Voltages           |	MSG_MENU_VOLTAGES
 //! 	@endcode
-//!
-//!
-//! 	| Experimental       |	c=18
 //! 
 //! 
 //! If DEBUG_BUILD is defined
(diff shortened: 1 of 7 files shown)


DRracer:
Addressing the broad discussion in PR #2784 - what you see here is the latest generally accepted solution reflecting also the community remarks (@josefprusa @gregsaun @vertigo235 FYI).

The technical idea is as follows:
We inhibit alternative fan detection only when the EEPROM is erased.
- That means our freshly manufactured machines will try to detect an alternative fan and record it into EEPROM upon start.
- Upgrading from a previous firmware will NOT activate/override fan speed autodetection - it will keep whatever has been set in the EEPROM (i.e. 100% or 50% PWM).

This shall solve the vast majority of firmware upgrades even on modded printers (i.e. will keep their fan speed at 100%).
- The only problem will arise in case of an erased EEPROM. In such a case the user of a printer with a different fan shall take extra steps to set the fan speed back to 100% if he wishes. 
- Please note we haven't found a reliable way of detecting a freshly manufactured Original Prusa i3 MK3S printer to inhibit the altfan detection only in such a case.

Generally (addressing the hot discussions in PR #2784 and PR #2692) :
1. The firmware 3.9.1 is not finished/released yet, we are looking for a quality solution of the whole fan issue.
2. If a PR gets merged into master, it means the (part of) solution was accepted in-house and was ready to be tested on printers internally. It may be subject to change after integration tests which are now running.
3. We are an open source company and we do listen to the community. 
We even listen to people running heavily modified HW.
If we weren't, this PR or any other discussion will not take place, right?
So please keep the discussion at a technical level. Offensive comments will not help anyone.
4. We are still considering a better PWM control for fans (quieter, safer with much more robust speed readings) for next FW updates.

Firmware/ultralcd.cpp
```diff
@@ -5730,6 +5714,25 @@ static void sheets_menu()
 
 void lcd_hw_setup_menu(void)                      // can not be "static"
 {
+    typedef struct
+    {// 2bytes total
+        int8_t status;
+        uint8_t experimental_menu_visibility;
```
3d-gussner, line 5720: Would it make sense to use bool here? If so also the eeprom.h documentation should be updated

DRracer (approved):

vertigo235:
Nice work Prusa team, and thank you for listening to feedback from us early bleeding edge testers.

PR #2813, opened 2020-08-25
Scale extruder motor current linearly with speed.

mkbel:
49% less heating when running at low speed and standstill, 4% more torque at maximum extrusion rate (15mm^3/s), 15% more torque in high speed movements (un/retractions).

StealthChop mode is used for low speeds (below 900mm/min)
spreadCycle is used above. Transition speed is well above maximum extrusion rate of 15mm^3/s (275mm/min) so mode transition is not expected to be visible on printed surface.

StealthChop is expected to improve printed surface quality (less artifacts).

Warning you can burn extruder motor if it is not the same impedance as original Prusa i3 Extruder stepper motor. There is no current feedback in low speed so lower impedance motor can be burned by over current.

Even there is no direct current feedback, there is no risk for original motor thermal runaway, as motor resistance increases with temperature, current decreases.

Standstill peak phase current is expected to be 500 mA and linearly increase with speed to 970 mA at 900mm/min where spreadCycle constant current regulation takes over and keeps peak current at 805 mA to maximum speed possible.

As motor heating increases with current squared, lowering low speed current from 700mA to 500mA decreases heating 49% in thate mode, where motor spends most of the time.

PFW-1234

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -1340,8 +1340,6 @@ void setup()
 #endif //TMC2130_VARIABLE_RESOLUTION
 
 #endif //TMC2130
-    // We enable E-cool mode for non-farm prints IFF the experimental menu is visible AND the EEPROM_ECOOL variable has
-    // a value of the universal answer to all problems of the universe
 	st_init();    // Initialize stepper, this enables interrupts!
   
 #ifdef TMC2130
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -1054,11 +1054,6 @@ void setup()
 		selectedSerialPort = 1;
 #endif //HAS_SECOND_SERIAL_PORT
 		MYSERIAL.begin(BAUDRATE);
-#ifdef TMC2130
-		//increased extruder current (PFW363)
-		tmc2130_current_h[E_AXIS] = 36;
-		tmc2130_current_r[E_AXIS] = 36;
-#endif //TMC2130
 #ifdef FILAMENT_SENSOR
 		//disabled filament autoload (PFW360)
 		fsensor_autoload_set(false);
@@ -1067,6 +1062,14 @@ void setup()
           if(!(eeprom_read_byte((uint8_t*)EEPROM_FAN_CHECK_ENABLED)))
                eeprom_update_byte((unsigned char *)EEPROM_FAN_CHECK_ENABLED,true);
 	}
+
+#ifdef TMC2130
+    if( FarmOrUserECool() ){
+		//increased extruder current (PFW363)
+		tmc2130_current_h[E_AXIS] = TMC2130_CURRENTS_FARM;
+		tmc2130_current_r[E_AXIS] = TMC2130_CURRENTS_FARM;
+    }
+#endif //TMC2130
 
     //saved EEPROM SN is not valid. Try to retrieve it.
     //SN is valid only if it is NULL terminated. Any other character means either uninitialized or corrupted
@@ -1339,13 +1342,12 @@ void setup()
 #endif //TMC2130
     // We enable E-cool mode for non-farm prints IFF the experimental menu is visible AND the EEPROM_ECOOL variable has
     // a value of the universal answer to all problems of the universe
-	enableECool = ( eeprom_read_byte((uint8_t *)EEPROM_ECOOL_ENABLE) == EEPROM_ECOOL_MAGIC_NUMBER ) && ( eeprom_read_byte((uint8_t *)EEPROM_EXPERIMENTAL_VISIBILITY) == 1 );
 	st_init();    // Initialize stepper, this enables interrupts!
   
 #ifdef TMC2130
 	tmc2130_mode = silentMode?TMC2130_MODE_SILENT:TMC2130_MODE_NORMAL;
 	update_mode_profile();
-	tmc2130_init(TMCInitParams(false, farm_mode || enableECool));
+	tmc2130_init(TMCInitParams(false, FarmOrUserECool() ));
 #endif //TMC2130
 #ifdef PSU_Delta
      init_force_z();                              // ! important for correct Z-axis initialization
@@ -2558,7 +2560,7 @@ void force_high_power_mode(bool start_high_power_section) {
 		cli();
 		tmc2130_mode = (start_high_power_section == true) ? TMC2130_MODE_NORMAL : TMC2130_MODE_SILENT;
 		update_mode_profile();
-		tmc2130_init(TMCInitParams(farm_mode || enableECool));
+		tmc2130_init(TMCInitParams(FarmOrUserECool()));
     // We may have missed a stepper timer interrupt due to the time spent in the tmc2130_init() routine.
     // Be safe than sorry, reset the stepper timer before re-enabling interrupts.
     st_reset_timer();
@@ -8427,6 +8429,7 @@ Sigma_Exit:
     /*!
 	### M907 - Set digital trimpot motor current in mA using axis codes <a href="https://reprap.org/wiki/G-code#M907:_Set_digital_trimpot_motor">M907: Set digital trimpot motor</a>
 	Set digital trimpot motor current using axis codes (X, Y, Z, E, B, S).
+    M907 has no effect when the experimental Extruder motor current scaling mode is active (that applies to farm printing as well)
 	#### Usage
     
         M907 [ X | Y | Z | E | B | S ]
@@ -8442,17 +8445,19 @@ Sigma_Exit:
     case 907:
     {
 #ifdef TMC2130
-        // See tmc2130_cur2val() for translation to 0 .. 63 range
-        for (int i = 0; i < NUM_AXIS; i++)
-			if(code_seen(axis_codes[i]))
-			{
-				long cur_mA = code_value_long();
-				uint8_t val = tmc2130_cur2val(cur_mA);
-				tmc2130_set_current_h(i, val);
-				tmc2130_set_current_r(i, val);
-				//if (i == E_AXIS) printf_P(PSTR("E-axis current=%ldmA\n"), cur_mA);
-			}
-
+        if( ! FarmOrUserECool() ){
+            // See tmc2130_cur2val() for translation to 0 .. 63 range
+            for (int i = 0; i < NUM_AXIS; i++){
+                if(code_seen(axis_codes[i]))
+                {
+                    long cur_mA = code_value_long();
+                    uint8_t val = tmc2130_cur2val(cur_mA);
+                    tmc2130_set_current_h(i, val);
+                    tmc2130_set_current_r(i, val);
+                    //if (i == E_AXIS) printf_P(PSTR("E-axis current=%ldmA\n"), cur_mA);
+                }
+            }
+        }
 #else //TMC2130
       #if defined(DIGIPOTSS_PIN) && DIGIPOTSS_PIN > -1
         for(int i=0;i<NUM_AXIS;i++) if(code_seen(axis_codes[i])) st_current_set(i,code_value());
@@ -11963,7 +11968,7 @@ void disable_force_z()
 #ifdef TMC2130
     tmc2130_mode=TMC2130_MODE_SILENT;
     update_mode_profile();
-    tmc2130_init(TMCInitParams(true, farm_mode || enableECool));
+    tmc2130_init(TMCInitParams(true, FarmOrUserECool()));
 #endif // TMC2130
 }
 
@@ -11977,7 +11982,7 @@ bEnableForce_z=true;
 #ifdef TMC2130
 tmc2130_mode=eeprom_read_byte((uint8_t*)EEPROM_SILENT)?TMC2130_MODE_SILENT:TMC2130_MODE_NORMAL;
 update_mode_profile();
-tmc2130_init(TMCInitParams(true, farm_mode || enableECool));
+tmc2130_init(TMCInitParams(true, FarmOrUserECool()));
 #endif // TMC2130
 
 WRITE(Z_ENABLE_PIN,Z_ENABLE_ON);                  // slightly redundant ;-p
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -8443,17 +8443,19 @@ Sigma_Exit:
     case 907:
     {
 #ifdef TMC2130
-        if( ! FarmOrUserECool() ){
-            // See tmc2130_cur2val() for translation to 0 .. 63 range
-            for (int i = 0; i < NUM_AXIS; i++){
-                if(code_seen(axis_codes[i]))
-                {
-                    long cur_mA = code_value_long();
-                    uint8_t val = tmc2130_cur2val(cur_mA);
-                    tmc2130_set_current_h(i, val);
-                    tmc2130_set_current_r(i, val);
-                    //if (i == E_AXIS) printf_P(PSTR("E-axis current=%ldmA\n"), cur_mA);
+        // See tmc2130_cur2val() for translation to 0 .. 63 range
+        for (uint_least8_t i = 0; i < NUM_AXIS; i++){
+            if(code_seen(axis_codes[i])){
+                if( i == E_AXIS && FarmOrUserECool() ){
+                    SERIAL_ECHORPGM(eMotorCurrentScalingEnabled);
+                    SERIAL_ECHOLNPGM(", M907 E ignored");
+                    continue;
                 }
+                long cur_mA = code_value_long();
+                uint8_t val = tmc2130_cur2val(cur_mA);
+                tmc2130_set_current_h(i, val);
+                tmc2130_set_current_r(i, val);
+                //if (i == E_AXIS) printf_P(PSTR("E-axis current=%ldmA\n"), cur_mA);
             }
         }
 #else //TMC2130
@@ -8680,7 +8682,7 @@ Sigma_Exit:
     case 350: 
     {
 	#ifdef TMC2130
-		for (int i=0; i<NUM_AXIS; i++) 
+		for (uint_least8_t i=0; i<NUM_AXIS; i++) 
 		{
 			if(code_seen(axis_codes[i]))
 			{
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -1337,13 +1337,14 @@ void setup()
 #endif //TMC2130_VARIABLE_RESOLUTION
 
 #endif //TMC2130
-
+    // @@TODO subject to discussion
+	// enableECool = eeprom_read_byte((uint8_t *)EEPROM_ECOOL);
 	st_init();    // Initialize stepper, this enables interrupts!
   
 #ifdef TMC2130
 	tmc2130_mode = silentMode?TMC2130_MODE_SILENT:TMC2130_MODE_NORMAL;
 	update_mode_profile();
-	tmc2130_init();
+	tmc2130_init(TMCInitParams(false, farm_mode || enableECool));
 #endif //TMC2130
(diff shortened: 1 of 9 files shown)


mkbel:
Test instructions: Use factory reset to be sure new increased spread cycle current is applied. (Factory reset is not needed for printers in farm mode, as spreadCycle current is now the same as in farm mode.) Do not interfere with current from gcode (Do not issue M907 in gcode.).

3d-gussner:
> Test instructions: Use factory reset to be sure new increased spread cycle current is applied. Do not interfere with current from gcode (Do not issue M907 in gcode.).

As Prusa Slicer settings [1.1.8](https://github.com/prusa3d/PrusaSlicer-settings/commit/dc0a93992b69bd05a0741a6a90ba3e84855686ae) added to start and end gcode `M907` commands, you should
create a modifiled Printer profile to test it without
Start g-code
```
; Don't change E values below. Excessive value can damage the printer.
{if print_settings_id=~/(0.05mm ULTRADETAIL @MK3|0.07mm ULTRADETAIL @MK3|0.10mm DETAIL @MK3|0.15mm QUALITY @MK3|0.20mm QUALITY @MK3)/}M907 E430 ; set extruder motor current{endif}
{if print_settings_id=~/(0.15mm SPEED @MK3|0.20mm SPEED @MK3|0.30mm DRAFT @MK3)/}M907 E538 ; set extruder motor current{endif}
```
End g-code
```
{if print_settings_id=~/.*(DETAIL @MK3|QUALITY @MK3|@0.25 nozzle MK3).*/}M907 E538 ; reset extruder motor current{endif}
```
Or comment out these lines with `;`

@mkbel correct?

mkbel:
> @mkbel correct?

Correct. That is why I mentioned it.

mkbel:
Extruder motor one phase current - printing long straight segments:
(left side faster extrusion, higher peak current) (right side slower extrusion, lower peak current)
SCR04
Extruder motor one phase current - printing short segments:
SCR03

mkbel:
@wavexx asked: Do you actually see/hear a jerk during the transition?
Transition is relatively smooth.
Current waveform when switching between stealthChop and spreadCycle:
(transition visible in first half-wave) G1 F6400 E-10
WA000001
(transition visible in last half-wave) G1 F6400 E-10
WA000002
(transition visible in first and last half-wave) G1 F6400 E-7
WA000003

mkbel:
@wavexx asked: When you set the transition speed between stealthchop and spreadcycle, did you try to do it at lower speeds? No I didn't try any other transition speed. I spend several hours tuning currents and I was very happy with first transition speed I tried. There are several constrains on transition speed. I don't want transition speed to be in speed used for extruding to avoid artifacts, so it must be higher than 400 mm/min. It must be also lower than 2200 mm/min as in that speed Marlin starts issuing two micro steps at once and it totally breaks voltage PWM regulation and motor phase current skyrockets to 1800 mA (nearly twice as rated current). Another constrain is, that spread cycle regulates constant phase current, so transition should be in area where voltage PWM gets similar current to have smooth transition.

mkbel:
@3d-gussner asked: How did you get the temperature and torque values? How looks your test rig?
I measured phase current with scope and current probe. Temperature rise is expected to be proportional to current squared. Torque is expected to be proportional to current.

mkbel:
@3d-gussner asked: Did you account the tolerances of the e motor? There are +-10% for resistance and +-20 for inductance in the data sheet.
Those effects are hard to predict, there must be some large scale test on our farm. I think it would be enough to check actual motor heating and if it decreased from previous firmware. And of course print quality. I don't worry that some current spikes - if present - could damage the motor. Short term over current is not damaging for the wire of motor coils. What is limiting for motor operation is maximum running temperature. We are currently very far from motor maximum operating temperature of 130°C and we hopefully will be even farther from it if this change is successful.
If extruder motor runs very slowly - which is what it does most of the time major effect on motor current has its resistance and it is also mostly the resistance which causes heating. In previous state there was constant current regulation, so current was expected to be constant and its variation between printers was affected only by accuracy of feedback resistor and reference voltage. If we assert that this variance was much smaller than motor resistance variance different users got  E motor increase in temperature which varies +-10% between printers before and constant torque. After the change, there is constant voltage, so dependence between heating power loses and resistance is inverted. Extruder motors with 10%+ resistance which heated most before will heat 10% less than average, but also its torque will be decreased 10% below average. Extruder motors with 10%- resistance which heated least before will heat 10% more than average (or even more if current gets into area where its steel becomes magnetically saturated), but also its torque will be increased 10% above average (if not saturated).

leptun:
Important note: If you change the steps/mm of the Extruder or change the microstepping of the motor you will have to adjust the TMC2130_TPWMTHRS_E value.

mkbel:
> Important note: If you change the microstepping of the motor you will have to adjust the TMC2130_TPWMTHRS_E value.

I am not sure about that, as TPWMTHRS compares to TSTEP and it is supposed to be "Actual  measured  time  between  two  1/256  micro-steps  derived from  the  step  input  frequency  in  units  of  1/fCLK." So it should be independent of micro-stepping settings according to datasheet.

But if you use higher micro-step resolution Marlin starts issuing two micro steps at once in lower speed and it totally breaks voltage PWM regulation and motor phase current skyrockets to 1800 mA.

mkbel:
> Important note: If you change the steps/mm of the Extruder motor you will have to adjust the TMC2130_TPWMTHRS_E value.

In such case you have probably different E motor and you can not use my PL at all. (Or you need to tune PWM_AMPL_E and PWM_GRAD_E yourself.)
Warning you can burn extruder motor if it is not the same impedance as original Prusa i3 Extruder stepper motor. There is no current feedback in low speed so lower impedance motor can be burned by over current.

DRracer:
**To prevent new conspiracy theories, which arose again on various social media...**
This PR is a purely experimental piece of code which might help lowering the E-motor's temperature. 
It is not meant to be merged in near future.

CacikeOtoao:
@DRacer

How fine tuned is this experimental feature in regards to impedance variance tolerance? +/- 10%? 20%?. I’m just trying to see if it is something that will have a wide adaptive curve for various motors or if it is more LDO/Prusa motor specific. Not bashing on anyone/anything, just want to prepare for it and test with various steppers.

thisiskeithb:
> Warning you can burn extruder motor if it is not the same impedance as original Prusa i3 Extruder stepper motor. There is no current feedback in low speed so lower impedance motor can be burned by over current.

Will there be a way to disable this feature for Bondtech BMG/LDO 0.9° pancake stepper upgrades or other "non-original" motors or will it require modifying firmware?

Panayiotis-git:
> This PR is a purely experimental piece of code which might help lowering the E-motor's temperature.
It is not meant to be merged in near future.

mkbel:
> Will there be a way to disable this feature for Bondtech BMG/LDO 0.9° pancake stepper upgrades or other "non-original" motors or will it require modifying firmware?

I would suggest adding menu item for that:
Rename existing menu item "mode" to "mode XYZ".
Add menu item "mode E" with options [normal] and [adaptive].
Make "mode E" [normal] default option. (After firmware upgrade and factory reset)
Add warning dialog when switching "mode E" from [normal] to [adaptive]:
"You can burn E motor and/or driver"
"if your E motor is lower impedance"
"or back EMF than ORIGINAL PRUSA"
"i3 Extruder Stepper motor."
"Do you wish to proceed?"
" Yes >No "

I have no final word if the whole extruder motor scaling thing will ever make it to official firmware and how it will be presented.

mkbel:
> How fine tuned is this experimental feature in regards to impedance variance tolerance? +/- 10%? 20%?. I’m just trying to see if it is something that will have a wide adaptive curve for various motors or if it is more LDO/Prusa motor specific. Not bashing on anyone/anything, just want to prepare for it and test with various steppers.

It is fine tuned to one specimen of ORIGINAL PRUSA i3 Extruder Stepper motor. It is not yet even known, if hard coded settings will be usable across parameter dispersion across same motor model. It was tested on two other printers with same motor model and no magical smoke was released and printers were actually printing. It is highly probable that it will work  across parameter dispersion across same motor model.

It is highly improbable that it could work with different motor model with different impedance and back EMF.

The whole feature is meant as workaround for too hot E motor. I personally never encountered problem related to too hot E motor. So I expect only minority of users are affected by this problem. I made this to make happy those minority users and I don't expect any use of this feature for most of the users. I expect the majority of users unaffected with hot E motor related problems will stay with constant current regulation as before.

leptun:
I’ve included these changes in my Klipper config and it worked correctly for me. My hardware was an original prusa stepper (high inductance) from the Mk2s era with an EinsyRambo.
The motor was warm, but not as hot as it was on my other Mk3 with stock firmware and current settings.

mkbel:
It was decided to:

- put this option to experimental menu
- always enable in farm_mode
- if enabled ignore M907 E gcode

mkbel:
Because of conflicting requirements to be always enabled in farm_mode and not to be possible to enable via gcode, additional safety mechanism (measure motor coil resistance) will be examined.

DRracer (approved):
Approving changes done by @mkbel

Firmware/ultralcd.cpp
```diff
@@ -8947,6 +8954,14 @@ void lcd_experimental_toggle()
...
 }
 
+void ECool_toggle(){
+    enableECool = ! enableECool;
+    // this is only called when the experimental menu is visible, thus the first condition for enabling of the ECool mode is met in this place
+    eeprom_update_byte((uint8_t *)EEPROM_ECOOL_ENABLE, enableECool ? EEPROM_ECOOL_MAGIC_NUMBER : EEPROM_EMPTY_VALUE);
+}
+bool ECool_get(){
+    return enableECool;
+}
```
mkbel: It is overly complicated and error prone to cache EEPROM in RAM, and it also costs RAM. It is more efficient to just read status from EEPROM inside ECool_get() and use ECool_get() consistently and remove global variable enableECool.

Also at this point enableECool variable is sometimes used directly and sometimes through ECool_get().

If you remove enableECool variable, you can also remove its initialization from setup(), making things less error prone.

mkbel: Also you can call TMC reinit from ECool_toggle(), so change is applied when the state of the variable changes. If experimental menu is not accessible during print, it is safe.

Firmware/Marlin_main.cpp
```diff
@@ -1337,13 +1337,15 @@ void setup()
...
+    // We enable E-cool mode for non-farm prints IFF the experimental menu is visible AND the EEPROM_ECOOL variable has
+    // a value of the universal answer to all problems of the universe
+	enableECool = ( eeprom_read_byte((uint8_t *)EEPROM_ECOOL_ENABLE) == EEPROM_ECOOL_MAGIC_NUMBER ) && ( eeprom_read_byte((uint8_t *)EEPROM_EXPERIMENTAL_VISIBILITY) == 1 );
 	st_init();    // Initialize stepper, this enables interrupts!
   
 #ifdef TMC2130
 	tmc2130_mode = silentMode?TMC2130_MODE_SILENT:TMC2130_MODE_NORMAL;
 	update_mode_profile();
-	tmc2130_init();
+	tmc2130_init(TMCInitParams(false, farm_mode || enableECool));
```
mkbel: It would be less error prone to hide (farm_mode || enableECool) condition inside ECool_get()

DRracer: ok, good point

Firmware/tmc2130.cpp
```diff
@@ -191,10 +188,22 @@ void tmc2130_init()
 	}
 	for (uint_least8_t axis = 3; axis < 4; axis++) // E axis
 	{
-		tmc2130_setup_chopper(axis, tmc2130_mres[axis], tmc2130_current_h[axis], tmc2130_current_r[axis]);
+		tmc2130_setup_chopper(axis,
+            tmc2130_mres[axis], // this is not changed with ECool
+            tmc2130_current_h[axis] + params.enableECool,
+            tmc2130_current_r[axis] + params.enableECool);
```
mkbel: Whiskey Tango Foxtrot?
Did you mean tmc2130_current_h[axis + params.enableECool] ?

mkbel (commented):
M907 E gcode is not ignored when E motor scaling current is enabled.

Firmware/tmc2130.h
```diff
@@ -1,12 +1,14 @@
...
+#include <stdint.h>
 
 //mode
 extern uint8_t tmc2130_mode;
+// Beware - the 4th index (counted from zero) is abused for the E-motor cool mode only used on the farm
 //holding and running currents
-extern uint8_t tmc2130_current_h[4];
-extern uint8_t tmc2130_current_r[4];
+extern uint8_t tmc2130_current_h[5];
+extern uint8_t tmc2130_current_r[5];
```
mkbel: This is not workable solution. If done this way, you would need to add condition to use index 3 or 4 on every place where this value is used, not only in tmc2130_init(). There are about 16 places in code where this value is used.

More workable solution would be to use approach already used in farm_mode (preferably use named constant instead of plain "36"):
```
		//increased extruder current (PFW363)
		tmc2130_current_h[E_AXIS] = 36;
		tmc2130_current_r[E_AXIS] = 36;
```

Firmware/tmc2130.cpp
DRracer: OMG, yes I did

Firmware/variants/1_75mm_MK3S-EINSy10a-E3Dv6full.h
```diff
@@ -267,7 +273,8 @@
 
 //new settings is possible for vsense = 1, running current value > 31 set vsense to zero and shift both currents by 1 bit right (Z axis only)
 #define TMC2130_CURRENTS_H {16, 20, 35, 30}  // default holding currents for all axes
-#define TMC2130_CURRENTS_R {16, 20, 35, 30}  // default running currents for all axes
+#define TMC2130_CURRENTS_FARM 36             // E 805 mA peak for ECool/farm mode
+#define TMC2130_CURRENTS_R {16, 20, 35, 30}  // default running currents for all axes, the last parameter is E 805 mA peak for ECool mode
```
mkbel: Comment after the comma `, the last parameter is E 805 mA peak for ECool mode`
is misleading now.

Firmware/Marlin_main.cpp
```diff
@@ -1337,13 +1340,14 @@ void setup()
 #endif //TMC2130_VARIABLE_RESOLUTION
 
 #endif //TMC2130
-
+    // We enable E-cool mode for non-farm prints IFF the experimental menu is visible AND the EEPROM_ECOOL variable has
+    // a value of the universal answer to all problems of the universe
```
mkbel: Orphaned comment.

Firmware/variants/1_75mm_MK3-EINSy10a-E3Dv6full.h
```diff
@@ -247,6 +252,7 @@
 #define TMC2130_PWM_CLK   (2 * TMC2130_FCLK / TMC2130_PWM_DIV) // PWM frequency (23.4kHz, 35.1kHz, 46.9kHz, 58.5kHz for 12MHz fclk)
 
 #define TMC2130_TPWMTHRS  0         // TPWMTHRS - Sets the switching speed threshold based on TSTEP from stealthChop to spreadCycle mode
+#define TMC2130_TPWMTHRS_E 403
```
mkbel: Suggest comment:
`// Switch extruder from StealthChop to SpreadCycle at around 900mm/min`

Firmware/variants/1_75mm_MK3S-EINSy10a-E3Dv6full.h
```diff
@@ -249,6 +254,7 @@
 #define TMC2130_PWM_CLK   (2 * TMC2130_FCLK / TMC2130_PWM_DIV) // PWM frequency (23.4kHz, 35.1kHz, 46.9kHz, 58.5kHz for 12MHz fclk)
 
 #define TMC2130_TPWMTHRS  0         // TPWMTHRS - Sets the switching speed threshold based on TSTEP from stealthChop to spreadCycle mode
+#define TMC2130_TPWMTHRS_E 403
```
mkbel: Suggest comment:
`// Switch extruder from StealthChop to SpreadCycle at around 900mm/min`

Firmware/ultralcd.cpp
```diff
@@ -8947,6 +8947,35 @@ void lcd_experimental_toggle()
     eeprom_update_byte((uint8_t *)EEPROM_EXPERIMENTAL_VISIBILITY, oldVal);
 }
 
+#ifdef TMC2130
+void UserECool_toggle(){
```
mkbel, line 8951: User? Did you mean Use?

Firmware/ultralcd.cpp
```diff
@@ -8947,6 +8947,35 @@ void lcd_experimental_toggle()
...
+}
+#endif
+
+/// Enable experimental support for cooler operation of the extruder motor
+/// Beware - REQUIRES original Prusa MK3/S/+ extruder motor with adequate maximal current
+/// Therefore we don't want to allow general usage of this feature in public as the community likes to
+/// change motors for various reasons and unless the motor is rotating, we cannot verify its properties
+/// (which would be obviously too late for an improperly sized motor)
+/// For farm printing, the cooler E-motor is enabled by default.
+bool UserECoolEnabled(){
```
mkbel, line 8970: User? Did you mean Use?

Firmware/ultralcd.h
```diff
@@ -133,6 +133,9 @@ extern uint8_t farm_mode;
 extern int farm_timer;
 extern uint8_t farm_status;
 
+extern bool UserECoolEnabled();
+extern bool FarmOrUserECool();
```
mkbel, line 137: I think it is misleading to have 2 functions there, someone could use wrong one.
I suggest remove FarmOrUserECool() and hide its logic inside UserECoolEnabled(). I don't know what is "User" there. Why it is not called UseECool(), or ECoolEnabled()? Effect will be, that in farm_mode hidden menu will always show ECool [On] and toggling will left it [On] but it is absolutely ok for me.

DRracer: I don't like the idea of merging the farm and user-selected (that's why I called it "User") E-cool mode, I'd like to keep it as it is.

Firmware/ultralcd.cpp
DRracer: nope, User - like _ordinary non-farm user-selected/activated_

Firmware/ultralcd.cpp
DRracer: nope, User - _like ordinary non-farm user-selected/activated_

Firmware/Marlin_main.cpp
```diff
@@ -8440,17 +8443,19 @@ SERIAL_PROTOCOLPGM("\n\n");
...
+                if(code_seen(axis_codes[i]))
+                {
+                    long cur_mA = code_value_long();
+                    uint8_t val = tmc2130_cur2val(cur_mA);
+                    tmc2130_set_current_h(i, val);
+                    tmc2130_set_current_r(i, val);
+                    //if (i == E_AXIS) printf_P(PSTR("E-axis current=%ldmA\n"), cur_mA);
+                }
+            }
+        }
```
mkbel: I think we want to disable only M907 E, not M907 XYZ.

dafi87:
With E Cool enabled, I feel like the extruder stepping got louder in stealth mode on my MK3S+. Was that just my ears or a bad print sample I picked, or did you guys notice similar behaviour?

Best,
Daniel

DRracer:
@dafi87 yes, the E-motor is "clicking" on fast changes of direction in this mode of operation... especially on retract/unretract moves.

dafi87:
Thanks for getting back so quickly! Retract/unretract is exactly the situations where I thought I heard the difference in noise level.

PR #2814, opened 2020-08-25
Use hardware TWI for the PAT9125 (optical) filament sensor

wavexx:
The PAT9125 on the MK3 is attached to the hardware SDA/SCL pins, but we're still using a software I2C implementation for some reason.

We add a new configuration option PAT9125_I2C (which is also enabled by default by this PR) to use the hardware TWI module in 400khz fast mode.

We save 170 bytes on the MK3S variant (which still requires probing to detect the current variant) and 268 on the MK3, plus a little bit of latency since despite the code still being synchronous, we can at least avoid having to wait for the stop condition.

PFW-1191


vintagepc:
Three guesses what prompted this change 🤣 
(and the first two don't count)

wavexx:
On Wed, Sep 09 2020, vintagepc wrote:
> Three guesses what prompted this change 🤣

I couldn't sleep well after finding that out, had to fix it ;)

vintagepc:
Did you end up testing it out by wiring up the hardware SPI sim? IIRC the SPIPeripheral class was written to support both bitbanged and "real" I2C connections and the fake PAT shouldn't care which it is wired to. 

Edit: for anyone else wondering about the context for this, see
https://github.com/vintagepc/MK404/issues/7#issuecomment-649170568_

Firmware/twi.c
```diff
@@ -0,0 +1,111 @@
...
+#define sbi(sfr, bit) (_SFR_BYTE(sfr) |= _BV(bit))
+#endif
+
+#include "twi.h"
+
+
+void twi_init(void)
+{
+  // activate internal pullups for twi.
+  digitalWrite(SDA, 1);
```
DRracer: Can we get rid of these Arduino functions and use fastio?

wavexx:
> Did you end up testing it out by wiring up the hardware SPI sim? IIRC the SPIPeripheral class was written to support both bitbanged and "real" I2C connections and the fake PAT shouldn't care which it is wired to.

Took a while, but getting back to this.
You mean the I2CPeripheral here?

At least without changes, the sensor doesn't seem to be detected right now. The SCL/SDA irqs do not seem to be generated in this case so the callbacks are never called.

vintagepc:
Ah, ok. If you wouldn't mind tossing up a MK404 issue with a .hex and I'll take a look. It's possible that something is still unimplemented.

Firmware/twi.c
wavexx: Ironically I cannot find a "stock" pin definition for SCL/SDA which is directly usable with fastio.
I'd like to avoid defining one again, since I'm in the process of removing SWI2C_SCL/SDA...

The SDA definition in fastio.h itself is prefix with DIO already, which is not compatible with the macros. This is ... ironic? (I don't get the purpose of these definitions if they cannot be used with fastio!)

Arduino's own PIN_WIRE_SDA is braced with parenthesis, which again doesn't work with the macros.

Overall, since this is only done during init, and we're already using digitalWrite elsewhere, I wouldn't care.

Otherwise we need to define these pins ourselves in the board definition.

There's also the issue of cbi/sbi being duplicated, since I don't want to include the whole "Marlin.h" for these. Once we merge https://github.com/prusa3d/Prusa-Firmware/pull/2690 we can do some cleanup.

leptun: As far as I understand the cbi and sbi macros are considered deprecated by avr-libc. Instead once should just do good old bitwise OR() and AND~().
Regardless of this, I think you can include something like <avr/io.h> which defines it as well.

leptun: See https://www.eit.lth.se/fileadmin/eit/courses/edi021/avr-libc-user-manual/group__avr__sfr.html#:~:text=Porting%20programs%20that%20use%20the%20deprecated%20sbi%2Fcbi%20macros&text=The%20sbi%20and%20cbi%20macros,sfr%20%7C%3D%20_BV(bit)%20.

wavexx: Yeah, for one use you're right. I'll remove it.

Firmware/pat9125.c
```diff
@@ -234,39 +254,48 @@ uint8_t pat9125_update_bs(void)
...
-	{
-		pat9125_PID1 = 0xff;
-		pat9125_PID2 = 0xff;
-		return;
-	}
-#endif //PAT9125_SWI2C
+        goto error;
+#elif defined(PAT9125_I2C)
+	if (twi_rw8(PAT9125_I2C_ADDR,TW_WRITE,&addr) ||
+        twi_rw8(PAT9125_I2C_ADDR,TW_READ,&data))
```
vintagepc: This isn't right - it doesn't actually write data to the device; it sets the address and then does a read on the same address - as evidenced by the printer startup:

```
PAT9125_RES_X=20. <-- should be 0
PAT9125_RES_Y=20. <-- should be 240. 
PAT9125_init:1.
```

Moreover, I don't think this is necessarily I2C compliant. When writing you need to send the data payload in the same start-stop frame as the address byte. Look at swi2c_writeByte_A8 to see this. (More context at https://github.com/vintagepc/MK404/issues/248)

vintagepc (changes requested):

Firmware/pat9125.c
wavexx: I probably shaved this while trying to save space, but you're definitely right.
I'll test this tomorrow on a real MK3.

vintagepc: Should be able to just modify twi_rw8 to take an additional pointer parameter to the byte to send or be read into. Would save the overhead of two successive function calls there and some i2c delays.

wavexx: *Should* be correct now. I've split the two functions back, which I think it's more readable anyway.

vintagepc: Yup, all seems to be as expected now!

```
<snip PID/VID check>
X_IN:00ea0100
START
TX_IN:0dea1000
ADDRESS 13
TX_IN:00ea1000
Set 13 to 0
TX_IN:01ea0200
STOP
TX_IN:00ea0100
START
TX_IN:0eea1000
ADDRESS 14
TX_IN:f0ea1000
Set 14 to 240
TX_IN:01ea0200
STOP
TX_IN:00ea0100
START
TX_IN:0dea1000
ADDRESS 13
TX_IN:00eb0100
START
TX_IN:ebeb2000
READ 13
TX_IN:01eb0200
STOP
PAT9125_RES_X=0.
TX_IN:00ea0100
START
TX_IN:0eea1000
ADDRESS 14
TX_IN:00eb0100
START
TX_IN:ebeb2000
READ 14
TX_IN:01eb0200
STOP
PAT9125_RES_Y=240.
PAT9125_init:1.
```

vintagepc (approved):
I don't have write access to the repo but I approve of this change regardless given its origins :wink:

DRracer:
@wavexx @vintagepc is this ready to get merged into MK3?

vintagepc:
I think it is, yes :) All was working as expected in MK404

DRracer:
OK, my only complaint is the usage of "digitalWrite()" - can we get rid of this?

vintagepc:
This is what it's doing, seems like it should be possible to just strip out the unnecessary bits and just do the actions we need to set it high:
```
void digitalWrite(uint8_t pin, uint8_t val)
{
	uint8_t timer = digitalPinToTimer(pin);
	uint8_t bit = digitalPinToBitMask(pin);
	uint8_t port = digitalPinToPort(pin);
	volatile uint8_t *out;

	if (port == NOT_A_PIN) return;

	// If the pin that support PWM output, we need to turn it off
	// before doing a digital write.
	if (timer != NOT_ON_TIMER) turnOffPWM(timer);

	out = portOutputRegister(port);

	uint8_t oldSREG = SREG;
	cli();

	if (val == LOW) {
		*out &= ~bit;
	} else {
		*out |= bit;
	}

	SREG = oldSREG;
}
```

leptun:
We can use the fastio macros here, can't we? Just fix the SCL and SDA definition in fastio so that it's fastio compatible

wavexx:
We can remove it, but is it necessary? digitalWrite is already used elsewhere, so we won't save any space AFAIK. It's also not on a hot code path.

wavexx:
Checking now if using FASTIO can lead to some space savings..

wavexx:
Using fastio would actually save 12 bytes :), however I need to add some crummy defines just to make it build: including fastio.h is not enough, which is something that would be nice to fix.

#2690 would make sense IMHO

If including #2690 makes sense in this release, I could proceed with something less invasive (another copy of CRITICAL_SECTION.. uugh) to be revisited later @DRracer

wavexx:
@DRracer now using fastio!

DRracer (approved):

PR #2828, opened 2020-09-07
Menu view return patch

### LINKED ISSUE #2620: [ENHANCEMENT]Main menu is shifted
i3 MK3S FW 3.9.0 RC1

Go to menu "Support" and click on "Date:". It will return you to the main menu but the "Support" line is not at the bottom even it is the last item in the list.

leptun:
Fixes #2620


DRracer (approved):
The fix looks correct, however we must test it properly since it influences every menu in the FW

PR #2835, opened 2020-09-18
Update new messages and their translations

3d-gussner:
As with FW3.9.1 some new messages have been introduced here an updated translation version.


DRracer (approved):

lang/lang_en_fr.txt
```diff
@@ -894,6 +902,10 @@
 "Please unload the filament first, then repeat this action."
 "SVP, dechargez le filament et reessayez."
 
+# c=20 r=4
+"Please check the IR sensor connection, unload filament if present."
+"SVP, verifiez la connexion du capteur IR et si le filament est decharge."
```
awenelo: I think that this should be `et decharge le filament`("and unload filament", instead of "check if the filament is unloaded")

DRracer: is it shorter than before ;) ?

DRracer: `SVP, verifiez la connexion du capteur IR et decharge le filament.`
this way?

awenelo: Yup!
65 vs 72 characters
```
New
SVP, verifiez la connexion du capteur IR et decharge le filament.
Old
SVP, verifiez la connexion du capteur IR et si le filament est decharge.
```

3d-gussner: Also used something similar in German translation as well as it makes it shorter and it makes sense
@awenelo Thanks again

awenelo: You're welcome!

PR #2836, opened 2020-09-20
Remove hard-coded X/Y jerk limits

wavexx:
For some unknown reason DEFAULT_XJERK and DEFAULT_YJERK not only apply as the default jerk after an EEPROM reset, but are also enforced as maximum limits, which shouldn't be!

No other print parameter is handled in the same way, so there's no reason jerk should be arbitrarily limited to the default variant value.

### DIFF
--- a/Firmware/ConfigurationStore.cpp
+++ b/Firmware/ConfigurationStore.cpp
@@ -247,10 +247,6 @@ bool Config_RetrieveSettings()
     {
 
         EEPROM_readData(reinterpret_cast<uint8_t*>(EEPROM_M500_base), reinterpret_cast<uint8_t*>(&cs), sizeof(cs), "cs");
-
-        
-		if (cs.max_jerk[X_AXIS] > DEFAULT_XJERK) cs.max_jerk[X_AXIS] = DEFAULT_XJERK;
-		if (cs.max_jerk[Y_AXIS] > DEFAULT_YJERK) cs.max_jerk[Y_AXIS] = DEFAULT_YJERK;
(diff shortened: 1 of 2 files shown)


DRracer:
So this basically reverts a commit: ac7c0621617bd9e9f8c111dd72f1f24ac6de81c0 done by @mkbel . I'm sure there must have been a reason for doing the jerk limit like this.

mkbel:
@DRracer https://github.com/prusa3d/Prusa-Firmware/commit/ac7c0621617bd9e9f8c111dd72f1f24ac6de81c0 is not changing any logic. It is just variable rename. Or to be more exact it is just clustering all M500 store-able parameters into single struct. You need to dig more into history to find out when
```
if (max_jerk[X_AXIS] > DEFAULT_XJERK) max_jerk[X_AXIS] = DEFAULT_XJERK;
if (max_jerk[Y_AXIS] > DEFAULT_YJERK) max_jerk[Y_AXIS] = DEFAULT_YJERK;
```
appeared for the first time

wavexx:
I'll dig into it. My suspicion initially was that this was done to handle uninitialized values.

mmirate:
The answer is amidst a large commit that does lots of other things:

https://github.com/prusa3d/Prusa-Firmware/commit/9aedaa8fe7b4cc30e48221960cc5440aae7e76b8#diff-7789dc366f36467d54a64d74868dedd992135a3a817cf88a8cc2320028cf709aR314-R315

... and is thus somewhat uninformative.

leptun (approved):

PR #2838, opened 2020-09-21
Implement M204 T (separate travel acceleration)

wavexx:
Allow to separate extrusion and travel acceleration settings using M204, as Marlin 1.1.x and 2.x does using `M204 T`.

This allows to reduce the number of instructions required during printing, since resetting the acceleration just for travel moves is no longer required and can be done a single time during the print.

Provision for this parameter was pre-existing, but not implemented.
M204 has two forms: the lagacy format (Marlin <1.1):

    M204 S[print-acc] T[retract-acc]

and the newer format:

    M204 P[print-acc] R[retract-acc] T[travel-acc]

The distinction in the MK3 FW is done based on the presence of the `P` parameter. If `P` is seen, the new format is *already* used.

In the new format however, `M204 T` was ignored until this change.

To keep backward compatibility, `M204 S[acc]` will set both print and travel acceleration, which is identical in behavior to recent versions of Marlin.

M500 and M503 will report settings using the "new format". A new field in the EEPROM is stored for the travel acceleration, and is initialized with the same value as the print acceleration. The initial default from a fresh eeprom is defined in the variant file, and again it is set to the same value.


wavexx:
I'm thinking now that the current travel acceleration *also* needs to be saved in eeprom during a power-panic, something which is currently missing here.

wavexx:
PR should be complete now. Rebased on top of the current master in order to handle new EEPROM offsets without conflicts.

In the last commit I handle _all_ acceleration settings (not just travel acceleration) during UVLO, which were completely skipped before.

Saving and restoring is not 100% correct with this PR, although is better than nothing. See comments on https://github.com/prusa3d/Prusa-Firmware/pull/2838/commits/3840fe31f220c841277460b68031fcb727f9e154 for details.

This needs to be handled correctly by saving/restore more limits in the future.

DRracer (approved):

PR #2842, opened 2020-09-23
:construction: Enable notification of changed EEPROM values

DRracer:
A prototype solution for discussion
Picked a few random but interesting settings to report to demonstrate the solution.

PFW-1193

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -1278,7 +1278,7 @@ void setup()
         // Once a firmware boots up, it forces at least a language selection, which changes
         // EEPROM_LANG to number lower than 0x0ff.
         // 1) Set a high power mode.
-	    eeprom_update_byte((uint8_t*)EEPROM_SILENT, SILENT_MODE_OFF);
+	    eeprom_update_byte_notify((uint8_t*)EEPROM_SILENT, SILENT_MODE_OFF);
 #ifdef TMC2130
         tmc2130_mode = TMC2130_MODE_NORMAL;
 #endif //TMC2130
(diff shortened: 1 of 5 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -4015,7 +4015,7 @@ eeprom_update_word((uint16_t*)EEPROM_NOZZLE_DIAMETER_uM,0xFFFF);
                strchr_pointer++;                  // skip 1st char (~ 's')
                strchr_pointer++;                  // skip 2nd char (~ 'e')
                nDiameter=(uint16_t)(code_value()*1000.0+0.5); // [,um]
-               eeprom_update_byte((uint8_t*)EEPROM_NOZZLE_DIAMETER,(uint8_t)ClNozzleDiameter::_Diameter_Undef); // for correct synchronization after farm-mode exiting
+               eeprom_update_byte_notify((uint8_t*)EEPROM_NOZZLE_DIAMETER,(uint8_t)ClNozzleDiameter::_Diameter_Undef); // for correct synchronization after farm-mode exiting
```
3d-gussner, line 4018: This can't be changed during print, so maybe leave it and let the host to read it
- after connect
- before starting print
- ???

Firmware/Marlin_main.cpp
```diff
@@ -5561,7 +5561,7 @@ if(eSoundMode!=e_SOUND_MODE_SILENT)
 		eeprom_update_byte((unsigned char *)EEPROM_FARM_MODE, farm_mode);
 		EEPROM_save_B(EEPROM_FARM_NUMBER, &farm_no);
           SilentModeMenu = SILENT_MODE_OFF;
-          eeprom_update_byte((unsigned char *)EEPROM_SILENT, SilentModeMenu);
+          eeprom_update_byte_notify((unsigned char *)EEPROM_SILENT, SilentModeMenu);
```
3d-gussner, line 5564: Not sure that you can change this during print

3d-gussner (commented):
I think we should only notify what can happen during print everything else can be read before.

Firmware/Marlin_main.cpp
DRracer: You can change Stealth<->Normal mode during printing (even though due to TMC behavior the print is temporarily stopped for a second and resumed after)

Firmware/Marlin_main.cpp
DRracer: It makes sense to report changes even when the user operates the LCD menu, hosts like Octoprint may update their UI in sync with what is happening directly on the printer. Moreover, we are not that limited in terms of CPU cycles when not printing (i.e. we may post more notifications).

PR #2851, opened 2020-10-07
LCD status changed

MartinPoupa:
When the lcd status changed it will be printed "LCD status changed" to serial line.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -8968,7 +8968,7 @@ void lcd_ignore_click(bool b)
 }
 
 void lcd_finishstatus() {
-  SERIAL_PROTOCOLLN("LCD status changed");
+  SERIAL_PROTOCOLLNRPGM(MSG_LCD_STATUS_CHANGED);
   int len = strlen(lcd_status_message);
   if (len > 0) {
     while (len < LCD_WIDTH) {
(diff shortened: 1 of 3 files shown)


Firmware/ultralcd.cpp
```diff
@@ -8968,6 +8968,7 @@ void lcd_ignore_click(bool b)
 }
 
 void lcd_finishstatus() {
+  SERIAL_PROTOCOLLN("LCD status changed");
```
leptun: Please change `SERIAL_PROTOCOLLN()` to `SERIAL_PROTOCOLLNPGM()`

MartinPoupa: ok, but why?

leptun: Also, please think of a more abbreviated string. Take this as an example:
`EEPROM value changed` -> `EEPROMChng`

leptun: The PGM attribute puts the string in a different memory space (PROGMEM) instead of RAM directly. As a result, it will use less of the already limited RAM at the cost of a few more instructions

MartinPoupa: I talked about it with Tomáš and we not find abbreviations in other messages from fw. I'm for abbreviations but we must declare some system of abbreviations.

PR #2866, opened 2020-10-21
M601 print OK

MartinPoupa:

PR #2832, opened 2020-09-15
Mk3 auto report temperatures

leptun:
Implement `AUTO_REPORT_TEMPERATURES` and M155 command. Octoprint automatically switches to using the M155 command instead of repeated M105 because of the capability line.

Requires #2786 to be merged first!

PFW-1139

PR #2870, opened 2020-10-26
cmdqueue: Patches to Serial communication, pausing and file completion check

### LINKED ISSUE #2854: [BUG]Errors during M860
**Printer type** - MK3S
**Printer firmware version**-  3.9.1

USB via Octoprint

**Describe the bug**
  For a couple of weeks when I compile the firmware myself and try to print something I no longer get "Recv: echo:busy: processing" messages instead I get "Communication timeout while printing, trying to trigger response from printer. Configure long running commands or increase communication timeout if that happens regularly on specific commands or long moves."  This eventually leads to the print failing with a rerequested line error.  The stock firmware does not do this and send the busy messages.  I've tried completely erasing the build environment and rebuilding from scratch with the same issue.  Prior to this I had no issues with my self-compiled firmware.  The first file is what happens with my firmware and the second with the stock firmware.
  
**To Reproduce**
Build the current firmware, start to print anything.  No busy messages instead communication timeouts.
  
**Expected behavior**
Should get busy messages and no errors.
  
[No Busy.txt](https://github.com/prusa3d/Prusa-Firmware/files/5358258/No.Busy.txt)
[Busy.txt](https://github.com/prusa3d/Prusa-Firmware/files/5358263/Busy.txt)

leptun:
This PR makes a lot of small changes all around the code regarding the cmdqueue.
Things fixed so far:
1) If for some odd reason octoprint decides to get a file list during an SD print (happened to me while I was running the firmware in the emulator), then the current filename will change to something else. I patched it so even if the long filename was for some reason broken, it would still send back to octoprint at least some filename for it to parse correctly the M27 response.
2) Fixed bug introduced with #2832 . Now keep-alive messages are sent correctly.
Closes #2854 
3) Fixed missing Ok in M601 and double Ok in M602.
Closes #2866 
4) M115 now sends the capabilities report only if it isn't provided any arguments. I found out later that M115 is also used for checking that the right firmware is used in gcode, so the report was sent to serial even when it shouldn't. Now only M115 without arguments prints the capabilities report.
5) Fixed a bug where if the last line in a SD file wasn't terminated by a EOL character, then it would leave the cmdqueue in an unfinished state (and possibly corrupted). Now if we encounter the end of file, we queue the line even if it was incomplete for some reason. This should allow file_check() to be able to parse a M84 at the end of the file correctly if it doesn't have a trailing '\n'
6) Instead of just dumping the queue with `cmdqueue_reset()`, now we pop the contents of the queue one after the other in order and check their source. In case we have Serial commands in the queue, we respond with Ok for each of them as to unclutch the Serial communication. If we didn't do this, then Octoprint or any other host for that matter would be waiting indefinitely for the Ok response for a command that no longer exists in the queue. With this, now fancheck works even better, without needing any Ok hacks like we did before.

Bonus: Now if a command is sent to the printer from Octoprint while starting a SD print from the lcd, it won't get discarded and it will be answered after the check_file finishes.

PFW-1150
PFW-1157

### DIFF
--- a/Firmware/cmdqueue.cpp
+++ b/Firmware/cmdqueue.cpp
@@ -97,8 +97,7 @@ void cmdqueue_reset()
 {
 	while (buflen)
 	{
-		uint8_t cmdqueue_current_type = *(cmdbuffer+bufindr);
-		// printf_P(PSTR("dumping: \"%s\" of type %hu\n"), cmdbuffer+bufindr+CMDHDRSIZE, cmdqueue_current_type);
+		// printf_P(PSTR("dumping: \"%s\" of type %hu\n"), cmdbuffer+bufindr+CMDHDRSIZE, CMDBUFFER_CURRENT_TYPE);
 		ClearToSend();
 		cmdqueue_pop_front();
 	}
--- a/Firmware/cmdqueue.cpp
+++ b/Firmware/cmdqueue.cpp
@@ -582,8 +582,6 @@ void get_command()
        ((serial_char == '#' || serial_char == ':') && comment_mode == false) ||
        serial_count >= (MAX_CMD_SIZE - 1) || n==-1)
     {
-      if(card.eof()) break;
-
       if(serial_char=='#')
         stop_buffering=true;
 
@@ -631,6 +629,9 @@ void get_command()
 
       comment_mode = false; //for new command
       serial_count = 0; //clear buffer
+    
+      if(card.eof()) break;
+    
       // The following line will reserve buffer space if available.
       if (! cmdqueue_could_enqueue_back(MAX_CMD_SIZE-1, true))
           return;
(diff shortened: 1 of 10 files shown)


Firmware/cmdqueue.cpp
```diff
@@ -91,14 +95,20 @@ bool cmdqueue_pop_front()
 
 void cmdqueue_reset()
 {
-    bufindr = 0;
-    bufindw = 0;
-    buflen = 0;
+	while (buflen)
+	{
+		uint8_t cmdqueue_current_type = *(cmdbuffer+bufindr);
```
3d-gussner: During compiling with PF-build.sh I get this below, is that correct? That is a new warning.
```
/Prusa-Firmware/Prusa-Firmware-build/sketch/cmdqueue.cpp:100:11: warning: unused variable 'cmdqueue_current_type' [-Wunused-variable]
   uint8_t cmdqueue_current_type = *(cmdbuffer+bufindr);
           ^
```

leptun: Ah, yes. Forgot about that one. It served a different purpose earlier, but then it was used only for the debugging line. Forgot to comment it out as well. Will probably insert it into the debug line.

DRracer (approved):
approving on behalf of successful tests (which took way longer than we expected)

PR #2901, opened 2020-11-24
Fix build

espr14:


Firmware/ultralcd.cpp
```diff
@@ -334,7 +334,7 @@ bool bSettings;                                   // flag (i.e. 'fake parameter'
 const char STR_SEPARATOR[] PROGMEM = "------------";
 
 
-static void lcd_implementation_drawmenu_sdfile_selected(uint8_t row, const char* filename, char* longFilename)
+static void lcd_implementation_drawmenu_sdfile_selected(uint8_t row, char* filename, char* longFilename)
```
DRracer, line 337: please don't touch this function, we know there is a problem and a correct solution is being prepared

PR #2907, opened 2020-12-02
Calibration: improved precision PFW-1162

espr14:


wavexx:
Could you provide an overview of what Z hop does?

espr14:
Z hop is MBL style measuring. Aka up, right, down, up, right, down,...

PR #2911, opened 2020-12-07
Add SuperPINDA Support for MK2.5/S

3d-gussner:
- Changed DETECT_SUPERPINDA to SUPERPINDA_SUPPORT as on miniRAMBo the thermistor readings below 30°C
  aren't accurate enough to determine if SUPERPINDA is connected or not
- Add LCD toggle menu Settings -> HW Setup -> SuperPINDA [Yes/No] to overwrite SuperPINDA detection
  - If EEPROM_PINDA_TEMP_COMPENSTATION is empty = 0xff then detect SuperPINDA by checking thermistor
  - If EEPROM_PINDA_TEMP_COMPENSTAION is 0 then force enable for temperature compensation menus and functions
  - If EEPROM_PINDA_TEMP_COMPENSATION is 1 then force disable for temperature compensation menus and functions

I have been testing quite a bit with MK404:

Starting MK404 with `--terminal` makes it quite easy to test the different scenarios:

_EEPROM values not set_
MK3/S will detect the SuperPINDA as before by checking the temperature
- Go to Calibration and check for "Temp. Calibration"
- Enter in MK404 terminal: `Thermistor2::Set(5)` and re-check

MK25/S will try to detect the SuperPINDA but this might be false detection with the "bad" temp readings on miniRAMBo
The MK404 starts with PINDA temp around 25°C the "Temp. Calibration" is not shown as the "PINDA_MINTEMP" is 30°C
- Enter in MK404 terminal: `Thermistor2::Set(35)` then you can see that the menu is shown.
- To overwrite the detection go to Settings -> HW Setup -> SuperPINDA and set it to [No] and recheck for the "Temp. Calibration" menu. It is always available even if the PINDA temp is above the 30°C limit of the miniRAMBo
Setting SuperPINDA to [Yes] overwrites the detection and disables the temperature compensation menus and functions.

Pros:
- SuperPINDA support for MK25/S printers
- Manual overwrite of SuperPINDA if the temps is below 10°C on MK3/S or 30°C on MK2.5S

Cons:
- Broken PINDAv2 signal cable or PINDAv1 will be detected as SuperPINDA if not set via LCD menu

Improvements ???:
- Add [Auto] to HW Setup -> SuperPINDA

This PR uses
- +0 bytes for MK3S build
- +408 bytes for MK25S build 
compared to the base MK3/MK25S build.

PFW-1167

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -5752,14 +5752,14 @@ void lcd_hw_setup_menu(void)                      // can not be "static"
         MENU_ITEM_SUBMENU_P(PSTR("Experimental"), lcd_experimental_menu);////MSG_MENU_EXPERIMENTAL c=18
     }
 
-#ifdef SUPERPINDA_SUPPORT
+#ifdef PINDA_TEMP_COMP
 	//! The SuperPINDA is detected when the PINDA temp is below its defined limit.
 	//! This works well on the EINSY board but not on the miniRAMBo board as
 	//! as a disconnected SuperPINDA will show higher temps compared to an EINSY board.
 	//! 
 	//! This menu allows the user to en-/disable the SuperPINDA manualy
 	MENU_ITEM_TOGGLE_P(_N("SuperPINDA"), eeprom_read_byte((unsigned char *)EEPROM_PINDA_TEMP_COMPENSATION) ? _T(MSG_YES) : _T(MSG_NO), lcd_pinda_temp_compensation_toggle);
-#endif //SUPERPINDA_SUPPORT
+#endif //PINDA_TEMP_COMP
 
     MENU_END();
 }
@@ -9263,6 +9263,7 @@ void lcd_experimental_menu()
     MENU_END();
 }
 
+#ifdef PINDA_TEMP_COMP
 void lcd_pinda_temp_compensation_toggle()
 {
 	uint8_t pinda_temp_compensation = eeprom_read_byte((uint8_t*)EEPROM_PINDA_TEMP_COMPENSATION);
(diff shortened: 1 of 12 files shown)


Firmware/ultralcd.cpp
```diff
@@ -5752,6 +5752,15 @@ void lcd_hw_setup_menu(void)                      // can not be "static"
         MENU_ITEM_SUBMENU_P(PSTR("Experimental"), lcd_experimental_menu);////MSG_MENU_EXPERIMENTAL c=18
     }
 
+#ifdef SUPERPINDA_SUPPORT
+	//! The SuperPINDA is detected when the PINDA temp is below its defined limit.
+	//! This works well on the EINSY board but not on the miniRAMBo board as
+	//! as a disconnected SuperPINDA will show higher temps compared to an EINSY board.
+	//! 
+	//! This menu allows the user to en-/disable the SuperPINDA manualy
+	MENU_ITEM_TOGGLE_P(_N("SuperPINDA"), eeprom_read_byte((unsigned char *)EEPROM_PINDA_TEMP_COMPENSATION) ? _T(MSG_YES) : _T(MSG_NO), lcd_pinda_temp_compensation_toggle);
```
DRracer: we'll probably show this menu item just for the MK25/S

DRracer: Moreover - a 3-way switch [Auto][Yes][No] would be a more suitable solution here (to cover the undefined/autodetection state as well)

3d-gussner: Changed to be only available on MK2.5/S

3d-gussner: It is [No][Yes]

DRracer:
@3d-gussner could we rebase this onto the MK3_3.9.3 branch?

3d-gussner:
Closed as it has been implemented with https://github.com/prusa3d/Prusa-Firmware/pull/2931

Issue #179, opened 2017-08-21
"Stop Printer" resets the extruder to E1. Please keep the last printing extruder.


m-Th:
"Stop Printer" resets the extruder to E1. Please keep the last printing extruder (the one which is loaded and printing when "Stop Printer" was chosen).

PR #2914, opened 2020-12-09
Extend M403 cmd with new parameters P, S and U giving extruder pulley and MMU load/unload speeds

mdjurfeldt:
The following extensions to the M403 command are important when loading soft filaments. Especially, the ability to reduce the extruder pulley speed substantially decreases the probability of failed loads of softer flex materials.

M403 E\<n\> F\<t\> P\<p\> S\<l\> U\<u\>

sets the following parameters for extruder \<n\>:

t: filament type (0, 1=FLEX or 2=PVA).
p: relative extruder pulley speed factor; defaults to load speed (l)
l: relative load speed factor, 1.0 is normal load speed; defaults to 1.0, or 0.5 for FLEX
u: relative unload speed factor; defaults to load speed (l)

The P, S and U parameters are optional.

This change depends on PR #179 in MM-control-01.

* New function mmu_set_filament_type() sets fil. type and speeds

* New function mmu_load_feedrate computes feedrate based on extruder
and speed factor.

* Macro MMU_LOAD_FEEDRATE calls mmu_load_feedrate

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -7764,8 +7764,15 @@ Sigma_Exit:
 			if(code_seen('E')) extruder = code_value();
 			if(code_seen('F')) filament = code_value();
 			if(code_seen('S'))
-			  relative_speed = relative_unload_speed = code_value();
-			if(code_seen('U')) relative_unload_speed = code_value();
+			  relative_speed = code_value();
+			else if (filament == 1)
+			  relative_speed = 0.5; // default for flex
+			else
+			  relative_speed = 1.0;
+			if(code_seen('U'))
+			  relative_unload_speed = code_value();
+			else
+			  relative_unload_speed = relative_speed;
 			mmu_set_filament_type(extruder,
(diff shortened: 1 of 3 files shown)


primuspingvin:
I think this addition to the firmware solves more of many peoples flex-material printing issues than any hardware thweak to the mk3s could ever do. 

In fact, I was surprised when I once found out it is currently not possible to adjust the loading speed - as basically anyone printing Flex 70A (I use the Recreus brand) *must* use a mod like this to slow down loading, as loading elsewise hardly ever works. Flex 82A and some fragile metal- and woodfills would benefit from slow loading as well. I implemented a rudimentary version for my own needs simply slowing down loading by a factor of two - what a relief to use this superior mod instead without need for re-compiling every firmware version! I can even do MMU-prints with 70A using slower loading!

Hans

DRracer:
All of the proposed changes can now be achieved with writing into MMU's registers (M708).
Thank you for your PR, I'd close it if you don't mind.

PR #2915, opened 2020-12-09
Save a few bytes of RAM by moving homing_feedrate to PROGMEM

mdjurfeldt:
Save a few bytes of RAM by moving homing_feedrate to PROGMEM.

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -276,7 +276,7 @@ void setPwmFrequency(uint8_t pin, int val);
 #endif
 
 extern bool fans_check_enabled;
-extern float homing_feedrate[];
+extern const float homing_feedrate[] PROGMEM;
 extern uint8_t axis_relative_modes;
 extern float feedrate;
 extern int feedmultiply;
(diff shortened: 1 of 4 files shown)


DRracer:
@mdjurfeldt thank you for your PR. It makes sense at the first look, even though there are changes all over Marlin_main.cpp.
My primary concern is: reading a byte from RAM takes 2 cycles, reading a byte from PROGMEM takes 3 cycles.

So a question for you - since some of the calls may be time critical (I know there is an extra call to a common implementation to save a huge amount of code), what is your opinion on slowing the code even further by these extra CPU cycles?

3d-gussner:
@mdjurfeldt As this PR is stale and no answer provided I close it.

mdjurfeldt:
@3d-gussner Right. I'm sorry but I never got the time to dig into this. On the top of my head I'm simply not competent enough to answer your question. I would have to do a lot of analysis. It seems like your judgement is that it could slow down code critically, so I won't object to your closing of this PR.

3d-gussner:
Thanks for your PR and understanding I closed it.
If you think it is still a valid PR reopen it and answer the question.

It was stale for so long and I am trying to get the open PRs down.

PR #2922, opened 2020-12-17
Added some extra instructions for building firmware with a Debian WSL install

mesakomarevich:
The process for building firmware with Debian on WSL is basically the same as Ubuntu, with the primary difference being that bash will not run `PF-build.sh` with `CRLF` line endings. I also updated the instructions to use `sudo` to avoid permission errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,10 +111,14 @@ _notes: Script and instructions contributed by 3d-gussner. Use at your own risk.
 - follow the Microsoft guide https://docs.microsoft.com/en-us/windows/wsl/install-win10
   You can also use the 'prepare_winbuild.ps1' powershell script with Administrator rights
 - Tested versions are at this moment
-  - Ubuntu other may different
+  - Ubuntu and Debian, other may different
   - After the installation and reboot please open your Ubuntu bash and do following steps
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -111,21 +111,25 @@ _notes: Script and instructions contributed by 3d-gussner. Use at your own risk.
...
 - Tested versions are at this moment
-  - Ubuntu other may different
+  - Ubuntu and Debian, other may different
   - After the installation and reboot please open your Ubuntu bash and do following steps
-  - run command `apt-get update`
-  - to install zip run `apt-get install zip`
+  - run command `sudo apt-get update`
+  - run command `sudo apt-get upgrade`
+  - to install zip run `sudo apt-get install zip`
+  - if using Debian:
```
3d-gussner: I think that is useful as well for other Linux distributions to have dos2unix so no need have that only in for Debian.

mesakomarevich: Valid point. Instructions updated.

DRracer (approved):

DRracer:
@mesakomarevich makes sense, thank you, approved, merging

PR #2926, opened 2020-12-18
Kinematics: Fix E mask

espr14:


Firmware/sm4.c
```diff
@@ -173,7 +173,7 @@ uint16_t sm4_line_xyze_ui(uint16_t dx, uint16_t dy, uint16_t dz, uint16_t de)
 		}
 		if (ce <= de)
 		{
-			sm |= 4;
+			sm |= 8;
```
DRracer, line 176: good catch, thanks god no-one was using it ;)

DRracer (approved):

PR #2932, opened 2020-12-25
Calibration: Turn off speed optimization due to MK2.5 PFW-1165

espr14:


DRracer (approved):
sadly, turning off the optimization is really necessary for the MK25/S
approving this PR based on test results

PR #2945, opened 2021-01-05
PFW-1168 Move Z up running xyz calibration

3d-gussner:
to prevent scratches on bed and sheet.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -2871,8 +2871,7 @@ bool gcode_M45(bool onlyZ, int8_t verbosity_level)
 	//set_destination_to_current();
 	int l_feedmultiply = setup_for_endstop_move();
 	lcd_display_message_fullscreen_P(_T(MSG_AUTO_HOME));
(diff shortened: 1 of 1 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -2871,6 +2871,9 @@ bool gcode_M45(bool onlyZ, int8_t verbosity_level)
 	//set_destination_to_current();
 	int l_feedmultiply = setup_for_endstop_move();
 	lcd_display_message_fullscreen_P(_T(MSG_AUTO_HOME));
+  current_position[Z_AXIS] += 0.8;
+  plan_buffer_line_curposXYZE(max_feedrate[Z_AXIS]);
```
espr14: Please, use
`        raise_z_above(MESH_HOME_Z_SEARCH);`
if possible. This is used in G28.

3d-gussner: Thanks for pointing it out. Run a test and it is working!

3d-gussner:
Close it here as it will not be included in FW3.9.3 release.
I will reopen it later for next release.

PR #2955, opened 2021-01-09
Add support for g-code compression/packing over serial connection

scottmudge:
# MeatPack - Easy, fast, effective g-code compression/packing

I've devised and implemented a g-code packing method which is very simple and has very little processing overhead on the MCU, but manages to reduce the size of g-code by a ratio of ~0.61. This is on-par with binary encoding, but it does not require any modifications to the parsing code, g-code files, or anything else. The MeatPack code/module is entirely self-contained, and it only required a few lines to be changed in `cmdqueue.c` to be integrated.

I'm not sure why the file-diff on github makes `cmdqueue.c` look so significantly modified (perhaps white-space?), but the changes are effectively limited to this (+ a closing bracket farther down):

image

This effectively increases the baud rate from 115,200 to 188,850, improving print quality for parts with very dense g-code.

I've also written a functional and tested OctoPrint plugin to work with this modification here:

https://github.com/scottmudge/OctoPrint-MeatPack/

The plugin repository better explains how the packing algorithm works. It's fairly simple and requires very little processing power, but leverages the limited nature of the character set used by most g-code.

The implementation with the firmware leaves the firmware functionally identical to how it already exists. The data packing mode is only enabled if a particular set of command bytes (well outside the ASCII range used by standard g-code) is sent to the MeatPack module by the associated OctoPrint plugin. Otherwise the firmware behaves exactly the same as it always does.

The plugin handles everything automatically. G-code files can remain completely unchanged. The plugin extends the standard PySerial Serial() object with additional functionality, packing whatever g-code data produced by OctoPrint right before it is sent to the firmware.

It also manages and synchronizes the MeatPack "state" of the firmware based on the active configuration of the plugin. If the user enables or disables g-code compression/packing, the plugin module will automatically send associated command bytes to the firmware, where the packing mode is dynamically enabled or disabled. The firmware also sends back confirmation of the enable/disable state. Overall the synchronization is very stable, and I have encountered no issues.

If the plugin is not installed or the user has disabled g-code compression in the plugin settings, the firmware will simply behave as it normally does. 

## How well has it been tested?

I've fully tested both firmware modifications and OctoPrint plugin. I've printed about 100-200 MB of various g-code (produced by both PrusaSlicer and Cura), which was effectively packed to a size of ~60 - 120 MB.

No errors or issues were encountered. And very dense g-code files which normally produced quality issues (zits, stuttering) when being printed at a high feed-rate over the serial connection no longer encountered the same quality issues when the data was packed with this module.

## More Information:

Plugin Repository: https://github.com/scottmudge/OctoPrint-MeatPack
Reddit Post: https://www.reddit.com/r/prusa3d/comments/ksuywv/i_created_a_new_octoprint_plugin_and_firmware_mod/


Cjkeenan:
This would be a very cool feature and very nice to extend the life of OctoPrint users that are on a Zero W as they are especially prone to microstutters. How would this work in relation to https://github.com/prusa3d/Prusa-Firmware/pull/2657 ? They seem like a very good combo for super efficient and compressed GCode.

scottmudge:
> How would this work in relation to #2657 ? They seem like a very good combo for super efficient and compressed GCode.

Both would work together nicely, reducing g-code dramatically. MeatPack doesn't really care what's being sent -- so long as it uses the g-code character set, it will pack it just fine. I think ArcWelder would behave as the OctoPrint-equivalent to the "MeatPack-OctoPrint" plugin for #2657, but some firmwares/printers have a hard time linearizing or planning arc-based motion (floating point arithmetic can be quite expensive). The nice thing about MeatPack is that it only requires a few operations. One or two bit shifts and an access to a short lookup table. 

Regardless, there's nothing incompatible between these two.

wavexx:
@scottmudge did you try to submit this (even just as a proposal to marlin) to 1.1.x/2?
I love this in principle, but it would be a shame if this was ad-hoc just for the Prusa FW.

I'd also be curious to get some feedback from some marlin veterans to see if there's anything we could improve.

DRracer:
@scottmudge that's a very interesting and simple idea, which could improve the throughput of serial printing - ~30-40% less bytes sent over serial can really make a difference.
I still haven't studied the code of yours, however several questions come to my mind:
- what is the code size increase on the AVR and memory requirements?
- how do you distinguish between "regular" G-code and the packed one?
- do you know any "standard" compression algorithms which behave in a similar way?
- any known limitations of your approach?

Firmware/meatpack.cpp
```diff
@@ -0,0 +1,298 @@
...
+        break;
+    }
+    return 0;
+}
+#endif
+
+
+// State variables
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+volatile uint8_t mp_active = 0;             // Defines if MeatPack is active
```
DRracer: Why are these vars volatile? Are you planning to have them accessed from an ISR?

Firmware/meatpack.cpp
```diff
@@ -0,0 +1,298 @@
...
+// Unpacking
+// low = (packed >> 4) & 0xF;
+// high = (packed & 0xF);
+
+//==========================================================================
+inline uint8_t mp_unpack_chars(const uint8_t pk, uint8_t* __restrict const chars_out) {
+    register uint8_t out = 0;
+
+#ifdef USE_LOOKUP_TABLE
+    // If lower 4 bytes is 0b1111, the higher 4 are unused, and next char is full.
```
DRracer, line 184: did you do some comparison in terms of performance gain of either of these methods? Lookup table vs. direct computation?

Please note the AVR cannot shift by 4 -> these will be 4 separate instructions

scottmudge: I loaded up both methods in both Ghidra and IDA, and the disassembly was byte-for-byte identical. 

The compiler optimizes part of the array away completely anyway, and figured out some math to compute an output byte. For the switch statement, it ends up substituting the case jumps for a memory access. In both cases, the end result looks to be faster than either explicit approach. The assembly indicates there is an array access, but not in the way intended by the c-code, and definitely not in the way described by conditional jumps. 

The compiler also "mashed" together all of my separate `mp_` methods into a single, large method. Overall the ability for the compiler to optimize was quite impressive, and indicated being too granular with c-level optimization is likely redundant or unnecessary.

Firmware/meatpack.cpp
scottmudge: 99% of my embedded C experience has been with Cortex processors, and I rarely ever used the volatile keyword outside of its traditional uses. However, for the Arduino compiler, documentation noted that using "volatile" was an expected and frequently used way to force the variables out of registers, which Arduino had noted often behave unpredictably (due to compiler optimization? or architectural limitations of MCUs? it wasn't clear). Given the critical nature of command parsing, I didn't find "unpredictable behavior" tolerable, and found yielding data integrity to "luck of the compiler" was not an option.

If I misinterpreted the Arduino documentation, feel free to modify the variable declarations. The Arduino documentation was particularly terse, so perhaps to avoid confusion between ISR-related memory access and standard memory access, they simply phrased it as "registers are unpredictable", and assumed the performance cost wasn't a concern for most Arduino applications. I did find it very vague.

scottmudge:
> * what is the code size increase on the AVR and memory requirements?

It's minor, less than 1 KB static memory, and only a few bytes of dynamic memory. But I will get you specific byte count. 

> * how do you distinguish between "regular" G-code and the packed one?

A state variable is maintained inside the unpacking module. By default unpacking is turned off until a control byte sequence is received to enable unpacking. There is nothing about the packed g-code (i.e., no bit mask or periodic indicator) that indicates it is packed, beyond the received bytes often lying outside of the standard ASCII range. The unpacker knows to unpack the data based on the value of the state variable.

A particular byte (0xFF) is used to indicate to the unpacker that a control command is expected. 0xFF has two purposes -- if received **once** in sequence while packing is enabled, the following two bytes are treated as full-width. If received once in sequence while packing is **disabled**, it is simply parsed as it normally is by the firmware (i.e., it is ignored, as it is not standard ASCII). But if it received twice consecutively *at any point* (i.e., regardless of the state of the packing enabled variable), the following byte received is treated as a command. Currently command bytes are very high value - ~248-254. This prevents standard g-code for being mistaken as commands. 

The commands are parsed inside the unpacking module, and perform the following functions:

* Set packing state variable to "enable"/1.
* Set packing state variable to "disable"/0.
* Toggle state packing variable from 1 to 0, or visa-versa.
* Send the state of the packing variable back to the PC over the serial connection.

Synchronization of the packing state is done over the serial line with these command sequences. The firmware module sends the state of the packing enabled variable (along with any other MeatPack-specific information) over the outgoing serial connection with an "[MP]" header on the outgoing line. The OctoPrint module parses incoming messages, looking for this "[MP]" header. It then looks for "on" or "off" to determine the state of the the "packing enabled" variable inside the firmware.

In this way, the OctoPrint plugin can safely synchronize the packing state between the PC and the device firmware.

If the state is inconclusive (i.e., the time between when the OctoPrint plugin is initialized and when the "[MP] on/off" message is received, or the time between toggling the Enable/Disable parameter inside the plugin and also receiving the "[MP] on/off" message in return from the firmware to indicate the state has been synchronized), the OctoPrint plugin will buffer outgoing data and send it as soon as the state is safely synchronized.

In this way, there is no chance packed data is received while the firmware is expecting unpacked, and visa-versa.

I've had no issues with state synchronization with this strategy.

> * do you know any "standard" compression algorithms which behave in a similar way?

Most standard compression uses some form of dictionary, but a vast majority use pattern recognition rather than standard data packing. This means the dictionary is dynamic.

In theory the plugin could analyze the g-code file and perform histogram analysis on it, and then send a custom lookup table to the firmware, but in my experience this is unnecessary. The compression ratio is universally ~0.61, regardless of g-code file. 

Because the library is dynamic with standard compression approaches, and because pattern recognition is so computationally expensive with these approaches, standard compression approaches would not be compatible with low-power AVR processors. 

> * any known limitations of your approach?

Nothing directly related to this particular approach. The computational overhead is so minimal, it should not impact the timing of any other process.

In terms of it solving my original problem of stuttering while printing over the serial connection, serial packing has helped, but I still encounter stuttering at very high print-speeds for prints with very dense curvature. But at this point I do not think it is due to limitations with the serial connection. Instead it is due to limitations with the `cmdqueue.c` module... checksum validation, a (universal) lack of multithreading, etc. Reading from the SD card, regardless of the speed of the bus being used (SPI), is simply faster because it's one-way during printing, while serial is bi-directional.

I did find I could alleviate some of this stuttering by removing the `MYSERIAL.checkRx()` calls inside `stepper.cpp`. It looks like these calls were added into the Marlin code-base 9 years ago (https://git.notfound.dk/mikkel/marlinfw/commit/f75f426dfae5190d3e637b247030d3a244968c2a), and the reasoning for doing so doesn't really make sense anymore. OctoPrint wasn't really a thing back then, and interrupting stepper routines with serial comms when printing at high speeds appears to cause issues. Removing these calls also significantly reduced the static memory requirements of the firmware (over 1 KB). Given the definition of this method was inside `MarlinSerial.h` rather than the associated source file, it looked like the compiler was implicitly treating it as an inlined method, meaning the code was unnecessarily duplicated.

Limiting the `cmdqueue.c` serial processing loop to to a certain number of iterations based on the number of full-commands received also seems to help a bit. In this way, it prevents a continually full serial buffer (indicated by `MYSERIAL.available()`) from preventing the `cmdqueue.c` loop from yielding control back to the primary `loop()` function. Limiting the serial parsing loop inside `cmdqueue.c` to 2 or 3 full serial commands seems to be sufficient. Normally it would yield control back when the command queue gets full, but on my experimental branch I've also increased the command buffer size from 4 to 8. So adding a separate loop-break condition was necessary.

Firmware/meatpack.cpp
scottmudge: Alright, after reading up more on this, the general online Arduino documentation was simply being overly vague for the sake of not confusing junior developers.

`volatile` is indeed treated as it normally is, so I've corrected the variable qualifiers. I've also removed the public interfaces to any methods which could have been potentially used in ISRs to modify internal variables. They weren't used anyway. Will commit shortly.

scottmudge:
> * what is the code size increase on the AVR and memory requirements?

Here are the final stats. I've also moved the `ENABLE_MEATPACK` preprocessor directive to "Configuration.h". Where it probably should be.

**WITHOUT MeatPack**:

```
Sketch uses 234030 bytes (92%) of program storage space. Maximum is 253952 bytes.
Global variables use 6437 bytes of dynamic memory.
```

**WITH MeatPack**:

```
Sketch uses 234780 bytes (92%) of program storage space. Maximum is 253952 bytes.
Global variables use 6479 bytes of dynamic memory.
```

I should note this is the MK3S build w/ EN_ONLY language, but given MeatPack is unaffected by language, the byte-count is going to be the same across all build configurations.

FormerLurker:
Ok, I have fixed the issue and ran a quick test, so feel free to take a look at the changes. 

Here is a rundown of what's changed:

1. I re-added n_arc_correction as a parameter.  This can be set via M214 N, where N is an integer between 0 and 255.  It is implemented as a uint8_t and the default value is 25 currently, and could likely be set much higher.
2. I re-added the small angle approximation, which was recently enhanced in Marlin V2.0.6 to include a second term.  This greatly enhances the accuracy.  A third term could be added if it is determined that very large arcs with very high values for n_arc_correction drift enough to make an impact on print quality.  In my testing, 2 terms is more than adequate for n_arc_correction = 25.
3. If n_arc_correction is 1 or 0, sin and cos are used to calculate an accurate value for each segment.  In this case, no further correction is needed, but it will take a bit more time for the first segment to go into the planner.  However, sin and cos are used for each Nth segment, where N = n_arc_correction.  If each segment is very small or if n_arc_correction is very small, this may negatively impact performance.

It will still be necessary to review the ConfigurationStore.cpp to ensure I did that correctly.  It seems to have worked for me, and I know that loading and saving works as expected.  It's the initial detection of an uninitialized configuration that I'm still not 100% sure I got right.  Also, the version may need to be incremented, but I'm not sure what that should be.  It is currently set as V2, but perhaps it should be V3 since these new values were added.

Please let me know if you need anything to help you evaluate these changes.  I can assist you in connecting arc welder to prusa slicer if you have any models you would like to test.  I can also provide before/after photos of gcode that contains G2/G3 showing large accuracy and quality improvements that can be had if this pull request is merged.  I am willing to make any changes you deem necessary, and would welcome other feedback for potential improvements.

Thank you in advance for your review and consideration.

DRracer:
@scottmudge I can see you really care about your proposal, that's great :+1: . Generally, the idea is simple and we find your implementation very good and considering it for testing now. The code size requirements are acceptable.

Another simple question - do you have some simple command line tool which would compress an existing G-code file? I know the algorithm is simple, but we'd like to perform some offline testing as well.

Thank you.

ellensp:
I note your packing spaces.  Why don't you just strip them out? I don't believe they are needed unless inside quotes.
Then you could add in 'Y'

scottmudge:
> I note your packing spaces. Why don't you just strip them out? I don't believe they are needed unless inside quotes.
> Then you could add in 'Y'

I may make that an extra option, as that would increase data throughput significantly more. The RS-274 standards indeed note spaces should be ignored, but there were cases of parsing not fully respecting these standards:

For instance: 
https://github.com/MarlinFirmware/Marlin/issues/2715

I'll add an optional dictionary shift and dictionary version command into the firmware to make this an optional change, just in case some firmwares encounter issues.

scottmudge:
> Another simple question - do you have some simple command line tool which would compress an existing G-code file? I know the algorithm is simple, but we'd like to perform some offline testing as well.

This was how I first tested my implementation. I did write a python utility (converted to an executable w/ PyInstaller for easier global scripting) which packed g-code, but found I needed to make more modifications to the firmware to ensure that the file-seek behavior performed when selecting a gcode file from the SD card performs correctly.

Because of the explicit requirement for sequential unpacking, by jumping to some random byte in a packed stream, the unpacker must traverse the previous 2 bytes to ensure none of the bytes at the current index are full-width (and thus not misinterpreted). 

I implemented this behavior into the firmware where the SD card file was read from a certain offset from the end (if it was longer than a pre-determined amount) to check for an M84 command. After these modifications, it could read packed g-code form the sd-card just fine.

However, I removed this code from the PR, as I figured it would be less desirable to modify a separate module in the code which would require further testing, and currently the sd-card interface does not present any transmission rate bottlenecks, like the serial port does. 

But if you plan to simply analyze the way bytes are organized, I will provide you this commandline utility. 

I should note (it may not be explained well in the original repo) that some of the characters in the byte stream are re-arranged to ensure no 4-bit window for packed data is wasted. I explain this here (https://github.com/scottmudge/OctoPrint-MeatPack/issues/3#issuecomment-759932511) but just in case you see some unexpected data organization in the gcode files, this is why.

scottmudge:
> I note your packing spaces. Why don't you just strip them out? I don't believe they are needed unless inside quotes.
> Then you could add in 'Y'

I implemented a version of this in both firmware and in the OctoPrint plugin, but it presented too many issues. First I had to strip and re-compute the command checksum after stripping out so many characters. Secondly, I think there's too many parsing rules which are in some ways expecting white space here and there. 

I hand-checked that all of the g-code was correct without white space, but when I tried to print, it ended up causing major issues. Crashed the head into the print bed due to errant speed, etc.

Ultimately it sound good in theory, but in practice it isn't going to be well supported.

ellensp:
That is a shame..  might do my own tests see if I can work out what breaks and get it fixed. Then maybe can have v2 of meatpack...

scottmudge:
> That is a shame.. might do my own tests see if I can work out what breaks and get it fixed. Then maybe can have v2 of meatpack...

Alright, I'll push a separate branch you can work on with all my current changes, for both the firmware and the OctoPrint plugin. You may need to debug the on the firmware end too.

scottmudge:
> That is a shame.. might do my own tests see if I can work out what breaks and get it fixed. Then maybe can have v2 of meatpack...

After looking into this more, it looks like some newer G-Code explicitly requires whitespaces.

For instance, M117, which OctoPrint uses to send updates to the LCD often sends whitespace along with the LCD message. 

I would need to implement a reactive parser to outgoing g-code, and dynamically disable whitespace omission for certain g-code . The white spaces in this instance would be full width.

Is there a list of g-code where white-space would need to be retained?

FormerLurker:
What about stripping white space only for G0, G1, G2, and G5?  These should account for the vast majority of codes sent, and are the ones where speed is most essential.

scottmudge:
> What about stripping white space only for G0, G1, G2, and G5? These should account for the vast majority of codes sent, and are the ones where speed is most essential.

Yes that is what I am testing now. It looks like a lot of the 'M' codes are implemented specifically for certain printers/firmwares, so I will restrict the whitespace removal to 'G' codes.

scottmudge:
Alright I attempted to remove all whitespace for "G" code operations only, but some commands are still being parsed incorrectly, leading to the extruder crashing into the end of the x-axis during the initial extruder wipe at Y=-3.0mm at the start of the print. It also never extruded anything. 

The extruder went all the way to the end of the build plate, didn't extrude anything, and never stopped. Some of the "G" commands used during print initialization are simply not being interpreted without whitespace.

When I added white space back in, the behavior stopped.

I may limit whitespace omission to G0 and G1 commands, as they are by far the most common. And they appear to behave normally. 

Even with adequate serial throughput, it looks like stuttering may still occur due to issues with motion planning (https://github.com/prusa3d/Prusa-Firmware/issues/1406), but it would be good to eliminate the serial port as a potential bottleneck.

FormerLurker:
Consider G2 and G3 perhaps?

scottmudge:
Yes all the single-digit G# commands will probably be fine. I think it's the extended set that are having trouble. I'll dig deeper to figure out which command(s)' parsers need to be refactored or looked at.

scottmudge:
Alright here's what I've discovered regarding whitespace omission:

* I've limited whitespace omission to the single-digit "G" commands -- G0 through G9. 
* I've implemented new configuration modes in the firmware to enable extra features like whitespace omission.
* I've implemented protocol versions to enable backward compatibility with older versions of the firmware (for OctoPrint plugin).
* When "No Whitespace Mode" is enabled, packed spaces are replaced with packed "E" characters.
* I've reflected the characters unpacked *on the firmware* back to the host via the serial connection. When "No Whitespace Mode" is enabled, the stripped "G" commands are properly received and unpacked, even the swapped packed-"E" characters, and the changed checksum is correct as well.
* **G1 commands** appear to be parsed **incorrectly** without whitespaces. I've haven't looked into why, but at the start of a print, when the extruder is being purged at the front of the build plate, nothing is being extruded, and the printhead/extruder never stops moving along the X-axis, eventually crashing into the end of the build volume, unless stopped. 

To me it appears as though the parser is waiting to receive a terminal whitespace character or something... I'll have to look into it more.

But for now, it looks like firmware-side issue. I've confirmed the actual serial data is sent and received correctly.

scottmudge:
Okay I figured out the issue.

When it's extracting the coordinates and values from a line like:
`G1 X119.159 Y116.538 E0.01227`

without whitespaces is:

`G1X119.159Y116.538E0.01227`

During this coordinate extraction, the function `strtod()` is used. In this case it gets the x-coordinate fine, but when it tries to extract the Y coordinate, it misinterprets "E" as an exponent delimiter. 

I'll need to refactor `void get_coordinates()` to buffer the pointers to each code letter, and then ensure `strtod()` doesn't parse beyond them.

---

Edit: It looks like the simplest option is to implement a version of `strtod()` which does not accept exponent notation.

Does anyone know if G-Code is explicitly meant to tolerate exponent notation?

scottmudge:
I have copied the `strtod()` source code for `avr-libc`. There are two options here:

1. Disable the section checking for exponents, *entirely*.
2. Disable the section checking for exponents from parsing *upper case 'E'*, leaving lower-case available. I don't think this is desirable, as it leads to ambiguity.

I don't see anything regarding a requirement to support exponents in RS-274, so I think option 1 is a viable option, unless anyone has a different opinion.

scottmudge:
Actually it looks like the original Marlin developers encountered this issue too. 

Except instead of implementing a proper version of `strtod()`, they just hacked their way around it:

image

Which works, but it isn't as performant, as it has to perform an unnecessary character scan, which happens inside of strtod() anyway. It also looks like they're only bypassing capital `E` characters, so I will use this convention with the custom strtod() method.

scottmudge:
Alright I made the patch and it corrected the issue. Here is the firmware change: https://github.com/scottmudge/Prusa-Firmware-MeatPack/commit/ebf039ca89872396783b041429c0b4054d488090

The firmware now correctly parses gcode without any whitespace, at least for the "G" commands. The compression ratio is now ~0.54. but it should be at or below 0.5 once I evaluate and enable the other g-code commands which don't explicitly require whitespace.

So far so good though. I will bundle an update in the next day or so after further testing.

scottmudge:
I posted this on the Marlin PR just now, but since this is the original PR, I will cross-post my results:

---

I've tested the following features, printing ~3 12-hour prints for a total of ~85MB raw gcode, ~46.75 MB effectively sent MB over serial:

* Normal packing method described by original MeatPack project.
* Removal of whitespace and automatic correction of lower-case to upper-case letters (in "G" commands only, as "M" commands sometimes receive arbitrary strings as parameters).
* Replacement of space ' ' character with 'E' in the packing table when white-space removal is active.
* Implementation of protocol versions to aid backward compatibility between firmware and plugin, in the event future features are added. By default if the MeatPack firmware module emits no protocol version, this indicates protocol version 0, where only the original packing method is supported. 

I've encountered no issues, but I did need to correct one aspect of the firmware outside of the MeatPack code. As described previously, the `strtod()` method used to convert the ASCII strings to floats was parsing upper-case "E" characters as exponent markers. So when whitespace was removed between the `Y` parameter and the `E` parameter for G0/G1 commands, both the parsed Y-component was incorrect, and the E-component was lost. 

To fix this, I implemented a modified version of `strtod()` which ignores **uppercase** 'E' characters. This allows for exponent notation to be used, but only with lower-case 'e' characters.

There was a hacky workaround already in the Prusa firmware code (possibly from its Marlin roots), but it was both not being used where it should have been, and it was inefficient. It was performing a secondary substring scan for the 'E' character, replacing it with \0 if it existed, extracting the float with standard `strtod()`, and then replacing the \0 with E before returning. It was horribly inefficient, so this has been corrected.

After all these fixes, I am getting an improved compression ratio of ~ 0.55:

image

Meaning the effective rate of the 115,200 baud serial connection is now ~210,000 baud. A significant improvement.

I plan to update the OctoPrint plugin and modified Prusa firmware release with these changes soon (from which I can work with your existing Marlin PR, but you will need to test it, as I do not have a Marlin-compatible MCU). Unfortunately I am quite busy with my other work, so this will likely happen in the next few days.

Thanks for the great ideas about culling the whitespace, it worked quite well after some debugging.

3d-gussner:
Very interesting project!
Have a look at https://github.com/vintagepc/MK404 which is a simulator for the MK3S with Prusa-Firmware
but also with Marlin 2.x on Einsy board.
I started a WIP Marlin 2.x for MK3S https://github.com/3d-gussner/Marlin/tree/MK3s_2.0.x which works in the sim and on real MK3S (still needs some work/fine tuning) and maybe you can test your idea with it.

scottmudge:
> Very interesting project!
> Have a look at https://github.com/vintagepc/MK404 which is a simulator for the MK3S with Prusa-Firmware
> but also with Marlin 2.x on Einsy board.
> I started a WIP Marlin 2.x for MK3S https://github.com/3d-gussner/Marlin/tree/MK3s_2.0.x which works in the sim and on real MK3S (still needs some work/fine tuning) and maybe you can test your idea with it.

Thanks for sharing. Does it somehow load a firmware image, or is it emulating the firmware entirely as a reimplemented host software layer?

I'm hesitant to invest too much time into modifying host-level code that is not immediately transferrable to an actual 8-bit AVR MCU (compiler optimizations will be different, etc). But if I could load a firmware image into a host-side emulator, that would be great. 

----

Edit: Nevermind, looks like it can load an ELF binary. Does this require Linux, or can it run on Windows? I can pull out a VM, but wanted to check before I start compiling.

leptun:
I’ve only used the sim on linux, but in theory it should also work with cygwin.
As for the files that can be loaded, it also accepts hex and bin files, not only elf. Elf is superior when you want to debug stuff.

scottmudge:
Awesome, this will help a lot with maintaining the integration with Marlin.

scottmudge:
I've updated the OctoPrint plugin to v1.4.1, and the modified version of the Prusa firmware to support the new whitespace removal feature.

Please build the firmware from the branch on this PR, or download the pre-built hex images from my fork's repo.

This update includes the new whitespace removal feature, bringing the compression ratio down to a reliable ~0.55. I've tested it with about 80 MB of g-code without issue.

## Note:

In order to correct the issue with the existing`strtod()` method ("E" characters being interpreted as exponents without prefixed whitespace), I implemented a modified version which ignores uppercase "E" characters when they trail a stream of numbers. 

**Lowercase** "e" characters are still interpreted as exponent markers.

The new `strtod()` method is a direct copy from AVR's libc implementation, with only minor changes, so it should perform exactly the same.

I also removed all references to the standard `strtod()` method, so it should not be linked against. Therefore the resulting binary size should not be appreciably different.

DRracer:
@scottmudge how does the overall print time keep up with the slicer's expected time to finish? Did you see any improvements using your approach - i.e. total print time closer to nominal print time - especially with short segments?

scottmudge:
At slow, "quality" print speeds the difference is less apparent, but at higher "draft" print speeds it does appear to keep up a bit better with polygon-dense objects when printing over USB/serial. There is still an upper ceiling if I really crank up the speed (~170-180% of the "quality" speed), but I think that's primarily due to limitations with the motion planner and the block buffer, which themselves are probably limited by the hardware. 

The difference is somewhat qualitative, but  the best quantitative data I have is that a model that would typically begin to stutter and have issues at 140% speed can print comfortably at 155-160% speed with MeatPack enabled.

Another benefit is that because multiple characters are being sent at each serial interrupt, the serial interrupt is going to be triggered less frequently, potentially improving timing or responsiveness of other processes.

fiveangle:
> Set packing state variable to "enable"/1.
> Set packing state variable to "disable"/0.
> Toggle state packing variable from 1 to 0, or visa-versa.
> Send the state of the packing variable back to the PC over the serial connection.

As since you have the ability to explicitly set the packer states to on or off plus can query the state at any time, the toggle seems extraneous and ripe for abuse by assuming state rather than stetting or confirming by a state query. Unless there was use case envisioned that I’m missing, couldn’t the toggle command be eliminated and replaced with an “Enable w/Whitespace Elimination”, so that it becomes a runtime option rather than compile-time option ? This way, someone trying to perform direct gcode controls or troubleshooting through the terminal can enjoy human readable output, while fully autonomous sender mode always gains the extra compression afforded white space elimination ? Of course the code size change would need to be characterized to ensure it’s acceptable. 

Total aside, I’m surprised E was more prevalent than Y but life is full of surprises until you look at the data :)

There were high hopes for the XON/XOFF buffered flow control a few years ago to mitigate the issue but it proved too problematic due to other moving parts. Nice little hack to help bandaid the long-standing problem with the ping-pong serial comms bottleneck issue 👍

scottmudge:
@fiveangle You're right, the toggle command is completely unused, I'll remove it. I added it as an initial thought at first, but after implementing the host/device state synchronization code, it became clear it wasn't going to be used. I don't want to change any of the pre-existing command byte values for forward/backward compatibility, but the toggle byte can be reused for something else later if needed. 

I'll do some minor refactoring to clean up unused functions.

fiveangle:
"Backward compatibility"

Oh, I thought this was a proposal and not yet merged.

scottmudge:
> "Backward compatibility"
> 
> Oh, I thought this was a proposal and not yet merged.

Backward compatibility for anyone who's already installed it into their OctoPrint and who is using the prebuilt firmware binaries on my firmware repo fork.

It is still a PR for the official repo.

scottmudge:
A brief update -- the OctoPrint plugin has been added to the official plugin repository, and Marlin has now merged MeatPack support. So it's had a few more eyes to review the code's safety and efficacy.

FormerLurker:
@scottmudge, I'll give it a go.  I'd like to see if it compliments ArcWelder too.

FormerLurker:
@scottmudge, do you mind turning on issues in your firmware repo?  I'd like to create one so as not to bog down the PR with unnecessary clutter:

image

scottmudge:
Sure, it's been enabled.

fiveangle:
>> It is still a PR for the official repo.

Right... so my proposal was to reuse it, specifically for enable/disable whitespace elimination.  It seems like it would be useful for debugging -- not debugging your tool, but in debugging other gcode issues such as confirming our existing octoprint plugins work with meatpacker in both configurations without having to recompile/reupload firmware just to test-fix-retest, back and forth.

Cjkeenan:
Is this still in consideration for official support in the Prusa firmwares? I would love to breathe new life into my ZeroW.

taxilian:
Is this going anywhere? I've been using it on one of my printers for awhile and it's pretty slick.

Cjkeenan:
> Is this going anywhere? I've been using it on one of my printers for awhile and it's pretty slick.

What firmware are you using? Did you compile it yourself? I have been looking to test both this out along with #2657 since I like using the RPi ZeroW due to its very convenient form factor, not to mention official support, but these days it just cannot keep up with the printer and it leads to quality degradation compared to the SD. I was hoping that this compression along with Arc-Welder's interpolation would significantly reduce the overhead on the ZeroW from the excess and uncompressed G-Code (I noticed severe quality issues on arcs in particular).

taxilian:
I compiled it myself, merging this into a bondtech version of the firmware to match the printer in question

scottmudge:
There are also precompiled versions of 3.10.0 on my branch, for both stock and Bear extruder.

GhostlyCrowd:
I'm running your compiled version from your fork. Its working great. Cant wait to see this in main. As well as Arc fixups.

Strat00s:
Will this ever see the light of day?

fiveangle:
@scottmudge, since it doesn't appear there will be an implementation of this included in official Prusa firmware anytime soon, are you familiar enough with the available CI tools that would allow you to auto-generate downloadable firmwares for the AVR based Prusa i3s? I know this would require you to rebase your code against every new official release, which may be non-trivial but as you point out, the changes are quite limited in scope. You would be doing a massive service to the entire Octoprint community who use Official Prusa i3 printers. And if it became too unwieldy, simply stop?

Please consider, but regardless, thank you for all the thoughtful work you've done on this. The implementation is really quite elegant and effective. You should be quite proud of it :)

AndrewLuebke:
I agree, I use it all of the time.  It's very useful and I wish Prusa would include it in the main line.

wavexx:
A lot of conflicts now, would it be possible to rebase with the recent cmdqueue changes?
Most of the big changes got merged so now it would make sense now to have a look again. The only constraint is really code size, since it's getting pretty tight.

scottmudge:
I'll take a look and rebase everything, when I measured the code size/memory impact last it was pretty minimal. @wavexx what sort of margins/limits are we looking at with the latest version of the FW?

wavexx:
Well, we had this sort of scheme where for each flash byte we needed we shaved it somewhere else ;)

scottmudge:
Updated PR from new branch equal w/ 3.12.2, cleaned up whitespace changes to make it easier to review. 

https://github.com/prusa3d/Prusa-Firmware/pull/4067

PR #2956, opened 2021-01-11
Workaround for skipping large comment blocks

DRracer:
If there are large blocks of comments in the G-code, the printer may get shot down by its own watchdog. Watchdog is generally set to 4s and updated only in manage_heaters (and some other spots in some specific cases).

So far, the code reading the file and feeding it into Marlin cycles indefinitely until it finds valid G-code lines and fills up the command queue. If the block is large enough, the printer cannot read it completely within those 4s.

A simple workaround - bail out after some consecutive empty/comment lines to enable other parts of code do their job (especially manage_heaters).

Tested on MK404, previous FW fails with 600KB of comment lines at the beginning, this patch survives. The printer even draws some update on its status screen before starting a real print.

Tested also on a real MK3S with a similar result (survives without a watchdog)

PFW-1175

### DIFF
--- a/Firmware/cmdqueue.cpp
+++ b/Firmware/cmdqueue.cpp
@@ -573,6 +573,7 @@ void get_command()
   // this character _can_ occur in serial com, due to checksums. however, no checksums are used in SD printing
 
   static bool stop_buffering=false;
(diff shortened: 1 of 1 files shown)


DRracer:
Closing in favour of PR #2967

PR #2957, opened 2021-01-11
Remove stray serial newlines in fsensor autoload messages

wavexx:


DRracer (approved):
makes sense, looks like a typo, may be after some refactoring from printf()

PR #2960, opened 2021-01-12
Calibration: more accelerations, fix scan, more robust circle search PFW-1169

espr14:


DRracer (approved):
the code needs some serious diet for the AVR, but the function is correct

PR #2969, opened 2021-01-14
:construction: Calibration: allow 400 step XY motors PFW-1181

espr14:
Dependent on PR #2968


nosRider1:
I tested it.
Correction is good up to 2 points, but at 3 points the problem arises.
Only the 3 point part is mistaken for Y coordinate as X coordinate.
Y aixs should move, but the X axis moves to the right, so 3 points cannot be checked.

Bed Points Map
+-----------------------+
| O (pass)     X (Failed)  |
|                                    |
| O (pass)     O (Pass)    |
+-----------------------+

3d-gussner:
Closing as we don't support 0.9° 400 step motors.

PR #2700, opened 2020-05-29
[WIP] SRAM usage optimization

leptun:
This really is [WIP]! Test it with caution. Anything could go wrong. A lot of changes have been made around the code and not really tested.
Each and every code change has to be checked that it still produces the same Serial output for backwards compatibility.

Resources used [MK3 FW3.9.0](https://travis-ci.com/github/prusa3d/Prusa-Firmware/builds/166731977) vs this [PR](https://travis-ci.com/github/prusa3d/Prusa-Firmware/builds/168728738):
|Printer| FW390 program space | FW390 mem | PR program space | PR mem|Diff program|Diff mem|
|-------|-------|-------|-------|-------|-------|-------| 
| MK3S | 248810 | 6486 | 248082 | 5835 | -728 | -651 |
| MK3 | 250396 | 6515 | 249538 | 5866 | -858 | -649 |
| MK25S 13a | 222264 | 6310 | 221612 | 5701 | -652 | -609 |
| MK25S 10a | 222144 | 6310 | 221118 | 5701 | -1026 | -609 |
| MK25 13a | 226274 | 6358 | 225442 | 5751 | -832 | -607 |
| MK25 10a | 226004 | 6358 | 225110 | 5751 | -894 | -607 |

PFW-1123

PR #2982, opened 2021-01-22
Reduce code size and RAM usage

DRracer:
This is an attempt to reclaim some of the code size lost during the new XYZ calibration and addition of new functions. More commits will come here as we manage to optimize and clean more code.

Kudos to @leptun for the code_seen_P()

Ideas will come from PR's #2688 and #2700


Firmware/ultralcd.cpp
```diff
@@ -5930,14 +5925,7 @@ void bowden_menu() {
...
 
-				lcd_set_cursor(0, 0);
-				lcd_print(' ');
-				lcd_set_cursor(0, 1);
-				lcd_print(' ');
-				lcd_set_cursor(0, 2);
-				lcd_print(' ');
-				lcd_set_cursor(0, 3);
-				lcd_print(' ');
+				lcd_puts_at_P(0, 0, PSTR(" \n \n \n "));
```
DRracer, line 5864: @leptun we may even extract duplicit strings into a variable and reuse them

Firmware/ultralcd.cpp
```diff
@@ -4240,23 +4170,23 @@ void lcd_pick_babystep(){
...
     
     lcd_set_cursor(12, 3);
     
-    lcd_print("4");
+    lcd_print('4');
     
     lcd_set_cursor(1, 2);
     
-    lcd_print(">");
+    lcd_print('>');
```
mkbel, line 4189: Change in commented out code.

Firmware/ultralcd.cpp
```diff
@@ -4287,20 +4217,20 @@ void lcd_pick_babystep(){
...
+                lcd_print(' ');
                 
                 if (cursor_pos < 3) {
                     lcd_set_cursor(1, cursor_pos+1);
-                    lcd_print(">");
+                    lcd_print('>');
                 }else{
                     lcd_set_cursor(10, cursor_pos-1);
-                    lcd_print(">");
+                    lcd_print('>');
```
mkbel, line 4233: Change in commented out code.

Firmware/ultralcd.cpp
mkbel: I would go even further -  extract code to clear first column 3 or 4 rows into separate function, as this code
`lcd_puts_at_P(0, 0, PSTR(" \n \n \n "));`
or
`lcd_puts_at_P(0, 1, PSTR(" \n \n "));`

Is repeated several times.

Firmware/xyzcal.cpp
```diff
@@ -18,6 +18,8 @@
 
 #define _PINDA ((READ(Z_MIN_PIN) != Z_MIN_ENDSTOP_INVERTING)?1:0)
 
+static const char endl[2] PROGMEM = "\n";
```
mkbel, line 21: I don't think array length needs to be stated here, it can be deduced from initialization string.

Firmware/xyzcal.cpp
```diff
@@ -840,26 +855,31 @@ void dynamic_circle(uint8_t *matrix_32x32, float &x, float &y, float &r, uint8_t
 
 		const float reducer = 32.f; ///< reduces speed of convergency to avoid oscillation
 		const float norm = 1.f / reducer;
-		x += CLAMP(median(shifts_x, blocks) * norm, -max_change, max_change);
-		y += CLAMP(median(shifts_y, blocks) * norm, -max_change, max_change);
-		r += CLAMP(median(shifts_r, blocks) * norm * .5f, -max_change, max_change);
+//		x += CLAMP(median(shifts_x, blocks) * norm, -max_change, max_change);
+//		y += CLAMP(median(shifts_y, blocks) * norm, -max_change, max_change);
+//		r += CLAMP(median(shifts_r, blocks) * norm * .5f, -max_change, max_change);
```
mkbel, line 860: Warning, commented out code.

Firmware/ultralcd.cpp
leptun: That could be done. Not sure if it will be as efficient, but I'll try.

Firmware/ultralcd.cpp
DRracer: this change was done automatically

Firmware/ultralcd.cpp
DRracer: this change was done automatically

Firmware/xyzcal.cpp
DRracer: could be, you are right

mkbel (approved):

Firmware/xyzcal.cpp
DRracer: this code will be removed after testing of this PR - saved 100B

PR #2983, opened 2021-01-22
Fix sm4.h line ending

leptun:


3d-gussner:
This may break https://github.com/prusa3d/Prusa-Firmware/pull/1263

DRracer (approved):
Approving - it is only a reformat from CR/LF to LF only to make it consistent with the rest of our repository, no code-change has been done. Affected PR's will be fixed later.

PR #2984, opened 2021-01-22
Do not send CR on the serial line

leptun:
Make the serial output format uniform by not sending CRLF in some places while only sending LF in others. From now on all Serial output should be LF only (just like in stdio).

PR is a draft since compatibility must be tested


DRracer (approved):
Approving conditionally - this will make the serial output consistent, however we may find obscure issues later on. For 3.10 - this is ok as the FW will be tested thoroughly

PR #2934, opened 2020-12-27
Modify output from M20 to include full filename rather than 8.3 format

drbobbob:
This changes the output from the M20 list command to include the full file names rather than an 8.3 formatted name.  This makes the files show up much nicer in octoprint.

If this is taken with #2731 then octoprint is much nicer when dealing with files on the SD card.

PR #2991, opened 2021-01-24
Modify output from M20 to include full filename rather than 8.3 (memory usage unchages

drbobbob:
Alternative to #2934 which does not increase memory usage.  This will only send the full filename, the path will still be in the 8.3 format.  This will not modify the stack usage as compared to master.  This is just using a variable that is already populated when sending the serial message

### DIFF
--- a/Firmware/cardreader.cpp
+++ b/Firmware/cardreader.cpp
@@ -117,7 +117,7 @@ void CardReader::lsDive(const char *prepend, SdFile parent, const char * const m
 					case LS_SerialPrint:
 						createFilename(filename, p);
 						SERIAL_PROTOCOL(prepend);
-						SERIAL_PROTOCOL(filename);
+						SERIAL_PROTOCOL(longFilename[0] == '\0' ? filename: longFilename);
 						MYSERIAL.write(' ');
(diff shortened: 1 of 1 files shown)


leptun:
Can you start a print from octoprint when you do this? I think you can’t since M23 accepts only SFN on our firmware

leptun:
Yes. I can confirm this. You can't start a print with this if the filename is a LFN:
```
Send: M23 /oldgco~1/3d_printer_test_mini.gcode
Recv: echo:Now fresh file: /oldgco~1/3d_printer_test_mini.gcode
Recv: oldgco~1R
Recv: open failed, File: oldgco~1R.
Recv: open failed, File: /oldgco~1/3d_printer_test_mini.gcode.
Recv: ok
```

drbobbob:
Well darn.  I was going to test octoprint this weekend, but hadn't gotten to it.  I generally start my prints from the LCD and not through octoprint.  I just use octoprint to monitor the status and get an estimate when the print will be done.

Since this breaks the ability to start a print via octoprint then I supposed that this should be closed.

DRracer:
@drbobbob ok, closing upon your request

PR #3000, opened 2021-01-28
 Optimization: Extract common call to plan_buffer_line_*XYZE with st_synchronize

DRracer:
Didn't bring any significant code size reduction, the impact on performance may not outweigh the code changes.
Just ~220B down...


DRracer:
closing after internal discussion due to impact on performance

PR #3003, opened 2021-01-29
Fix/improve lcd_implementation_drawmenu_sdfile_selected

wavexx:
Fix a warning in ``lcd_implementation_drawmenu_sdfile_selected`` casting ``filename`` to non-const:

```
sketch/ultralcd.cpp: In function 'void lcd_implementation_drawmenu_sdfile_selected(uint8_t, const char*, char*)':
sketch/ultralcd.cpp:348:22: warning: invalid conversion from 'const char*' to 'char*' [-fpermissive]
         longFilename = filename;
```

Modify lcd_implementation_drawmenu_* functions to never modify input parameters. Truncation in these functions was never needed.

Partially rewrite ``lcd_implementation_drawmenu_sdfile_selected`` to be more efficient, and also avoid scrolling one character past the end of the long filename. The rewrite doesn't require modifying the input parameter as a result, fixing the warning as well.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -332,59 +332,58 @@ bool bSettings;                                   // flag (i.e. 'fake parameter'
 const char STR_SEPARATOR[] PROGMEM = "------------";
 
 
-static void lcd_implementation_drawmenu_sdfile_selected(uint8_t row, const char* filename, char* longFilename)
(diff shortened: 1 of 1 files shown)


wavexx:
Note that this conflicts with #2405

wavexx:
Closing in favor of #2405

PR #3005, opened 2021-01-29
Allow all microstep resolutions for all axes

wavexx:
The current FW limits the microstep resolution depending on the motor.
XYZ are limited to 32 usteps, while E requires 8 or higher. This is sensible, but prevents debugging or custom configurations.

This PR simplifies the check in order to require a valid µstep resolution without imposing any restriction on the resolution.
Some combinations might not be that useful or work properly, but I believe it's up to the user to decide.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -8530,7 +8530,13 @@ Sigma_Exit:
 			if(code_seen(axis_codes[i]))
 			{
 				uint16_t res_new = code_value();
+#ifdef ALLOW_ALL_MRES
 				bool res_valid = res_new > 0 && res_new <= 256 && !(res_new & (res_new - 1)); // must be a power of two
+#else
+				bool res_valid = (res_new == 8) || (res_new == 16) || (res_new == 32); // resolutions valid for all axis
+				res_valid |= (i != E_AXIS) && ((res_new == 1) || (res_new == 2) || (res_new == 4)); // resolutions valid for X Y Z only
+				res_valid |= (i == E_AXIS) && ((res_new == 64) || (res_new == 128)); // resolutions valid for E only
+#endif
 				if (res_valid)
 				{
 					st_synchronize();
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -8530,9 +8530,7 @@ Sigma_Exit:
 			if(code_seen(axis_codes[i]))
 			{
 				uint16_t res_new = code_value();
-				bool res_valid = (res_new == 8) || (res_new == 16) || (res_new == 32); // resolutions valid for all axis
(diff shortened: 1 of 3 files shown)


DRracer:
_... it's up to the user to decide._ ... that's a fair point, however our FW should be safe for general usage/users in the first place. Allowing too many degrees of freedom will make the maintainability of the FW harder, especially in terms of testing.

We are still thinking about merging this, however some discussion of potential impact on user experience and the printers themselves shall be done first.

bubmann:
> 
> 
> _... it's up to the user to decide._ ... that's a fair point, however our FW should be safe for general usage/users in the first place. Allowing too many degrees of freedom will make the maintainability of the FW harder, especially in terms of testing.
> 
> We are still thinking about merging this, however some discussion of potential impact on user experience and the printers themselves shall be done first.

maybe an "advanced/expert" menu can be fitted into the eeprom? just like in the slicer

leptun (approved):

Firmware/Marlin_main.cpp
```diff
@@ -8530,9 +8530,7 @@ SERIAL_PROTOCOLPGM("\n\n");
 			if(code_seen(axis_codes[i]))
 			{
 				uint16_t res_new = code_value();
-				bool res_valid = (res_new == 8) || (res_new == 16) || (res_new == 32); // resolutions valid for all axis
-				res_valid |= (i != E_AXIS) && ((res_new == 1) || (res_new == 2) || (res_new == 4)); // resolutions valid for X Y Z only
-				res_valid |= (i == E_AXIS) && ((res_new == 64) || (res_new == 128)); // resolutions valid for E only
+				bool res_valid = res_new > 0 && res_new <= 256 && !(res_new & (res_new - 1)); // must be a power of two
```
DRracer, line 8534: nice trick :+1: 
`!(res_new & (res_new - 1)); // must be a power of two`

DRracer (changes requested):
Please add some #define to enable additional ustep resolutions.
`#ifdef ALLOW_ALL_MRES` or something similar
Experienced users can then enable such a mode + recompile the FW while others stay on the safe side with the previous explicit list of supported mres.

DRracer:
> maybe an "advanced/expert" menu can be fitted into the eeprom? just like in the slicer

@bubmann I'm afraid adding a menu item + related infrastructure to enable runtime switching between the safe/advanced mres check is just way too heavy for our limited CPU space. It can be done, but we'd like to devote the remaining few KB of memory to more important features.

leptun:
Wow.
```
Sketch uses 229716 bytes (90%) of program storage space. Maximum is 253952 bytes.
Global variables use 6436 bytes of dynamic memory.
```
The ram usage has gone down significantly with the optimizations PRs. (756B of ram saved overall). Keep in mind that the base of this branch is from many releases ago. But still, the efficiency was grearly improved in the meantime.

leptun (approved):
Requested changes have been done. Approving again

DRracer (approved):
:+1:

PR #3006, opened 2021-01-29
Babystep insertion fixes

wavexx:
This PR prevents the user to enter Live Z (through long-press or via the menu) while performing mesh bed leveling or issuing individual Z probes.

Obviously these interfere with each other. We do this by setting/checking for the mesh_bed_leveling/homing flag when relevant.

We also cleanup some of the code around babystep insertion (notably, the alternative to BABYSTEP_LOADZ_BY_PLANNER is broken and has been removed).

PFW-1210

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -377,7 +377,7 @@ extern uint16_t gcode_in_progress;
 extern LongTimer safetyTimer;
 
 #define PRINT_PERCENT_DONE_INIT   0xff
-#define PRINTER_ACTIVE (IS_SD_PRINTING || is_usb_printing || isPrintPaused || (custom_message_type == CustomMsg::TempCal) || saved_printing || (lcd_commands_type == LcdCommands::Layer1Cal) || mmu_print_saved)
+#define PRINTER_ACTIVE (IS_SD_PRINTING || is_usb_printing || isPrintPaused || (custom_message_type == CustomMsg::TempCal) || saved_printing || (lcd_commands_type == LcdCommands::Layer1Cal) || mmu_print_saved || homing_flag || mesh_bed_leveling_flag)
 
 //! Beware - mcode_in_progress is set as soon as the command gets really processed,
 //! which is not the same as posting the M600 command into the command queue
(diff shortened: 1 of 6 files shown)


DRracer:
@wavexx what is the preferred procedure of testing this PR? It makes sense to get merged as it improves the safety of operation with the printer, however we must be sure it doesn't break anything else.

wavexx:
So while coming up with testing scenarios I was able to find even more unhandled situations which make the printer fail in various way.. I'll address these as well.

wavexx:
The fix turned out slightly different than anticipated, however it covers 3 related scenarios involving axis moves where there shouldn't be any.

To reproduce (and verify):

Scenario 1: Move Z can be triggered during homing/leveling/single Z probe, making it fail.
To reproduce:
- Reset
- Calibration -> First Layer Cal -> PLA
- Long-press encoder to enter Move Z
- Wait there until MBL starts
- Turn the encoder to make MBL fail

Scenario 2: The same with Live Z. This needs an SD or USB print to reproduce, but similarly:
```
G28 W
G4 S10 ; long-press encoder here and wait
G80 ; turn encoder while probing
```
MBL will not fail due to the small shifts, however the resulting mesh is completely wrong.

We fix these two by kicking the user out of Move/Live Z menu when homing/MBL is performed. Since the long-press function (and the other code paths) already check for it, it's not possible to enter again until MBL is fully complete.

This is better than just waiting (as incorrectly done previously), since the problem is that it's not guaranteed that the long-press function will be the same when the homing/mbl is over. Example: long press while preheating is usually "Move Z", but changes to "Live Z" as soon as printing starts. Previously this would allow to insert huge Z jumps while printing.

Scenario 3: Insert random moves during a print (several possible entry points).
To reproduce (easy):
- Calibration -> First Layer Cal -> PLA
- Enter Settings -> Move Axis
- Wait until MBL starts
- Start moving the encoder
Second:
- While recovering a print, during homing:
- Settings -> Disable steppers

We now disable these two entries using the PRINTER_ACTIVE check, which now also includes homing and MBL.
I removed one extra commit about the Tune/Preheat menu constantly flashing which was fixed in PR #3015

DRracer (approved):

PR #3009, opened 2021-02-01
"M27 P" support. Get current file absolute filename (file path)

leptun:
Refactored M27 as to be able to actually get `SD print paused` and `Print saved` as responses.
It works like this now:
`M27` without any arguments works just like before (same response format):
- When a file is being printed, the LFN (but without directory structure) is reported back alongside progress and time since start.
- When an SD print is paused, `SD print paused` is returned instead of `Not SD printing`
- When a USB print is paused, `Print saved` is returned.
- When USB printing or not printing at all, `Not SD printing` is returned.

`M27 P` works the same as without the `P` parameter, with the only exception that the the filename is SFN only and it also includes the full directory structure up to the printed file (file path).

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -5816,6 +5816,12 @@ if(eSoundMode!=e_SOUND_MODE_SILENT)
 
     /*!
 	### M27 - Get SD status <a href="https://reprap.org/wiki/G-code#M27:_Report_SD_print_status">M27: Report SD print status</a>
+    #### Usage
+	
+	      M27 [ P ]
+	
+	#### Parameters
+	  - `P` - Show full SFN path instead of LFN only.
     */
     case 27:
       card.getStatus(code_seen('P'));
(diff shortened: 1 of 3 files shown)


DRracer:
@leptun  Having M27 P (as path) will probably make more sense, otherwise it is ready to be merged.

Firmware/Marlin_main.cpp
```diff
@@ -5816,9 +5816,15 @@ if(eSoundMode!=e_SOUND_MODE_SILENT)
 
     /*!
 	### M27 - Get SD status <a href="https://reprap.org/wiki/G-code#M27:_Report_SD_print_status">M27: Report SD print status</a>
+    #### Usage
+	
+	      M27 [ P ]
+	
+	#### Parameters
+	  - `P` - Show full SFN path instead of LFN only.
```
3d-gussner, line 5824: Add @todo Update RepRap

DRracer (approved):

PR #3010, opened 2021-02-01
Fix M23

leptun:
Right now M23 is kinda broken. There are MANY issues with it, some more significant than others.
First issue that got fixed in this PR is proper string termination in diveSubfolder. This is what would happen before the fix:
```
M23 /oldgco~1/prusabox/printed/psucli~1.gco                                         
echo:Now fresh file: /oldgco~1/prusabox/printed/psucli~1.gco                        
oldgco~1�                                                                           
open failed, File: oldgco~1�.                                                       
File opened: /oldgco~1/prusaboxprinted/psucli~1.gco Size: 579005                    
File selected
```
```
echo:Now fresh file: /oldgco~1/prusabox/printed/psucli~1.gco
oldgco~1B
.pen failed, File: oldgco~1B
File opened: /oldgco~1/prusabox/printed/psucli~1.gco Size: 579005
File selected
```

~~Stuff still left to fix:~~ Stuff that was fixed as well in the second commit:
- Find out why `diveSubfolder` breaks after 1 folder depth (Hint: in `SdBaseFile::open(SdBaseFile*, const char*, uint8_t)`, `isOpen()` check returns failure). -> The problem was that the code tried to use the exact same file object recursively. We use chdir now instead which avoids this problem and also populates `workDirParents` and `dir_names` correctly.
- Find out why `diveSubfolder` is needed in the first place, since files open just fine even without it. -> We now keep the M23 path and the SD menu in sync. We need `diveSubfolder` in order to accomplish this.
- Fix `dir_names[3][9]` array overrun if you exceed 3 subdirs on the LCD SD menu: https://photos.app.goo.gl/UpmVjtBDn232NndC9 -> resized the array and made it stay true to the MAX_DIR_DEPTH macro
- Fix wrong filename displayed when starting a SD print from octoprint.
- Limit M20 max recursion level. Prevents static data overwritten.
- Prevent the SD menu from going over MAX_DIR_DEPTH.
- Moved some global variables to `CardReader`
- Made M23 and M30 safer by aborting the open/remove action in case of error in `diveSubfolder`
- M23 now prints the full path (/SFN/SFN/LFN.gcode) instead of just SFN.GCO
- M27 P also has /SFN/SFN/LFN.gcode format

### DIFF
--- a/Firmware/cardreader.cpp
+++ b/Firmware/cardreader.cpp
@@ -309,6 +309,18 @@ void CardReader::getAbsFilename(char *t)
   else
     t[0]=0;
 }
+
+void CardReader::printAbsFilenameFast()
+{
+    SERIAL_PROTOCOL('/');
+    for (uint8_t i = 0; i < getWorkDirDepth(); i++)
+    {
+        SERIAL_PROTOCOL(dir_names[i]);
+        SERIAL_PROTOCOL('/');
+    }
+    SERIAL_PROTOCOL(LONGEST_FILENAME);
+}
+
 /**
  * @brief Dive into subfolder
  *
@@ -425,16 +437,16 @@ void CardReader::openFile(const char* name,bool read, bool replace_current/*=tru
   {
     if (file.open(curDir, fname, O_READ)) 
     {
+      getfilename(0, fname);
       filesize = file.fileSize();
       SERIAL_PROTOCOLRPGM(_N("File opened: "));////MSG_SD_FILE_OPENED
-      SERIAL_PROTOCOL(fname);
+      printAbsFilenameFast();
       SERIAL_PROTOCOLRPGM(_n(" Size: "));////MSG_SD_SIZE
       SERIAL_PROTOCOLLN(filesize);
       sdpos = 0;
       
       SERIAL_PROTOCOLLNRPGM(MSG_FILE_SELECTED);
       lcd_setstatuspgm(MSG_FILE_SELECTED);
-      getfilename(0, fname);
     }
     else
     {
@@ -508,10 +520,8 @@ void CardReader::getStatus(bool arg_P)
     {
         if (arg_P)
         {
-            SERIAL_PROTOCOL('/');
-            for (uint8_t i = 0; i < getWorkDirDepth(); i++)
-                printf_P(PSTR("%s/"), dir_names[i]);
-            puts(filename);
+            printAbsFilenameFast();
+            SERIAL_PROTOCOLLN();
         }
         else
             SERIAL_PROTOCOLLN(LONGEST_FILENAME);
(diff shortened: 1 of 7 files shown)


Firmware/cardreader.cpp
```diff
@@ -340,9 +340,9 @@ void CardReader::diveSubfolder (const char *fileName, SdFile& dir)
             {
                 const size_t maxLen = 12;
                 char subdirname[maxLen+1];
-                subdirname[maxLen] = 0;
                 const size_t len = ((static_cast<size_t>(dirname_end-dirname_start))>maxLen) ? maxLen : (dirname_end-dirname_start);
                 strncpy(subdirname, dirname_start, len);
+                subdirname[len] = 0;
```
DRracer, line 367: makes sense - we need to terminate the copied dir name after the copy

Firmware/cardreader.cpp
```diff
@@ -69,12 +70,15 @@ char *createFilename(char *buffer,const dir_t &p) //buffer>12characters
 +*/
 
 void CardReader::lsDive(const char *prepend, SdFile parent, const char * const match/*=NULL*/) {
+	static uint8_t recursionCnt = 0;
```
DRracer, line 73: This will get initialized only once - the first time you enter this function.
I'm afraid there will be scenarios when this variable will be left in non-zero state thus breaking successive attempts to perform lsDive()

Moreover - please use a RAII pattern to make sure every return decrements the recursionCnt

Firmware/cardreader.cpp
```diff
@@ -317,19 +334,17 @@ void CardReader::getAbsFilename(char *t)
  * @param[in,out] fileName
  *  expects file name including path
  *  in case of absolute path, file name without path is returned
- * @param[in,out] dir SdFile object to operate with,
- *  in case of absolute path, curDir is modified to point to dir,
- *  so it is not possible to create on stack inside this function,
- *  as curDir would point to destroyed object.
  */
-void CardReader::diveSubfolder (const char *fileName, SdFile& dir)
+bool CardReader::diveSubfolder (const char *&fileName)
```
DRracer, line 345: combining a pointer and a reference is very hard to read/verify in the code, please use a pointer to pointer if you really need an in-out parameter

DRracer (changes requested):
@leptun please address my comments in the code - one of them is serious - the recursionCnt

Firmware/cardreader.cpp
leptun: Updated the PR to use a RAII principle for the incrementer. There shouldn't exist any case where recursionCnt is non-0 after the last return unless the stack got corrupted

Firmware/cardreader.cpp
leptun: This was the initial implementation of diveSubfolder according to the comments above, so I'd prefer if I kept it like this. It's 100% C++ complinat, but not C compliant. It's good that we don't care about that scenario in this case.

DRracer (approved):

PR #3011, opened 2021-02-01
Fix few translations introduced with #2982

3d-gussner:
I messed up some translation containing too many/few spaces and wrong upper/lower cases.
Found these by comparing the `not_used_*.txt` files before and after.

This PR fixes these.

Note to myself update https://github.com/3d-gussner/Prusa-Firmware/blob/MK3_Dutch/lang/translations.md to compare these files.


Firmware/messages.c
```diff
@@ -16,7 +16,7 @@ const char MSG_BED_DONE[] PROGMEM_I1 = ISTR("Bed done"); ////
 const char MSG_BED_HEATING[] PROGMEM_I1 = ISTR("Bed Heating"); ////
 const char MSG_BED_LEVELING_FAILED_POINT_LOW[] PROGMEM_I1 = ISTR("Bed leveling failed. Sensor didnt trigger. Debris on nozzle? Waiting for reset."); ////c=20 r=5
 const char MSG_BED_SKEW_OFFSET_DETECTION_FITTING_FAILED[] PROGMEM_I1 = ISTR("XYZ calibration failed. Please consult the manual."); ////c=20 r=8
-const char MSG_BELT_STATUS[] PROGMEM_I1 = ISTR("Belt Status");////c=18
+const char MSG_BELT_STATUS[] PROGMEM_I1 = ISTR("Belt status");////c=18
```
DRracer, line 19: why changing "S" -> "s"? We have "Bed Heating" and similar menu items

Firmware/ultralcd.cpp
```diff
@@ -8390,7 +8390,7 @@ static int lcd_selftest_screen(TestScreen screen, int _progress, int _progress_s
 	if (screen == TestScreen::EndStops) lcd_puts_P(_i("Checking endstops"));////MSG_SELFTEST_CHECK_ENDSTOPS c=20
 	if (screen == TestScreen::AxisX) lcd_puts_P(_T(MSG_CHECKING_X));
 	if (screen == TestScreen::AxisY) lcd_puts_P(_T(MSG_CHECKING_Y));
-	if (screen == TestScreen::AxisZ) lcd_puts_P(_i("Checking Z axis  "));////MSG_SELFTEST_CHECK_Z c=20
+	if (screen == TestScreen::AxisZ) lcd_puts_P(_i("Checking Z axis"));////MSG_SELFTEST_CHECK_Z c=20
```
DRracer, line 8393: did you double-check, that this change doesn't leave any previous characters "under" the 2 last spaces?

DRracer (changes requested):

Firmware/messages.c
3d-gussner: Because it was before Belt status with lower case "s". When I added the `MSG_BELT_STATUS` to `messages.cpp/.h` I wrote it like I am used to it with upper "S", but the lang files had still "Belt status" with lower "s" and so have not been translated.

See https://github.com/prusa3d/Prusa-Firmware/commit/377a59c624ba332a82e72c5497d08cbb6776f25a

I am fine to change it to upper case "S" if you like that more.

Firmware/ultralcd.cpp
3d-gussner: Just double checked it during Selftest and it works without any issues at least in English.
Will check if any other longer translation may create an issue here.

3d-gussner: Forced MSG_CHECKING_Y to `axe` and the previous message has been overwritten correctly

Firmware/messages.c
DRracer: @3d-gussner you are right, lower case "s" fits the Support menu better, other items also have lower case starting letters of second words.

DRracer (approved):

PR #3012, opened 2021-02-02
M552 - Printer IP address

leptun:
You can set the printer IP address with M552 P<ip_addr>. Eg:
`M552 P192.168.1.14`
It will be visible in the support menu:
image

PFW-1173

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -2057,8 +2057,8 @@ static void lcd_support_menu()
 	{	// 22bytes total
 		int8_t status;                 // 1byte
 		bool is_flash_air;             // 1byte
-		uint8_t ip[4];                 // 4bytes
-		char ip_str[3*4+3+1];          // 16bytes
+		uint32_t ip;                   // 4bytes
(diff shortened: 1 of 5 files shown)


Firmware/ultralcd.cpp
```diff
@@ -2057,8 +2057,8 @@ static void lcd_support_menu()
 	{	// 22bytes total
 		int8_t status;                 // 1byte
 		bool is_flash_air;             // 1byte
-		uint8_t ip[4];                 // 4bytes
-		char ip_str[3*4+3+1];          // 16bytes
+		uint32_t ip;                   // 4bytes
+		char ip_str[IP4_STR_SIZE];     // 16bytes
```
DRracer, line 2061: do you really need to store the formatted string? Is printing 4 numbers so slow?

leptun: No. We store stuff in memory since that's how the stupid LCD function works. I could probably avoid using this, but it's menu_data so it's not taking any extra space

DRracer (approved):

PR #3013, opened 2021-02-02
DIR depth adjustments

leptun:
Changed MAX_DIR_DEPTH from 10 to 6. It makes sense to reduce it a bit due to limitations in the cmdqueue.
fixed the dir_names array definition to stay true to the macro.


DRracer (approved):
this is a conservative but correct solution, approving

PR #3014, opened 2021-02-02
Ignore button long-press event, if not allowed

Panayiotis-git:
The button of the printer can register two kind of events:
- the click, which is triggered when the button is released after a short press
- the long-press, which at certain screens is triggered when the button is pressed continuously for some time.

At the screens where the long-press event is not allowed, currently the code executes the menu_back() method. This seems to breaks the functionality of some screens. For example, if the button is long-pressed during the preheating phase of the Load filament menu (which is a modal screen), then the user is returned to the previous menu and is not any more able to Cancel the preheat, not even by selecting again the Load filament menu.

This PR, simply ignores the long-press event where it is not allowed. This is a more natural behavior, since the button always triggers a click event (independently of the keypress duration), except for the screens where we want to trigger the long keypress functionality.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -8048,7 +8048,7 @@ void menu_lcd_longpress_func(void)
     // explicitely listed menus which are allowed to rise the move-z or live-adj-z functions
     // The lists are not the same for both functions, so first decide which function is to be performed
     if ( (moves_planned() || IS_SD_PRINTING || usb_timer.running() )){ // long press as live-adj-z
-        if(( current_position[Z_AXIS] < Z_HEIGHT_HIDE_LIVE_ADJUST_MENU ) // only allow live-adj-z up to 2mm of print height
+        if (( current_position[Z_AXIS] < Z_HEIGHT_HIDE_LIVE_ADJUST_MENU ) // only allow live-adj-z up to 2mm of print height
         && ( menu_menu == lcd_status_screen // and in listed menus...
           || menu_menu == lcd_main_menu
           || menu_menu == lcd_tune_menu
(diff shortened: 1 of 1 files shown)


wavexx:
I didn't test it yet, but I approve the idea. The preheating issue is that there's no real/proper "modal" menu handling while waiting for preheat.

I'd actually keep the "lcd_quick_feedback" call, and change the previous menu_back() to quick_feedback instead. It gives the user some feedback the that the longpress was detected, but was ignored on purpose.

DRracer:
@Panayiotis-git fixing functionality by removing code - I like that :)

Panayiotis-git:
The button long-press event is handled in several different ways depending on the current screen:
1. go to Move Z axis
2. go to Live adjust
3. enable a hidden functionality
4. return to previous menu (sometimes it could be interpreted as Cancel choice, while selecting a numeric value)
5. remain at the current screen (event is skipped)
6. not triggered at all and behave as a normal click on button release (when a multiple choice screen is implemented)

While navigating at the menus and functions of the printer, I usually find my self unable to determine what a long-press would result to. Even the Move Z axis function is triggered at some places (the Setting menu for example), but not at other places without an obvious reason (e.g. the Calibration menu).

This is my initial try to solve the bug at the loading filament and make the handling less confusing. Obviously by cutting functionality... ;)

Anyway, I have posted a fix for the loading filament preheat bug at the PR https://github.com/prusa3d/Prusa-Firmware/pull/2335, because it could not stand as a separate PR.

wavexx:
I tried this and I think it's working fine, but I'd keep the lcd_quick_feedback() to indicate that it has been seen. Without it I have the impression the longpress didn't register.

Panayiotis-git:
I've added the lcd_quick_feedback() to notify the user that a longpress was registered but was blocked, if it was not allowed.

wavexx:
What's your personal take on this? (do you think the feedback is useful or not?).
Since this PR  I tried to spam the longpress a bit everywhere. It's definitely more broken than not.

Panayiotis-git:
I find the feedback really useful. I would prefer though a different pitch sound to differentiate the case when the event is skipped from the case when the event was handled by the printer (cases 1...3 from the list at the above post).

wavexx:
@leptun I would consider this for the 3.12

Looks like there's currently a trivial merge conflict due to some whitespace changes.

leptun:
@wavexx I agree, this can go into 3.12. It is already tagged. Still need to see how it works on the printer before approving, but it looks fine to me

leptun (approved):
Seems to work just fine. Menus just get redrawn when the long-press event happens

PR #3016, opened 2021-02-03
Fix additional gcc 7.3 warnings

wavexx:
This is a no-code-change PR to silence more warnings caught by gcc 7.3.
They're divided by category through each commit.


wavexx:
Remaining warnings are cleared through https://github.com/prusa3d/Arduino_Boards/pull/9

I consider suppressing the portability warnings superior than changing several existing (and correct) defines, since we won't be switching to a different compiler.

Firmware/Marlin_main.cpp
```diff
@@ -3133,11 +3136,13 @@ bool gcode_M45(bool onlyZ, int8_t verbosity_level)
 				bool result = sample_mesh_and_store_reference();
 				if (result)
 				{
-					if (calibration_status() == CALIBRATION_STATUS_Z_CALIBRATION)
-						// Shipped, the nozzle height has been set already. The user can start printing now.
-						calibration_status_store(CALIBRATION_STATUS_CALIBRATED);
-						final_result = true;
-					// babystep_apply();
+                    if (calibration_status() == CALIBRATION_STATUS_Z_CALIBRATION)
+                    {
```
DRracer, line 3504: And which one seems to be correct? With final_result=true inside the block or not? 
Not considering the previous misleadingly indented state of code...

thess: `final_result` needs to be set outside the block as done by this patch. Otherwise, this function would return _false_ when an _onlyZ_ home is successfully executed.

DRracer (approved):

PR #3026, opened 2021-02-10
Remove incorrect const declarations on filename strings.

thess:
Filename strings are not const. 
Compiler warnings are generated when assigning a const to a non-const pointer - as in:
```
Firmware/ultralcd.cpp: In function 'void lcd_implementation_drawmenu_sdfile_selected(uint8_t, const char*, char*)':
Firmware/ultralcd.cpp:343:24: warning: invalid conversion from 'const char*' to 'char*' [-fpermissive]

          longFilename = filename;                                                                   
                         ^~~~~~~~ 
```

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -275,8 +275,8 @@ static void lcd_delta_calibrate_menu();
 
 
 /* Different types of actions that can be used in menu items. */
-static void menu_action_sdfile(const char* filename);
-static void menu_action_sddirectory(const char* filename);
+static void menu_action_sdfile(char* filename);
+static void menu_action_sddirectory(char* filename);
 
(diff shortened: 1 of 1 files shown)


wavexx:
I had a similar one (see #3003) which I closed in favor of #2405 which is more comprehensive and also seems to be already tagged for the next release

thess:
I merged #2405 into a clean MK3 branch and it still has a couple of const conversion warnings when built with 7.3.0. Filename strings being passed around ultralcd sdfile functions are NOT const.

wavexx:
When I last tried it was ok, but I'd suggest to comment on #2405 directly

thess:
Closing and deferring to #2405 (nice work BTW). Hopefully when 3.10 is cut the devs will be using current tools and cleanup warnings :wink:

PR #3029, opened 2021-02-13
Add support to build a specific language

jbrunner:
Allows to build a single language only using `./build.sh en`


wavexx:
FIY we overhauled completely the language infrastructure.

With the new cmake build you can select a language during configuration: ``cmake -DMAIN_LANGUAGES=[code] -DCOMMUNITY_LANGUAGES=``. For the common english-only build, you can configure as usual and just "make ALL_ENGLISH" for the english-only build that has no multi-language support.

The old build system still allows to configure the build-in languages by exporting the ``LANGUAGES`` and ``COMMUNITY_LANGUAGES`` environment variables prior to build. See ``lang/config.sh``.

Do you think this is still necessary?

Issue #2954, opened 2021-01-08
[BUG]Flashing 3.9.3 to MK3S fails


hlipka:
I have a MK3S (bought in September, no MMU), currently at 3.9.2.
Today I tried to flash the new 3.9.3 firmware, but it fails at the end of upload. For the avrdude log see below. When this happens, the mainboard seems to be stuck in a boot loop - I looks as if the LCD is refreshed several times in a row with thew boot message, and then there is a short pause (like a reset / reboot), then it starts again.
I also tried 3.9.3RC1, which also fails. Instead of the constant LCD refresh I just see an empty LCD.
Flashing was done via first via Octoprint (using the Firmware updater). When this failed, I used PrusaSlicer connected directly to go back to 3.9.2 (which worked), the used PrusaSlicer again to try 3.9.3 and 3.9.3RC1.
(I did the update to 3.9.2 successfully before so FW updates as such work).
So it seems I'm stuck with 3.9.2 (falling back to it worked fine both times) and cannot upgrade to 3.9.3. Any other things I can try?

avrdude output (from PrusaSlicer):
```
avrdude-slic3r -v -p atmega2560 -c wiring -P /dev/ttyACM0 -b 115200 -D -U flash:w:0:/home/hli/Downloads/prusa3d_fw_MK3S_MK3S+_3_9_3_3556.hex:i

avrdude-slic3r: Version 6.3-20160220-prusa3d, compiled on Dec 22 2020 at 16:48:28
                Copyright (c) 2000-2005 Brian Dean, http://www.bdmicro.com/
                Copyright (c) 2007-2014 Joerg Wunsch


                Using Port                    : /dev/ttyACM0
                Using Programmer              : wiring
                Overriding Baud Rate          : 115200
                AVR Part                      : ATmega2560
                Chip Erase delay              : 9000 us
                PAGEL                         : PD7
                BS2                           : PA0
                RESET disposition             : dedicated
                RETRY pulse                   : SCK
                serial program mode           : yes
                parallel program mode         : yes
                Timeout                       : 200
                StabDelay                     : 100
                CmdexeDelay                   : 25
                SyncLoops                     : 32
                ByteDelay                     : 0
                PollIndex                     : 3
                PollValue                     : 0x53
                Memory Detail                 :

                                         Block Poll               Page                       Polled
                  Memory Type Mode Delay Size  Indx Paged  Size   Size #Pages MinW  MaxW   ReadBack
[...]

PR #3033, opened 2021-02-17
Make watchdogReset() force_inline

DRracer:
It makes no sense keeping watchdogReset as a separate function which must be called, since it only contains one instruction: `wdr`.

Not only was the code larger by 32 bytes in total, but also much slower (call+ret take 4 cycles together for no reason in this case).

Surprisingly, doing just this on FW 3.9.3 solves issue #2954 on the one affected EINSY board, even though it makes not much sense (there must be some other timing issue on that one board itself).

### DIFF
--- a/Firmware/optiboot_w25x20cl.cpp
+++ b/Firmware/optiboot_w25x20cl.cpp
@@ -7,6 +7,7 @@
 #include "w25x20cl.h"
 #include "stk500.h"
 #include "bootapp.h"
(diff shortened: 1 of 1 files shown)


leptun (approved):
For the near future: replace this reimplementation of WDT control with the avr-libc one (`<avr/wdt.h>`)

XPila (approved):

PR #3034, opened 2021-02-18
PFW-960 Improve M0/M1/M117

3d-gussner:
# `M0/M1` are **NOT** pause gcode commands 

### Gcode `M1` must have a string while `M0` it is optional
As the `M0/M1` moved to the beginning of the parser
 - parser would not be able to "find" `M1nn` command if the query was `M1` instead of `M1 ` (a space behind)
 - to be able to "stop/halt" without sending a string and display default message use gcode `M0`
 - as there are no `M0nn` gcodes the parser can query `M0` without additional space needed as in `M1 `
 - `M0/M1` with parameters `P or S` will **NOT** show any text as mentioned in https://github.com/prusa3d/Prusa-Firmware/issues/2055#issuecomment-713090311

### In Octoprint `M0, M1` should not be in "Pausing commands" as these are **NOT** pause commands, they stop/halt/sleep the printer.

Addressing few requests from https://github.com/prusa3d/Prusa-Firmware/issues/2055

Thanks to @vojtech-pavlik and his PR #1900 which I merged to newer version and improved one or two things.


3d-gussner:
Replaces PR #1900

DRracer (approved):
After fixing some indentation this looks correct to me. Approving.

Issue #3037, opened 2021-02-19
[BUG] M73 does not show any remaining time


fuso2001:
**Printer type** - MK3S+
**Printer firmware version**- 3.9.3 commit 3556

**MMU Upgrade** - MMU2S

**SD card or USB/Octoprint**
  Print via USB/Octoprint

**Describe the bug**
  M73 Commando only shows the completed percentages in the display. The remaining time display remains at 0. The messages via the USB interface towards the host are OK and contain the new "C" object. Time to change is a "-1" value.
  
**To Reproduce**
  Start a print and watch the display.
  
**G-code**
  G-Code was generated with PrusaSlicer and contains M73 commands

3d-gussner:
@fuso2001 Thanks for reporting. Please add next time that you compiled the firmware in the **Printer firmware version** part.

fuso2001:
> **Printer type** - MK3S+
> **Printer firmware version**- 3.9.3 commit 3556
> 
> **MMU Upgrade** - MMU2S
> 
> **SD card or USB/Octoprint**
> Print via USB/Octoprint
> 
> **Describe the bug**
> M73 Commando only shows the completed percentages in the display. The remaining time display remains at 0. The messages via the USB interface towards the host are OK and contain the new "C" object. Time to change is a "-1" value.
> 
> **To Reproduce**
> Start a print and watch the display.
> 
> **G-code**
> G-Code was generated with PrusaSlicer and contains M73 commands

> @fuso2001 Thanks for reporting. Please add next time that you compiled the firmware in the **Printer firmware version** part.

> Okay, thank you for the answer. I now set starttime to _millis () in the M862.3 command. It works now.

3d-gussner:
Should be fixed with https://github.com/prusa3d/Prusa-Firmware/pull/3067

JohnnyDeer:
I'm closing this issue as solved by mentioned #3067.

PR #3039, opened 2021-02-20
Fix issue #3037

3d-gussner:
`starttime` is only set for SD prints via gcode `M24 and M32`


3d-gussner:
Better solution is to have also `starttime` for USB prints. Other wise without `M73` the printer uptime is shown and not the print time during USB prints.

3d-gussner:
Moved "Better solution" to new branch. Let's fix this issue first.

DRracer (approved):

PR #3046, opened 2021-02-28
Fix PR3032

3d-gussner:
With PR#3032 the Welcome message on LCD got broken.
Also updated the EEPROM documentation adding the `Service prep`

Before this fix
Before_MK3_Fix_PR3032
and here with this fix
After_MK3_Fix_PR3032


DRracer (approved):
thanks @3d-gussner

PR #3048, opened 2021-03-03
Fix flashing languages with inline wdr instructions

DRracer:
A fairly mysterious situation happened recently in the MK3 branch.
After merging PR #3033 (change watchdogReset() into a single inline wdr instruction)
we were unable to flash languages.

Since it looked similarly suspicious like issue #2954 we started investigating deeply.
The problem was in the code as described in the comment in this PR.

### DIFF
--- a/Firmware/optiboot_w25x20cl.cpp
+++ b/Firmware/optiboot_w25x20cl.cpp
@@ -126,6 +126,7 @@ uint8_t optiboot_w25x20cl_enter()
     }
     selectedSerialPort = 0; //switch to Serial0
     MYSERIAL.flush(); //clear RX buffer
+    int SerialHead = rx_buffer.head;
     // Send the initial magic string.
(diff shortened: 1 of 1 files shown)


dragomirecky (approved):
Nice 👍

I think we should consider making the global variable `rx_buffer` volatile instead of leaving the possibility that something like this will happen again somewhere else in the codebase.

DRracer:
@dragomirecky good point, we'll look into this deeper in the next FW. Two things:
- there is space for optimization of rx_buffer
- rx_buffer is used throughout the whole FW so making this instance volatile completely may break things we haven't touched in this release - therefore my fix is just a local one.

XPila (approved):

DRracer:
@leptun thanks for the SerialHead fix, I shouldn't have optimized that hard after having a working code

PR #3052, opened 2021-03-04
Echo "paused for user" if print is paused in usb printing mode

zeroflow:
Octoprint times out on M601.
This patch at least sends "paused for user" if the print is paused.


zeroflow:
Redundant in 3.10.0-RC1, as now M601 works well without this patch

PR #3055, opened 2021-03-07
Add satin sheets

### LINKED ISSUE #3051: [ENHANCEMENT] add default sheets for Satin
**Printer type** - MK3S+
**MMU Upgrade** - N/A

  There is no steel sheet named after Satin sheets so you have to use Custom sheets
  
Add a few slots for Satin1, Satin2, et cetera.

awenelo:
This changes the `Custom1` and `Custom2` sheets to be `Satin 1` and `Satin 2`. Additionally, as there are now only the numbers `1` and `2` at the end of sheets, this replaces the switch that was used to determine sheet numbers, now using modulo to assign the sheet a `1` or `2`. I believe that I have updated the table in `eeprom.h` correctly, if I haven't, let me know. 

To have the sheet names change, the sheet must be reset.

### New:
Index | Name
---|---
0 | Smooth1
1 | Smooth2
2 | Textur1
3 | Textur2
4 | Satin 1
5 | Satin 2
6 | Custom1
7 | Custom2

### Old:
Index | Name
---|---
0 | Smooth1
1 | Smooth2
2 | Textur1
3 | Textur2
4 | Custom1
5 | Custom2
6 | Custom3
7 | Custom4

Picture of new names

Resolves #3051 

PFW-1240

### DIFF
--- a/Firmware/eeprom.h
+++ b/Firmware/eeprom.h
@@ -276,19 +276,19 @@ static_assert(sizeof(Sheets) == EEPROM_SHEETS_SIZEOF, "Sizeof(Sheets) is not EEP
 | 0x0D71 3441		| uint16	| ^										| 00 00h 0		| ff ffh 65535			| 4th sheet - Z offset 								| ^				| D3 Ax0d71 C2	
 | 0x0D73 3443		| uint8		| ^										| 00h 0			| ffh 255				| 4th sheet - bed temp 								| ^				| D3 Ax0d73 C1	
 | 0x0D74 3444		| uint8		| ^										| 00h 0			| ffh 255				| 4th sheet - PINDA temp 							| ^				| D3 Ax0d74 C1	
-| 0x0D75 3445		| char		| _5th Sheet block_						| 437573746f6d31| ffffffffffffff		| 5th sheet - Name: 	_Custom1_					| ^				| D3 Ax0d75 C7
+| 0x0D75 3445		| char		| _5th Sheet block_						| 437573746f6d31| ffffffffffffff		| 5th sheet - Name: 	_Satin 1_					| ^				| D3 Ax0d75 C7
 | 0x0D7C 3452		| uint16	| ^										| 00 00h 0		| ff ffh 65535			| 5th sheet - Z offset 								| ^				| D3 Ax0d7c C2	
 | 0x0D7E 3454		| uint8		| ^										| 00h 0			| ffh 255				| 5th sheet - bed temp 								| ^				| D3 Ax0d7e C1	
 | 0x0D7F 3455		| uint8		| ^										| 00h 0			| ffh 255				| 5th sheet - PINDA temp 							| ^				| D3 Ax0d7f C1	
-| 0x0D80 3456		| char		| _6th Sheet block_						| 437573746f6d32| ffffffffffffff		| 6th sheet - Name: 	_Custom2_					| ^				| D3 Ax0d80 C7
+| 0x0D80 3456		| char		| _6th Sheet block_						| 437573746f6d32| ffffffffffffff		| 6th sheet - Name: 	_Satin 2_					| ^				| D3 Ax0d80 C7
 | 0x0D87 3463		| uint16	| ^										| 00 00h 0		| ff ffh 65535			| 6th sheet - Z offset 								| ^				| D3 Ax0d87 C2	
 | 0x0D89 3465		| uint8		| ^										| 00h 0			| ffh 255				| 6th sheet - bed temp 								| ^				| D3 Ax0d89 C1	
 | 0x0D8A 3466		| uint8		| ^										| 00h 0			| ffh 255				| 6th sheet - PINDA temp 							| ^				| D3 Ax0d8a C1	
-| 0x0D8B 3467		| char		| _7th Sheet block_						| 437573746f6d33| ffffffffffffff		| 7th sheet - Name: 	_Custom3_					| ^				| D3 Ax0d8b C7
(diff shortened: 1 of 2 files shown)


Firmware/eeprom.h
```diff
@@ -276,19 +276,19 @@ static_assert(sizeof(Sheets) == EEPROM_SHEETS_SIZEOF, "Sizeof(Sheets) is not EEP
...
 | 0x0D7E 3454		| uint8		| ^										| 00h 0			| ffh 255				| 5th sheet - bed temp 								| ^				| D3 Ax0d7e C1	
 | 0x0D7F 3455		| uint8		| ^										| 00h 0			| ffh 255				| 5th sheet - PINDA temp 							| ^				| D3 Ax0d7f C1	
-| 0x0D80 3456		| char		| _6th Sheet block_						| 437573746f6d32| ffffffffffffff		| 6th sheet - Name: 	_Custom2_					| ^				| D3 Ax0d80 C7
+| 0x0D80 3456		| char		| _6th Sheet block_						| 536174696e2032| ffffffffffffff		| 6th sheet - Name: 	_Satin 2_					| ^				| D3 Ax0d80 C7
 | 0x0D87 3463		| uint16	| ^										| 00 00h 0		| ff ffh 65535			| 6th sheet - Z offset 								| ^				| D3 Ax0d87 C2	
 | 0x0D89 3465		| uint8		| ^										| 00h 0			| ffh 255				| 6th sheet - bed temp 								| ^				| D3 Ax0d89 C1	
 | 0x0D8A 3466		| uint8		| ^										| 00h 0			| ffh 255				| 6th sheet - PINDA temp 							| ^				| D3 Ax0d8a C1	
-| 0x0D8B 3467		| char		| _7th Sheet block_						| 437573746f6d33| ffffffffffffff		| 7th sheet - Name: 	_Custom3_					| ^				| D3 Ax0d8b C7
+| 0x0D8B 3467		| char		| _7th Sheet block_						| 437573746f6d33| ffffffffffffff		| 7th sheet - Name: 	_Custom1_					| ^				| D3 Ax0d8b C7
```
3d-gussner: Custom1 hex value should end with `31` for `1`

Firmware/eeprom.h
```diff
@@ -276,19 +276,19 @@ static_assert(sizeof(Sheets) == EEPROM_SHEETS_SIZEOF, "Sizeof(Sheets) is not EEP
...
 | 0x0D92 3474		| uint16	| ^										| 00 00h 0		| ff ffh 65535			| 7th sheet - Z offset 								| ^				| D3 Ax0d92 C2	
 | 0x0D94 3476		| uint8		| ^										| 00h 0			| ffh 255				| 7th sheet - bed temp 								| ^				| D3 Ax0d94 C1	
 | 0x0D95 3477		| uint8		| ^										| 00h 0			| ffh 255				| 7th sheet - PINDA temp 							| ^				| D3 Ax0d95 C1	
-| 0x0D96 3478		| char		| _8th Sheet block_						| 437573746f6d34| ffffffffffffff		| 8th sheet - Name: 	_Custom4_					| ^				| D3 Ax0d96 C7
+| 0x0D96 3478		| char		| _8th Sheet block_						| 437573746f6d34| ffffffffffffff		| 8th sheet - Name: 	_Custom2_					| ^				| D3 Ax0d96 C7
```
3d-gussner: Custom2 hex value should end with `32` for `2`

3d-gussner (commented):
Small change needed in the eeprom.h documentation

3d-gussner:
@awenelo Did you run the doxygen to check that the documentation of the EEPROM Table is still correct?

Firmware/eeprom.h
awenelo: Fixed!

Firmware/eeprom.h
awenelo: Fixed!

awenelo:
Yep, it looks to still be correct.

doxygen_eeprom_sheets

DRracer (approved):

DRracer:
@awenelo very nice :+1: 
There is only one pitfall to this solution - the names are only generated once (which is correct - we don't want to overwrite the names at every start) so for a user to get "new" names at indices 4 and 5 one has to reset the sheet profile at that index. This may be hard to explain to ordinary users

awenelo:
There are two ways that I can think to solve that, both of which have problems:
1. Check the z offset of all sheets on boot (instead of the sheet name) and overwrite the sheet names if the z offset is the default value. The two problems I see with this are:
    <ol><li>If the user has changed the z offset, then changed it back to 0 (as if they weren't using the sheet anymore, but didn't re-setting the sheet), the name wouldn't be updated. </li><li>If they had re-named a sheet and were planning on using it in the future (hadn't changed the Z offset), the name would be re-set.</li></ol>

2. On boot, compare the names of sheets 4-7 and see if any of them match the old names (`custom1-4`), and replace the names if they also have the default z offset. This solves problem 2 in way number 1, but takes up more space, as we're now still storing all the old names.

The best choice would probably be way number 1, as that would solve the problem for the majority of the people. However, I don't know if either one would offset the number of new issues that people would file because they didn't re-set a sheet, or re-named a sheet and lost the re-named sheet.

awenelo:
I have a working solution to way number 1. I feel like the number of people who would be affected by either of the problems is small enough that it's negligible.

I couldn't find a way to do way number 2 without eating up a lot more space (at least 220 bytes more than way 1, without even checking z offsets)

Should I merge way 1 into this branch, or create a second PR?

DRracer:
@awenelo we are still considering the ways how to change the names seamlessly even for existing printers - you wrote some of the options.
For new printers (unitialized EEPROM) your current solution would definitely work correctly

DRracer:
@awenelo good news, our dev team finally got an approval, it's been an interesting internal discussion "what to to if....". Merging as it is right now with no additional rename-if code. Thank you for your PR.

PR #3067, opened 2021-03-17
Attempt to workaround the M73 C0|D0 visual issue

DRracer:
This is an attempt to enable alternation of time to print finish and time to color change even in the last minute of time to color change, i.e. be able to print "0:00C".

The proposed solution leverages the capability of the current FW to read float values from the C|D parameter. This could have the raw benefit of being able still to alternate this time on the LCD as "0:00C" (or "<1min") if the slicer sends a non-zero but <1 time right before the color change.

PFW-1219

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -6389,7 +6389,6 @@ Sigma_Exit:
         if(code_seen('R')) print_time_remaining_normal = code_value();
         if(code_seen('Q')) print_percent_done_silent = code_value();
         if(code_seen('S')) print_time_remaining_silent = code_value();
-#if 1
         if(code_seen('C')){
             float print_time_to_change_normal_f = code_value_float();
             print_time_to_change_normal = ( print_time_to_change_normal_f <= 0 ) ? PRINT_TIME_REMAINING_INIT : print_time_to_change_normal_f;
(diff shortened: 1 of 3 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -6389,14 +6389,24 @@ SERIAL_PROTOCOLPGM("\n\n");
...
         if(code_seen('D')) print_time_to_change_silent = code_value();
-
-    {
-        const char* _msg_mode_done_remain = _N("%S MODE: Percent done: %d; print time remaining in mins: %d; Change in mins: %d\n");
-        printf_P(_msg_mode_done_remain, _N("NORMAL"), int(print_percent_done_normal), print_time_remaining_normal, print_time_to_change_normal);
-        printf_P(_msg_mode_done_remain, _N("SILENT"), int(print_percent_done_silent), print_time_remaining_silent, print_time_to_change_silent);
-    }
+#endif
+        {
+            const char* _msg_mode_done_remain = _N("%S MODE: Percent done: %d; print time remaining in mins: %d; Change in mins: %d\n");
```
3d-gussner: `...Change in mins: %d\n"` should be `Change in mins: %.1f\n"` otherwise we get negative output.

Firmware/Marlin_main.cpp
```diff
@@ -6389,14 +6389,24 @@ SERIAL_PROTOCOLPGM("\n\n");
         if(code_seen('R')) print_time_remaining_normal = code_value();
         if(code_seen('Q')) print_percent_done_silent = code_value();
         if(code_seen('S')) print_time_remaining_silent = code_value();
+#if 1
+        if(code_seen('C')){
+            float print_time_to_change_normal_f = code_value_float();
```
3d-gussner, line 6393: Also
https://github.com/prusa3d/Prusa-Firmware/blob/MK3_3.10.0/Firmware/Marlin_main.cpp#L319-L320
and
https://github.com/prusa3d/Prusa-Firmware/blob/MK3_3.10.0/Firmware/Marlin.h#L369-L370
should be float

3d-gussner (commented):
Fix serial output and external use

Firmware/Marlin_main.cpp
DRracer: no, that's what I want to avoid

Firmware/Marlin_main.cpp
DRracer: we can handle the serial output separately

3d-gussner (commented):
Tested with MK404 and works as expected.

XPila (approved):

PR #3071, opened 2021-03-18
Reprint function implemented for MK3S/MK3S+

jfestrada:
We have created a new menu option that is displayed only when the previous print ends successfully. 
To support this functionality we have added a **extern bool variable** which can be read/set from **cmdqueue.cpp** and **ultralcd.cpp** so when the card reader reaches the EOF the variable isPrintFinished will be set to true and the main menu will show the **REPRINT** option. 
How do we know which file we need to reprint? We have read the code and we saw that all the information needed to perform the reprint was stored on the EEPROM in case of a power panic and we recover this information in order to start the print from SD again, this feature has been implemented on the function reprint_from_eeprom().

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -79,7 +79,7 @@ static uint8_t lcd_commands_step = 0;
 CustomMsg custom_message_type = CustomMsg::Status;
 unsigned int custom_message_state = 0;
 
-
+bool isPrintFinished = false;
 bool isPrintPaused = false;
 uint8_t farm_mode = 0;
 int farm_timer = 8;
@@ -6559,6 +6559,11 @@ static void lcd_main_menu()
     MENU_ITEM_FUNCTION_P(PSTR("power panic"), uvlo_);
 #endif //TMC2130_DEBUG
 
+	if ( (!PRINTER_ACTIVE) && isPrintFinished)
+  	{
+		MENU_ITEM_SUBMENU_P(_i("Reprint"), reprint_from_eeprom);
+  	}
+
     if ( ( IS_SD_PRINTING || is_usb_printing || (lcd_commands_type == LcdCommands::Layer1Cal)) && (current_position[Z_AXIS] < Z_HEIGHT_HIDE_LIVE_ADJUST_MENU) && !homing_flag && !mesh_bed_leveling_flag) {
         MENU_ITEM_SUBMENU_P(_T(MSG_BABYSTEP_Z), lcd_babystep_z);//8
     }
@@ -8957,6 +8962,43 @@ void lcd_experimental_menu()
 #endif //EXTRUDER_ALTFAN_DETECT
 
     MENU_END();
+}
+
+void reprint_from_eeprom() {
+	char cmd[30];
+	char filename[13];
+	uint8_t depth = 0;
+	char dir_name[9];
+
+	isPrintFinished=false;
+
+	//cmdqueue_reset();
+
+	depth = eeprom_read_byte((uint8_t*)EEPROM_DIR_DEPTH);
+	
+	MYSERIAL.println(int(depth));
+	for (int i = 0; i < depth; i++) {
+		for (int j = 0; j < 8; j++) {
+			dir_name[j] = eeprom_read_byte((uint8_t*)EEPROM_DIRS + j + 8 * i);
+		}
+		dir_name[8] = '\0';
+		MYSERIAL.println(dir_name);
+		// strcpy(dir_names[i], dir_name);
+		card.chdir(dir_name, false);
+	}
+
+	for (int i = 0; i < 8; i++) {
+		filename[i] = eeprom_read_byte((uint8_t*)EEPROM_FILENAME + i);
+	}
+	filename[8] = '\0';
+
+	MYSERIAL.print(filename);
+	strcat_P(filename, PSTR(".gco"));
+	sprintf_P(cmd, PSTR("M23 %s"), filename);
+	enquecommand(cmd);
+  	sprintf_P(cmd, PSTR("M24"));
+	enquecommand(cmd);
+	lcd_return_to_status();
 }
 
 #ifdef PINDA_TEMP_COMP
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -6559,10 +6559,13 @@ static void lcd_main_menu()
     MENU_ITEM_FUNCTION_P(PSTR("power panic"), uvlo_);
 #endif //TMC2130_DEBUG
 
-	if ( (!PRINTER_ACTIVE) && isPrintFinished)
+	if ( (!PRINTER_ACTIVE) && isPrintFinished && card.cardOK)
   	{
 		MENU_ITEM_SUBMENU_P(_i("Reprint"), reprint_from_eeprom);
-  	}
+  	}else if (!card.cardOK)
+	  {	  //If the user remove the SD card the reprint will be disabled because you can't be sure that the gcode file will remain in the SD
+		  isPrintFinished = false; 
+	  }
 
     if ( ( IS_SD_PRINTING || is_usb_printing || (lcd_commands_type == LcdCommands::Layer1Cal)) && (current_position[Z_AXIS] < Z_HEIGHT_HIDE_LIVE_ADJUST_MENU) && !homing_flag && !mesh_bed_leveling_flag) {
         MENU_ITEM_SUBMENU_P(_T(MSG_BABYSTEP_Z), lcd_babystep_z);//8
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -8970,6 +8970,7 @@ void lcd_experimental_menu()
 void reprint_from_eeprom() {
 	char cmd[30];
 	char filename[13];
+	char altfilename[13];
 	uint8_t depth = 0;
 	char dir_name[9];
 
@@ -8994,10 +8995,21 @@ void reprint_from_eeprom() {
 		filename[i] = eeprom_read_byte((uint8_t*)EEPROM_FILENAME + i);
 	}
 	filename[8] = '\0';
-
-	MYSERIAL.print(filename);
-	strcat_P(filename, PSTR(".gco"));
-	sprintf_P(cmd, PSTR("M23 %s"), filename);
+	
+	strcpy(altfilename,filename);
+	if (!card.FileExists(altfilename))
+	{
+		strcat_P(filename, PSTR(".gco"));
+		if (card.FileExists(filename))
+		{
+			strcpy(altfilename,filename);
+		}else
+		{
+			strcat_P(altfilename, PSTR(".g"));
+		}
+	}
+	MYSERIAL.print(altfilename);
+	sprintf_P(cmd, PSTR("M23 %s"), altfilename);
 	enquecommand(cmd);
   	sprintf_P(cmd, PSTR("M24"));
 	enquecommand(cmd);
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -79,7 +79,7 @@ static uint8_t lcd_commands_step = 0;
 CustomMsg custom_message_type = CustomMsg::Status;
 unsigned int custom_message_state = 0;
 
-bool isPrintFinished = false;
+bool enableReprint = false;
 bool isPrintPaused = false;
 uint8_t farm_mode = 0;
 int farm_timer = 8;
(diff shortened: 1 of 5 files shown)


DRracer:
@jfestrada thank you for your contribution, we are currently looking at your PR

jfestrada:
> @jfestrada thank you for your contribution, we are currently looking at your PR

If you need to clarify something just ask to me.

DRracer:
@jfestrada we found just one issue - if you remove the SD card after print, the Reprint option is still being offered to the user.

jfestrada:
> @jfestrada we found just one issue - if you remove the SD card after print, the Reprint option is still being offered to the user.

That's right I didn't thought on that case of use, so I will try to solve it as soon as possible.

jfestrada:
Hi @DRracer, 

I have made the changes in order to deactivate the "REPRINT" in the case that the user removes the SD Card because:

* If there is no SD card obviously there is nothing to reprint
* If the user inserts a different SD card maybe the file to reprint is not located on the SD

I have tested the changes with MK404 simulator because currently, I have my MK3S busy.

## TEST

https://user-images.githubusercontent.com/7290345/115211651-0e7bd300-a100-11eb-8e26-6ad0031d5e32.mp4

Firmware/ultralcd.cpp
```diff
@@ -8959,6 +8967,43 @@ void lcd_experimental_menu()
...
+		card.chdir(dir_name, false);
+	}
+
+	for (int i = 0; i < 8; i++) {
+		filename[i] = eeprom_read_byte((uint8_t*)EEPROM_FILENAME + i);
+	}
+	filename[8] = '\0';
+
+	MYSERIAL.print(filename);
+	strcat_P(filename, PSTR(".gco"));
```
DRracer: - are you sure the printed file always ends with a ".gco"?
- how about "1.g"?

Firmware/ultralcd.cpp
```diff
@@ -8959,6 +8967,43 @@ void lcd_experimental_menu()
...
+			dir_name[j] = eeprom_read_byte((uint8_t*)EEPROM_DIRS + j + 8 * i);
+		}
+		dir_name[8] = '\0';
+		MYSERIAL.println(dir_name);
+		// strcpy(dir_names[i], dir_name);
+		card.chdir(dir_name, false);
+	}
+
+	for (int i = 0; i < 8; i++) {
+		filename[i] = eeprom_read_byte((uint8_t*)EEPROM_FILENAME + i);
```
DRracer, line 8995: you can use a block reading function from the EEPROM, especially when you know the size

Firmware/ultralcd.cpp
```diff
@@ -8959,6 +8967,43 @@ void lcd_experimental_menu()
...
+
+	isPrintFinished=false;
+
+	//cmdqueue_reset();
+
+	depth = eeprom_read_byte((uint8_t*)EEPROM_DIR_DEPTH);
+	
+	MYSERIAL.println(int(depth));
+	for (int i = 0; i < depth; i++) {
+		for (int j = 0; j < 8; j++) {
```
DRracer, line 8985: how can you be sure that each dir has 8 characters and then there is '/'?

Firmware/ultralcd.cpp
```diff
@@ -79,7 +79,7 @@ static uint8_t lcd_commands_step = 0;
 CustomMsg custom_message_type = CustomMsg::Status;
 unsigned int custom_message_state = 0;
 
-
+bool isPrintFinished = false;
```
DRracer: Isn't there already some variable with similar meaning/semantics? Adding more of these vars only makes the code more "spaghetti-like"

DRracer (changes requested):
I looked at the code closer and my primary complaint is about the brittle implementation of recovering the filepath from the EEPROM.
Could you please address my comments?

Firmware/ultralcd.cpp
jfestrada: * I use the same logic that the one developed on the function **restore_print_from_eeprom()**
* As far as I know the eeprom value **EEPROM_FILENAME** only can contain 8 chars, and these chars are stored from the function **menu_action_sdfile()** on Marlin_main.cpp, where it takes the filename and assume always that the extension is _".gco"_ 

> ```
> static void menu_action_sdfile(const char* filename)
> {
>   loading_flag = false;
>   char cmd[30];
>   char* c;
>   bool result = true;
>   sprintf_P(cmd, PSTR("M23 %s"), filename);
>   for (c = &cmd[4]; *c; c++)
>     *c = tolower(*c);
> 
>   const char end[5] = ".gco";
> 
>   //we are storing just first 8 characters of 8.3 filename assuming that extension is always ".gco"
>   for (uint_least8_t i = 0; i < 8; i++) {
> 	  if (strcmp((cmd + i + 4), end) == 0) {
> 		  //filename is shorter then 8.3, store '\0' character on position where ".gco" string was found to terminate stored string properly
>  		  eeprom_write_byte((uint8_t*)EEPROM_FILENAME + i, '\0');
> 		  break;
> 	  }
> 	  else {
> 		  eeprom_write_byte((uint8_t*)EEPROM_FILENAME + i, cmd[i + 4]);
> 	  }
>   }
> ```

* There is a similar situation when the printer tries to recover from a power disruption because the function **restore_print_from_eeprom()** always assume that the extension will be _".gco"_  and concatenate the extension _".gco"_ with the filename 
* How about 1.g? The reprint will not work and I suppose that the recovery from a power outage will not work also because will not be able to find the file.

Firmware/ultralcd.cpp
jfestrada: I'm using the same logic that I have seen when the code is trying to read or write the EEPROM_FILENAME, I suppose that should be easy to recover a block of 8 positions starting from the EEPROM_FILENAME address but I have not tested it.

Firmware/ultralcd.cpp
jfestrada: That's the first thing that I have tried to find, but I didn't find any variable that could I use to know that the printer has finished a print job successfully, and because of that I change the status when the card finds the eof. Of course, if there is another existing variable that I could use for that purpose II will suggest the deletion of mine because that doesn't make sense, even more, when the resources are limited like in the 8 bits architecture. Please let me know if there is another vars combination that I could use, anyway I will try to find it harder.

Firmware/ultralcd.cpp
jfestrada: Because the EEPROM_DIRS has 80 chars and EEPROM_DIR_DEPTH has a limitation of 10 levels so each _dir name_ can be only 8 chars, I didn't read how the code manage when a dirname is smaller than 8 chars, I suppose that will fill the empty positions with "\0" but I didn't dig on it, as far as I know, the card.dirname[][] store the dirnames associated with the depth of each dir so the function **menu_action_sdfile()**  store in EEPROM_DIRS all the names together and in order to compose the path you need to do the same taking blocks of 8 chars for the dirname. 

```
static void menu_action_sdfile(const char* filename)
{
  loading_flag = false;
  char cmd[30];
  char* c;
  bool result = true;
  sprintf_P(cmd, PSTR("M23 %s"), filename);
  for (c = &cmd[4]; *c; c++)
    *c = tolower(*c);

  const char end[5] = ".gco";

  //we are storing just first 8 characters of 8.3 filename assuming that extension is always ".gco"
  for (uint_least8_t i = 0; i < 8; i++) {
	  if (strcmp((cmd + i + 4), end) == 0) {
		  //filename is shorter then 8.3, store '\0' character on position where ".gco" string was found to terminate stored string properly
 		  eeprom_write_byte((uint8_t*)EEPROM_FILENAME + i, '\0');
		  break;
	  }
	  else {
		  eeprom_write_byte((uint8_t*)EEPROM_FILENAME + i, cmd[i + 4]);
	  }
  }

  uint8_t depth = (uint8_t)card.getWorkDirDepth();
  eeprom_write_byte((uint8_t*)EEPROM_DIR_DEPTH, depth);

  for (uint_least8_t i = 0; i < depth; i++) {
	  for (uint_least8_t j = 0; j < 8; j++) {
		  eeprom_write_byte((uint8_t*)EEPROM_DIRS + j + 8 * i, card.dir_names[i][j]);
	  }
  }
```

Also is the same logic is used on **restore_print_from_eeprom()** so if there is a problem to retrieve the path from the reprint function maybe will have the same behaviour on the recovery from a power outage.

jfestrada:
> I looked at the code closer and my primary complaint is about the brittle implementation of recovering the filepath from the EEPROM.
> Could you please address my comments?
Yes, of course, I have answered all the questions, I'm with you that the filepath is the brittlest point and the reprint will not work if the filename extension isn't like _".gco*"_ maybe a solution should be to check if the file exists with the ".gco" and without it, but still not being the more elegant solution. What do you think?

Firmware/ultralcd.cpp
DRracer: Oh, good point, now I got why you made it this way.
Honestly, as the new Reprint function will be immediately available to users (read much more used than Power Panic), it should really work with all of the supported file extensions ... that would also probably fix the issue in Power Panic as well.
I'm not sure how to fix restore_print_from_eeprom() correctly yet.

Firmware/ultralcd.cpp
DRracer: ok, you are right

Firmware/ultralcd.cpp
DRracer: How far is card.sdprinting from the semantics of your isPrintFinished?

jfestrada: I will check it, the thing is that you need to know if there was a previous print wich was not cancelled in order to offer the reprint to the user.

DRracer: good point, probably not - card.sdprinting is only true when there is some SD print going on
... probably if you rename `isPrintFinished` to `enableReprint` it would be better ... because there is a bunch of isPrintFinished-like vars in the FW and each means something different.

jfestrada: Ok, that makes sense. Definitely, I will do the change.

Firmware/ultralcd.cpp
jfestrada: Maybe a solution could be to check if the file exists on the SD with the _".gco"_ and if the file is not found try to locate it without the extension. This will work if the filename is less than 8 chars but if the name is longer I'm not sure about the behaviour, is something that should be tested. What do you think?

Firmware/ultralcd.cpp
jfestrada: The variable **_isPrintFinished_** has been renamed to **_enableReprint_** and the change has been tested successfully.

Firmware/ultralcd.cpp
jfestrada: @DRracer I was thinking about this issue and maybe a solution could be to use a compression library in order to compress the string of the whole path (including the filename) before to store on the eeprom and decompress the string on a reprint situation or a recovery from a power outage.  The bad side is that needs compute efforts and free space for the library, also must be sure that the compressed string does not exceed the size reserved on the eeprom to store it. 
There are some libraries like https://github.com/siara-cc/Unishox_Arduino_Progmem_lib which could be studied.

wavexx: I think we can fix all these cases (reprint, restore and menu select from sd) by adding an additional function that starts the print via M24 starting from a file prefix. How we fix the extension (either by listing or by trying a second time with .g) would be trivial then

DRracer:
@jfestrada ok, enableReprint is fine ... and what shall we do about the (short) file name?

jfestrada:
> @jfestrada ok, enableReprint is fine ... and what shall we do about the (short) file name?

@DRracer the reprint function now works for every ".gcode" and ".g" file extensions also with a short or long name. In order to check if a file exists from the SD card, I have created a new function that will return true or false in case of the file exits or not.

Firmware/ultralcd.cpp
jfestrada: By now I have implemented the following solution: 

```
	strcpy(altfilename,filename);
	if (!card.FileExists(altfilename))
	{
		strcat_P(filename, PSTR(".gco"));
		if (card.FileExists(filename))
		{
			strcpy(altfilename,filename);
		}else
		{
			strcat_P(altfilename, PSTR(".g"));
		}
	}
	MYSERIAL.print(altfilename);
	sprintf_P(cmd, PSTR("M23 %s"), altfilename);
```

I know that this is not the more elegant solution but will work with .g and .gcode extensions. In order to know if the file exists or not I have implemented the _card.FileExists()_ function.

jfestrada:
@DRracer what do you think about implementing the same solution for the **restore_print_from_eeprom()** function in order to be able to recover **.g** files?

adamwolf:
Hi folks! I would absolutely love to see this, and am wondering if there's something I can do to help out.  Additional testing, rebasing, code changes....?

jfestrada:
Hi, you can test it using my version of the firmware, here is the latest release https://github.com/ComunidadMaker3D/CM3D-Prusa-Firmware/releases/tag/cm3d_3.11.0_v2.1

jfestrada:
> Hi folks! I would absolutely love to see this, and am wondering if there's something I can do to help out.  Additional testing, rebasing, code changes....?

Hi, you can test it using my version of the firmware, here is the latest release https://github.com/ComunidadMaker3D/CM3D-Prusa-Firmware/releases/tag/cm3d_3.11.0_v2.1

3d-gussner:
@jfestrada Hi, long time ago you made this PR. Do you think you could rebase it to the current MK3 branch?
We are also want to add `// action:reprint` to notify the HOST to re-print last host print from LCD.
Also a Host reprint SD card print option is something we want to implement.

jfestrada:
> @jfestrada Hi, long time ago you made this PR. Do you think you could rebase it to the current MK3 branch? We are also want to add `// action:reprint` to notify the HOST to re-print last host print from LCD. Also a Host reprint SD card print option is something we want to implement.

Hi, yes, that is something that I use to do with the [CM3D fork of the firmware ](https://github.com/ComunidadMaker3D/CM3D-Prusa-Firmware) so I will create a new PR from that repository.

3d-gussner:
@jfestrada Sorry it took so long to get more attention to this PR.
We are trying to get better and faster in reviews of community PRs.
The community PRs are very appreciated :hugs: .

jfestrada:
I have created a new PR #4462 we can continue the discussion there, the code has been integrated on the current MK3 branch and I added the translations for the MSG_REPRINT.

PR #3075, opened 2021-03-19
New G-Code (M880) in order to check the Steel Sheet installed

jfestrada:
**Summary**

Suddenly I forgot about to change the steel sheet selected, obviously, I'm not using the same sheet with all the materials, so I use to get in a hurry to change the sheet before my PEI got damage by the nozzle, sadly this is something that sometimes happens and my PEI knows very well. 

**Why a gcode?**

I decided to generate a G-Code instead of modifying another piece of code because I understand that not everyone will like to have this reminder on the LCD screen before starting the print. So you can decide if you will use it on your "start G-Code" of your printer profile or not. 

**Why the gcode M880?**

I don't really sure how to choose a g-code, so I was looking at the g-codes that perform checks and I take a free slot near to them. 

**How it works?**

The M880 will check if you are printing from USB or not, if not will check how many steel sheets do you have stored on the EEPROM, if there are 2 or more steel sheets stored will recover the name of the one that is currently active and will show a fullscreen message with the question "Continue?". If there is no answer after 30 seconds will cancel the print. 

This is an example of the message:

```
----------------------------------
| Check the steel sheet installed|
| Texture1, Continue?            |
----------------------------------
```

**Example of start g-code on the printer profile:**
```

M115 U3.9.3 ; tell printer latest fw version
M880; check the steel sheet
M83 ; extruder relative mode
```

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -8380,6 +8380,17 @@ Sigma_Exit:
                }
     break;
 
+  /*!
+  ### M880 - Check the Steel Sheet, the M880 code should be set before to start the print in order to ask the user about
+  the viability of use the current configured steel sheet and try to minimize the risk of use a sheet with the configuration 
+  of another sheet
+  ### Usage
+      M880
+  */
+    case 880:
+      steel_sheet_check();
+    break;
+
 #ifdef LIN_ADVANCE
     /*!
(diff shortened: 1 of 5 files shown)


jfestrada:
I have improved this feature a lot, so I will close this PR due that is based on a repository where I do not have access anymore, and I will open the new one with the improved feature.

PR #3078, opened 2021-03-21
Nozzle fan and print fan are swapped on lcd_extruder_info() function

jfestrada:
Maybe I'm lost on translation, in that case, I would suggest changing the name of the FANs in order to avoid these mistakes. 

**HotFix**

This is a HotFix for the lcd_extruder_info() function because the nozzle fan is showing the RPM coming from position 0 of the array fan_speed[] but the correct position for this fan is the 1. So I swapped it to the correct assignment.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -1442,7 +1442,7 @@ void lcd_menu_extruder_info()                     // NOT static due to using ins
     char nozzle[maxChars], print[maxChars];
     pgmtext_with_colon(_i("Nozzle FAN"), nozzle, maxChars);  ////c=10 r=1
     pgmtext_with_colon(_i("Print FAN"), print, maxChars);  ////c=10 r=1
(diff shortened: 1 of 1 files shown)


3d-gussner:
Have you tested this PR? I think this PR breaks the output instead of fixing something.

I have tested the FW3.10.0-RC1 and the Fan speeds are **correct.**
- Connect via USB with Octoprint/Pronterface
- Open Support-> Extruder info
- Send gcode `M106 S255`
- Watch Print FAN: xxxx RPM ~4800
- Send gcode `M106 S0`
- Print FAN: 0 RPM
- Heat up nozzle to 60°C using gcode `M104 S60`
- Watch Nozzle FAN: yyyy RPM ~3400 RPM as soon above ~50°C
- Cool down nozzle `M104 S0`
- Nozzle FAN: 0 RPM as soon nozzle temp is below 50°C

3d-gussner (commented):
Dis-approve

jfestrada:
Then I suppose I was wrong with the translation.

3d-gussner:
Please test the changes before you create a PR.

Still like it that you are creating PRs, please keep doing it.

PR #3084, opened 2021-03-25
:construction: Detect motor type

mkbel:
Experiment to detect connected motor as safety mechanism for https://github.com/prusa3d/Prusa-Firmware/pull/2813

PFW-1234


mkbel:
It is possible to distinguish different motors based on PWM amplitude scaler, but motor must be moving because:
PWM_SCALE is updated only when the motor is moving.
The register value freezes once motor has stopped and it even can not be detected, whether motor moved before, or it is initial value of the register.
Register value for two specimens of motors I want to distinguish differs only when the motor is moving at some speed, but is the same once the register freezes in motor standstill.

PR #3089, opened 2021-03-28
Show affected axes in crash detection messages

wavexx:
Explicitly show which axes are affected on the status line when a crash
has been detected.

When the crash recovery prompt is triggered, also show all axes which
have been affected during the detection interval (currently X, Y, or
both).

The timer detection has been simplified to match the MK3 optical
filament timeout handling, which makes more sense: a crash immediately
following a crash recovery prompt will trigger the prompt again. The
timer needs to properly expire in order to reset the crash count.

This helps crash detection tuning (and help users reporting issues with
crash detection) without having to monitor the serial line.

This currently abuses the MSG_CRASH_DETECTED message by prepending the
[X][Y] labels to the message, which is ok in english but might not
translate that well for all languages without tuning the translations.

On English it's now: ``X Crash detected.``, optionally followed by ``Resume print?``.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -594,6 +594,14 @@ void crashdet_restore_print_and_continue()
 //	babystep_apply();
 }
 
+void crashdet_fmt_error(char* buf, uint8_t mask)
+{
+    if(mask & X_AXIS_MASK) *buf++ = axis_codes[X_AXIS];
+    if(mask & Y_AXIS_MASK) *buf++ = axis_codes[Y_AXIS];
+    *buf++ = ' ';
+    strcpy_P(buf, _T(MSG_CRASH_DETECTED));
+}
+
 void crashdet_detected(uint8_t mask)
 {
 	st_synchronize();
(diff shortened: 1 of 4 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -629,26 +624,45 @@ void crashdet_detected(uint8_t mask)
 	{
 		eeprom_update_byte((uint8_t*)EEPROM_CRASH_COUNT_X, eeprom_read_byte((uint8_t*)EEPROM_CRASH_COUNT_X) + 1);
 		eeprom_update_word((uint16_t*)EEPROM_CRASH_COUNT_X_TOT, eeprom_read_word((uint16_t*)EEPROM_CRASH_COUNT_X_TOT) + 1);
+		strcat(msg, "X");
```
leptun: Please use the progmem variant for there

Firmware/Marlin_main.cpp
```diff
@@ -629,26 +624,45 @@ void crashdet_detected(uint8_t mask)
...
 	}
-    
-
 
 	lcd_update_enable(true);
 	lcd_update(2);
-	lcd_setstatuspgm(_T(MSG_CRASH_DETECTED));
+
+    // prepare the status message with the _current_ axes stauts
+    strcat(msg, " ");
```
leptun: maybe you can use some printf function for this. It seems wasteful to me to call various string functions instead of just one call with the FMT in the progmem

Firmware/Marlin_main.cpp
```diff
@@ -629,26 +624,45 @@ void crashdet_detected(uint8_t mask)
...
 	gcode_G28(true, true, false); //home X and Y
 	st_synchronize();
 
 	if (automatic_recovery_after_crash) {
 		enquecommand_P(PSTR("CRASH_RECOVER"));
 	}else{
 		setTargetHotend(0, active_extruder);
-		bool yesno = lcd_show_fullscreen_message_yes_no_and_wait_P(_i("Crash detected. Resume print?"), false);////MSG_CRASH_RESUME c=20 r=3
+
+        // notify the user of *all* the axes previously affected, not just the last one
```
leptun, line 652: Is this outdated code by any chance?

wavexx:
Rebased and updated a bit to reduce code duplication. This saves 2 bytes of ram, 18 of flash.

Firmware/Marlin_main.cpp
wavexx: This is still relevant. It's also used in pr #3090 which saves some extra bytes too.

leptun (approved):
Uses 110B of flash and 1B of ram. I say it's optimized enough.
I wish the lcd code for the fullscreen menu didn't need a custom implementation, but considering it's not using too many resources, I say it's good enough as it is now.
Tested on my MK3S and the prompts / messages were correct

DRracer (approved):

PR #3090, opened 2021-03-28
Optimize fullscreen prompt handling

wavexx:
Modify ``lcd_show_multiscreen_message_two_choices_and_wait_P`` to also handle single-screen or empty (no-clear) prompts, making other functions redundant.

Saves 76 bytes.

This is based on top of PR #3089 (there's just one additional commit). It was split off because it changes the layout of existing Yes/No of messages: the previous prompt would use the last two lines of the LCD, while the new prompt is using just the last line of the LCD instead.

Before:
2021-03-28T175413

After:
2021-03-28T175118

Translation do not require updates, since the Yes/No translation was already the same in both implementations and thus should always fit already.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -3178,7 +3178,7 @@ lcd_wait_for_click_delay(0);
 }
 
 //! @brief Show multiple screen message with yes and no possible choices and wait with possible timeout
-//! @param msg Message to show
+//! @param msg Message to show. If NULL, do not clear the screen and handle choice selection only.
 //! @param allow_timeouting if true, allows time outing of the screen
 //! @param default_yes if true, yes choice is selected by default, otherwise no choice is preselected
 //! @retval 1 yes choice selected by user
@@ -3188,8 +3188,21 @@ int8_t lcd_show_multiscreen_message_yes_no_and_wait_P(const char *msg, bool allo
 {
     return lcd_show_multiscreen_message_two_choices_and_wait_P(msg, allow_timeouting, default_yes, _T(MSG_YES), _T(MSG_NO));
 }
-//! @brief Show multiple screen message with two possible choices and wait with possible timeout
-//! @param msg Message to show
+//! @brief Show a two-choice prompt on the last line of the LCD
+//! @param first_selected Show first choice as selected if true, the second otherwise
+//! @param first_choice text caption of first possible choice
+//! @param second_choice text caption of second possible choice
+void lcd_show_two_choices_prompt_P(bool first_selected, const char *first_choice, const char *second_choice)
+{
+    lcd_set_cursor(0, 3);
+    lcd_print(first_selected? '>': ' ');
+    lcd_puts_P(first_choice);
+    lcd_set_cursor(7, 3);
+    lcd_print(!first_selected? '>': ' ');
+    lcd_puts_P(second_choice);
+}
+//! @brief Show single or multiple screen message with two possible choices and wait with possible timeout
+//! @param msg Message to show. If NULL, do not clear the screen and handle choice selection only.
 //! @param allow_timeouting if true, allows time outing of the screen
 //! @param default_first if true, fist choice is selected by default, otherwise second choice is preselected
 //! @param first_choice text caption of first possible choice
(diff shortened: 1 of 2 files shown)


Firmware/ultralcd.cpp
```diff
@@ -3472,8 +3472,21 @@ int8_t lcd_show_multiscreen_message_yes_no_and_wait_P(const char *msg, bool allo
...
+//! @brief Show a two-choice prompt on the last line of the LCD
+//! @param first_selected Show first choice as selected if true, the second otherwise
+//! @param first_choice text caption of first possible choice
+//! @param second_choice text caption of second possible choice
+void lcd_show_two_choices_prompt_P(bool first_selected, const char *first_choice, const char *second_choice)
+{
+    lcd_set_cursor(0, 3);
+    lcd_print(first_selected? '>': ' ');
+    lcd_puts_P(first_choice);
+    lcd_set_cursor(7, 3);
```
3d-gussner: Can you change it to `lcd_set_cursor(10, 3);` so the `Yes/No` both translations can be up to 8 chars?

wavexx: We need to ensure that all the possible translation for "choice 2" (not just no - but all the wizard stuff) fits into 8 characters though.

Otherwise that's pretty trivial

wavexx:
We now allow to position the second prompt on an arbitrary column, allowing choices to be flexible in spacing.
As a a side-effect, we can now balance the space of Yes/No prompts:

2021-04-05T212107

sketch size didn't seem to change, so we just got a free feture? ;)

Firmware/ultralcd.cpp
wavexx: Done!

3d-gussner:
Reminder after it got merged:
- Review translations to un-shorten these again to make it more readable.

3d-gussner:
@wavexx I have found few other messages that could be improved:
- Stop SD card print
- Selftest FANs
So if you like please change it as well.

```
void lcd_sdcard_stop()
{

	lcd_puts_at_P(0, 0, _T(MSG_STOP_PRINT));
	lcd_puts_at_P(2, 2, _T(MSG_NO));
	lcd_puts_at_P(2, 3, _T(MSG_YES));
	lcd_putc_at(0, 2, ' ');
	lcd_putc_at(0, 3, ' ');

....
	lcd_puts_at_P(1, 2, _T(MSG_SELFTEST_FAN_YES));
	lcd_putc_at(0, 3, '>');
	lcd_puts_at_P(1, 3, _T(MSG_SELFTEST_FAN_NO));
....
```

Firmware/messages.c
```diff
@@ -26,7 +26,7 @@ const char MSG_COMMUNITY_MADE[] PROGMEM_I1 = ISTR("Community made"); ////c=18
 const char MSG_CONFIRM_NOZZLE_CLEAN[] PROGMEM_I1 = ISTR("Please clean the nozzle for calibration. Click when done."); ////c=20 r=8
 const char MSG_COOLDOWN[] PROGMEM_I1 = ISTR("Cooldown"); ////c=18
 const char MSG_CRASH[] PROGMEM_I1 = ISTR("Crash"); ////c=7
-const char MSG_CRASH_DETECTED[] PROGMEM_I1 = ISTR("Crash detected."); ////c=20
+const char MSG_CRASH_DETECTED[] PROGMEM_I1 = ISTR("Crash detected."); ////c=17 r=1
```
3d-gussner: Why c=17?

3d-gussner (changes requested):

3d-gussner:
I was thinking if we can improve changed correctly as well.
```
01234567890123456789
Changed correctly and
Color is correct? 
>Yess>Nooo>Extr. more
```
We have a limit of c=4 for MSG_YES and MSG_NO which means we could have triple choice with c=9
<space or `>`:1><`YES`:4><space or `>`:1><`NO`:4><space or `>`:1><third option>
20-1-4-1-4-1=9

Firmware/messages.c
wavexx: This is because "XY " can/is added at the beginning of the string now, so we have a little less space.

3d-gussner:
Tried to crash Y axis and the output is odd
image

leptun:
@3d-gussner I think the branch is outdated. I already fixed that bug. See https://github.com/prusa3d/Prusa-Firmware/pull/3415

3d-gussner (approved):
Works and looks fine. Approved

Content approved the layout changes and @gudnimg is building on this one for the error messags.

PR #3093, opened 2021-03-30
Remove FW version parsing

DRracer:
... as it can be done at compile time. Code size dropped by >800 bytes.

### DIFF
--- a/Firmware/Configuration.h
+++ b/Firmware/Configuration.h
@@ -16,8 +16,13 @@ extern uint16_t nPrinterType;
 extern PGM_P sPrinterName;
 
 // Firmware version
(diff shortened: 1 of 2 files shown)


3d-gussner:
Created https://github.com/DRracer/Prusa-Firmware/pull/11 as `PF-build.sh` otherwise will fail.
Also added "FLAVORs" for "DEBUG, DEVEL, ALPHA, BETA and RC" to `Configuration.h`

DRracer:
Postponing to 3.11 as we are quite ok with code size in the upcoming 3.10.

XPila (approved):

PR #3100, opened 2021-04-06
Fix PowerPanic and Crash Detection during homing and bed leveling

wavexx:
This PR fixes recovery from either power panic or crash detection that happen during bed leveling or homing.

In the current master, a crash detection during MBL will recover incorrectly, first by moving to the last MBL probe point, then level, and then start to extrude in a random spot of the bed prior to resume. If a crash happens during the second MBL run, a stack overflow is likely to happen. A similar situation happens if power panic triggers.

During crash detection, the MBL function needs to correctly check if the moves are aborted during st_synchronize() and unroll the stack if that's the case (that's true for any function that performs multiple moves - sadly a condition which is hardly checked-for currently).

In both PP and Crash detection, we now handle G28 and G80 specially by marking the current XY coordinates and saved target as invalid. When resuming we check is the saved coordinates are valid or not, and if not we avoid restoring them entirely. This is correct, because at a higher level both crash detection and power panic will always re-run the entirety of the last gcode instruction. This avoids both the useless move towards the last probed point and also the incorrect extrusion immediately following.

In d2be404 we also fix another race condition that can happen during PP recovery that I managed to trigger while debugging this issue. It's similar in behavior to previous #2469: a missing call to clamp_to_software_endstops during initialization can generate a phantom move, except that in this case it can result in an immediate axis crash right before homing.

PFW-1170

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3036,6 +3036,8 @@ static void gcode_G80()
     plan_buffer_line_curposXYZE(XY_AXIS_FEEDRATE);
     // Wait until the move is finished.
     st_synchronize();
+    if (waiting_inside_plan_buffer_line_print_aborted)
+        return;
 
     uint8_t mesh_point = 0; //index number of calibration point
     int Z_LIFT_FEEDRATE = homing_feedrate[Z_AXIS] / 40;
@@ -3105,6 +3107,8 @@ static void gcode_G80()
         //printf_P(PSTR("after clamping: [%f;%f]\n"), current_position[X_AXIS], current_position[Y_AXIS]);
         plan_buffer_line_curposXYZE(XY_AXIS_FEEDRATE);
         st_synchronize();
+        if (waiting_inside_plan_buffer_line_print_aborted)
+            return;
 
         // Go down until endstop is hit
         const float Z_CALIBRATION_THRESHOLD = 1.f;
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -10851,8 +10851,9 @@ void uvlo_()
     }
 
     // save the global state at planning time
+    bool pos_invalid = XY_NO_RESTORE_FLAG;
     uint16_t feedrate_bckp;
-    if (current_block)
+    if (current_block && !pos_invalid)
     {
         memcpy(saved_target, current_block->gcode_target, sizeof(saved_target));
         feedrate_bckp = current_block->gcode_feedrate;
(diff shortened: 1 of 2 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -11282,10 +11288,13 @@ void restore_print_from_eeprom(bool mbl_was_active) {
...
     // without shifting Z along the way. This requires performing the move without mbl.
-	sprintf_P(cmd, PSTR("G1 X%f Y%f F3000"),
-              eeprom_read_float((float*)(EEPROM_UVLO_CURRENT_POSITION + 0)),
-              eeprom_read_float((float*)(EEPROM_UVLO_CURRENT_POSITION + 4)));
-	enquecommand(cmd);
+    float pos_x = eeprom_read_float((float*)(EEPROM_UVLO_CURRENT_POSITION + 0));
+    float pos_y = eeprom_read_float((float*)(EEPROM_UVLO_CURRENT_POSITION + 4));
+    if (pos_x != X_COORD_INVALID)
+    {
+        sprintf_P(cmd, PSTR("G1 X%f Y%f F3000"), pos_x, pos_y);
```
DRracer, line 11295: are we sure we never run out of the cmd buffer with different float format?

wavexx: You mean something which could be potentially broken?

x/y were always stored during uvlo_ (and during crash detection, which is handled the same just in a different function), so it's highly unlikely that you'll be able to flash the fw just after a power panic, but before the printer resumes (changing the value is not an issue either).

The same method is also used to mark the "saved target" (for multipath G1 commands) as invalid.
This is less annoying that adding a new global+eeprom value to mark the value as invalid IMHO.

I'm currently defining X_COORD_INVALID as X_MIN-1, which is guaranteed to not pass the frontend in marlin (it's going to be clamped to X_MIN-X_MAX).

DRracer: No, I meant if there is a possibility of creating `G1 X<some float> Y<some float> F3000` which runs of those 30 bytes allocated for the cmd buffer. Adding some limitations like `%4.2f` could solve the problem (avoiding the need for snprinf())

wavexx: Got it, I'm going to check that

wavexx: So there's indeed a buffer overflow, which is caused by the subsequent sprintf.
I expanded the buffer to fix the maximum practical limit.

However, sprintf doesn't allow to specify a maximum field width for floats that I know of, it's always a minimum which doesn't help to ensure we're not going to overflow the buffer.

We would need to check all float values manually, which feels expensive code-wise.

DRracer (approved):

PR #3120, opened 2021-04-21
Fix partial redraw during filament autoload

wavexx:
If the printer was already being pre-heated but didn't reach the target
temperature yet and a new filament is being inserted, the LCD used to
display a "Preheating to load" message to block the loading until the
extruder is hot.

This message is currently missing, and the ">Cancel" option doesn't
display immediately either, depending on the extruder height.

This PR fixes this behavior, which was broken during an earlier update.

We now force-update LCD updates during the first (and _only_) time the
screen is setup, and push all messages _before_ the carriage is
eventually raised, so that ">Cancel" is shown immediately as well.

PFW-1220

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -2201,14 +2201,15 @@ void mFilamentItem(uint16_t nTemp, uint16_t nTempBed)
     }
     else
     {
-        lcd_set_cursor(0, 0);
-        lcdui_print_temp(LCD_STR_THERMOMETER[0], (int) degHotend(0), (int) degTargetHotend(0));
-
         if (!bFilamentWaitingFlag)
         {
             // First run after the filament preheat selection:
             // setup the fixed LCD parts and raise Z as we wait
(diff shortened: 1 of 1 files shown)


DRracer:
@wavexx any ideas which modes of operation this change/fix can influence? I.e. what is your proposed test scenario?

wavexx:
The main issue here was that at some point we optimized the screen redraw to just setup the screen once, with the exception of temperature. The screen used to redraw "correctly" before, but just because the second redraw/s would fix it ;)

The main difference here is that lcd_draw_update was already set only when accessing this menu _from_ the main menu.
When the temp is already set, we jump directly to the wait screen, explaining the difference.

The scenarios I tested were:

- reset printer, insert filament without preheat first (you need to go through the material menu)
- reset printer, preheat first (don't wait for temp), then insert filament
- reset, home first, then insert filament (changes the raising behavior for ">Cancel")

DRracer (approved):

Issue #3077, opened 2021-03-21
[BUG] Printer fails to print certain models after firmware update to 3.10.0-RC1


lhecker:
**Printer type** - MK3S+
**Printer firmware version**-  3.10.0-RC1
**SD card or USB/Octoprint** - stock SD card
**PrusaSlicer** - 2.3.0

**Describe the bug**
My MK3S+ fails to print the popular [Cam Clip](https://www.prusaprinters.org/prints/27216-bag-clip-print-in-place-cam) model using the 3.10.0-RC1 firmware. It works fine under 3.9.3.
The initial nozzle flush completes successfully. Afterwards it just prints the first 2-3mm of the skirt and stops entirely. It doesn't pause, but rather aborts the print at that point (there's no pause/stop print menu point anymore). The SD card menu point reads "no SD card", but if you click on it, it'll open just fine.
This issue didn't happen with any of the models I've printed in the days before (on 3.10.0-RC1 as well).
  
**To Reproduce**
Flash the 3.10.0-RC1 firmware on a MK3S+ and try to print the attached .gcode file using Prusament PLA.
As mentioned above, it'll abort after printing the first 2-3mm of the skirt.
  
**Expected behavior**
The print shouldn't abort.
  
**G-code**
[cam_clip.zip](https://github.com/prusa3d/Prusa-Firmware/files/6177520/cam_clip.zip)

P.S.: I hope this ticket isn't a duplicate. The other, similar issues seemed to be unrelated to this.

Barnabasek:
I am able to reproduce this bug with your g-code from that zip file. But not with some of my own g-codes.
- run the print
- stop the printing via the menu during the first layer
- run the print again right away
- it then stops after the purge line showing no SD card in the menu (but you can access the files) and the printer just keeps heating while showing the main screen. Only reset helps.

DRracer:
@lhecker we managed to reproduce the issue with your G-code, interestingly the behavior of the printer varies.

Hauzman:
I had similar issues with one of my gcode making the brim is strange way and it move the extruder almost in home position and the mesh leveling wash on the lcd calibrating z 49 point and the up down  movement was super slow almous i think the printer was stoped but after a couples of tryouts and reslice the model all was ok again
Here is a video 
https://www.dropbox.com/s/x1jwtozyljqhznr/20210325_082711.mp4?dl=0

Pyroschnueffler:
[...]

PR #3125, opened 2021-04-23
Fix M105 from SD

leptun:
Fixes https://github.com/prusa3d/Prusa-Firmware/issues/3077#issuecomment-821963501

The M105 command would accidentally pop two commands from the cmdqueue. This did not affect the USB printing because there is ever only one command in the cmdqueue when USB printing. However, when SD printing, the cmdqueue is most of the time full, so the command after M105 would be popped as well. With this patch only one command is popped by the M105 handling.

PFW-1238


3d-gussner:
Tested it before and it is working, even I couldn't see any difference in Putty out between this PR and removing previous FIX `cmdqueue_pop_front(); //prevent an ok after the command since this command uses an ok at the beginning.`

DRracer (approved):

PR #3155, opened 2021-05-21
Improve filament sensor PCB upgrade logic 

wavexx:
The current code surrounding the IR filament sensor board upgrade is problematic in a few ways. It still checks for a PCB upgrade even though the FW already knows the PCB is at the newest version, wasting cycles. It allows downgrading the version after a previous successful detection, which doesn't make much sense.

In this PR:

- Ensure the upgrade logic only runs if the previous PCB version is unknown or if an upgrade is actually possible.
- Ensure the check only runs until a reliable detection can be made, then stop checking.
- Ensure the version detection doesn't run forever: limit the number of total autodetection attempts in a single session.
- Simplify and clarify the code surrounding the check.


Phetdanai:
#4312

PR #3158, opened 2021-05-22
Fix the eeprom address documentation

Panayiotis-git:
A small fix at the eeprom address documentation for the setting EEPROM_ECOOL_ENABLE.


Panayiotis-git:
In the same comment block, it is written that

> _Pronterface_ does __not__ support D-codes

The Pronterface, starting from the latest version [2.0.0rc8,](https://github.com/kliment/Printrun/releases) supports the Debug codes. So, `D3` will return the content of the EEPROM.

But, unfortunately, it also converts the user commands to uppercase letters before sending them to the printer, so it will fail to execute the command `D3 Ax0d04 C1`.

wavexx:
> But, unfortunately, it also converts the user commands to uppercase letters before sending them to the printer, so it will fail to execute the command `D3 Ax0d04 C1`.

I wonder if we should allow uppercase-hex when parsing. Have you tried to file a bug in proterface for this? Uppercasing gcode by default sounds sensible in most cases for convenience, but it's not universally ok to do.

PR #3157, opened 2021-05-22
Add extra debugging functions & crash analysis

wavexx:
D2 allows to read/write memory, but it's currently disabled behind DEBUG_DCODES.
This PR allows to enable D2 selectively by defining DEBUG_DCODE2, as similarly done for DEBUG_DCODE3.

We also fix the function to read the entire address space (SRAM starts at 0x200). SRAM also becomes the default starting address, so that we avoid reading from bidirectional registers by default.

DEBUG_DCODES2 "costs" 436 bytes when enabled.
I think it would be possible to save some space by unifying D2 and D3, which are mostly identical.

PR #3162, opened 2021-05-24
Add several low-level debugging tools

wavexx:
In this PR we add several low-level debugging tools to inspect/poke/debug the Prusa MK3 firmware. All the tools are added into the ``tools/`` directory, along with a README containing summary documentation.

Most of these are new and were prompted due to the need to investigate recent issues and thus are not perfect yet. They are designed to be used in conjunction with Prusa Dcodes (see #3157).

Some are quite useful (``elf_mem_map``), while others are mostly convenience wrappers, but nevertheless they can be quite handy. Hopefully there will be more in the future. The list currently includes:

### ``dump_eeprom``

Dump the content of the entire EEPROM using the D3 command.
Requires ``printcore`` from [Pronterface].

### ``dump_sram``

Dump the content of the entire SRAM using the D2 command.
Requires ``printcore`` from [Pronterface].

### ``elf_mem_map``

Generate a symbol table map starting directly from an ELF firmware with DWARF2 debugging information (which is the default using the stock board definition).

When used along with a memory dump obtained from the D2 g-code, show the value of each symbol which is within the address range.

This assumes the running firmware generating the dump and the elf file are the same.
Requires Python3 and the [pyelftools](https://github.com/eliben/pyelftools) module.

### ``update_eeprom``

Given one EEPROM dump, convert the dump to update instructions that can be sent to a printer.

Given two EEPROM dumps, produces only the required instructions needed to update the contents from the first to the second. This is currently quite crude and assumes dumps are aligned (starting from the same address or same stride).

Optionally writes the instructions to the specified port (requires ``printcore`` from [Pronterface]).

### ``noreset``

Set the required TTY flags on the specified port to avoid reset-on-connect for *subsequent* requests (issuing this command might still cause the printer to reset).

[Pronterface]: https://github.com/kliment/Printrun


wavexx:
Rebased onto MK3_3.10.1

DRracer (approved):
nice :+1:

PR #3167, opened 2021-05-27
Remove ignored/incorrect PROGMEM

wavexx:
This PROGMEM is currently ignored by gcc, but even if it wasn't it
wouldn't be correct since the following code is expecting to read "item"
without fetching the array itself from PROGMEM.


leptun:
Oops. This was my fault. Is there any way / benefit to moving the pointer array to progmem?

wavexx:
Considering the size of the array, I don't think so. Looks fine to me as is

DRracer (approved):

PR #3170, opened 2021-06-01
Refactored TMC2130 crash detection, save 12 bytes of dynamic mem, optimize performance

scottmudge:
Edit: See comment below for explanation on the majority of changes.

---

Within the TCM2130 ISR, it loops through each axis inclusive of the z-axis, despite the Z-axis driver's diagnostic bit flags never being read or checked in `tmc2130_sample_diag()`, where it has been commented away.

Thus, iterating over the z-axis in the associated ISR is unnecessary. 

After performing a diff on the compiled elf binaries (with and without z-axis included in loop), It looks like the compiler has already optimized away the unused loop anyway, but for explicit clarity, it may be best to exclude the unchecked axis unless it is added back in the associated diagnostic check.

Binary diff:
image

### DIFF
--- a/Firmware/tmc2130.cpp
+++ b/Firmware/tmc2130.cpp
@@ -251,7 +251,8 @@ void tmc2130_st_isr()
 	uint8_t crash = 0;
 	uint8_t diag_mask = tmc2130_sample_diag();
 //	for (uint8_t axis = X_AXIS; axis <= E_AXIS; axis++)
-	for (uint8_t axis = X_AXIS; axis <= Z_AXIS; axis++)
+//	for (uint8_t axis = X_AXIS; axis <= Z_AXIS; axis++)
+    for (uint8_t axis = X_AXIS; axis <= Y_AXIS; axis++)
 	{
 		uint8_t mask = (X_AXIS_MASK << axis);
 		if (diag_mask & mask) tmc2130_sg_err[axis]++;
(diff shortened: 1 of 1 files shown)


scottmudge:
This is an update.

After further refactoring, it was possible to optimize away and shrink the size of a few counters, saving 12 bytes of dynamic memory, and 94 bytes of program storage space. The behavior is identical. The counters are now hidden behind preprocessor switches only active when the error counters were being debugged, as this was the only time they were being used.

The counters were also decreased from 2 bytes to 1 byte, as the maximum threshold value was `64`. There is a preprocessor switch automatically increasing these counter variable sizes if the predefined threshold is ever increased.

General optimizations were done as well.

Before refactoring:

`Sketch uses 233146 bytes (91%) of program storage space. Maximum is 253952 bytes.
Global variables use 6120 bytes of dynamic memory.`

After refactoring:

`Sketch uses 233052 bytes (91%) of program storage space. Maximum is 253952 bytes.
Global variables use 6108 bytes of dynamic memory.`

leptun:
This PR might not be needed in the end if #3395 is reliable. Can you check this PR and see if you have issues with crash detection?

leptun:
@scottmudge It has been decided that the sfilt PR (#3395) will be used instead. That one will be merged towards the end of the next week after a larger scale test is done internally. I'm removing the 3.12 tag from this PR as it is not needed anymore.

scottmudge:
That's fine, that is probably the more appropriate solution anyway. Thanks.

PR #3178, opened 2021-06-05
Cleanup

gudnimg:
Fixed indentation mostly. I don't want to make too many changes. This file needs a lot more work.


gudnimg:
Updated the branch with rebase to include latest merges

DRracer (commented):
**Please, don't**. Such changes do not bring any improvement and will break other existing pull requests.

The only correct solution to indentation problems/inconsistencies is to introduce clang-formatting rules like we have in the Buddy firmware including the PR-checking and reformatting the whole code base at once.

PR #3181, opened 2021-06-07
memory and flash optimizations for 3.12

leptun:


leptun:
Still left to do:
- [x] Make WELCOME_MSG non-translated progmem (the translation was dropped many years ago)
- [ ] Remove_unused_variables if there are any left.
- [ ] change the maaaaany 2byte ints to uint8_t/int8_t where appropriate.
- [ ] Merge hotend and bed PID loops. Just like with thermal runaway where the bed is index 0 and the hotends are index>=1
- [x] Pack these flags into two bytes (endstops and endstops_old) and use bitmasks for access:
```
5d2     1       old_z_max_endstop
5d3     1       z_max_endstop
5d4     1       old_z_min_endstop
5d5     1       z_min_endstop
5d6     1       old_y_max_endstop
5d7     1       y_max_endstop
5d8     1       old_y_min_endstop
5d9     1       y_min_endstop
5da     1       old_x_max_endstop
5db     1       x_max_endstop
5dc     1       old_x_min_endstop
5dd     1       x_min_endstop
```
- [x] Wth is this? Let's get rid of it
```
25c     1       firstrun
```
- [x] These three can be merged.
```
4da     1       endstop_y_hit
4db     1       endstop_x_hit
endstop_z_hit
```
- [ ] Should be ifdefed out with SNMM flag:
```
21d     2       bowden_length
```
- [ ] These two look redundant. Should be investigated:
```
26b     4       max_pos
277     4       min_pos
```
- [x] Is this really needed?
```
519     4       endstops_trigsteps
```
- [ ] We should change the Timer.cpp/h to be more useful for cases like these:
```
283     4       safetytimer_inactive_time
287     4       stepper_inactive_time
2a1     4       prev_busy_signal_ms
57f     4       _usb_timer
5b9     4       mmu_last_request
5bd     4       mmu_last_response
6f7     4       lcd_next_update_millis
702     4       NcTime
814     4       max_inactive_time
14a1    4       extruder_autofan_last_check
1598    4       kicktime
159c    4       PingTime
```
- [ ] PID autotune should be improved. Creating a secondary main loop is stupidly dangerous and flash hungry. The state machines should be done better. With that, these global variables become redundant:
```
6a4     4       _Kd
6a8     4       _Ki
6ac     4       _Kp
6b0     1       pid_tuning_finished
```
- [ ] Considering all the memory corruption that can happen, I feel pretty unsafe to have this flag in SRAM. Maybe we can use some IO register for this bit (eg, `GPIOR0`. It is <=0x3F, so the bit manipulation instructions work on this)
```
18e7    1       bedPWMDisabled
```
- [ ] This flag can be removed since the timer already has a flag built-in:
```
bIRsensorStateFlag
```

gudnimg:
> * change the maaaaany 2byte ints to uint8_t/int8_t where appropriate.

I think all uses of `fanSpeedBckp` and `fanSpeed` can be changed from `int` to `uint8_t`. 

Especially changing `fanSpeed` to `uint8_t` in `Marlin.h` and `Marlin_main.cpp` saves `126` bytes of flash and `1` byte of dynamic memory on my end using the code in this PR.

Edit:
Changing this line to `uint8_t` saves additional `16` bytes of flash and `16` bytes of dynamic memory.
https://github.com/prusa3d/Prusa-Firmware/blob/a8a253c0ae115d9e4ec75adf62018436019a50d6/Firmware/planner.h#L108

leptun:
@gudnimg Yes. These fan speeds that are uint16_t can be changed all around the code. There are a lot more scenarios like this for other variables, so I just generalized all of them into this (from the todo list):
> - change the maaaaany 2byte ints to uint8_t/int8_t where appropriate.

gudnimg:
Adding a comment here since I think its best to handle this in this PR

I noticed `backlightLevel_HIGH` and `backlightLevel_LOW` are declared as `int16_t` but we're storing them into memory as `uint8_t`.

```cpp
void backlight_save() //saves all backlight data to eeprom.
{
    eeprom_update_byte((uint8_t *)EEPROM_BACKLIGHT_LEVEL_HIGH, (uint8_t)backlightLevel_HIGH); // <--- uint8_t
    eeprom_update_byte((uint8_t *)EEPROM_BACKLIGHT_LEVEL_LOW, (uint8_t)backlightLevel_LOW); // <--- uint8_t
    eeprom_update_byte((uint8_t *)EEPROM_BACKLIGHT_MODE, backlightMode);
    eeprom_update_word((uint16_t *)EEPROM_BACKLIGHT_TIMEOUT, backlightTimer_period);
}

```
Changing these two variables to `uint8_t` saves `10` bytes of Flash and `2` bytes of dynamic memory.

Also `backlightTimer_period` should be declared `uint16_t` and not `int16_t` (for consistency).

gudnimg:
More potential savings:

`homeaxis()` function: change the parameter type of axis to `uint8_t`
```c
#ifdef TMC2130
void homeaxis(int axis, uint8_t cnt, uint8_t* pstep)
#else
void homeaxis(int axis, uint8_t cnt)
#endif //TMC2130
```

There was another function somewhere which just changing the type of the parameters saved a bit of flash. I can't remember where I found it, I will edit this comment if I find it again. 😃 

Edit:  found it

If we change 

```c
static void lcd_selftest_screen_step(int _row, int _col, int _state, const char *_name_PROGMEM, const char *_indicator)
```
to this:
```c
static void lcd_selftest_screen_step(uint8_t _row, uint8_t _col, uint8_t _state, const char *_name_PROGMEM, const char *_indicator)
```
We save 62 bytes of flash (at least on my end).

gudnimg:
Is there any way for me to help with this PR. Is it possible for me to create PRs for it? 😊

leptun:
Yes, of course. PRs to this one are welcome. I don’t think there can be multiple PRs for this in the main repo directly since the memory optimizations will be merged into MK3 at once, not in parts.
I’d probably start with the easy fixes, such as converting ints to uint8_ts wherever possible.
Something I can recommend is keeping a log of the compiled output size after each change. The code size usually goes down, but other times it becomes way larget for no reason. If you can’t make it smaller again, note it down so that it can be checked again at the end.

Firmware/tmc2130.cpp
```diff
@@ -963,7 +963,7 @@ void tmc2130_set_wave(uint8_t axis, uint8_t amp, uint8_t fac1000)
 		else
 			reg >>= 1;
 //		printf("%3d\t%3d\t%2d\t%2d\t%2d\t%2d    %08x\n", i, vA, dA, b, w[s], s, reg);
-	}
+	} while (i++ != 255);
```
leptun, line 966: @gudnimg please double check if it behaves correctly now

gudnimg: Was the code not working before? 🤔 

I have a vague idea of what the code does but I don't see any issues. We start at `i = 0` and loop through `i = 255` (microstep number 256) which I believe your code will do.

leptun: The old code would loop indefinitely since uint8_t is always smaller than 256.

gudnimg: Ah I see, can't believe I missed that.

leptun:
Hmmm… there seems to be some issue with the commits. I don’t remember having those commits at the beginning.

scottmudge:
There is a bug in `Marlin_main.cpp`.

At line ~`5679`:

```
for (int y = MESH_NUM_Y_POINTS-1; y >= 0; y--) {
```

was changed to 

```
for (uint8_t y = MESH_NUM_Y_POINTS-1; y >= 0; y--) {
```

to save a couple dynamic bytes. However, the test/break condition `y >= 0` will never be satisfied (it will always be true) if y is declared unsigned. 

In other areas, similar loops use `int8_t`, which is what I assume you meant to use.

gudnimg:
> There is a bug in `Marlin_main.cpp`.
> 
> At line ~`5679`:
> 
> ```
> for (int y = MESH_NUM_Y_POINTS-1; y >= 0; y--) {
> ```
> 
> was changed to
> 
> ```
> for (uint8_t y = MESH_NUM_Y_POINTS-1; y >= 0; y--) {
> ```
> 
> to save a couple dynamic bytes. However, the test/break condition `y >= 0` will never be satisfied (it will always be true) if y is declared unsigned.
> 
> In other areas, similar loops use `int8_t`, which is what I assume you meant to use.

@scottmudge Yup you're right, good catch. 👍 

EDIT:
Interesting, changing this from `uint8_t` to `int8_t` increases flash usage by 212 bytes on my end.

leptun:
Closing in favor of #3365

PR #3182, opened 2021-06-07
Implement/use xflash_rd_data_atomic to read from XFLASH 

wavexx:
When reading from the main loop the stepper ISR is potentially running and expects the SPI bus to be free.

As such any time we need to read from the XFLASH we need to do so with interrupts disabled to ensure we don't read garbled data. Implement xflash_rd_data_atomic to do this.

lang_get_count and lang_get_code get called in the menus while interrupts are enabled. Use the new xflash_rd_data_atomic to perform the reads.

### DIFF
--- a/Firmware/language.c
+++ b/Firmware/language.c
@@ -108,13 +108,12 @@ uint8_t lang_get_count()
 	if (pgm_read_dword(((uint32_t*)(_PRI_LANG_SIGNATURE))) == 0xffffffff)
 		return 1; //signature not set - only primary language will be available
 #ifdef XFLASH
-	XFLASH_SPI_ENTER();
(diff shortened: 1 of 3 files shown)


leptun:
What exactly would try to use the spi bus from an interrupt? Because if that’s the case, the SD is unsafe as well.

wavexx:
When reading the stallguard registers periodically when the printer is in normal mode.

wavexx:
.. Or maybe not. I noticed this is done inside st_synchronize(), but that is only called within the ISR during a PAT9125 filament sensor recheck, which is probably not correct behavior anyway.

wavexx:
Ok, changed my mind, this doesn't seem to be exceedingly useful.

PR #3187, opened 2021-06-08
Adapt the flow of first layer calibration taking care of the installed nozzle

jfestrada:
# Summary
Currently, the G-Code for the first layer calibration does not take care of the nozzle installed on the printer, so I propose to change that behavior in order to get an accurate first layer calibration with the different nozzle size, selecting the correct flow so you don't need to recalculate the whole G-Code just get a good approximation of the flow using the M221 to set the approximate flow.

## Flow assumption
* Nozzle **0.25** --> 62% of the flow
* Nozzle **0.4** --> 100% of the flow
* Nozzle **0.6** --> 150% of the flow
* Nozzle **0.8** --> 200% of the flow

The M221 will be enqueued before starting the intro line. 

**Example with 0.6 nozzle:**

n06_flow150

PFW-1009


wavexx:
This makes a ton of sense..

3d-gussner:
Changing it to `uint8_t nDiameter; ` and read `EEPROM_NOZZLE_DIAMETER` you could save a bit.
Not sure if you can enque a formula so it looks like `M221 S<calculated_value>`
calculated_value=EEPROM_NOZZLE_DIAMETER/40*100

This would save even more bytes.

Tested `M221 S62,5` which is the calculated value for 25 nozzle and here the output:
```
Send: M221 S62,5
Recv: ok
[...]
Send: M221
Recv: 62%
Recv: ok
```

jfestrada:
> Changing it to `uint8_t nDiameter; ` and read `EEPROM_NOZZLE_DIAMETER` you could save a bit.
> Not sure if you can enque a formula so it looks like `M221 S<calculated_value>`
> calculated_value=EEPROM_NOZZLE_DIAMETER/40*100
> 
> This would save even more bytes.
> 
> Tested `M221 S62,5` which is the calculated value for 25 nozzle and here the output:
> 
> ```
> Send: M221 S62,5
> Recv: ok
> [...]
> Send: M221
> Recv: 62%
> Recv: ok
> ```

Thanks for the tip I have updated the code and now the flow value is dynamically calculated and saves 22 bytes from the sketch size. 

```
void lay1cal_set_flow()
{    
    uint8_t nDiameter;
    nDiameter=eeprom_read_word((uint16_t*)EEPROM_NOZZLE_DIAMETER);

    char cmd[12];
    char buffnr[6];
    dtostrf((double)(nDiameter/0.4),3,2,buffnr);
    sprintf_P(cmd, PSTR("M221 S%s"), buffnr);
    enquecommand(cmd);   
}
```

3d-gussner:
Looking at the travis build your new and nicer code uses
https://travis-ci.com/github/prusa3d/Prusa-Firmware/builds/229444871#L558-L559
```
Sketch uses 251104 bytes (98%) of program storage space. Maximum is 253952 bytes.
Global variables use 6103 bytes of dynamic memory.
```

while the previous version used
https://travis-ci.com/github/prusa3d/Prusa-Firmware/builds/228387501#L558-L559
```
Sketch uses 251070 bytes (98%) of program storage space. Maximum is 253952 bytes.
Global variables use 6103 bytes of dynamic memory.
```

So the idea to make it more elegant uses more resources. Sometimes it "sucks" trying make it nicer and hoping to get it smaller but at the end it turns out that it uses more resources.

BTW. you still use `eeprom_read_word` and `uinit16_t` while `eeprom_read_byte` and `uinit8_t` is correct for the eeprom value.

jfestrada:
I have switched back to the `switch` version and also I changed from `uint16_t` to `uint8_t` and `eeprom_read_byte`, you were totally right. Again, thanks for the tips.

Firmware/first_lay_cal.cpp
```diff
@@ -212,3 +212,40 @@ void lay1cal_square(char *cmd_buffer, uint8_t i)
...
+
+//! @brief Set the flow in order to the installed nozzle
+//!
+//! This function must be called before to start to print in order to 
+//! enqueue the correct flow depending the nozzle size stored on the 
+//! eeprom
+//!
+void lay1cal_set_flow()
+{    
+   uint8_t nDiameter;
```
wavexx: I would change this to ClNozzleDiameter from util.h, cast the return of eeprom_read_byte to this type, and handle all the various diameters (ClNozzleDiameter::_Diameter_250, etc) also removing the "default" case.

This way when we add a new diameter type in the future we'll get a warning that a new switch case is needed.

DRracer:
@jfestrada nice and elegant PR, definitely worth some more work for the next FW version.

Regarding your attempt to dynamically calculate the flow - have you actually seen the assembly output of your attempt?
```
    char cmd[12];
    char buffnr[6];
    dtostrf((double)(nDiameter/0.4),3,2,buffnr);
    sprintf_P(cmd, PSTR("M221 S%s"), buffnr);
    enquecommand(cmd);  
```
Obviously this must have taken a ton of CPU FLASH, because the floating point operations are costly. However, all this can be computed in integers with some clever approach, as we do not really care about flow being 125 or 124.8. 

Btw. forget about doubles in 8bit FW, we only have floats.

DRracer:
@jfestrada one more question - where do you reset the flow factor after the 1st layer calibration is done?

jfestrada:
> @jfestrada nice and elegant PR, definitely worth some more work for the next FW version.
> 
> Regarding your attempt to dynamically calculate the flow - have you actually seen the assembly output of your attempt?
> 
> ```
>     char cmd[12];
>     char buffnr[6];
>     dtostrf((double)(nDiameter/0.4),3,2,buffnr);
>     sprintf_P(cmd, PSTR("M221 S%s"), buffnr);
>     enquecommand(cmd);  
> ```
> 
> Obviously this must have taken a ton of CPU FLASH, because the floating point operations are costly. However, all this can be computed in integers with some clever approach, as we do not really care about flow being 125 or 124.8.
> 


I have tested the bellow code (using integers), and still being 16 bytes expensive than the code which uses the switch.

_Start to build Prusa Firmware ...
Using variant 1_75mm_MK3S-EINSy10a-E3Dv6full
Sketch uses **251030** bytes (98%) of program storage space. Maximum is 253952 bytes.
Global variables use 6103 bytes of dynamic memory._

```
    uint8_t nDiameter;
    nDiameter=eeprom_read_byte((uint8_t*)EEPROM_NOZZLE_DIAMETER);

    char cmd[9];
    sprintf_P(cmd, PSTR("M221 S%i"), (nDiameter*100)/40);
    enquecommand(cmd);

```

The code with the switch has 16 bytes less compared with the previous which concatenates the strings, and there is less computational cost, so in my opinion, the below code is efficienter.

_Start to build Prusa Firmware ...
Using variant 1_75mm_MK3S-EINSy10a-E3Dv6full
Sketch uses **251014** bytes (98%) of program storage space. Maximum is 253952 bytes.
Global variables use 6103 bytes of dynamic memory._

```
    uint8_t nDiameter;
    nDiameter=eeprom_read_byte((uint8_t*)EEPROM_NOZZLE_DIAMETER);

    switch (nDiameter)
    {
    case 25:
        enquecommand_P(PSTR("M221 S62"));
        break;

    case 40:
        enquecommand_P(PSTR("M221 S100"));
        break;
    
    case 60:
        enquecommand_P(PSTR("M221 S150"));
        break;

    case 80:
        enquecommand_P(PSTR("M221 S200"));
        break; 

    default:
        enquecommand_P(PSTR("M221 S100"));
        break;
    }
```

> Btw. forget about doubles in 8bit FW, we only have floats.

Thanks for the tip, it makes sense.

jfestrada:
> @jfestrada one more question - where do you reset the flow factor after the 1st layer calibration is done?

I didn't reset the flow factor because the gcode files generated with Prusaslicer and the Prusa profiles include the M221 G-Code, and if the printer is reset the flow factor will be reset anyway. So I don't see the use case which needs to reset the flow factor. Anyway, could be included as the last instruction on the **lay1cal_square()** function but will have an extra cost on the term of bytes.

Firmware/first_lay_cal.cpp
jfestrada: I made those changes, as you said I removed the "default" case but I had to add the case for "ClNozzleDiameter::_Diameter_Undef:" in order to avoid the warning, so now if there is any new nozzle size there will be a warning at compilation time.

```
case ClNozzleDiameter::_Diameter_Undef:
        enquecommand_P(PSTR("M221 S100"));
        break;
```
Now is 4 bytes expensive but it worth the extra bytes because now use the defined values on the enum class ClNozzleDiameter.

Firmware/first_lay_cal.cpp
```diff
@@ -212,3 +213,33 @@ void lay1cal_square(char *cmd_buffer, uint8_t i)
...
+//!
+void lay1cal_set_flow()
+{    
+    ClNozzleDiameter nDiameter=(ClNozzleDiameter)eeprom_read_byte((uint8_t*)EEPROM_NOZZLE_DIAMETER);
+
+    switch(nDiameter){
+    case ClNozzleDiameter::_Diameter_250:
+        enquecommand_P(PSTR("M221 S62"));
+        break;
+    case ClNozzleDiameter::_Diameter_400:
```
wavexx, line 233: This could be:

```
case ClNozzleDiameter::_Diameter_Undef:
case ClNozzleDiameter::_Diameter_400:
```

so that we fold the undefined case with the .4 nozzle.

jfestrada: I don't know why, but the solution that you propose is expensive in bytes, the use of program storage space is `251058` 

```
Start to build Prusa Firmware ...
Using variant 1_75mm_MK3S-EINSy10a-E3Dv6full
Sketch uses **251058** bytes (98%) of program storage space. Maximum is 253952 bytes.
Global variables use 6103 bytes of dynamic memory.
```
But with the current solution, the bytes used are `251018`.

```
Start to build Prusa Firmware ...
Using variant 1_75mm_MK3S-EINSy10a-E3Dv6full
Sketch uses **251018** bytes (98%) of program storage space. Maximum is 253952 bytes.
Global variables use 6103 bytes of dynamic memory.
```
I suppose that the compiler identifies the same static string and just links to that string and because of that consume fewer bytes because the string is not really duplicated, anyway, the best approach is the current one, what do you think?
Maybe I'm wrong, I need to investigate a little bit about that behavior.

wavexx: I just tried with the following:

```
    switch(nDiameter){
    case ClNozzleDiameter::_Diameter_250:
        enquecommand_P(PSTR("M221 S62"));
        break;
    case ClNozzleDiameter::_Diameter_Undef:
    case ClNozzleDiameter::_Diameter_400:
        enquecommand_P(PSTR("M221 S100"));
        break;
    case ClNozzleDiameter::_Diameter_600:
        enquecommand_P(PSTR("M221 S150"));
        break;
    case ClNozzleDiameter::_Diameter_800:
        enquecommand_P(PSTR("M221 S200"));
        break;
    }
```

and it shrinks by ~14 bytes on my end

jfestrada: You were right, I have pushed the changes just to compare the Travis build logs. 

The las build with your suggested code:

```
Sketch uses 251412 bytes (98%) of program storage space. Maximum is 253952 bytes.

```

The previous build from 10 days ago:

```
Sketch uses 251482 bytes (99%) of program storage space. Maximum is 253952 bytes.

Global variables use 6106 bytes of dynamic memory.
```

So the code consumes fewer bytes now.

DRracer:
@jfestrada Thank you for your effort, it's been inspiring. We decided to merge another implementation #3312 which in our opinion calculates the extrusion amount more correctly.

Would you mind closing this PR of yours?

PR #3189, opened 2021-06-17
Second attempt at retrieving the SN from the 32u2 IC

leptun:
Sometimes an incorrect SN was saved to the EEPROM because the host software tried connecting to the printer while the mega2560 was talking to the mega32u2. This resulted in the saved SN to look something like this: `N0 M11CZPX0919X004X`.

Now the SN is checked on startup more thoroughly. A valid SN starts with `CZPX`, anything else being considered invalid.
The `get_PRUSA_SN()` function was also improved. The serial line is flushed before trying to get the SN and the retrieval process is attempted multiple times until a valid SN is received (with a 250ms timeout).

PFW-1256

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -457,7 +457,7 @@ static void gcode_M105(uint8_t extruder);
 static void temp_compensation_start();
 static void temp_compensation_apply();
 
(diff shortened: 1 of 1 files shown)


DRracer (approved):
this really looks like improving the reliability of reading the SN

PR #3192, opened 2021-06-17
PFW-1271 PF-build v2.0.0

3d-gussner:
PF-build.sh version 2.0.0

Added support for [MK404](https://github.com/vintagepc/MK404) see also [MK404 wiki](https://github.com/vintagepc/MK404/wiki)
- MK404-build.sh is only supported on LINUX at this moment
  - Checks for dependencies to build the MK404 SIM
  - Compiles the MK404 and some default SD cards.

Added ELF files being saved next to the hex files for debugging with FW3.10.1 and newer

PF-build.sh got few new arguments
- to start automatically the MK404 SIM with compiled Prusa-Firmware
  - `-m` 1 = Printer, 2 = Printer+MMU2
  -  `-g` 1 = lite, 2 = fancy graphics, 3 = lite with Quad_HR and 4=fancy with Quad_HR
  - `-x` 8/64 ram size for the virtual atmel404 in MK404
  - `-y` 256/384/512/1024/32M flash size for the virtual atmel404 in MK404
- Verbose mode
  - `-j` 1 = verbose output

Depends on https://github.com/prusa3d/Prusa-Firmware/pull/3093

ToDo before final PR:
- [x] Set MK404 `--debugcore` and `--bootloader-file` depending when extended RAM and FLASH size.

- [x] Remove MK404 copy of lang file as it has been fixed in MK404


vintagepc:
I approve but perhaps I'm slightly biased here :rofl:

MK404-build.sh
```diff
@@ -221,3 +312,115 @@ if [[ ! -f "$MK404_BUILD_PATH/Prusa_MK3S_SDcard.bin" || "$force_flag" == "1" ]];
...
+        MK404_options="--debugcore"
+    fi
+    if [[ ! -z $MK404_DEBUG && "$MK404_DEBUG" == "atmega404_no_bootloader"  || ! -z $BOARD_FLASH && "$BOARD_FLASH" != "0x3FFFF" ]]; then
+        MK404_options='--debugcore --bootloader-file ""'
+    fi
+
+# Run MK404 with grafics
+    if [ ! -z "$graphics_flag" ]; then
+        if [ ! -z "$MK404_options" ]; then
+            MK404_options="${MK404_options} --colour-extrusion --extrusion Quad_HR -g "
```
vintagepc: Not sure if we want to make Quad_HR the default; perhaps better as an option too since the HRs can put a pretty hefty toll on the system for larger test prints.

MK404-build.sh
```diff
@@ -0,0 +1,546 @@
...
+sleep 2
+}
+#### End: Check if OSTYPE is supported
+
+#### Start: Check MK404 dependencies
+check_packages()
+{
+packages=(
+"libelf-dev"
+"gcc-7"
```
DRracer, line 318: @vintagepc do we really need to stick with gcc-7? At least gcc-10 works well

DRracer (approved):

MK404-build.sh
vintagepc: If it compiles without issue there's no reason we have to stick to 7.5; it's just what my system and the build runners use. I can upgrade my system but I remember having to backtrack on the build runners because it was griping about something at the time. 

Nothing to it other than being a known constant for consistency reasons :)

PR #3194, opened 2021-06-19
Remove useless function EEPROM_read_st

wavexx:
This function is both questionable and unused...

### DIFF
--- a/Firmware/stepper.cpp
+++ b/Firmware/stepper.cpp
@@ -1559,17 +1559,6 @@ void digitalPotWrite(int address, int value) // From Arduino DigitalPotControl e
 }
 #endif
 
-void EEPROM_read_st(int pos, uint8_t* value, uint8_t size)
-{
-    do
-    {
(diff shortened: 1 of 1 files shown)


DRracer (approved):

gudnimg:
I found the exact same function in ultralcd.cpp.

https://github.com/prusa3d/Prusa-Firmware/blob/53fcd6fc8f8d9545e6d069e457b70a6a059dae49/Firmware/ultralcd.cpp#L4256

```cpp
void EEPROM_read(int pos, uint8_t* value, uint8_t size)
{
  do
  {
    *value = eeprom_read_byte((unsigned char*)pos);
    pos++;
    value++;
  } while (--size);
}
```

just above it there is also this function:

```cpp
void EEPROM_save(int pos, uint8_t* value, uint8_t size)
{
  do
  {
    eeprom_write_byte((unsigned char*)pos, *value);
    pos++;
    value++;
  } while (--size);
}
```

PR #3195, opened 2021-06-19
Remove redundant `extern` variables and add `#ifdef` when using `extern` variables

gudnimg:
* Remove redundant `extern` already included with `temperature.h`.
* Add `#ifdef` in `Dcodes.cpp` when using `extern` variables. We don't want to use them if they're not defined.
* Removed one `extern` variable in `cmdqueue.cpp` which is already included by `cardreader.h`
* `lcd_change_fil_state` had two identical `extern`'s in `Marlin.h`, since only one is needed I removed one.
* Remove redundant `extern` variable `is_usb_printing` from `tmc2130.cpp`. It is included with `Marlin.h`
* Remove redundant `extern` variable `lcd_encoder` from `menu.cpp`. It is included with `lcd.h`

### DIFF
--- a/Firmware/Dcodes.cpp
+++ b/Firmware/Dcodes.cpp
@@ -538,26 +538,20 @@ const char* dcode_9_ADC_name(uint8_t i)
 	return 0;
 }
 
-#ifdef AMBIENT_THERMISTOR
-extern int current_temperature_raw_ambient;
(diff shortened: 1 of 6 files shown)


gudnimg:
Updated the branch with rebase to include latest merges

DRracer (approved):

Issue #3175, opened 2021-06-02
[BUG] Crazy high nozzle temperature display with 3.10.0


tg73:
**MK3S**
**3.1.10**

**MMU2S**
**1.0.6**

**SD card or USB/Octoprint**
Both, but in this case I think Octoprint was not connected.

**Describe the bug**
Generally, I've noticed some weird behaviour since updating to 3.10.0. Hard to pin down, but includes sometimes getting stuck at the boot screen and needing a power cycle. And also ended up in this state somehow, I think after doing some MMU load/unload from the menu after cancelling a print initiated from octoprint and also disconnecting in octoprint, but I can't be sure. It was a few days ago and I didn't have time to log the bug at the time, just took a photo. I remember that the temperature was gradually decreasing, so it looks like some weird multiplier got involved. Along with #3174 which I've just logged, and the changes announced in 3.10.0, the erratic behaviour smells to me (a professional software engineer) of memory corruption. Nothing like this has happened on my machine pre 3.10.0.

![image](https://user-images.githubusercontent.com/1101693/120541455-cf6cbd00-c3e1-11eb-9b5d-8e39cc1bc22d.png)

leptun:
Yes, we are chasing some memory corruption issue. Probably caused by the same bug that generates #3151 #3147

wavexx:
Indeed. The high temp which slowly reduces looks like some memory spraying happening in the PID regulation, which is another pointer going towards the temperature ISR. Which we also enable earlier in 3.10, which could also explain getting stuck early during boot.

Do you remember exactly when it got stuck? Can you reproduce it with the serial attached? (serial messages will show a little more info and could help pinpoint the moment).

tg73:
Sorry, I can't remember for sure the exact circumstances for it getting stuck during boot. Quite possibly it was when clicking "connect" in octoprint - so the printer does I guess a soft reboot (or at least, that's what it looks like). I'm not sure how to connect serial at the same time as diagnosing an interaction with octoprint which is connected via USB (running on an separate RPi).

wavexx:
In that case, what's the output of the serial log shown in octoprint? Completely empty?

wavexx:
Also.. @tg73, how easily can you reproduce this lock-on-boot issue? Does it happen >50% of the times you connect?

tg73:
[...]

Issue #3147, opened 2021-05-13
[BUG] 3.10 Preheat menu loops when power on with loaded filament


Lyr3x:
**Printer type** - [e.g. MK3S, MK3, MK2.5S, MK2.5, MK2S, MK2]
Mk3s+
**Printer firmware version**-  [e.g. 3.8.1, 3.8.1-RC1, ...]
3.10 / 3.9.3

**SD card or USB/Octoprint**
SD

**Describe the bug**
  When power on the printer with loaded filament the preheat menu is appearing again and again (inconsistent delay between). This only occurs if filament is loaded before power on the printer. As soon as I remove the filament once the issue is gone until the next power switch. Filament sensor is detected correct and the lever etc. is working flawless. 
Flashed 3.9.3 back and the issue is gone. I repeated the flash process multiple times.
I also tried multiple materials and colors.

Tried to factory reset before and after flashing 3.10 with no success. Sensor voltages and states are correct. When unplugging the filament sensor the pop up does not appear. 

I can't identify a hardware issue as with 3.9.3 everything is working without any issues. The preheat menu appears also if autoload is disabled. 

When unloading and reloading the filament while the printer is powered on, the sensor is working as expected. 
  
**To Reproduce**
1. Load filament
2. Power off the printer
3. Power on the printer
4. Preheat/Autoload Menu appears again and again
  
**Expected behavior**
1. Load filmanet
2. power off -> on
3. Nothing

If you need any more information, i am happy to provide them ASAP.

wavexx:
@Lyr3x can you make a video of the behavior?

If I understand correctly, the Preheat menu shows up, you *exit* the menu, but after a few seconds it pops up again?

Lyr3x:
Sure. Follow the link: https://photos.app.goo.gl/trkSfidprWqBfEJdA

I printed over the weekend with 3.9.3 without the issue. Flashed `prusa3d_fw_MK3S_MK3S+_3_10_0_4481.hex` for the video and as you can see, the issue is instantly present.

wavexx:
Might be an issue with the fsensor version detection?

Start from reset with the filament unloaded. Under the support support, just under the date what does the "Fil. sensor v.:" say? Then go to voltages and note down the value. Then insert the filament just to trigger the sensor (ignore autoload - go back to voltages). What does it read now?

Then re-do the same, but with the filament inserted and reset. Detected version under support? Initial voltage?

Lyr3x:
[...]

PR #3202, opened 2021-06-23
Work-around GCC LTO codegen bug in process_commands()

wavexx:
When building with GCC 4.9.2 (bundled with PF-build-env-1.0.6.*), -Os and LTO enabled, PID_autotune gets automatically inlined into process_commands() as it's the only call location.

Sadly, due to the massive size of process_commands(), it results in codegen bug doing a partial stack overwrite in process_commands() itself, manifesting as random behavior depending on the timing of interrupts and the codepath taken inside the merged function.

I believe this is the main reason behind #3147, #3151, #3175 and probably others. I verified the codegen issue was also present in 3.9.3 already and probably older FW versions.

In most cases, this results in a frameshift of the stack pointer. Due to the sheer size of process_commands and the many different branches it *usually* has no effect, except when run the first time. Since the time-to-first-run is influenced by the serial and some external ICs such as the xflash, it can cause absurd behavior depending on how the stack is built, such as the one seen in #3147.

I admit I didn't really read the _entire_ disassembly of the process_commands() function, it's just too large to keep track manually of all the registers and the stack allocations.

When I discovered the issue I tried a few approaches to avoid the bug. Shrinking process_commands() itself to at least half the size (splitting M/G handling blocks) seem to reduce the chances of hitting this (hinting to a register allocation issue as the basis), but it's not always reliable. Marking PID_autotune as noinline is more sensible and avoids the issue completely even when other code in process_commands() is modified.

Given how old this gcc version is there's not much to report upstream. I also don't think there's much we can do besides manual verification in the future to avoid hitting this again. I tried to manually trigger this issue on other large functions within process_commands(), but there's something specific to PID_autotune and it's position.

I didn't verify if this also applies to more recent versions of gcc that we could potentially use.
Nasty.

PFW-1254

### DIFF
--- a/Firmware/temperature.cpp
+++ b/Firmware/temperature.cpp
@@ -287,8 +287,10 @@ bool checkAllHotends(void)
     return(result);
 }
 
-  void PID_autotune(float temp, int extruder, int ncycles)
-  {
+// WARNING: the following function has been marked noinline to avoid a GCC 4.9.2 LTO
+//          codegen bug causing a stack overwrite issue in process_commands()
+void __attribute__((noinline)) PID_autotune(float temp, int extruder, int ncycles)
+{
   pid_number_of_cycles = ncycles;
   pid_tuning_finished = false;
   float input = 0.0;


DRracer:
@wavexx wow, great job isolating this compiler bug ... so we should really have a look at process commands and keep it just as one big switch but with every case statement as a separate non_inline function.

DRracer (approved):

PR #3204, opened 2021-06-24
Change `boolean` to `bool` everywhere

### LINKED ISSUE #3199: [BUG] `boolean` typedef is used a few times, replace it with `bool` for consistency
Not a bug exactly, but I noticed that the Arduino typedef `boolean` is used in very few places in the code. I suggest we change them all to `bool` for consistency.

```cpp
typedef bool boolean;
```

One example:

https://github.com/prusa3d/Prusa-Firmware/blob/eb4cf1a77f57b0825b72c5209d48fceaa8c0ec58/Firmware/cmdqueue.h#L52

gudnimg:
This closes #3199 

The use of `boolean` instead of `bool` are due to the Arduino framework, there is no benefit in using `boolean` since it is just a typedef for the built-in `bool` type.

Edit:

This change was also implemented in Marlin 2 in 2017: https://github.com/MarlinFirmware/Marlin/commit/fa26767efe50d278d4dfa66df9f48831f14c7af2

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -375,7 +375,7 @@ bool target_direction;
 //Insert variables if CHDK is defined
 #ifdef CHDK
 unsigned long chdkHigh = 0;
-boolean chdkActive = false;
+bool chdkActive = false;
 #endif
 
 //! @name RAM save/restore printing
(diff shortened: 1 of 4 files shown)


DRracer (approved):
Thank you, it makes sense to remove some relicts of the past.

PR #3206, opened 2021-06-24
Remove four functions `EEPROM_read`, `EEPROM_save`, `EEPROM_read_B`, `EEPROM_save_B`

gudnimg:
`EEPROM_read()` is exactly the same as the function removed in https://github.com/prusa3d/Prusa-Firmware/pull/3194 and since it is only used to read data that is 1 byte we can use `eeprom_read_byte()` directly.

Functions `EEPROM_read_B()` and `EEPROM_save_B()` are also removed for the same reason.

The changes save 312 bytes of flash memory on Arduino IDE 1.8.19. No changes in SRAM footprint.


wavexx:
Who-ho.. good catch.

Firmware/Marlin_main.cpp
```diff
@@ -1502,7 +1502,9 @@ void setup()
 #ifdef SNMM
 	if (eeprom_read_dword((uint32_t*)EEPROM_BOWDEN_LENGTH) == 0x0ffffffff) { //bowden length used for SNMM
 	  int _z = BOWDEN_LENGTH;
-	  for(int i = 0; i<4; i++) EEPROM_save_B(EEPROM_BOWDEN_LENGTH + i * 2, &_z);
+	  for(uint8_t i = 0; i < 4; i++) {
+	  	eeprom_update_word((uint16_t*)EEPROM_BOWDEN_LENGTH + i * 2, _z);
```
DRracer: Isn't this a different address than before? 
I'm not sure about the casting and pointer arithmetics rules in this case, but I'm afraid the result will produce addresses:
say we have EEPROM_BOWDEN_LENGTH = 0 for simplicity;
`(uint16_t*)EEPROM_BOWDEN_LENGTH + i * 2 -> (0 + sizeof(uint16_t) * i * 2)`
and not the original:
`(0 + i * 2)`

DRracer (changes requested):
Please address my comment about the pointer arithmetics results.

Otherwise, saving code size by removing unnecessary code is always welcome

Firmware/Marlin_main.cpp
gudnimg: That’s an interesting point. I actually am not sure so I would need to research to confirm. My instinct says this isn’t an issue, but I want to be sure. 

Am I understanding you correctly that you refer to the casting from `uint8_t` to `uint16_t*`?

It is desireable to use `uint16_t*` to use the eeprom function correctly. https://www.nongnu.org/avr-libc/user-manual/group__avr__eeprom.html


`void eeprom_update_word (uint16_t *__p, uint16_t __value)`

gudnimg:
@DRracer You were right! I created a quick Arduino sketch to test this. I write the value `433` as a word to the EEPROM in four words in a row. Then I verify the result by reading one byte at a time. The expected output from a word is `0x1B1` = 433.

<details>
<summary> Code: </summary>

```cpp
void setup() {
  // put your setup code here, to run once:
  Serial.begin(115200);

  while(!Serial);
}

void loop() {
  // put your main code here, to run repeatedly:
  
  for(uint8_t i = 0; i < 4; i++) {
     Serial.print("Word ");
     Serial.println(i);
     eeprom_update_word((uint16_t*)4007 + i * 2, 433);
     Serial.print("    Byte 1 (");
     Serial.print(4007 + i * 2);
     Serial.print(") --> ");
     Serial.println(eeprom_read_byte((uint8_t*)4007 + i * 2), HEX);
     Serial.print("    Byte 2 (");
     Serial.print(4007 + i * 2 + 1);
     Serial.print(") --> ");
     Serial.println(eeprom_read_byte((uint8_t*)4007 + i * 2 + 1), HEX);
     Serial.println(" ");
  }

  Serial.flush();
  abort();
}
```
</details>


Serial output:
```
Word 0
    Byte 1 (4007) --> B1
    Byte 2 (4008) --> 1
 
Word 1
    Byte 1 (4009) --> FF
    Byte 2 (4010) --> FF
 
Word 2
    Byte 1 (4011) --> B1
    Byte 2 (4012) --> 1
 
Word 3
    Byte 1 (4013) --> FF
    Byte 2 (4014) --> FF
```

Notice how Word 1 and Word 3 are skipped. Now... if I change this line

```cpp
eeprom_update_word((uint16_t*)4007 + i * 2, 433);
```

to this:

```cpp
eeprom_update_word((uint16_t*)4007 + i, 433);
```
I get the expected output:

```
Word 0
    Byte 1 (4007) --> B1
    Byte 2 (4008) --> 1
 
Word 1
    Byte 1 (4009) --> B1
    Byte 2 (4010) --> 1
 
Word 2
    Byte 1 (4011) --> B1
    Byte 2 (4012) --> 1
 
Word 3
    Byte 1 (4013) --> B1
    Byte 2 (4014) --> 1

```

Firmware/Marlin_main.cpp
gudnimg: @DRracer fixed :)

leptun:
@gudnimg Is the arithmetic correct now? Please also fix the merge conflict

gudnimg:
> @gudnimg Is the arithmetic correct now? Please also fix the merge conflict

@leptun Yeb it should be correct now. The Arduino sketch in my comment was enough to verify it. (Had an Arduino Mega /ATmega2560 to test on)

leptun (approved):

leptun:
I don't think the stats are correct. In my testing, it only saved 226B of flash and no SRAM improvements at all. Maybe it has something to do with avr-gcc 4.9.2 vs avr-gcc 7.3.0

gudnimg:
Current stats with my setup on Arduino IDE 1.8.19:

Before:
```
Sketch uses 227356 bytes (89%) of program storage space. Maximum is 253952 bytes.
Global variables use 5662 bytes of dynamic memory.
```

After:
```
Sketch uses 227044 bytes (89%) of program storage space. Maximum is 253952 bytes.
Global variables use 5662 bytes of dynamic memory.
```

I only see 312 byte of flash memory savings and no SRAM difference. I will update the PR description.

EDIT: Just for reference, I use `LANG_MODE = 0`, with `LANG_MODE = 1` I see 318 byte difference.

leptun:
@gudnimg Wow. With arduino 1.8.19 the code is almost 1KB smaller than with 1.8.5. Maybe we should switch sooner than later. @DRracer @3d-gussner fyi

As for 1.8.13 vs 1.8.19, the code size is identical for me. So maybe we could update to 1.8.19 directly so we are on the newest version

DRracer (approved):
looks correct to me now

PR #3208, opened 2021-06-25
hwtim atomic operations

leptun:
All around safer operations regarding timers by making all timer changes atomic. Might fix a few existent or non-existent race conditions regarding the stepper interrupt.


Firmware/tone04.c
```diff
@@ -41,20 +41,24 @@ void timer4_set_fan0(uint8_t duty)
...
 	{
 		// We use digital logic if the duty cycle is 0% or 100%
-		TCCR4A &= ~(1 << COM4C1);
-		OCR4C = 0;
-		WRITE(EXTRUDER_0_AUTO_FAN_PIN, duty);
+		ATOMIC_BLOCK(ATOMIC_RESTORESTATE)
+		{
+			TCCR4A &= ~(1 << COM4C1);
+			OCR4C = 0;
+			WRITE(EXTRUDER_0_AUTO_FAN_PIN, duty);
```
DRracer: the EXTRUDER_0_AUTO_FAN_PIN update can be moved out of the critical section

Firmware/xyzcal.cpp
```diff
@@ -13,8 +13,8 @@
 #define XYZCAL_PINDA_HYST_MAX 100 //250um
 #define XYZCAL_PINDA_HYST_DIF 5   //12.5um
 
-#define ENABLE_FANCHECK_INTERRUPT()  EIMSK |= (1<<7)
-#define DISABLE_FANCHECK_INTERRUPT() EIMSK &= ~(1<<7)
+#define ENABLE_FANCHECK_INTERRUPT() ATOMIC_BLOCK(ATOMIC_RESTORESTATE) EIMSK |= _BV(INT7)
+#define DISABLE_FANCHECK_INTERRUPT() ATOMIC_BLOCK(ATOMIC_RESTORESTATE) EIMSK &= ~_BV(INT7)
```
DRracer, line 17: if this expands into a single instruction (hypothetically), we don't have to do the critical section

leptun: It doesn't. It expands to 3 instructions (IN, OR/AND, OUT). Only IO registers from 0x00 to 0x1F are atomic RMW.

Firmware/tone04.c
leptun: Yes. I will move it now

Firmware/stepper.cpp
```diff
@@ -1301,62 +1301,67 @@ void st_init()
...
+    // Set the timer pre-scaler
+    // Generally we use a divider of 8, resulting in a 2MHz timer
+    // frequency on a 16MHz MCU. If you are going to change this, be
+    // sure to regenerate speed_lookuptable.h with
+    // create_speed_lookuptable.py
+    TCCR1B = (TCCR1B & ~(0x07<<CS10)) | (2<<CS10);
+
+    // Plan the first interrupt after 8ms from now.
+    OCR1A = 0x4000;
+    TCNT1 = 0;
```
wavexx, line 1326: I need to re-verify this, but we should be able to close the atomic brace here, like we do in st_reset_timer.

Firmware/tone04.c
```diff
@@ -41,20 +41,24 @@ void timer4_set_fan0(uint8_t duty)
...
-		CRITICAL_SECTION_START;
-		// Enable the PWM output on the fan pin.
-		TCCR4A |= (1 << COM4C1);
-		OCR4C = (((uint32_t)duty) * ((uint32_t)((TIMSK4 & (1 << OCIE4A))?OCR4A:255))) / ((uint32_t)255);
-		CRITICAL_SECTION_END;
+		ATOMIC_BLOCK(ATOMIC_RESTORESTATE)
+		{
+			// Enable the PWM output on the fan pin.
+			TCCR4A |= _BV(COM4C1);
+			OCR4C = (((uint32_t)duty) * ((uint32_t)((TIMSK4 & (1 << OCIE4A))?OCR4A:255))) / ((uint32_t)255);
```
wavexx, line 60: This is the only part I've seen where the value is not constant, so I would compute the value outside the critical section, then just do the assignment here.

Firmware/stepper.cpp
leptun: It scares me that planner variables are still initialized after the logic analyzer stuff. Doesn't seem healthy to start the stepper ISR with these values not initialized.

Firmware/tone04.c
leptun: Yes, that's a good idea, but I depend on a bit in TIMSK4. I'm afraid of checking it non-atomically, although that could be safe. Please keep in mind that this function is called extremely rarely, so I wouldn't worry that much about the atomic operation delay.

wavexx: is it a problem even if the time between read/write is not atomic? I just worry that this is multiply+divide of a 32bit value, it's quite some cycles even if that's called rarely.

Firmware/stepper.cpp
wavexx: I would move the whole LIN_ADVANCE block just before the timer initialization then, so that we can keep the atomic block just for the timer setup.

Firmware/tone04.c
leptun: ok. I'll do the math separately an just write atomically

Firmware/stepper.cpp
leptun: Ok. will do that. Just not right now

PR #3209, opened 2021-06-26
Remove unused `Sound_Save()` function declaration

gudnimg:
I noticed there is no defined function called `Sound_Save()`

### DIFF
--- a/Firmware/sound.h
+++ b/Firmware/sound.h
@@ -19,7 +19,6 @@ extern eSOUND_MODE eSoundMode;
 
 extern void Sound_Init(void);
 extern void Sound_Default(void);
-extern void Sound_Save(void);
 extern void Sound_CycleState(void);
 extern void Sound_MakeSound(eSOUND_TYPE eSoundType);
 extern void Sound_MakeCustom(uint16_t ms,uint16_t tone_ ,bool critical);


gudnimg:
Why are all the functions `extern`? I noticed this in other files also like `ultralcd.h`, it looks like we can drop the `extern` keyword everywhere... am I missing something?

leptun:
Is there any difference between c and c++ when it comes to the `extern` keyword in headers? I mean, will c break if we remove the `extern`?

gudnimg:
> Is there any difference between c and c++ when it comes to the `extern` keyword in headers? I mean, will c break if we remove the `extern`?

There is no difference I’m aware of between C and C++ when it comes to the `extern` keyword. 

`extern` is usually used only for variables to make them available to other files as global. When the function declaration is included with a header file then the `extern` keyword (on a function) is not needed since external linkage is the default behaviour. So long as we have all the right `#include`’s I don’t see how removing `extern` can break things. 🤔 

But if there is no header file between two `.cpp` files then it is my understanding that `extern` can be used but I think that should be avoided if possible.

wavexx:
The default has always been "extern" on gcc. There's no difference between c/c++ in this regard, except when combined with the "inline" keyword, which is a weird thing to do anyway. And const globals in c++ if I'm not mistaken, which is not an issue normally.

You can change the default using -fvisibility or some pragma, so that all functions/classes behave as "static", and you explicitly need to use extern instead to make the function visible outside the current module.

I think extern was probably used by some with experience with some other compiler/os.

DRracer (approved):
Confirmed, there is no implementation for Sound_Save(), it makes sense to remove this relict of the past.

PR #3214, opened 2021-06-29
Print temperatures only if filament loading is still active

Panayiotis-git:
A fix to a small glitch I've found.

Test case:
1. Disable filament auto-load
2. Reset printer
3. Select to Load filament and choose a material
4. During the initial Z axis raise, press the button to Cancel
5. The printer returns to the previous menu
6. After the Z axis raising has finished, the printer although inside the menu, it displays at the top row of the screen the hotend temperature.

With this fix, at the step 6, the printer skips the printing of the temperature.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -2317,8 +2317,10 @@ void mFilamentItem(uint16_t nTemp, uint16_t nTempBed)
             }
         }
 
-        lcd_set_cursor(0, 0);
-        lcdui_print_temp(LCD_STR_THERMOMETER[0], (int) degHotend(0), (int) degTargetHotend(0));
+        if (bFilamentWaitingFlag) {
+            lcd_set_cursor(0, 0);
+            lcdui_print_temp(LCD_STR_THERMOMETER[0], (int) degHotend(0), (int) degTargetHotend(0));
+        }
 
         if (lcd_clicked())
         {


Panayiotis-git:
@DRracer I think that this fix should be included in the release 3.10.1. The issue does not exist at the release 3.10. It was introduced afterwards.

DRracer:
@Panayiotis-git could you please rebase this PR onto MK3_3.10.1? Or cherry pick against MK3_3.10.1? Thank you.
A simple change of bases in github web interface would have introduced way too may changes.

3d-gussner (approved):
It is working as described.

Here some tests I did:
- Before this PR https://youtu.be/O3CqFaO7xPk
- With this PR https://youtu.be/o54S7BDIZ7Q

There is still a small issue that `Tune` stays until you turn the knob see https://youtu.be/gmjLBm1vxcA

DRracer:
@Panayiotis-git @3d-gussner thank you, approving and merging #3303

Panayiotis-git:
>There is still a small issue that Tune stays until you turn the knob see https://youtu.be/gmjLBm1vxcA

You are right about this issue @3d-gussner, but I think it is a more generic issue and has to do with how the menu is redrawn. I've seen this happening at other points too. I would not even call it an issue, rather a known behaviour.

DRracer:
I've removed the 3.10.1 milestone tag from this PR in favour of #3303 which has been merged instead.
@Panayiotis-git please consider closing this PR, the mk3 branch will get all the changes from MK3_3.10.1 once the FW gets released (so your contribution will not be forgotten :wink: )

PR #3215, opened 2021-06-30
PFW-1297: lang-add.sh improvements

wavexx:
This PR adds a few improvements to the ``lang/lang-add.sh`` script so that it can, in a good amount of cases, extract the string metadata directly from the source files.

This is currently **absolutely crude**, but it will recognize translatable strings (by using the macros _i/ISTR) and fetch the associated data when the comment at the end of the same line is using the syntax ``////ID c=columns r=rows``` (to be more exact, anything following the four slashes really).

Not only this eases the pain when adding new strings, but we can use this to re-verify/update the string data by doing the following:

- zeroing lang_en.txt (``echo > lang_en.txt``) and lang_en_* files
- running ``fw-build.sh`` to update the ``not_tran.txt`` file with **all** the strings
- copy ``not_tran.txt`` to ``lang_add.txt``
- run ``lang-add.sh`` to regenerate ``lang_en.txt`` with the string data copied from the source files.
- diff each against the git version (``git diff lang_en.txt``)

It's still bad, but not *as* bad as the "spot the difference" game as done currently.
There are _many_ changes to be updated... so marking this as a draft.

@3d-gussner

### DIFF
--- a/lang/lang-add.sh
+++ b/lang/lang-add.sh
@@ -26,6 +26,7 @@ echo "fw-clean languages:$LANGUAGES" >&2
 
 # insert single text to english dictionary
 # $1 - text to insert
+# $2 - metadata
 insert_en()
 {
 	#replace '[' and ']' in string with '\[' and '\]'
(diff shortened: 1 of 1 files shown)


lang/lang-add.sh
```diff
@@ -7,7 +7,12 @@
 #  lang_add.txt
 # Updated files:
 #  lang_en.txt and all lang_en_xx.txt
-#
+
+# List of supported languages
+LANGUAGES="cz de es fr it pl"
+
+# Community languages
+LANGUAGES+=" nl" #Dutch
```
3d-gussner, line 15: I have changed that a bit see https://github.com/prusa3d/Prusa-Firmware/blame/MK3_3.11.0/lang/lang-import.sh#L45-L59

3d-gussner (approved):

PR #3245, opened 2021-07-19
Change `nrFiles` from `int16_t` to `uint16_t`

gudnimg:
Change `nrFiles` from `int16_t` to `uint16_t`

It is never given a negative value and always used as an `uint16_t`. The variable can have twice the value now.

I propose this tiny change merely to keep things consistant. 👍


DRracer (approved):
I also didn't find any situation where a negative value of nrFiles was needed/used, approving

PR #3247, opened 2021-07-21
Remove unused ifdef `MK1BP` and `MAX_SILENT_FEEDRATE`

gudnimg:
`MK1BP` is never defined anywhere. So I think it is a good idea to remove it as cleanup.


gudnimg:
Added another commit which removes another old ifdef `MAX_SILENT_FEEDRATE`. It was added by this commit https://github.com/prusa3d/Prusa-Firmware/commit/e09f0c64c792e869364679bddba73f6aac133206#diff-a2e2d6e9dab1e37ea9a2882433e2b8f70af751b3c3445d5e997d11a93d4b28b3 in July 3rd 2017.

It was no longer used after November 7th 2017: https://github.com/prusa3d/Prusa-Firmware/blob/1e81861e0c9e52487054bf8bcc48aa5ac6f9a474/Firmware/variants/1_75mm_MK3-EINY10a-E3Dv6full.h

Thought it would be easier to add the commit here than to create another PR

leptun (approved):

DRracer (approved):

PR #3251, opened 2021-07-26
Optimize `farm_mode` initialisation in `setup()`

gudnimg:
This saves 28 bytes of flash memory. It looks like to me `farm_mode` is initialised twice in the `setup()` function. I also added an `eeprom_update_byte()` call in case the EEPROM memory has not been set. That way if `farm_mode` is never activated, this is only called on the first boot-up instead of on every boot-up.

Before:
Sketch uses **231746** bytes (91%) of program storage space. Maximum is 253952 bytes.
Global variables use **6103** bytes of dynamic memory.

After:
Sketch uses **231718** bytes (91%) of program storage space. Maximum is 253952 bytes.
Global variables use **6103** bytes of dynamic memory.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -1101,8 +1101,10 @@ void setup()
 	setup_powerhold();
 
 	farm_mode = eeprom_read_byte((uint8_t*)EEPROM_FARM_MODE); 
-	if (farm_mode == 0xFF) 
+	if (farm_mode == 0xFF) {
 		farm_mode = false; //if farm_mode has not been stored to eeprom yet and farm number is set to zero or EEPROM is fresh, deactivate farm mode
+  	eeprom_update_byte((uint8_t*)EEPROM_FARM_MODE, farm_mode);
+  }
 	if (farm_mode)
 	{
 		no_response = true; //we need confirmation by recieving PRUSA thx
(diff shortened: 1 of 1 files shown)


leptun (approved):
This looks correct to me. Haven't tested it however

vintagepc (approved):
Looks good to me

Firmware/Marlin_main.cpp
```diff
@@ -1440,13 +1440,6 @@ void setup()
     enable_z();
 #endif
 
-	farm_mode = eeprom_read_byte((uint8_t*)EEPROM_FARM_MODE);
-	if (farm_mode == 0xFF) farm_mode = false; //if farm_mode has not been stored to eeprom yet and farm number is set to zero or EEPROM is fresh, deactivate farm mode
-	if (farm_mode)
-	{
-		prusa_statistics(8);
```
DRracer, line 1447: Removing the call to `prusa_statistics(8);` from this spot may result in problems with remote monitoring of the printer at the farm. The farm monitoring SW may accidentally expect 2 messages of "printer started" to consider a printer working.

`prusa_statistics(8);` sends "printer started", so it may be better to keep the call at this spot.

Firmware/Marlin_main.cpp
```diff
@@ -1116,9 +1116,9 @@ void setup()
 		//disabled filament autoload (PFW360)
 		fsensor_autoload_set(false);
 #endif //FILAMENT_SENSOR
-          // ~ FanCheck -> on
-          if(!(eeprom_read_byte((uint8_t*)EEPROM_FAN_CHECK_ENABLED)))
-               eeprom_update_byte((unsigned char *)EEPROM_FAN_CHECK_ENABLED,true);
+		// ~ FanCheck -> on
+		if(!(eeprom_read_byte((uint8_t*)EEPROM_FAN_CHECK_ENABLED)))
```
DRracer: May be, when you already touched these lines, setting the `EEPROM_FAN_CHECK_ENABLED` to true always in farm mode may save a few bytes as well.
`eeprom_update_byte` checks the EEPROM cell's content and wites a new value only if the previous was different.

DRracer (changes requested):
Please address my comments in the code.

Firmware/Marlin_main.cpp
gudnimg: @DRracer I will add this back. Is the comment below OK? A comment explaining the second call to `prusa_statistics(8)` is better than no explanation :)

```c
if (farm_mode) {
    // The farm monitoring SW may accidentally expect 
    // 2 messages of "printer started" to consider a printer working.
    prusa_statistics(8);
}
```

Firmware/Marlin_main.cpp
gudnimg: It saves 12 bytes of flash 👍

leptun (approved):
The addressed changes look correct to me. Approving again

DRracer (approved):
:+1:

PR #3257, opened 2021-07-31
Remove redundant operations on LCD CGRAM

gudnimg:
This Pull Request reduces the number of reads/writes to the LCD's CGRAM (Character Generator RAM). 

----

### An overview of how the FW uses CGRAM

There can only be 8 special characters stored into the CGRAM at any given time. In 99% of cases these 8 symbols are defined by Prusa firmware:

| Special Character  | What does the character represent? |
| --------------------| --------------------------------------- |
| 0                           | 🄷 Bed Temperature                              |
| 1                           | ° Degree symbol                                 |
| 2                           | 🌡 Thermometer (for extruder temperature) |
| 3                           | ⬏ Upwards arrow to back up a menu  |
| 4                           | 🗘 Refresh symbol                                |
| 5                           | 🗀 Folder icon for SD card                    |
| 6                           | » Feedrate symbol                           |
| 7                           |🕑Clock symbol                                   |

When the firmware needs to use other special characters than defined above, we need to change the CGRAM table. Currently there are only two cases when this is needed:

1. When the LCD shows a full screen message
2. When the LCD shows a multi-screen message. 

This is done by calling  `lcd_set_custom_characters_nextpage()` which replaces Special Characters 1 and 2 such that the CGRAM table looks like below:

| Special Character  | What does the character represent? |
| --------------------| --------------------------------------- |
| 0                           | 🄷 Bed Temperature                              |
| **1**                     | **↡  Double arrow pointing down (multi-screen message)**                              |
| **2**                     | **✔Confirm symbol when at the bottom of a multi-screen message** |
| 3                           | ⬏ Upwards arrow to back up a menu  |
| 4                           | 🗘 Refresh symbol                                |
| 5                           | 🗀 Folder icon for SD card                    |
| 6                           | » Feedrate symbol                               |
| 7                           |🕑Clock symbol                                   |

Then when we exit the full-screen or multi-screen message we need to revert this change by writing back into CGRAM the Degree and Thermometer symbols. This is done by calling `lcd_set_custom_characters()`.

Each call to `lcd_set_custom_characters_nextpage()` must also have it's respective pair `lcd_set_custom_characters()`.

----

#### Function `lcd_set_custom_characters_degree()` is removed
 
This function only stores the Degree symbol (Special Character 1) back into CGRAM. But this is not good enough. Because if `lcd_set_custom_characters_nextpage()`  was called, then the Thermometer symbol (Special Character 2) will be missing.  It is sufficient to call `lcd_set_custom_characters()` instead to ensure both symbols are present after a full-screen or multi-screen message.

----

#### Function `lcd_set_custom_characters_arrows()` is removed: 
This function replaces the Degree symbol (Special Character 1) with an arrow pointing downwards (not the same as one used for the multi-line messages) but this symbol is never actually used. Thus this function is currently wasting CPU cycles whenever it is called. 

----

#### Removed `lcd_charsetup_func` and `menu_lcd_charsetup_func`
The firmware needs to somehow call `lcd_set_custom_characters()` when the printer is booted up. This is done by `lcd_refresh()` in `ultralcd_init()`. The function `menu_lcd_charsetup_func` calls `lcd_set_custom_characters_degree()` which as I explained above is not needed.

----

#### Other changes

* Removed two unused symbols: `lcd_chardata_arrup` and `lcd_chardata_arrdown`
* Removed calls to `lcd_set_custom_characters()` where there was no call to `lcd_set_custom_characters_nextpage()` beforehand. As described above, then there is no need to touch the CGRAM table.

I used the following datasheet for reference
[RK-10290_410.pdf](https://github.com/prusa3d/Prusa-Firmware/files/6910346/RK-10290_410.pdf)
which states that each byte we write to the LCD takes around 29-53μs.

-----

#### Memory
The current changes reduce flash usage by 206 bytes and SRAM by 6 bytes.


leptun:
https://github.com/prusa3d/Prusa-Firmware/pull/2064/files
Take a look at this PR. There I’ve improved also the custom character handling. There are many ways that PR can be made more efficient. Maybe you can help me with that :)

gudnimg:
> https://github.com/prusa3d/Prusa-Firmware/pull/2064/files
> Take a look at this PR. There I’ve improved also the custom character handling. There are many ways that PR can be made more efficient. Maybe you can help me with that :)

I will take a look at it. I see some of the removals I have in this PR are in yours. I'm thinking it would be easier to continue with my PR right? Since I am only removing stuff and not adding anything. Might help slightly with the review of your large PR :P

gudnimg:
Huh... my PR changes worked yesterday... but now it is clear I need to investigate more 😆 
image

**Edit:** Never mind... I forgot to put `LANG_MODE` to `0`. I wouldn't  have expected a crash though... probably not related to this PR.

leptun:
@gudnimg 
This is somewhat of a proof of concept at the moment. It works, but it could be much more efficient and in some parts rewritten entirely.
The basic idea is this: instead of writing to the lcd directly, the ultralcd code instead pushes the characters to a mirror buffer of the display. It also updates some flags in a bitmap so that the ISR can know which characters got updated and which stayed the same. In a best case scenario, the ISR does all the talking to the lcd, including the longer sequences such as the init and character map updating, but I never got around to implementing this part.
It also improves another aspect: the two nibbles don’t actually need the regular delay between them. They can be pushed almost instantaneously and only after that you need the 100us delay.
The idea behind the custom character banking is this: insead of switching the banks automatically, let the lcd code switch them automatically as needed. To differentiate the two different banks, the “second set” of custom characters are used (from 0x08 to 0x0F). From the perspective of the lcd, these characters are the same as <0x00-0x07>, but from the perspective of the code, we can actually know which bank is needed for the character. A single byte is used for storing which bank is currently loaded on the lcd, so each bit in this byte represents the selection of bank0 or bank2 for that slot. The bank checking and switching is done in the function that takes characters and where \n is handled.
As a side note, since \n overlaps one slot in bank1, it is reserved and can’t be used, so only 15 custom characters can really be defined.

There were some plans to get this PR ready for 3.11, but I doubt that’s going to happen. If you want, you can try making a more efficient implementation of this, especially when it comes to the ISR and the mirror buffer. I for one don’t really want to touch this code again after all the headaches it’s given me 😁

gudnimg:
Marking as ready for review. I've been printing most of the weekend and testing the menus. I have as of yet not seen any issues. I don't want to add more changes to this PR so it is easier for maintainers to review it.

I believe this PR is a good step in the right direction for more changes in the future. 👍

gudnimg:
I've updated the description. Hopefully it is more clear to read.

leptun:
The refresh symbol is in fact used. Wdym "unused"? See `MSG_REFRESH`.

Btw, there are more palces in the firmware which hardcode the lcd characters. We should find those and use the correct define instead so that we can easily find them

gudnimg:
> The refresh symbol is in fact used. Wdym "unused"? See `MSG_REFRESH`.
> 
> Btw, there are more palces in the firmware which hardcode the lcd characters. We should find those and use the correct define instead so that we can easily find them

Thanks, I didn't see it when searching through the firmware.

gudnimg:
@leptun 

Changing

`lcd_printf_P(_N("%3.2f\x01"), _deg(angleDiff));`

To this:

`lcd_printf_P(_N("%3.2f%c"), _deg(angleDiff), LCD_STR_DEGREE[0]);`

Adds 12 bytes of flash on my end

leptun:
Try: `lcd_printf_P(_N("%3.2f" LCD_STR_DEGREE), _deg(angleDiff));`

gudnimg:
> Try: lcd_printf_P(_N("%3.2f" LCD_STR_DEGREE), _deg(angleDiff));

I get the same result. That is 12 byte increase.

leptun:
Are you sure? That changes nothing compared to the original. It literally concatenates the strings before putting them to progmem, so it should result in the same string in progmem. Try cleaning the build dir before building again

gudnimg:
That seems to do the trick! :)

gudnimg:
@leptun Seems I can't include `lcd.h` into `messages.c` due to `lcd.h` using C++ specific stuff like `class`. Is there any reason to keep `messages.c` as a C file? The compile works fine if I change `messages.c` to `messages.cpp`.

leptun:
I assume this is what saves 64B? `chars = lcd_puts_P(_N(LCD_STR_CLOCK "--:-- "));`

> Is there any reason to keep messages.c as a C file

I see no problem in changing that to .cpp. After all, there are only variables there, so it should make no difference.
I think if you do that you can remove also this since c++ has bool, true and false:
```
//this is because we need include Configuration_prusa.h (CUSTOM_MENDEL_NAME)
#define bool char
#define true 1
#define false 0
```
Also, no clue why the didn't `#include <stdbool.h>` instead

gudnimg:
```c
#define bool char
#define true 1
#define false 0
```
I didn't see these definitions being used anywhere in messages or in the included configuration file. So I just removed them.

All the hardcoded hex values should now be using defines from lcd.h

leptun:
> I didn't see these definitions being used anywhere in messages or in the included configuration file. So I just removed them.

I also can't see any reason for that.

> All the hardcoded hex values should now be using defines from lcd.h

I don't think that's true. Search for `lcd_print` and you'll see at least one '\xFF' being printed. Add `#define LCD_STR_SOLID_BLOCK "\xFF" //from the default character set` to lcd.h and change the hardcoded values accordingly

leptun:
@gudnimg Please rebase this branch onto the MK3 branch so that the branch happens at a more recent commit

gudnimg:
@leptun it should now be in sync with MK3 branch

leptun:
I've reverted a commit that breaks the status screen. Please leave it like this for now. I have plans for it for later where this change will be applied again

leptun:
@gudnimg I've reverted yet another commit unfortunately. This one was reverted because the space character had a purpose. It fills the last column of the lcd. Otherwise it never gets set/updated. It might look right, but I'd still prefer if it was like before.

leptun:
I've also fixed the merge conflict.
Btw, it's very important that we don't forget about https://github.com/prusa3d/Prusa-Firmware/commit/e62bc4686f452f3222860146d28b7769b06200fd. It contains some mandatory hardcoded->define transitions that I need for a prototype I'm working on. Here's a sneak peek. Please note that the font was changed a bit in the meantime, so it looks better now.
image

Firmware/ultralcd.cpp
```diff
@@ -3246,33 +3245,30 @@ bool lcd_wait_for_pinda(float temp) {
 		lcd_print(ftostr3(current_temperature_pinda));
 		lcd_print('/');
 		lcd_print(ftostr3(temp));
-		lcd_print(LCD_STR_DEGREE);
+		lcd_print(LCD_STR_DEGREE[0]);
```
DRracer, line 3248: :+1: even though it is not completely obvious at the first look, forcing a call to `lcd_print` with just a character argument (instead of an array of chars = string) makes the compiler generate better code.

DRracer (approved):
nice :+1: ... and very well documented

Issue #3262, opened 2021-08-09
[BUG] After inserting SD-card, the file list is sometimes automatically dismissed


wavexx:
**Printer type** - MK3+
**Printer firmware version**-  3.10.1-RC1

**MMU Upgrade** - No
**MMU upgrade firmware version** - N/A

**SD card or USB/Octoprint**
SD card

**Describe the bug**
When inserting an SD card, the file list is displayed for a fraction of a second, but then it returns immediately to the status screen before you have time to do anything.

This is a new issue that got introduced recently (3.10.1-RC1 for sure, but could be 3.10.0). Can't determine exactly, since it's somewhat rare and doesn't reproduce consistently at all.

IMHO this happens due to an imperfect SD contact while being sled-in, *probably*. But we detect this late, and dismiss the menu. 
Or maybe we simply forgot to reset the "return to status" timer after sorting?

I never worked too closely to the SD code, but maybe there's a flag we should reset? @leptun 

Because it looks like the file list is perfect so *clearly* we could read it just fine. If I return to the SD menu manually, everything is working - so SD hasn't been removed either. I'm certain the contacts are clean, make good contact as well, and do not disconnect even with vibrations.

**To Reproduce**
Common way to reproduce (reproduces 1 time in ~20 tries):

* Remove SD card
* Reset printer
*  Wait ~10 seconds
* Insert SD card, not too fast
* File sorting starts, then shows file list
* Less than a second later, the menu automatically returns to the status screen
  
**Expected behavior**
After inserting SD card, we shouldn't go back to the status screen immediately.
  
**G-code**
N/A

**Video**
Couldn't record it (yet)

wavexx:
Also pinging @gudnimg - did you notice this happening to you?

gudnimg:
> Also pinging @gudnimg - did you notice this happening to you?

No I haven’t noticed it before. But I will try to reproduce it later today after work.

just for clarification:
> Reset printer

Do you mean to turn the printer on and off. Or factory reset? Or are you referencing the reset button on the PCB?

Panayiotis-git:
It is happening for me always at the first SD card insertion, after a reset or power on.

gudnimg:
I can't seem to be able to reproduce this. Though I would expect the issue to be somehow in this function `menu_lcd_lcdupdate_func()`

Edit: Spoke too soon, managed to reproduce it one time now. Will look at this a bit more.

gudnimg:
[...]

PR #3268, opened 2021-08-21
Remove one call to `lcd_timeoutToStatus.start`

gudnimg:
While looking into #3262 I noticed that `lcd_timeoutToStatus.start()` is called multiple times in `lcd_buttons_update()` before finally being called once in `menu_lcd_lcdupdate_func()`. (See image below of serial stream)

I noticed this because I added a serial message everywhere `lcd_timeoutToStatus` is used just to get a better picture of when it is started and stopped.

`lcd_timeoutToStatus.start()` is called from `menu_lcd_lcdupdate_func()` once when the knob is clicked. The change reduces flash memory usage by 8 bytes. This change affects the temperature ISR because it calls `lcd_buttons_update()`.

I have not found any issues so far on my printer.

image

### DIFF
--- a/Firmware/lcd.cpp
+++ b/Firmware/lcd.cpp
@@ -741,7 +741,6 @@ void lcd_buttons_update(void)
 
     if (READ(BTN_ENC) == 0)
     { //button is pressed
-        lcd_timeoutToStatus.start();
         if (!buttonBlanking.running() || buttonBlanking.expired(BUTTON_BLANKING_TIME)) {
             buttonBlanking.start();
             safetyTimer.start();


wavexx:
I suspect this might be intentional. The idea is that if you rotate the knob the timeout is extended.

gudnimg:
> I suspect this might be intentional. The idea is that if you rotate the knob the timeout is extended.

In my testing, this line takes care of the rotation: https://github.com/prusa3d/Prusa-Firmware/blob/400f673fe0b898d0e3ec8c43fd2011f9320cf20c/Firmware/ultralcd.cpp#L8881

The call this PR removes is only executed when the knob is clicked/pushed. And for that, this line seems to take care of it: https://github.com/prusa3d/Prusa-Firmware/blob/400f673fe0b898d0e3ec8c43fd2011f9320cf20c/Firmware/ultralcd.cpp#L8887

wavexx:
You're right. Then indeed looks good to me.

leptun (approved):
After seeing the following comment, it looks correct to me as well
https://github.com/prusa3d/Prusa-Firmware/pull/3268#issuecomment-903145234

vintagepc (approved):

DRracer (approved):

PR #3269, opened 2021-08-21
Fix issue when the SD card is inserted and the file menu may sometimes close immediately due to a timeout

gudnimg:
In this PR I propose to make sure `lcd_timeoutToStatus.start()` is called in the specific case when the SD card is inserted into the printer after it was removed beforehand.

### DIFF
--- a/Firmware/cardreader.cpp
+++ b/Firmware/cardreader.cpp
@@ -1030,7 +1030,6 @@ void CardReader::presort() {
 
 	lcd_update(2);
 	KEEPALIVE_STATE(NOT_BUSY);
-	lcd_timeoutToStatus.start();
 }
 
 void CardReader::flush_presort() {
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -8858,6 +8858,7 @@ void menu_lcd_lcdupdate_func(void)
 			LCD_MESSAGERPGM(_T(WELCOME_MSG));
 			bMain=false;                       // flag (i.e. 'fake parameter') for 'lcd_sdcard_menu()' function
 			menu_submenu(lcd_sdcard_menu);
+			lcd_timeoutToStatus.start();
 		}
 		else
 		{


wavexx:
IMHO is this also more logical as well. The timeout should start when any menu is entered, not before.
I'm surprised the timeout is not automatically reset when entering _any_ menu instead of having to reset the timeout by hand...

wavexx:
The big question is whether this fixes the issue ;). I tried to reduce the lcd_timeoutToStatus to smaller values to trigger the problem, but I couldn't make it happen :/

gudnimg:
I agree, it looks to me that 99% of cases when entering a menu are covered by the user clicking the knob. But for this menu there is no rotation or click 🤔

wavexx:
Tested this so far, on a couple of printers, never had the issue. IMHO the change is good irregardless, I would remove the "draft"

leptun (approved):
Makes sense to me as well

DRracer (approved):

DRracer (approved):

DRracer:
@gudnimg could you please rebase this PR onto MK3_3.10.1? Or cherry pick against MK3_3.10.1? Thank you.
A simple change of bases in github web interface would have introduced way too may changes.

gudnimg:
> @gudnimg could you please rebase this PR onto MK3_3.10.1? Or cherry pick against MK3_3.10.1? Thank you.
> A simple change of bases in github web interface would have introduced way too may changes.

I’ll take a look at it tonight :)

3d-gussner:
Here some test I did:
- Before this PR https://youtu.be/WYpBOBojLiQ watch at the 2nd `Remount` ... the Folder/File list disappears quite fast
- With this PR is working fine with multiple `Unmout/Remounts` https://youtu.be/PTgNOyCf650

gudnimg:
I created a new PR with cherry pick here https://github.com/prusa3d/Prusa-Firmware/pull/3305

DRracer:
@gudnimg great, merged #3305 into 3.10.1

PR #3271, opened 2021-08-22
Fix typos & small grammar changes

dweekly:
Applied some gentle love to make this read better to a native English speaker. No meaningful change in content.


DRracer (approved):
Thank you for polishing the text a bit :+1:

PR #3292, opened 2021-10-12
Fix build error if `PINDA_THERMISTOR` is not defined

gudnimg:
`temp_cal_active` variable is not defined in this scope.

Also using `#ifndef PINDA_THERMISTOR` for these lines of code does not make sense to me. Is this dead code perhaps? 

I fixed the build failure in case it helps later on.


leptun (approved):
Looks correct to me. `temp_cal_active` doesn't exist anymore. `EEPROM_TEMP_CAL_ACTIVE` flag is used instead

DRracer (approved):

PR #3307, opened 2021-11-25
Added extruder current to avoid skipping problems

JR-Y:


DRracer:
@JR-Y I'm afraid we can't merge your PR. Increased current automatically means increase of motor's temperature, which in case of MK3/S/+ would have been deadly for PLA prints. We are already on the limit when the ambient temperature is high.

devsnowy:
What about increasing it temporarily would that be possible?

3d-gussner:
@JR-Y Thanks, but extra current means 2 things -> overheating the motor + grinding the filament. Closing PR

PR #3310, opened 2021-12-07
My solution to solve the watchdog issues with custom sound generation

bmuessig:
As discussed issue #2045, long sound durations often cause the watchdog timer to reset the printer, as the sound generation is currently using regular delays that don't feed the watchdog timer.
I am proposing a hybrid solution that uses regular delays for short beeps and a modified version of `delay_keep_alive` that doesn't update the LCD for long delays that exceed 1.1 seconds.

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -391,7 +391,9 @@ extern void calculate_extruder_multipliers();
 
 // Similar to the default Arduino delay function, 
 // but it keeps the background tasks running.
-extern void delay_keep_alive(unsigned int ms);
+void _delay_keep_alive(unsigned int ms, bool do_update_lcd);
+#define delay_keep_alive(ms) _delay_keep_alive((ms), true)
+#define delay_keep_alive_headless(ms) _delay_keep_alive((ms), false)
 
 extern void check_babystep();
 
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -391,9 +391,7 @@ extern void calculate_extruder_multipliers();
 
 // Similar to the default Arduino delay function, 
 // but it keeps the background tasks running.
-void _delay_keep_alive(unsigned int ms, bool do_update_lcd);
-#define delay_keep_alive(ms) _delay_keep_alive((ms), true)
-#define delay_keep_alive_headless(ms) _delay_keep_alive((ms), false)
+extern void delay_keep_alive(unsigned int ms, bool do_update_lcd = true);
 
 extern void check_babystep();
 
--- a/Firmware/sound.cpp
+++ b/Firmware/sound.cpp
@@ -68,7 +68,7 @@ void Sound_Delay(uint16_t ms)
     if (ms < (SOUND_KEEPALIVE_MIN_MS))
         _delay(ms);
     else
-        delay_keep_alive_headless(ms);
+        delay_keep_alive(ms, false);
 }
 
 // Makes a custom sound
--- a/Firmware/sound.cpp
+++ b/Firmware/sound.cpp
@@ -13,13 +13,13 @@
 // !?! eSOUND_MODE eSoundMode; v ultraldc.cpp :: cd_settings_menu() se takto jevi jako lokalni promenna
 eSOUND_MODE eSoundMode; //=e_SOUND_MODE_DEFAULT;
 
-
 static void Sound_SaveMode(void);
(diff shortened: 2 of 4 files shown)


bmuessig:
Having hit PR number 3310 in a sound-related PR, the classic Nokia 3310 ringtone came to mind: https://www.youtube.com/watch?v=h6iIFVTwKeE

Firmware/Marlin.h
```diff
@@ -391,7 +391,9 @@ extern void calculate_extruder_multipliers();
 
 // Similar to the default Arduino delay function, 
 // but it keeps the background tasks running.
-extern void delay_keep_alive(unsigned int ms);
+void _delay_keep_alive(unsigned int ms, bool do_update_lcd);
```
leptun: Function default arguments from c++ can be used here instead of macros.
See https://en.cppreference.com/w/cpp/language/default_arguments

Firmware/sound.cpp
```diff
@@ -62,36 +62,38 @@ switch(eSoundMode)
 Sound_SaveMode();
 }
 
-//if critical is true then silend and once mode is ignored
-void Sound_MakeCustom(uint16_t ms,uint16_t tone_,bool critical){
+// Waits for the duration and uses the most appropriate delay type
+void Sound_Delay(uint16_t ms)
+{
+    if (ms < (SOUND_KEEPALIVE_MIN_MS))
```
leptun, line 68: Not sure if doing something like this is really needed. You could just use delay_keep_alive(noUIupdates) regardless of duration. As long as no sounds are generated, delay_keep_alive() is just fine at all times, with the added bonus of managed heaters and inactivity messages on uart so that the host doesn't time out.

Firmware/sound.h
```diff
@@ -13,6 +13,9 @@ typedef enum
 typedef enum
      {e_SOUND_CLASS_Echo,e_SOUND_CLASS_Prompt,e_SOUND_CLASS_Confirm,e_SOUND_CLASS_Warning,e_SOUND_CLASS_Alert} eSOUND_CLASS;
 
+// The minimum number of milliseconds of sound duration, to use the keep-alive method instead of the default delay
+// This is used to allow the printer to retain control in long-running sound output
+#define SOUND_KEEPALIVE_MIN_MS 1100
```
leptun, line 18: Not really relevant considering the points above, but just a quick observation:
Instead of doing `if (delay < 1100) delay(); else delay_keep_alive();`, do `if (delay > 1000) delay_keep_alive(); else delay();`

leptun (changes requested):

Firmware/sound.cpp
bmuessig: I was not sure how much delay the update routine adds (which might also vary), so I thought it would make sense to keep the length of short beeps more accurate. But if this is not required, this could be simplified indeed :+1:

Firmware/Marlin.h
bmuessig: Oh yes, good point. I wasn't sure if I should modify the parameter list of the original `delay_keep_alive` method.
This is certainly a more elegant solution.

bmuessig:
Alright, I have now combined the multiple methods into one with the second optional parameter controlling whether to update the LCD.

Now, what's the best minimum threshold to trigger `delay_keep_alive` instead of the regular delay?
We could lower it from the current 1100 ms to e.g. 260 or 510 ms, but I want to avoid that regular beeps vary in length.
Many long running update operations might produce an audible difference, so I would suggest keeping the regular delay method for the short default beeps.

Firmware/sound.h
bmuessig: I didn't just choose 1100 ms to exclude one second long beeps, but I added 100 ms in case someone somewhere used a delay around 1 second, but not exactly.

leptun:
I'll look at the updated PR soonTM (maybe tomorrow). In the meantime, here is a gcode with music:
[WII.zip](https://github.com/prusa3d/Prusa-Firmware/files/7672312/WII.zip)
You can try playing with the treshold and see if it negatively affects the file. Please also use the stock firmware first to get a baseline (it's kinda crappy already).

I think `delay>1000` is a good enough threshold instead of >1100. It doesn't really matter at that point if it's 1000 or 1050ms. This threshold might only improve things when quick beeps are produced. WII.gco includes some of these in a quick sequence in the second half of the song.

PR #3311, opened 2021-12-07
Ensure THERMAL RUNAWAY / temperature errors stay on-screen

wavexx:
The current handling of the thermal runaway is sub-optimal. This is the only temperature-related warning that is recoverable by the user (the print can be restarted without a reset), which requires some extra bits of care.

There is a rare possibility that during thermal runaway for the LCD status message to get overwritten during shutdown, resulting in the error message being lost.

This can happen if the thermal runaway happens to interrupt before a print is actually started (before the check_file() function is run and/or running - due to a missing check) or any fan check happens to detect a fan error in combination with a thermal runaway (causing the fan checks to reset the alert message).

For this reason we use the existing ``lcd_status_message_level`` to introduce alert severity levels, reserving higher values for temperature-related issues. This ensures a multitude of hardware failure always results in showing the first most-critical error on the LCD display.

We also rework the Stop() function to always ensure the waiting-for-heaters loops are interrupted by moving the cancel flag to the disable_heater() call instead of littering various other parts of the code, as well as ensuring Stop() moves back to the main status message at each call, ensuring a repeated thermal runaway (for example, attempting to perform filament autoload twice in a row) dismisses the modal prompt.

Sadly, this is still not perfect. ``lcd_print_stop()`` shouldn't perform any motion directly after calling ``planner_abort_hard()``: the remainder of the code should be moved as a "command step" to be performed on the next ``loop()`` iteration to ensure the marlin command loop is truly free. This should be fixed separately...


DRracer (approved):
approving:
- the code changes look reasonable
- successful feature tests

PR #3312, opened 2021-12-10
Implement first layer cal with calculated extrusion amounts

GilesBathgate:
This is a fix for #2177 (and #2238 ) It calculates the extrusion amount based on the extrusion width and layer height making at least as wide as the nozzle_diameter.

The extrusion width for the original hardcoded implementation was reverse-engineered using the following formula:

image

This was then used as the basis for checking the calculated values.

Notes for release:

* First layer calibration height changed from 0.15 to 0.2

* First layer calibration is calculated based on nozzle dia. If you use a custom nozzle different from the default 0.4mm, set nozzle dia in HW setup menu before running the calibration.

* First layer calibration square uses slightly more advanced flow math based on Prusa Slicer math https://manual.slic3r.org/advanced/flow-math

PFW-1009

### DIFF
--- a/Firmware/first_lay_cal.cpp
+++ b/Firmware/first_lay_cal.cpp
@@ -60,8 +60,8 @@ bool lay1cal_load_filament(char *cmd_buffer, uint8_t filament)
     if (MMU2::mmu2.Enabled())
     {
         enquecommand_P(PSTR("M83"));
-        enquecommand_P(PSTR("G1 Y-3.0 F1000.0"));
-        enquecommand_P(PSTR("G1 Z0.4 F1000.0"));
+        enquecommand_P(PSTR("G1 Y-3 F1000"));
+        enquecommand_P(PSTR("G1 Z0.4 F1000"));
 
         uint8_t currentTool = MMU2::mmu2.get_current_tool();
         if(currentTool == filament ){
@@ -90,16 +90,16 @@ bool lay1cal_load_filament(char *cmd_buffer, uint8_t filament)
 //! @param extrusion_width the width of the extrusion layer 
 void lay1cal_intro_line(bool extraPurgeNeeded, float layer_height, float extrusion_width)
 {
-    static const char cmd_intro_mmu_3[] PROGMEM = "G1 X55.0 E29.0 F1073.0";
-    static const char cmd_intro_mmu_4[] PROGMEM = "G1 X5.0 E29.0 F1800.0";
-    static const char cmd_intro_mmu_5[] PROGMEM = "G1 X55.0 E8.0 F2000.0";
-    static const char cmd_intro_mmu_6[] PROGMEM = "G1 Z0.3 F1000.0";
-    static const char cmd_intro_mmu_7[] PROGMEM = "G92 E0.0";
-    static const char cmd_intro_mmu_8[] PROGMEM = "G1 X240.0 E25.0  F2200.0";
-    static const char cmd_intro_mmu_9[] PROGMEM = "G1 Y-2.0 F1000.0";
-    static const char cmd_intro_mmu_10[] PROGMEM = "G1 X55.0 E25 F1400.0";
-    static const char cmd_intro_mmu_11[] PROGMEM = "G1 Z0.20 F1000.0";
-    static const char cmd_intro_mmu_12[] PROGMEM = "G1 X5.0 E4.0 F1000.0";
+    static const char cmd_intro_mmu_3[] PROGMEM = "G1 X55 E29 F1073";
+    static const char cmd_intro_mmu_4[] PROGMEM = "G1 X5 E29 F1800";
+    static const char cmd_intro_mmu_5[] PROGMEM = "G1 X55 E8 F2000";
+    static const char cmd_intro_mmu_6[] PROGMEM = "G1 Z0.3 F1000";
+    static const char cmd_intro_mmu_7[] PROGMEM = "G92 E0";
+    static const char cmd_intro_mmu_8[] PROGMEM = "G1 X240 E25  F2200";
+    static const char cmd_intro_mmu_9[] PROGMEM = "G1 Y-2 F1000";
+    static const char cmd_intro_mmu_10[] PROGMEM = "G1 X55 E25 F1400";
+    static const char cmd_intro_mmu_11[] PROGMEM = "G1 Z0.2 F1000";
+    static const char cmd_intro_mmu_12[] PROGMEM = "G1 X5 E4 F1000";
 
     static const char * const intro_mmu_cmd[] PROGMEM =
     {
@@ -127,7 +127,7 @@ void lay1cal_intro_line(bool extraPurgeNeeded, float layer_height, float extrusi
     else
     {
         char cmd_buffer[30];
-        static const char fmt1[] PROGMEM = "G1 X%d E%-.3f F1000.0";
+        static const char fmt1[] PROGMEM = "G1 X%d E%-.3f F1000";
         sprintf_P(cmd_buffer, fmt1, 60, count_e(layer_height, extrusion_width * 4.f, 60));
         enquecommand(cmd_buffer);
         sprintf_P(cmd_buffer, fmt1, 100, count_e(layer_height, extrusion_width * 8.f, 40));
@@ -138,12 +138,12 @@ void lay1cal_intro_line(bool extraPurgeNeeded, float layer_height, float extrusi
 //! @brief Setup for printing meander
 void lay1cal_before_meander()
 {
-    static const char cmd_pre_meander_0[] PROGMEM = "G92 E0.0";
+    static const char cmd_pre_meander_0[] PROGMEM = "G92 E0";
     static const char cmd_pre_meander_1[] PROGMEM = "G21"; //set units to millimeters TODO unsupported command
     static const char cmd_pre_meander_2[] PROGMEM = "G90"; //use absolute coordinates
     static const char cmd_pre_meander_3[] PROGMEM = "M83"; //use relative distances for extrusion TODO: duplicate
-    static const char cmd_pre_meander_4[] PROGMEM = "G1 E-1.50000 F2100.00000";
-    static const char cmd_pre_meander_5[] PROGMEM = "G1 Z5 F7200.000";
+    static const char cmd_pre_meander_4[] PROGMEM = "G1 E-1.5 F2100";
+    static const char cmd_pre_meander_5[] PROGMEM = "G1 Z5 F7200";
     static const char cmd_pre_meander_6[] PROGMEM = "M204 S1000"; //set acceleration
     static const char cmd_pre_meander_7[] PROGMEM = "G1 F4000";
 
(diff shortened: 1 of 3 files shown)


GilesBathgate:
Here is a test. This is with a 0.6mm dia nozzle.

image

DRracer:
@GilesBathgate nice :+1: . 

Could you please check your code doesn't eat up too much CPU flash? I'm especially afraid of the equation. Several counter measures may be considered if the code consumption is high:
- pre-computed values for the nozzles we use
- rewriting the formula of possible
- ...

GilesBathgate:
@DRracer :+1: 

> Could you please check your code doesn't eat up too much CPU flash?

Do you mean CPU, or flash? I.e optimise for memory or speed? How do I measure this?

> I'm especially afraid of the equation.

Is the concern pauses between buffer loads? The motion seemed smooth during testing.
also bear in mind the code already had:

```c++
//! @brief Count extrude length
//!
//! @param layer_height layer height in mm
//! @param extrusion_width extrusion width in mm
//! @param extrusion_length extrusion length in mm
//! @return filament length in mm which needs to be extruded to form line
static constexpr float count_e(float layer_height, float extrusion_width, float extrusion_length)
{
    return (extrusion_length * layer_height * extrusion_width / (M_PI * pow(1.75, 2) / 4));
}
```

it has been updated to:

```c++
//! @brief Count extrude length
//!
//! @param layer_height layer height in mm
//! @param extrusion_width extrusion width in mm
//! @param extrusion_length extrusion length in mm
//! @return filament length in mm which needs to be extruded to form line
static constexpr float count_e(float layer_height, float extrusion_width, float extrusion_length, float filament_diameter=1.75f)
{
    return (extrusion_length * ((M_PI * pow(layer_height, 2)) / 4 + layer_height * (extrusion_width - layer_height))) / ((M_PI * pow(filament_diameter, 2)) / 4);
}
```

>Several counter measures may be considered if the code consumption is high:

Yes although it is a constexpr, it's possibly not being evaluated at compile time if its not being passed const parameters. Trying to get it all const would be ideal, but as you say this requires precomputing all the widths for various nozzle sizes.

DRracer:
@GilesBathgate I mean program memory of the CPU (we are looking for the shortest possible code) - i.e. we are not afraid of speed but code size.
The simplest way of checking the code size is to compile the FW without your changes (note the code size), then compile the FW with your changes and report the difference.

3d-gussner:
You can also compare the travis build to the previous merged one.
You can find in the [checks](https://github.com/prusa3d/Prusa-Firmware/pull/3312/checks) and then [The build](https://app.travis-ci.com/github/prusa3d/Prusa-Firmware/builds/243269729)

Before your PR:
https://app.travis-ci.com/github/prusa3d/Prusa-Firmware/builds/235939660#L558-L559
```
Sketch uses 250784 bytes (98%) of program storage space. Maximum is 253952 bytes.
```
With your PR:
https://app.travis-ci.com/github/prusa3d/Prusa-Firmware/builds/243269729#L558-L559
```
Sketch uses 251328 bytes (98%) of program storage space. Maximum is 253952 bytes.
```

So your PR uses  additional 251328 - 250784 = 544 bytes

GilesBathgate:
@3d-gussner Thats odd, because I remove quite a few static strings:

https://github.com/prusa3d/Prusa-Firmware/pull/3312/files#diff-7713c9d58aa3e1510b52d8369b439c4484f5ab00f0763c1c5ea8f596ab8286cfL149-L164

Maybe someone can give some hints as to what is using the memory?

gudnimg:
> @3d-gussner Thats odd, because I remove quite a few static strings:
> 
> https://github.com/prusa3d/Prusa-Firmware/pull/3312/files#diff-7713c9d58aa3e1510b52d8369b439c4484f5ab00f0763c1c5ea8f596ab8286cfL149-L164
> 
> Maybe someone can give some hints as to what is using the memory?

Without looking too much into it, but I would guess `pow()` calls take up some memory. It should also be faster to not use when the power is 2. I've noticed that at least on the Arduino platform.

As an example I think this: `layer_height * layer_height ` is more efficient than this: `pow(layer_height, 2)`

Also another thing:
When dividing by 4, maybe bit shift is faster or more memory efficient? `pow(layer_height, 2)) / 4` is equal to `pow(layer_height, 2)) >> 2` since 4 is a power of 2. Division is really slow on ATmega chips.

GilesBathgate:
@gudnimg Replacing `pow(layer_height,2)` with `(layer_height*layer_height)` makes 0 bytes difference. I guess there is a template specifically for 2, and the function gets inlined.

GilesBathgate:
@DRracer I have reduced the code size as much as I think I can within the scope of this PR, but see also #3322 there are further space savings to be made there.

DRracer:
@GilesBathgate very good, I can see you managed to improve the code beyond the scope of the original intent :+1: . Now I think the PR is good enough to make it into 3.11 and tuned further if needed.

DRracer:
@gudnimg actually the shifting optimization is kind of tricky on the AVR due to:
- the AVR can only shift by 1 bit (unlike the x86)
- the AVR can only shift 8bit registers (unlike the x86)

So one must think twice before rewriting division into shifting, especially if code size is preferred over speed (which is our case here). To divide something by 4 means 2 shift instructions which is pretty close to a call to a _div_ subroutine.

In terms of speed there are no doubts of course, a _div_ subroutine is much slower than a single-cycle shift instruction.

GilesBathgate:
> Now I think the PR is good enough to make it into 3.11 and tuned further if needed.

@DRracer Great 👍.  I think more testing would be good. There is only so much I can do with MK404 and my old MK2 (its not stock hardware either as you can see from the photo)

wavexx:
Related previous PR: https://github.com/prusa3d/Prusa-Firmware/pull/3187

(also nothing that the result of the formula is .42, so it must be the answer ;))

gudnimg:
@GilesBathgate Hi 👋 There's a good window now to merge this PR into 3.13 as we now have some spare flash memory to work with, do you have time to rebase the PR with respect to MK3 branch? If not I could do it for you.

I will spend some time on reviewing this PR over the weekend.


Memory usage I'm seeing (3.13 Multilang MK3S build):

Before:
```
[build] Program:  250734 bytes (95.6% Full)
[build] 
[build] Data:       5638 bytes (68.8% Full)
```

After:
```
[build] Program:  251438 bytes (95.6% Full)
[build] 
[build] Data:       5638 bytes (68.8% Full)
```

704 byte increase.

gudnimg:
Adding `__attribute__((noinline))` to `static constexpr float count_e()` seems to shave off 298 bytes on my end. Bringing the total increase down to **406 bytes**.

GilesBathgate:
@gudnimg I think this is right, please review and test.

gudnimg:
@GilesBathgate Thanks for the quick response! I’ll run tests tomorrow morning.

gudnimg:
Tested on MK3S+ with MMU2S and stock 0.4mm nozzle

BEFORE (3.13 MK3 branch)
==========

I can feel the surface is a bit rough and running this I would think the nozzle is too close to the bed.

before


AFTER (With PR 3312)
==========

I think this is the best first layer calibration I've seen. Z baby step value is unchanged at -1.117mm.

I am very impressed by these results. The square's surface is incredibly smooth and the picture may not show it well:
image

Firmware/first_lay_cal.cpp
```diff
@@ -10,6 +10,32 @@
...
+#include <math.h>
+
+//! @brief Count extrude length
+//!
+//! @param layer_height layer height in mm
+//! @param extrusion_width extrusion width in mm
+//! @param extrusion_length extrusion length in mm
+//! @return filament length in mm which needs to be extruded to form line
+__attribute__((noinline))
+static constexpr float count_e(float layer_height, float extrusion_width, float extrusion_length, float filament_diameter=1.75f)
```
gudnimg: Minor nitpick but I would write `static constexpr float __attribute__((noinline))` to be consistent with other code in the project. The attribute is applied currently, so this isn't an issue :)

GilesBathgate: @gudnimg No worries, I will change it,  I am fond of consistency myself.

gudnimg (approved):
Approving this PR as it works very well on my side.

On multilang build (which uses the most resources) this PR only adds 408 bytes. I think this is acceptable as we have around 3KB free currently.

GilesBathgate:
> I am very impressed by these results. The square's surface is incredibly smooth and the picture may not show it well:

I am pleased it works well for you. Dare you risk taking off the stock nozzle and trying with 0.6 (If you don't re-tighten properly that way there be dragons :dragon:  :joy: )

gudnimg:
> I am pleased it works well for you. Dare you risk taking off the stock nozzle and trying with 0.6 (If you don't re-tighten properly that way dragons lead 🐉 😂 )

It's a good opportunity to try our new Nozzle Change menu 👍 I don't think I have 0.6mm nozzle for the MK3S but I have some nozzles leftover from my time with Ender 3, if it fits I'll try it.

GilesBathgate:
> It's a good opportunity to try our new Nozzle Change menu +1 I don't think I have 0.6mm nozzle for the MK3S but I have some nozzles leftover from my time with Ender 3, if it fits I'll try it.

There is of course [MK404](https://github.com/vintagepc/MK404) to test it with, I don't know if it simulates accurately the filament for the requirement of this test however.

DRracer:
@GilesBathgate in you original PR there is a statement:
_First layer calibration height changed from 0.15 to 0.2_
Does that affect the live-z value somehow? Or is it just a recomputation of material flow for the increased height?
Do you expect any impact on documentation of the First layer calibration process?

Firmware/ultralcd.cpp
```diff
@@ -857,78 +857,57 @@ void lcd_commands()
 	if (lcd_commands_type == LcdCommands::Layer1Cal)
 	{
 		char cmd1[30];
+		const uint16_t nozzle_dia = eeprom_read_word((uint16_t*)EEPROM_NOZZLE_DIAMETER_uM);
+		const float extrusion_width = (nozzle_dia + 20)/1000.0f;
+		const float layer_height = 0.2f;
```
GilesBathgate, line 829: @DRracer The value I am referring to is here. It makes the layer cal gcode put the nozzle 0.2 above the bed, and since the formula also uses this value to determine the extrude amount, all it means is that it extrudes a little more for the given height, which produces the correct results. As the tests by @gudnimg show, no changes to live-z and the cal results are good.

GilesBathgate: >Does that affect the live-z value somehow? Or is it just a recomputation of material flow for the increased height?

As above its just a recomputation of material flow and nozzle position.

>Do you expect any impact on documentation of the First layer calibration process?

I can't find anything that specifically refers to the current height of 0.15 in the documentation, so no amendments needed:

https://help.prusa3d.com/article/first-layer-calibration-i3_112364

GilesBathgate: @DRracer Also worth noting in the existing code the values here:

https://github.com/prusa3d/Prusa-Firmware/blob/85d27e6e5efe3fd8b0896b5868af4a3cb7894f68/Firmware/first_lay_cal.cpp#L154

And here:

https://github.com/prusa3d/Prusa-Firmware/blob/85d27e6e5efe3fd8b0896b5868af4a3cb7894f68/Firmware/first_lay_cal.cpp#L162

Did never line up. So the choice was to make them both as 0.2 as it gives a larger margin for error.

DRracer:
@GilesBathgate thank you very much for your effort keeping this PR up to date and fixing stuff. Our team has finally decided to approve and merge :+1:

DRracer (approved):

gudnimg:
Better late than never… I moved to a 0.6mm nozzle (from 0.4mm) and can confirm the results are perfect with the same Z-adjust settings as before. 

Surface is perfectly smooth 👌

This is with MK3S+ and MMU. PETG

IMG_3221

PR #3324, opened 2021-12-21
:construction: MK3 3.10.1 languages

3d-gussner:
This is a darft PR


3d-gussner:
This PR build uses

```
Sketch uses 251416 bytes (99%) of program storage space. Maximum is 253952 bytes.
```
vs. MK3_3.10.1 final
```
Sketch uses 251262 bytes (98%) of program storage space. Maximum is 253952 bytes.
```
Which is 154byte more
- each language string allocates some space

PR #3327, opened 2021-12-25
Remove `TEMP_STAT_LEDS` from firmware

### LINKED ISSUE #3314: [BUG]Build fails if `TEMP_STAT_LEDS` is enabled
**Describe the bug**
If `TEMP_STAT_LEDS` preprocessor definition is enabled in `Configuration.h` the build will fail. This is because the two preprocessor symbols `STAT_LED_RED` and `STAT_LED_BLUE` are not defined anywhere in the project. I suspect this is some leftover code from Marlin? We can potentially remove this code alltogether to reduce code *noise*.

image

  
**To Reproduce**
Enable `TEMP_STAT_LEDS` preprocessor definition.
  
**Expected behavior**
Build succeeds without any warnings or errors.

gudnimg:
This isn't used by MK3 printers.
It seems to be leftover code from Marlin

Motivation for removing this code
====
 
* Reduce code size
* Reduce dead code in firmware
* Resolve https://github.com/prusa3d/Prusa-Firmware/issues/3314

This Pull Request fixes issue https://github.com/prusa3d/Prusa-Firmware/issues/3314 where if `TEMP_STAT_LEDS` is defined manually then the firmware will not build because the definitions `STAT_LED_RED` and `STAT_LED_BLUE` do not exit and are intended to describe hardware pins. None of the Prusa printers have a dedicated Red or Blue LED for Temperature Indication as far as I know.

### DIFF
--- a/Firmware/Configuration.h
+++ b/Firmware/Configuration.h
@@ -489,11 +489,6 @@ your extruder heater takes 2 minutes to hit the target on heating.
 // Increase the FAN pwm frequency. Removes the PWM noise but increases heating in the FET/Arduino
 //#define FAST_PWM_FAN
 
-// Temperature status LEDs that display the hotend and bet temperature.
-// If all hotends and bed temperature and temperature setpoint are < 54C then the BLUE led is on.
-// Otherwise the RED led is on. There is 1C hysteresis.
-//#define TEMP_STAT_LEDS
-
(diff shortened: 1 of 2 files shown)


leptun (approved):

vintagepc (approved):
Looks good

PR #3336, opened 2022-01-03
Prepare for more Community languages

3d-gussner:
Adding following community languages:

- Swedish -- Svenska
- Danish -- Dansk
- Slovanian -- Slovenscina
- Hugarian -- Magyar
- Luxembourgish -- Letzebuergesch
- Croatian -- Hrvatski

### DIFF
--- a/PF-build.sh
+++ b/PF-build.sh
@@ -1303,7 +1303,7 @@ create_multi_firmware()
         echo "$(tput setaf 3)"
         ./lang-build.sh || failures 25
         # build community languages
-        ./lang-community.sh || failures 25
+        #./lang-community.sh || failures 25
         # Combine compiled firmware with languages 
         ./fw-build.sh || failures 25
         cp not_tran.txt not_tran_$VARIANT.txt
--- a/PF-build.sh
+++ b/PF-build.sh
@@ -56,7 +56,7 @@
 #   Some may argue that this is only used by a script, BUT as soon someone accidentally or on purpose starts Arduino IDE
 #   it will use the default Arduino IDE folders and so can corrupt the build environment.
 #
-# Version: 2.0.0-Build_63
+# Version: 2.0.0-Build_66
 # Change log:
 # 12 Jan 2019, 3d-gussner, Fixed "compiler.c.elf.flags=-w -Os -Wl,-u,vfprintf -lprintf_flt -lm -Wl,--gc-sections" in 'platform.txt'
 # 16 Jan 2019, 3d-gussner, Build_2, Added development check to modify 'Configuration.h' to prevent unwanted LCD messages that Firmware is unknown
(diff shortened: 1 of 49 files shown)


PF-build.sh
```diff
@@ -1303,7 +1303,7 @@ create_multi_firmware()
         echo "$(tput setaf 3)"
         ./lang-build.sh || failures 25
         # build community languages
-        ./lang-community.sh || failures 25
+        #./lang-community.sh || failures 25
```
DRracer: I assume this is just a temporary comment

lang/lang-export.sh
```diff
@@ -16,18 +24,22 @@ if [ -z "$LNG" ]; then LNG=all; fi
...
- ./lang-export.sh cz
- ./lang-export.sh de
- ./lang-export.sh es
- ./lang-export.sh fr
- ./lang-export.sh it
- ./lang-export.sh pl
+  for lang in $LANGUAGES; do
+   ./lang-export.sh $lang
+  done
+ #./lang-export.sh cz
```
DRracer: we can probably remove the commented lines

lang/lang-import.sh
```diff
@@ -171,6 +188,55 @@ if [ "$LNG" = "nl" ]; then
  sed -i 's/\xc3\x85/A/g' $LNG'_filtered.po'
 fi
 
+if [ "$LGN" = "sv" ]; then
+#repace '�' with 'Aa'
+sed -i 's/\xc3\x85/Aa/g' $LNG'_filtered.po'
```
DRracer, line 180: since we are translating one input character into 2 output ones, can't we run into some formatting issues?

DRracer (commented):
Looks good - my only concern is the automatic conversion of 1 input character into 2 output ones (Danish, Swedish). This may break formatting of LCD messages in some cases.

If we check the converted messages afterwards, there is no problem afterall. Please explain.

lang/lang-export.sh
3d-gussner: Sure we can I had these just as a fall back. on it

lang/lang-import.sh
3d-gussner: We have these for `öäuß` in German as well. This part in `lang-import.sh` is a safety net to avoid characters that cannot be shown.
The `lang-check.py` will find too long messages and if we like, we can even prevent from building the firmware to avoid LCD glitches caused by too long messages.

lang/lang-export.sh
3d-gussner: https://github.com/prusa3d/Prusa-Firmware/pull/3336/commits/6dd7681f6b2e5595c1a265be538dd64ed57c4466

PF-build.sh
3d-gussner: No, it is not needed anymore.

The updated lang scripts handle this.
`config.sh` is checking which community languages are active in `config.h` and sets the additional languages to be created and added:
```
config.sh started
 Arduino main folder: OK
 Arduino builder: OK
 AVR gcc tools:
   objcopy OK
   objdump OK
 Output folder: OK
 Objects folder: OK
 Generated elf file: OK
 Generated hex file: OK
 Languages: cz de es fr it pl
 Community languages: nl sv da sl hu lb hr 
config.sh finished with success
``` 
See https://github.com/prusa3d/Prusa-Firmware/pull/3336/files#diff-b94c0240e4353fc9197fdb3a55217a7092036fb62d84aa30c35dfe648b52a89dR39-R55

DRracer (approved):
approving after explanation, documentation of the whole process shall come next

PR #3339, opened 2022-01-04
sound: Fix default beep frequency

espr14:


wavexx:
@leptun your thoughts on this?

Default frequency of M300 on marlin is 260hz. Both current and this value seem wrong?

leptun:
My opinion on this issue is as follows:
M300 with no arguments shouldn’t actually have a default frequency. This is because the actual tone is horrible on active buzzers. I propose that M300 without a frequency argument should just hold the BEEPER pin high while beeping so that the native frequency of the buzzer is used. As soon as a frequency is provided, tone should be used (with less than ideal results).

leptun:
Closing in favor of #3542

PR #3341, opened 2022-01-06
Fix crash when the 32u2 has no SN saved

leptun:
Even though the 32u2 had the eeprom erased, it still sent 19 0x7F characters back as a supposed SN. There was an issue where the last ";S" attempt before the timeout actually didn't flush the uart rx buffer, so those 0x7F would then get queued by the cmdqueue once the firmware finished booting up. Starting a print would result in a crash because of that incomplete uart command in the buffer (the cmdqueue waits for a complete uart command to be received before it allows enqueuing commands from the SD card). Since the cmdqueue was locked up, check_file would also get stuck and trigger a watchdog reset trying (and failing) to feed data to the cmdqueue. With these changes, the cmdqueue should be clean after a failed SN get attempt on startup.

PFW-1284

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3958,17 +3958,17 @@ static uint8_t get_PRUSA_SN(char* SN)
         {
             if (timeout.expired(250u))
                 goto exit;
-            if (MYSERIAL.available() > 0)
-            {
-                SN[rxIndex] = MYSERIAL.read();
-                rxIndex++;
-            }
+            int c = MYSERIAL.read();
+            if (c >= 0)
+                SN[rxIndex++] = (char)c;
         }
         SN[rxIndex] = 0;
         // printf_P(PSTR("SN:%s\n"), SN);
         SN_valid = (strncmp_P(SN, PSTR("CZPX"), 4) == 0);
     }
 exit:
+    _delay(100); // flush any remaining characters from the 32u2. Prevents enqueuing garbage data
+    MYSERIAL.flush(); //clear RX buffer
     selectedSerialPort = selectedSerialPort_bak;
     return !SN_valid;
 }


leptun:
Marked as draft. It is not ready yet. Boards still have issues sometimes. More tests should follow next week to identify why those boards fail.

Firmware/Marlin_main.cpp
```diff
@@ -3958,17 +3958,17 @@ static uint8_t get_PRUSA_SN(char* SN)
...
+            if (c >= 0)
+                SN[rxIndex++] = (char)c;
         }
         SN[rxIndex] = 0;
         // printf_P(PSTR("SN:%s\n"), SN);
         SN_valid = (strncmp_P(SN, PSTR("CZPX"), 4) == 0);
     }
 exit:
+    _delay(100); // flush any remaining characters from the 32u2. Prevents enqueuing garbage data
+    MYSERIAL.flush(); //clear RX buffer
```
gudnimg, line 3971: I’m not sure if `flush()` clears the input buffer. It is my understanding that it will only block the program until active transmission is done and not guarantee an empty RX buffer.  

Perhaps using the MCU registers is more safe? Like this from Atmega2560 datasheet: 
8C892175-6934-4012-BBCA-0A1F51122377

leptun:
```
void MarlinSerial::flush()
{
  // don't reverse this or there may be problems if the RX interrupt
  // occurs after reading the value of rx_buffer_head but before writing
  // the value to rx_buffer_tail; the previous value of rx_buffer_head
  // may be written to rx_buffer_tail, making it appear as if the buffer
  // were full, not empty.
  rx_buffer.head = rx_buffer.tail;
}
```

In this firmware flush has a different meaning. It just means clear software RX buffer.
We could also delete the hardware buffer, but that should be cleared automatically by the RX interrupt.
Transmission is always blocking in the firmware (until the last byte is inserted in the TX hardware buffer, not really when that one is transmitted).

leptun: I'm also a little scared by the flush function. The comment above it doesn't help since it contains conflicting information. As for the action itself, it should do the flush correctly, but I find it a bit risky. Maybe flush should be in a critical section so that the assignment is atomic

PR #3345, opened 2022-01-13
Expand maximum language count

leptun:
Changed the bootapp split:
- 4 bits of lang id
- 4 bits of state
Now it is:
- 5 bits of lang id
- 3 bits of state

As a result, the maximum single language size is 32K instead of 64K. Right now LANG_SIZE_RESERVED is 12K, so it is safe.
Now up to 30 translations can be downloaded from xflash (compared to the previous 14 translations)

PFW-1283

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -919,8 +919,8 @@ void update_sec_lang_from_external_flash()
 {
 	if ((boot_app_magic == BOOT_APP_MAGIC) && (boot_app_flags & BOOT_APP_FLG_USER0))
 	{
-		uint8_t lang = boot_reserved >> 4;
-		uint8_t state = boot_reserved & 0xf;
+		uint8_t lang = boot_reserved >> 3;
+		uint8_t state = boot_reserved & 0x07;
 		lang_table_header_t header;
 		uint32_t src_addr;
 		if (lang_get_header(lang, &header, &src_addr))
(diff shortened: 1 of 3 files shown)


Firmware/config.h
```diff
@@ -58,7 +58,7 @@
 //#define LANG_MODE              0 // primary language only
 #define LANG_MODE              1 // sec. language support
 
-#define LANG_SIZE_RESERVED     0x3000 // reserved space for secondary language (12288 bytes)
+#define LANG_SIZE_RESERVED     0x3000 // reserved space for secondary language (12288 bytes). Maximum 23768 bytes
```
DRracer: Really 23768? Shouldn't it be 32768?

Firmware/Marlin_main.cpp
```diff
@@ -919,16 +919,16 @@ void update_sec_lang_from_external_flash()
...
+		uint8_t state = boot_reserved & 0x07;
 		lang_table_header_t header;
 		uint32_t src_addr;
 		if (lang_get_header(lang, &header, &src_addr))
 		{
 			lcd_puts_at_P(1,3,PSTR("Language update."));
 			for (uint8_t i = 0; i < state; i++) fputc('.', lcdout);
 			_delay(100);
-			boot_reserved = (state + 1) | (lang << 4);
+			boot_reserved = (boot_reserved & 0xF8) | ((state + 1) & 0x07);
```
DRracer, line 931:
```
state=boot_reserved & 0x7; // up to value of 7, right?
... then nothing changes state
boot_reserved=(boot_reserved & 0xF8) | ((state + 1) & 0x07);
```
in the last line:
- state is still up to 7
- state+1 is 8
- 8 & 7 is zero (i.e. nothing is added to `boot_reserved & 0xf8`)

is that correct?

leptun: It won't matter. State 8 won't ever run. Right now we only use up to 4 states anyway. It will indeed save state 0 instead of 8, but it will also finish the whole language on state 7 at most

DRracer: ok

DRracer (approved):
after some explanation of the principles I wasn't aware of, I'm approving this PR

3d-gussner (approved):
Test have been successful.

leptun:
I got two approves. Merging

PR #3348, opened 2022-01-16
Fix the c limit for a few translations

leptun:
- "Filament sensor" was changed from `MSG_FILAMENT_SENSOR` with c=20 to the correct `MSG_SELFTEST_FILAMENT_SENSOR` with c=17. This shorter limit conflicts with the following languages: ES, FR. Those translations need to be fixed.
- "Unload filament" c limit was increased from 16 to 18. The 16 limit was imposed by the SNMM, but as that is no longer supported, c=18 is possible. I also needed it for the RO translation

PFW-1283


3d-gussner (changes requested):
Please fix the broken `lang_en_es.txt` and `lang_en_fr.txt`

3d-gussner (approved):

PR #3350, opened 2022-01-18
Add a uart RX timeout

leptun:
This is an alternative to #3341

That PR doesn't seem to be foolproof and sometimes, even after waiting 500ms after the last ;S, the cmdqueue still receives some garbage characters.
So, instead of trying to prevent garbage from reaching and blocking the queue, instead this approach handles garbage data over the uart in a different way.
The way the cmdqueue is structured results in a lock-up whenever a serial string is being added to the cmdqueue. This is needed to avoid queuing incomplete commands from the uart while SD printing. The garbage characters from the 32u2 would look like a perfectly fine, but incomplete command from the uart, so the code would just wait indefinitely for an EOL character to be reached (which would never come).
With this PR, if the command is incomplete and 2 seconds have passed from the last received character, the command is considered broken and is discarded, unblocking the other cmdqueue sources (such as SD or internal commands). Whenever this happens, `RX timeout` is sent over serial to the host.
This approach was actually also used for the farm mode. I just refreshed the code and changed some dangerous behaviour. Farm mode has a 800ms timeout (kind of random if you ask me, but I kept is as it was before) and it used to queue the incomplete command on timeout. That behaviour was dangerous and broken in subtle ways, so I made it dump instructions instead. Also, shortTimer was used instead of 2 global uint32_t counters. It has the added benefit of only taking up 3B of ram instead of 8.

Example command getting dumped on startup due to 32u2 timeout + garbage characters:
```
start
;S
;S
;S
;S
;S
SN update failed
echo: 3.10.1-4697
echo: Last Updated: Jan 18 2022 17:47:47 | Author: (none, default config)
Compiled: Jan 18 2022
echo: Free Memory: 2049  PlannerBufferBytes: 1760
echo:Hardcoded Default Settings Loaded
adc_init
Extruder fan type: NOCTUA
CrashDetect DISABLED
FSensor ENABLED (sensor board revision:unknown state)
Sending 0xFF
echo:SD card ok
LCD status changed
echo:busy: paused for user
RX timeout              <-----------------This is where the timeout happens (2 seconds after the last garbage character)
echo:busy: paused for user
MMU => 'start'
MMU <= 'S1'
echo:busy: paused for user
echo:busy: paused for user
echo:busy: paused for user
MMU => '106ok'
MMU <= 'S2'
MMU => '372ok'
MMU version valid
MMU <= 'M1'
MMU - ENABLED
echo:busy: paused for user
```

As for disadvantages with this approach, it might cause problems when directly typing a command by hand in a terminal such as putty which sends every character as soon as it is typed. That being said, I think 2 seconds between characters is more than enough for sending quick commands if you already know what to type (and it's not like you don't get a message telling you that the command was discarded). This disadvantage could also be considered an advantage for when a mistake is done while writing a command by hand. This way the command can be aborted by simply waiting for a bit compared to previously where you had no way of altering what was already sent.

PFW-1284

### DIFF
--- a/Firmware/cmdqueue.cpp
+++ b/Firmware/cmdqueue.cpp
@@ -24,8 +24,7 @@ int serial_count = 0;  //index of character read from serial line
 bool comment_mode = false;
 char *strchr_pointer; // just a pointer to find chars in the command string like X, Y, Z, E, etc
 
-unsigned long TimeSent = _millis();
-unsigned long TimeNow = _millis();
+ShortTimer serialTimeoutTimer;
 
 long gcode_N = 0;
 long gcode_LastN = 0;
@@ -400,8 +399,7 @@ void get_command()
         MYSERIAL.write(serial_char); // for debuging serial line 2 in farm_mode
         selectedSerialPort = 1; 
     } */ //RP - removed
-      TimeSent = _millis();
-      TimeNow = _millis();
+      serialTimeoutTimer.start();
 
     if (serial_char < 0)
         // Ignore extended ASCII characters. These characters have no meaning in the G-code apart from the file names
@@ -535,22 +533,11 @@ void get_command()
     }
   } // end of serial line processing loop
 
-    if(farm_mode){
-        TimeNow = _millis();
-        if ( ((TimeNow - TimeSent) > 800) && (serial_count > 0) ) {
-            cmdbuffer[bufindw+serial_count+CMDHDRSIZE] = 0;
-            
-            bufindw += strlen(cmdbuffer+bufindw+CMDHDRSIZE) + (1 + CMDHDRSIZE);
-            if (bufindw == sizeof(cmdbuffer))
-                bufindw = 0;
-            ++ buflen;
-            
-            serial_count = 0;
-            
-            SERIAL_ECHOPGM("TIMEOUT:");
-            //memset(cmdbuffer, 0 , sizeof(cmdbuffer));
-            return;
-        }
+    if (serial_count > 0 && serialTimeoutTimer.expired(farm_mode ? 800 : 2000)) {
+        comment_mode = false;
+        serial_count = 0;
+        SERIAL_ECHOLNPGM("RX timeout");
+        return;
     }
 
   #ifdef SDSUPPORT
--- a/Firmware/cmdqueue.h
+++ b/Firmware/cmdqueue.h
@@ -52,9 +52,6 @@ extern int serial_count;
 extern bool comment_mode;
 extern char *strchr_pointer;
 
-extern unsigned long TimeSent;
-extern unsigned long TimeNow;
-
 extern long gcode_N;
 extern long gcode_LastN;
 extern long Stopped_gcode_LastN;


leptun:
@gudnimg Since you also commented on my other PR that tried to fix this exact issue, what do you think of this approach? Do you see any downside to merging this PR instead?

gudnimg:
> @gudnimg Since you also commented on my other PR that tried to fix this exact issue, what do you think of this approach? Do you see any downside to merging this PR instead?

In this PR I don't see where exactly the garbage data is cleared. Doesn't it end up in a `flush()` call somewhere? I don't see a disadvantage with your solution besides *maybe* the 2 second timeout. But I do like that this PR may be a better solution in that it looks to be more general and not specific to reading the SN.

Regarding the other PR, after looking more into the code  I do agree that `flush()` "clears" the RX buffer. It seems that it only matters that the `head` and `tail` are the same integer value so that `read()` and `available()` and other functions use that condition to determine the buffer is "empty".

If the `flush()` truly worked then something must have sent the garbage character after you cleared it🤔 Is there possibility other hardware connected to the einsy board is interfering?

leptun:
@gudnimg In this PR I change the philosophy. Instead of flushing the RX buffer at the end of SN_get, I just discard the corrupted data after a timeout later on.

The other hardware you're reffering to is the atmega32u2. It has some sort of bug where it keeps sending characters if the SN is uninitialized (possibly also at an incorrect baud rate). The wrong baud rate might explain why it takes so long for the 20B SN to be sent. If the baud rate is in fact correct, we'd still only receive 19 0x7F characters and a '\n' at the end. If that was the case, the SN would be found to be invalid and the process would retry, but nothing would get in the cmdqueue later on, so something is definitely wrong.

If I just discard those incomplete lines later on in the cmdqueue, then the bug should be gone as that line would get aborted and the cmdqueue unblocked for gcode from UI. The thing is nobody tested these changes yet besides me, so it is unknown if the bug got fixed for the other boards which had the bug even after the first PR. Do you see any way this approach could introduce other bugs?

gudnimg:
> Do you see any way this approach could introduce other bugs?

I honestly can't think of any situation. I think this change does more good than harm. If it helps I can test this PR over the weekend.

I am curious though if this could be an issue for Octoprint users, but I don't have that setup. It is my understanding that it uses the Serial stream to send and receive data.

leptun:
I was also thinking it could cause issues with hosts. Or at least it might change the transaction at the beginning of the connection. If you can, please do some quick tests with octoprint. Maybe try both single and multi-lang builds as the startup timing is very different between the two. Of course, if you have time.

Btw, the exact crash scenario is this:
Have no SN in the atmega32u2 eeprom
Start the firmware, start a print from the LCD. A crash due to watchdog reset would happen when checking the file

gudnimg:
> If you can, please do some quick tests with octoprint

I don't have that setup, sorry.

> Have no SN in the atmega32u2 eeprom

I would need to re-flash it to trigger this (manually erase the SN)? Is that firmware available on Github?

I've been trying to trigger the RX timeout, it seems to be impossible with Arduino IDE. That's good.

leptun:
No octoprint: ok. I’ll find another way to test it myself.

32u2 fw: not sure if it’s available. I can send you the flash and eeprom of my 32u2. You need a programmer for this. There’s no other way to erase the eeprom of the 32u2.

Arduino buffers lines, so it must be ok. Pronterface does the same I think.

leptun:
I don't understand why this PR has no travis build triggered. Weird

3d-gussner (approved):

PR #3355, opened 2022-01-21
Change Travis to Ubuntu 20.04 LTS (focal)

3d-gussner:
Travis runs on Ubuntu 14.0.5 LTS (trusty) which
- Isn't supported anymore
- EOF soon
- causes issues with Python3 script `lang-check.py`
- Doesn't represent the environment of
  - PF-build.sh
  - build server

As we compile with Arduino IDE the different gcc version between trusty and focal don't affect the firmware build.
Only the `test.sh` uses the newer gcc version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: trusty
+dist: focal
 before_install:
   - sudo apt-get install -y ninja-build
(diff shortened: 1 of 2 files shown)


DRracer (approved):
The build seems to work, no issues found, approving the update

Issue #3352, opened 2022-01-20
[FEATURE REQUEST] Allow cancellation of Live-Z WITHOUT saving adjustment


alexwhittemore:
I think this might be a duplicate of https://github.com/prusa3d/Prusa-Firmware/issues/1968, but I honestly can't tell.

If it makes sense, enter what type of printer or upgrade the feature request applies to.
**Printer type** - Mk2.5, but I think all of the Mk<=3 series
**MMU Upgrade** - N/A

I'm a power-user of "press and hold knob to change Z height." I.e. I very regularly finish a print, then press and hold to move the nozzle away. Or more often, cancel a failed print, then press and hold to move the nozzle away.

Helpfully, a few firmware updates ago introduced the ability to press-and-hold the knob to get to the Live Z adjustment while in-print. That's cool, since changing the Z height in a running print makes no sense.

Unhelpfully, the definition of "in a print" is VERY fuzzy, and I literally-every-time I cancel a print try to jog the extruder up, only to accidentally enter the live Z menu and mess it all up.

Since there's no option to cancel a live Z adjustment, my only recourse is to reset the printer at this point.
  
**Describe the solution you'd like**
In the Adjusting Live Z modal, I'd LOVE a "press and hold to cancel." I.e. tapping the knob commits the new live Z, where long-pressing cancels out without saving the value.

Panayiotis-git:
The long-press functionality as implemented at the moment is inconsistent. Look at the https://github.com/prusa3d/Prusa-Firmware/pull/3014#issuecomment-776993759 for more details. I believe that this event requires a clear definition, so a user is always sure what to expect from a long-press.

gudnimg:
Just to highlight the inconsistencies mentioned by @Panayiotis-git below is the menu when the printer is not printing anything. Obviously not all the menus.

| Menu                 |  Long-Press | What does it do? |
| ----------------- | -------------------- | ---------------  |
| Status      | X | Open move Z-axis menu   |
| Preheat | X | Open move Z-axis menu   |
| Print from SD card | X | Open move Z-axis menu   |
| AutoLoad filament| X | Exit menu   |
| Unload filament|  |  |
| Settings  | X | Open move Z-axis menu   |
| Calibration | X | Exit menu    |
| Statistics | X | Exit menu    |
| Fail stats| X | Exit menu    |
| Support | X | Open move Z-axis menu   |
| Settings->Temperature  | X | Open move Z-axis menu   |
[...]

PR #3360, opened 2022-01-22
:construction: Add Long press event to cancel Live Z adjustments

gudnimg:
My attempt at resolving https://github.com/prusa3d/Prusa-Firmware/issues/3352

This is just a draft currently. I need to test the menus more. But the cancelation seems to be working on my end.

Problems I saw while looking into this:

1. The `current_position` array does not take into account the value of the Z offset.
2. When the Z offset is adjusted in the menu, the value is not saved until the user clicks the knob. But if we want to cancel the operation we must grab this value.

Change in memory footprint:
Flash: +258 bytes
SRAM: +4 bytes

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -73,7 +73,7 @@ CustomMsg custom_message_type = CustomMsg::Status;
 uint8_t custom_message_state = 0;
 
 bool isPrintPaused = false;
-
+float unsaved_live_z_adjust = 0;
 static ShortTimer display_time; //just timer for showing pid finished message on lcd;
 static uint16_t pid_temp = DEFAULT_PID_TEMP;
 
@@ -2637,6 +2637,7 @@ static void lcd_babystep_z()
         else babystepsTodoZadd(lcd_encoder);
 
 		_md->babystepMemMMZ = _md->babystepMemZ/cs.axis_steps_per_unit[Z_AXIS];
+        unsaved_live_z_adjust = _md->babystepMemMMZ; // Store value in case user wants to cancel changes
 		_delay(50);
 		lcd_encoder = 0;
 		lcd_draw_update = 1;
@@ -7442,10 +7443,12 @@ uint8_t get_message_level()
 
 void menu_lcd_longpress_func(void)
 {
+    lcd_longpress_trigger = 0;
     // Wake up the LCD backlight and,
     // start LCD inactivity timer
     lcd_timeoutToStatus.start();
-    if (homing_flag || mesh_bed_leveling_flag || menu_menu == lcd_babystep_z || menu_menu == lcd_move_z || menu_block_mask != MENU_BLOCK_NONE || Stopped)
+    backlight_wake();
+    if ((homing_flag || mesh_bed_leveling_flag) && (menu_menu == lcd_babystep_z || menu_menu == lcd_move_z || menu_block_mask != MENU_BLOCK_NONE  || Stopped))
     {
         // disable longpress during re-entry, while homing, calibration or if a serious error
         lcd_quick_feedback();
@@ -7456,6 +7459,13 @@ void menu_lcd_longpress_func(void)
         // only toggle the experimental menu visibility flag
         lcd_quick_feedback();
         eeprom_toggle((uint8_t *)EEPROM_EXPERIMENTAL_VISIBILITY);
+        return;
+    }
+
+    if (menu_menu == lcd_babystep_z) {
+        // User wants to cancel the adjustment to Z-offset.
+        babystep_revert();
+        menu_back();
         return;
     }
 
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -9481,7 +9481,6 @@ void manage_inactivity(bool ignore_stepper_queue/*=false*/) //default argument s
       if (lcd_longpress_func && lcd_update_enabled)
       {
           lcd_longpress_func();
-          lcd_longpress_trigger = 0;
       }
   }
 
--- a/Firmware/mesh_bed_calibration.cpp
+++ b/Firmware/mesh_bed_calibration.cpp
@@ -3034,6 +3034,23 @@ static void shift_z(float delta)
     plan_set_z_position(current_position[Z_AXIS]);
 }
 
+/**
+ * @brief Similar to shift_z() but this function is
+ * intended to revert changes in the Z-offset from
+ * "Live Z Adjustment" menu.
+ * @param delta the difference between Z offset stored in EEPROM
+ * and Z offset that was adjusted in the "Live Z Adjustment" menu but
+ * was not saved.
+ */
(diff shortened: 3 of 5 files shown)


Firmware/ultralcd.cpp
```diff
@@ -86,6 +86,7 @@ uint8_t farm_mode = 0;
 int farm_timer = 8;
 uint8_t farm_status = 0;
 bool printer_connected = true;
+float unsaved_live_z_adjust = 0;
```
leptun, line 76: maybe we can reload the value from eeprom instead

Firmware/ultralcd.cpp
```diff
@@ -8750,21 +8752,31 @@ uint8_t get_message_level()
...
+    if ( (homing_flag || mesh_bed_leveling_flag) && (menu_menu == lcd_babystep_z || menu_menu == lcd_move_z))
     {
         // disable longpress during re-entry, while homing or calibration
+        MYSERIAL.println("disable longpress during re-entry, while homing or calibration");
         lcd_quick_feedback();
         return;
     }
     if (menu_menu == lcd_hw_setup_menu)
     {
+        MYSERIAL.println("menu_menu == lcd_hw_setup_menu");
```
leptun: Use `puts_P(PSTR("The message"));`
It does the same thing but it keeps things in progmem instead of valuable ram. You might soon overflow the stack in certain situations

gudnimg: Those prints were not supposed to be in the commit 👍 But good to know

gudnimg:
One thing I am a bit confused by is that using this code:

```c
static void shift_z(float delta)
{
    plan_buffer_line(current_position[X_AXIS], current_position[Y_AXIS], current_position[Z_AXIS] - delta, current_position[E_AXIS], homing_feedrate[Z_AXIS]/40, active_extruder);
    st_synchronize();
    plan_set_z_position(current_position[Z_AXIS]);
}
```

Namely `plan_buffer_line()` and `st_synchronize()`, just didn't work when printing. I saw shifts in the Y-axis for whatever reason. So I ended up using only `plan_set_z_position()` in my new function `shift_baby_z()`. That seems to work well, but I'm curious if someone familiar with those functions could see a potential danger here?

~~I moved the clearing of `lcd_longpress_trigger` because of `st_synchronize()` but now that I don't call it, I can revert that change.~~ This actually saves 2 bytes so probably better to not revert it.

alexwhittemore:
I guess I'm not the only one bothered by this! What a hero :)

wavexx:
I wanted to give this a shot, but I wanted to understand the logic behind it =)

What's the rationale? You start live Z adjust, make some tweaks, but realize the result is worse than before.
At this point long-press just moves back to the original live Z value?

What if the original Z value is big? (for example, cancelling a 1.5mm gap during the first layer calibration)

gudnimg:
> What's the rationale? You start live Z adjust, make some tweaks, but realize the result is worse than before.
At this point long-press just moves back to the original live Z value?

Exactly right. When we enter the Z Adjust menu, the value shown on the screen is stored with `unsaved_live_z_adjust` (at least in this PR). This is needed because as the Z value is changed, the Z-axis will move immediately no matter what, so we need to track the number shown on the LCD screen somehow. Because of this -- if the user decides to cancel the Z-adjust changes with a long-press -- we will later use `unsaved_live_z_adjust` to revert the actual movement made to the Z-axis so that it is the same as when we entered the Z Adjust menu. This is done with `babystep_revert()`

> What if the original Z value is big? (for example, cancelling a 1.5mm gap during the first layer calibration)

Then 1.5mm will be used again and any change made to the Z-axis position will be reverted since the change in Z value is not saved. I don't see a reason this would be a problem if the user explicitly set 1.5mm before. We could always add a warning message or abort the change with a maximum threshold of Z value but I don't see a reason to.

----

As I am writing this reply, I got an idea... perhaps with the long-press... instead of backing out of the menu we should instead show the "reverted" value on the screen. 

1. Say we changed 1.5mm to -0.900mm
2. User Long-presses the knob
3. The screen will show 1.5mm now instead of -0.900mm; behind the scenes the printer will revert the Z-axis position to be consistent with 1.5mm. 
 
In this scenario we would make the user click the knob (no long-press) to actually exit the menu and also confirm the Z value shown on the screen is correct.

alexwhittemore:
@gudnimg That seems reasonable, but perhaps overcomplicated (the logic-on-cancel I mean). 

My entire reason for requesting the feature is that, 90% of the time I enter the live-adjust-z menu, it's entirely accidental, and I was just trying to jog the Z axis UP after a print (seems to take a weirdly long time post-print before long-press becomes jog Z instead of live adjust z). 

In other words, while using this PR to cancel LAZ, it would bug me almost never to "accidentally long-press and lose my adjustment" vs it would bug me almost always to "go through yet another step to just get out of here and keep my original LAZ"

wavexx:
I had this happen too.
Having live Z show the delta is also an interesting proposal - I like that too

Panayiotis-git:
I'm still undecided if we need one more use for the long-press event.

> ... 90% of the time I enter the live-adjust-z menu, it's entirely accidental, and I was just trying to jog the Z axis UP...

Exactly! The problem is that most of the times we are unsure of the outcome or something else happens when we choose the long-press functionality.

I flirt with the idea that the long-press should be a shortcut to a menu with a couple or more of the most useful possible actions at that time.

Something like:
Back
Cancel action (if applicable)
Adjust Live-Z (if applicable)
Adjust Z-axis (if applicable)
Adjust hotend temperature
Pause print
Show hidden/Hide menu (if applicable)
etc.

gudnimg:
I'm closing this PR _for now_ because we don't have much space for this feature in 3.12. I might tackle this again later.

wavexx:
You can also leave this open. IMHO just because it doesn't apply now, as long as it's still relevant it's fine to stay.

wavexx:
I think it's time to revisit this now.

Panayiotis-git:
The printer has only one button and there are implemented two events for the user to interact with it: the click and the long-press. The click event means "Select/Confirm", while the long-press in general activates a "Shortcut" function.

In general, as a user interface approach, the one button has the meaning of the "Select" and all the functionality is based on screens with options, where the user scrolls to the appropriate option and clicks to select.

For the "Live adjust Z" page and the request for a Cancel functionality, I can think of the following solutions:
1. As @gudnimg proposes, to redefine for this case only the meaning of the long-press event.
2. After setting the Live Z value and clicking the button, the user goes to a "Do you want to keep the changes? [Yes]  [No]" screen and select with the button the preferred option. This approach is more aligned with the universal one button user interface.
3. Implement a double-click event, and consider it as a "Cancel/Revert" for an edit field. 😋

gudnimg:
Rebased branch with current MK3 (3.13)

KOBALT3D:
It could be useful to use this feature in menu as **Back Button** on all settings which does not have option for cancel. Maybe optionally enabled/disabled in Settings.

PR #3364, opened 2022-01-28
Crash if pullups get enabled on the thermistor inputs

leptun:
WIth the hopes of tracking down why this happened


vintagepc (approved):
Looks good for what it intends to do

PR #3365, opened 2022-01-30
memory and flash optimizations for 3.12 PR2

leptun:
Supersedes the broken PR https://github.com/prusa3d/Prusa-Firmware/pull/3181
Comments from there still apply.

Stats as of https://github.com/prusa3d/Prusa-Firmware/pull/3365/commits/f068c3258f477a114d4b59ad36f83f6ae4f023c7 for MK3S build:
- flash: 232522 -> 229386 = 3136B
- ram: 6115 -> 5679 = 436B


leptun:
@gudnimg Now we need to check that nothing got broken during the rebase. To begin, let's make it build again

Firmware/cardreader.h
```diff
@@ -130,13 +130,12 @@ class CardReader
...
   //int16_t n;
-  unsigned long autostart_atmillis;
+  ShortTimer autostart_atmillis;
   uint32_t sdpos ;
 
   bool autostart_stilltocheck; //the sd start is delayed, because otherwise the serial cannot answer fast enought to make contact with the hostsoftware.
   
-  uint16_t nrFiles; //counter for the files in the current directory and recycled as position counter for getting the nrFiles'th name in the directory.
-  char* diveDirName;
+  int16_t nrFiles; //counter for the files in the current directory and recycled as position counter for getting the nrFiles'th name in the directory.
```
gudnimg: `nrFiles` should be `uint16_t`. It must have been reverted to `int16_t` 🤔

gudnimg: https://github.com/prusa3d/Prusa-Firmware/pull/3245

leptun: I hope other already merged PRs weren't affected during the rebase. If we see anything suspicious in this PR, we must investigate

gudnimg:
@leptun is the PR locked or is it Ok to add more optimisations 🙃? I may have found a way to optimise memory by removing static keywords on local function variables. Also found another int which can be uint8_t

leptun:
@gudnimg You can still add stuff 👍

leptun:
@gudnimg I managed to optimize MultiU16X8toH16 more.
With arduino 1.8.5 I optimized 14B.
With arduino 1.8.13 I optimized 32B since the compiler is not hot garbage. I've had the code increase (!!!) with GCC 5.4.0 at times when I was allowing the compiler to choose what register to use instead of R26. Now that I don't use R26, the code is somewhat more optimized even with the old compiler. The new compiler was optimizing correctly even when I still used R26, so it's just the old compiler that fails to optimize correctly.

leptun:
@gudnimg Let's try and find any places in the code that could use __uint24 instead of uint32_t/long/whatever. I'm pretty sure some place in the stepper routines could use 24bit integers instead of 32bit (the acceleration variables)

gudnimg:
I managed to save a good chunk of bytes by looking over how we use `code_value()`. A bit overkill converting a `float` to `uint8_t` in a few cases :P

leptun:
Got 200B of ram and 44B of flash by changing how the sorting works. I became aware of a limitation of fat32 which allowed me to change from uint32_t addressing to uint16_t for file entries in directories. Since 400B are thrown away for soring, 200B are saved when we halve that cache

leptun:
@gudnimg We'll probably merge this PR to MK3 on Monday. Unless you have some other changes already made, I think we should stop optimizing for now and verify the PR over the weekend.

gudnimg:
@leptun I have no changes locally.

leptun:
I forgot to push this commit. Now it should be final

vintagepc (approved):
Looks reasonable to me, I don't see any glaring issues.

vintagepc (approved):

gudnimg:
This tiny change saves 64 bytes on my end image

leptun:
Please add it 👍

gudnimg:
Added another commit for an optimization I found last night since it saves > 100 bytes. 

For reference here is the description of `hypot()`:
image

leptun:
Are you 100% sure this has no side effects (like loss in precision or something like that)?
I'd like to stop with the changes for now so that we can finally test and merge this PR tomorrow. I have ~9 PRs that I want to merge tomorrow

gudnimg:
> Are you 100% sure this has no side effects (like loss in precision or something like that)?
> I'd like to stop with the changes for now so that we can finally test and merge this PR tomorrow. I have ~9 PRs that I want to merge tomorrow

@leptun I don’t want to promise it without testing it in more detail. But I would be surprised if it affects the precision. Feel free to revert the commit, I realise this may be a risky commit so close to the expected merge date.

leptun:
I'll copy this commit to another branch (as in part 3 of this PR) so that we don't lose it. Then I'll drop it

leptun:
See https://github.com/leptun/Prusa-Firmware/tree/MK3_3.12_Optimizations_extra

Firmware/Marlin_main.cpp
```diff
@@ -6120,7 +6120,7 @@ if(eSoundMode!=e_SOUND_MODE_SILENT)
         int8_t pin_number = LED_PIN;
         if (code_seen('P'))
           pin_number = code_value();
-        for(int8_t i = 0; i < (int8_t)(sizeof(sensitive_pins)/sizeof(int8_t)); i++)
+        for(int8_t i = 0; i < (int8_t)(sizeof(sensitive_pins)/sizeof(*sensitive_pins)); i++)
```
DRracer: please use
`sizeof(sensitive_pins)/sizeof(sensitive_pins[0])`

I know it has the same effect, but its intent is much more obvious

Firmware/planner.cpp
```diff
@@ -459,10 +459,7 @@ void plan_init() {
   #ifdef LIN_ADVANCE
   memset(position_float, 0, sizeof(position_float)); // clear position
   #endif
-  previous_speed[0] = 0.0;
-  previous_speed[1] = 0.0;
-  previous_speed[2] = 0.0;
-  previous_speed[3] = 0.0;
+  memset(previous_speed, 0, sizeof(previous_speed));
```
DRracer, line 462: I'm surprised the compiler couldn't optimize the float assignments into a single block of "memset"
Have you checked the assembly from this optimization? Does it call the `memset` or has it been inlined in a form of a short zeroing cycle?

Firmware/Marlin_main.cpp
```diff
@@ -2827,7 +2827,7 @@ static void gcode_G28(bool home_x_axis, long home_x_value, bool home_y_axis, lon
 
       enable_endstops(true);
 
-      memcpy(destination, current_position, sizeof(destination));
+      set_destination_to_current();
```
DRracer, line 2820: :+1:

Firmware/Marlin.h
```diff
@@ -275,6 +275,17 @@ FORCE_INLINE unsigned long millis_nc() {
 void setPwmFrequency(uint8_t pin, int val);
 #endif
 
+enum class HeatingStatus : uint8_t
```
DRracer, line 280: :+1:

Firmware/Marlin_main.cpp
```diff
@@ -5623,7 +5623,7 @@ if(eSoundMode!=e_SOUND_MODE_SILENT)
                 SERIAL_PROTOCOLPGM("\nZ search height: ");
                 SERIAL_PROTOCOL(MESH_HOME_Z_SEARCH);
                 SERIAL_PROTOCOLLNPGM("\nMeasured points:");
-                for (uint8_t y = MESH_NUM_Y_POINTS-1; y >= 0; y--) {
+                for (int16_t y = MESH_NUM_Y_POINTS-1; y >= 0; y--) {
```
DRracer: this looks like something else has not been optimized - it doesn't make sense to keep the int16_t at this spot if `MESH_NUM_Y_POINTS` never exceeds 255.

Firmware/mesh_bed_calibration.cpp
```diff
@@ -2872,14 +2872,14 @@ bool sample_mesh_and_store_reference()
 		}
         mbl.set_z(0, 0, current_position[Z_AXIS]);
     }
-    for (int8_t mesh_point = 1; mesh_point != MESH_MEAS_NUM_X_POINTS * MESH_MEAS_NUM_Y_POINTS; ++ mesh_point) {
+    for (uint8_t mesh_point = 1; mesh_point != MESH_MEAS_NUM_X_POINTS * MESH_MEAS_NUM_Y_POINTS; ++ mesh_point) {
```
DRracer, line 2876: please add a `static_assert(MESH_MEAS_NUM_X_POINTS * MESH_MEAS_NUM_Y_POINTS < 255, "overflow.....")`

Firmware/Marlin_main.cpp
```diff
@@ -11130,7 +11130,7 @@ void uvlo_()
     eeprom_update_dword((uint32_t*)(EEPROM_FILE_POSITION), sd_position);
 
     // Store the mesh bed leveling offsets. This is 2*7*7=98 bytes, which takes 98*3.4us=333us in worst case.
-    for (int8_t mesh_point = 0; mesh_point < MESH_NUM_X_POINTS * MESH_NUM_Y_POINTS; ++ mesh_point) {
+    for (uint8_t mesh_point = 0; mesh_point < MESH_NUM_X_POINTS * MESH_NUM_Y_POINTS; ++ mesh_point) {
```
DRracer, line 11138: again - please add a static_assert that MESH_NUM_X_POINTS * MESH_NUM_Y_POINTS never exceeds 255

Firmware/cardreader.h
```diff
@@ -135,7 +135,7 @@ class CardReader
 
   bool autostart_stilltocheck; //the sd start is delayed, because otherwise the serial cannot answer fast enought to make contact with the hostsoftware.
   
-  int16_t nrFiles; //counter for the files in the current directory and recycled as position counter for getting the nrFiles'th name in the directory.
+  uint16_t nrFiles; //counter for the files in the current directory and recycled as position counter for getting the nrFiles'th name in the directory.
```
DRracer, line 138: are you absolutely sure that `nrFiles` is never abused with a negative value?

Firmware/speed_lookuptable.h
```diff
@@ -8,30 +8,36 @@ extern const uint16_t speed_lookuptable_slow[256][2] PROGMEM;
...
+    "mul %B2, %A1" "\n\t"
+    "movw %0, r0" "\n\t"
+    "mul %A2, %A1" "\n\t"
+    "lsl r0" "\n\t"         //push MSB to carry for rounding
+    "adc %A0, r1" "\n\t"    //add with carry (for rounding)
+    "clr r1" "\n\t"         //make r1 __zero_reg__ again
+    "adc %B0, r1" "\n\t"    //propagate carry of addition (add 0 with carry)
+    : "=&r" (out)
+    : "r" (x), "r" (y)
+    : "r0", "r1"            //clobbers: Technically these are either scratch registers or always 0 registers, but I'm making sure the compiler knows just in case.
```
DRracer, line 37: nice :+1: ... looks correct to me

Firmware/ConfigurationStore.cpp
```diff
@@ -32,7 +32,7 @@ static bool EEPROM_writeData(uint8_t* pos, uint8_t* value, uint8_t size, const c
 #endif //DEBUG_EEPROM_WRITE
 {
 #ifdef DEBUG_EEPROM_WRITE
-	printf_P(PSTR("EEPROM_WRITE_VAR addr=0x%04x size=0x%02hhx name=%s\n"), pos, size, name);
+	printf_P(PSTR("EEPROM_WRITE_VAR addr=0x%04x size=0x%02x name=%s\n"), pos, size, name);
```
DRracer, line 35: please use formatting flags defines from `inttypes.h` and be _very specific_ about the data type of the attributes
https://manned.org/PRIu8

leptun: In variadic functions, uint8_t always gets promoted to 16bit. hh/h on avr doesn’t do anything for printf. It only does stuff for scanf.

Firmware/Marlin_main.cpp
```diff
@@ -6511,9 +6511,9 @@ SERIAL_PROTOCOLPGM("\n\n");
     */
     case 73: //M73 show percent done, time remaining and time to change/pause
     {
-        if(code_seen('P')) print_percent_done_normal = code_value();
+        if(code_seen('P')) print_percent_done_normal = code_value_uint8();
         if(code_seen('R')) print_time_remaining_normal = code_value();
```
DRracer, line 6513: please rename `code_value` to `code_value_float` so that we don't run into such an issue in the future.

Firmware/cardreader.cpp
```diff
@@ -689,11 +689,11 @@ void CardReader::getfilename(uint16_t nr, const char * const match/*=NULL*/)
   
 }
 
-void CardReader::getfilename_simple(uint32_t position, const char * const match/*=NULL*/)
+void CardReader::getfilename_simple(uint16_t entry, const char * const match/*=NULL*/)
 {
 	curDir = &workDir;
 	nrFiles = 0;
-	curDir->seekSet(position);
+	curDir->seekSet((uint32_t)entry << 5);
```
DRracer, line 696: nice :+1:

Firmware/eeprom.cpp
```diff
@@ -100,6 +100,11 @@ if (eeprom_read_byte((uint8_t*)EEPROM_PINDA_TEMP_COMPENSATION) == 0xff) eeprom_u
 
 	if (eeprom_read_dword((uint32_t*)EEPROM_JOB_ID) == EEPROM_EMPTY_VALUE32)
 		eeprom_update_dword((uint32_t*)EEPROM_JOB_ID, 0);
+
+    if (eeprom_read_byte((uint8_t *)EEPROM_TOTALTIME) == 255 && eeprom_read_byte((uint8_t *)EEPROM_TOTALTIME + 1) == 255 && eeprom_read_byte((uint8_t *)EEPROM_TOTALTIME + 2) == 255 && eeprom_read_byte((uint8_t *)EEPROM_TOTALTIME + 3) == 255) {
```
DRracer, line 104: Wasn't this the way to "create" negative filament usage in stats? In case the user skipped the wizard or something?

Firmware/ultralcd.cpp
```diff
@@ -772,21 +771,6 @@ void lcdui_print_status_screen(void)
 // Main status screen. It's up to the implementation specific part to show what is needed. As this is very display dependent
 void lcd_status_screen()                          // NOT static due to using inside "Marlin_main" module ("manage_inactivity()")
 {
-	if (firstrun == 1)
```
DRracer, line 776: I'm not completely convinced that the removal of this code is correct. Please prepare a detailed feature test for this scenario.

Firmware/stepper.cpp
```diff
@@ -134,9 +129,7 @@ static uint16_t OCR1A_nominal;
 static uint8_t  step_loops_nominal;
 
 volatile long endstops_trigsteps[3]={0,0,0};
-static volatile bool endstop_x_hit=false;
-static volatile bool endstop_y_hit=false;
-static volatile bool endstop_z_hit=false;
+static volatile uint8_t endstop_hit = 0;
```
DRracer, line 135: Please DON'T - these flags are used in time critical code and updating a single bit takes more CPU cycles than writing a single byte.

Firmware/speed_lookuptable.h
```diff
@@ -39,61 +39,70 @@ FORCE_INLINE uint16_t MUL8x16R8(uint8_t x, uint16_t y) {
...
+    "adc %B0, r26 \n\t"
+    "mul %C1, %A2 \n\t"
+    "add r27, r0 \n\t"
+    "adc %A0, r1 \n\t"
+    "adc %B0, r26 \n\t"
+    "mul %B1, %A2 \n\t"
+    "add r27, r1 \n\t"
+    "adc %A0, r26 \n\t"
+    "adc %B0, r26 \n\t"
+    "lsl r27 \n\t"
```
DRracer, line 85: please explain the change from `lsr` to `lsl`

DRracer (changes requested):
I went through each commit separately. Most of them ale completely correct and contribute to lowering the code size greatly.
But I found some changes which I don't find completely correct. At least they need explanation and/or fixing.

Firmware/speed_lookuptable.h
leptun: @DRracer This is my conversation with Yuri about this

--leptun
Hello :wave:
I was looking at the speed lookup table asm functions and noticed some weird behavior:
```
	MultiU16X8toH16(x, 100, 42); // 0x1068 >>8 => 0x10
	MultiU16X8toH16(x, 101, 42); // 0x1092 >>8 => 0x10
	MultiU16X8toH16(x, 101, 43); // 0x10f7 >>8 => 0x11 ???
	MultiU16X8toH16(x, 95, 41);  // 0x0f37 >>8 => 0x10 ???
	MultiU16X8toH16(x, 95, 42);  // 0x0f96 >>8 => 0x0f ???
```
Why are odd numbers rounded up and even numbers rounded down? That really makes no sense. Is the function trying to round the >>8 but using the lsb instead of the msb by mistake?

--Yuri D'Elia
Huh, good question. No idea, we really need to check the asm. This sounds almost intentional :thinking_face:
Were you trying to use them for some other multiplication?

--leptun
No. I noticed I could optimize it a bit more
And while doing that I also noticed this odd behaviour
Considering this function is used for stepper timer calculation, it made me wonder if this weird rounding is introducing artifacts in accelerations

--Yuri D'Elia
These come from marlin 1.x (even before the MK2 fork)

--leptun
So they might be wrong

Yuri D'Elia
It could be. But at some point when I added the "c" code I did some for loops comparing the outputs and I didn't notice this odd rounding
I remember definitely the comment above the function was wrong. And I only understood what was it attempted to do by cloning marlin and trying to look at the history

--leptun
the c code does absolutely no rounding. So that's what's even weirder
It does a multiply and then right shifts the result. But it also fiddles with this least significant bit for some reason
Btw, here is my proposed optimization:
image
Do you see anything that could go wrong by doing so?
The old code was adding r1 to the result and then adding that rounding as the carry of "lsr". r26 was 0, so they were just adding the carry to the result
My optimization simply adds the carry at the same time r1 is added to the result, avoiding those "adc r26" commands
Both my code and the old code exhibit this weird rounding issue
If instead of "lsr r0" I do "lsl r0", then that will get the most significant bit of r0, so it actually rounds the output correctly
lsl and lsr are logical shift btw. They push either the msb or the lsb to the carry and shift the register by 1.
image
image

Comparison:
```
	asm("nop\n\t");//                             lsr       lsl
	MultiU16X8toH16(x, 100, 42); // 0x1068 >>8 => 0x10      0x10
	MultiU16X8toH16(x, 101, 42); // 0x1092 >>8 => 0x10      0x11
	MultiU16X8toH16(x, 101, 43); // 0x10f7 >>8 => 0x11 ???  0x11
	MultiU16X8toH16(x, 95, 41);  // 0x0f37 >>8 => 0x10 ???  0x0f
	MultiU16X8toH16(x, 95, 42);  // 0x0f96 >>8 => 0x0f ???  0x10
```
We could also technically disable the rounding completely and do just like the C code, but I think doing the rounding correctly might help, unlike the broken rounding which could oscillate between two values with the right conditions

--leptun
Look how it oscillates in this scenario (with lsr like before)
image
And this is the same, but with lsl instead of lsr
image

MultiU24X24toH16 also oscillates
image
And, not surprising at all, changing the lsr there to lsl also fixes those oscillations:
image

--leptun
I did two test prints, the surface looks the same :confused:

leptun: Please keep in mind that I managed to optimize the 16x8 function even more in the meantime, but the stuff mentioned above still holds

Firmware/planner.cpp
leptun: Before:
```
  previous_speed[0] = 0.0;
   3152a:	sts	0x0532, r1	; 0x800532 <_ZL14previous_speed.lto_priv.327>
   3152e:	sts	0x0533, r1	; 0x800533 <_ZL14previous_speed.lto_priv.327+0x1>
   31532:	sts	0x0534, r1	; 0x800534 <_ZL14previous_speed.lto_priv.327+0x2>
   31536:	sts	0x0535, r1	; 0x800535 <_ZL14previous_speed.lto_priv.327+0x3>
  previous_speed[1] = 0.0;
   3153a:	sts	0x0536, r1	; 0x800536 <_ZL14previous_speed.lto_priv.327+0x4>
   3153e:	sts	0x0537, r1	; 0x800537 <_ZL14previous_speed.lto_priv.327+0x5>
   31542:	sts	0x0538, r1	; 0x800538 <_ZL14previous_speed.lto_priv.327+0x6>
   31546:	sts	0x0539, r1	; 0x800539 <_ZL14previous_speed.lto_priv.327+0x7>
  previous_speed[2] = 0.0;
   3154a:	sts	0x053A, r1	; 0x80053a <_ZL14previous_speed.lto_priv.327+0x8>
   3154e:	sts	0x053B, r1	; 0x80053b <_ZL14previous_speed.lto_priv.327+0x9>
   31552:	sts	0x053C, r1	; 0x80053c <_ZL14previous_speed.lto_priv.327+0xa>
   31556:	sts	0x053D, r1	; 0x80053d <_ZL14previous_speed.lto_priv.327+0xb>
  previous_speed[3] = 0.0;
   3155a:	sts	0x053E, r1	; 0x80053e <_ZL14previous_speed.lto_priv.327+0xc>
   3155e:	sts	0x053F, r1	; 0x80053f <_ZL14previous_speed.lto_priv.327+0xd>
   31562:	sts	0x0540, r1	; 0x800540 <_ZL14previous_speed.lto_priv.327+0xe>
   31566:	sts	0x0541, r1	; 0x800541 <_ZL14previous_speed.lto_priv.327+0xf>
```
After:
```
  memset(previous_speed, 0, sizeof(previous_speed));
   3152c:	ldi	r30, 0x32	; 50
   3152e:	ldi	r31, 0x05	; 5
   31530:	movw	r26, r30
   31532:	st	X+, r1
   31534:	dec	r24
   31536:	brne	.-6      	; 0x31532 <main+0xae0>
```

Firmware/Marlin_main.cpp
leptun: this was fixed in a later commit:
```
                for (uint8_t y = MESH_NUM_Y_POINTS; y-- > 0;) {
                    for (uint8_t x = 0; x < MESH_NUM_X_POINTS; x++) {
```

Firmware/Marlin_main.cpp
leptun: One should be enough

Firmware/mesh_bed_calibration.cpp
leptun: done

Firmware/Marlin_main.cpp
leptun: done

Firmware/cardreader.h
leptun: Just looked again. Nothing uses it as a signed number. Only as a file count / file index (0 indexed)

Firmware/eeprom.cpp
leptun: I don't understand. I simply moved the initialization of a eeprom variable from the status screen to eeprom_init(). I don't see the issue in doing this.

Firmware/ultralcd.cpp
leptun: I'll do a factory reset later and look at the statistics. i already tested the welcome message change, so that is fine, but haven't tested the statistics init yet. Even though that should be fine, I'll test it anyway a bit later

Firmware/stepper.cpp
leptun: I disagree. Operations on multiple bits at a time are much faster.

This is quicker:
```
endstop_hit = 0;
   3262e:	sts	0x0574, r1	; 0x800574 <_ZL11endstop_hit.lto_priv.323>
###################
 if( endstop_hit) {
   3253a:	lds	r24, 0x0574	; 0x800574 <_ZL11endstop_hit.lto_priv.323>
   3253e:	and	r24, r24
```
Single bit read is Just as fast as before because it had to turn a byte into a bool. Now it uses sbrs instead of tst:
```
   if(endstop_hit & _BV(X_AXIS)) {
   32554:	lds	r24, 0x0574	; 0x800574 <_ZL11endstop_hit.lto_priv.323>
   32558:	sbrs	r24, 0
```

Single bit changes are indeed slower by default. I see where it could be a problem in stepper_check_endstops(), but I just managed to save 192B of flash just by caching volatile values in registers while inside stepper_check_endstops (perfectly safe because of the interrupt part). It's in fact now a lot faster than before. even with caching since I read and write to memory only once and otherwise just do ORI and ANDI on a register

Tested the functionality again using Home, Belt test, Z calibration and MBL. All worked correctly.

leptun:
Also fixed merge conflict

Firmware/mesh_bed_calibration.cpp
DRracer: I thought you'd add a reasonable message as well why the static assert fails :wink: 
Something like:
_MESH_MEAS_NUM_X_POINTS * MESH_MEAS_NUM_Y_POINTS > 255: it would overflow the uint8_t range - the following for-cycle would misbehave_

Firmware/Marlin_main.cpp
leptun: We already have code_value_float with a workaround for E notation. How should we proceed?

Firmware/stepper.cpp
```diff
@@ -230,7 +230,9 @@ bool endstops_hit_on_purpose()
 bool endstop_z_hit_on_purpose()
 {
   bool hit = endstop_hit & _BV(Z_AXIS);
+CRITICAL_SECTION_START;
```
DRracer, line 238: Nope, I'm strictly against it
The endstop_hit* flags must remain as standalone variables due to being used in time critical code. Any additional instructions are to be avoided.
Have a look at the assembly output if you don't believe :wink:

Firmware/speed_lookuptable.h
DRracer: ok, good enough, thank you

Firmware/stepper.cpp
DRracer: explained, closing this comment

DRracer (approved):

leptun:
@gudnimg Well done with this PR 🎉 Now the intense testing shall begin with what we have so far in 3.12.0 https://github.com/prusa3d/Prusa-Firmware/milestone/1?closed=1

Firmware/ConfigurationStore.cpp
leptun: Reference: https://www.nongnu.org/avr-libc/user-manual/group__avr__stdio.html#:~:text=The%20hh%20length%20modifier%20is%20ignored%20(char%20argument%20is%20promouted%20to%20int).%20More%20exactly%2C%20this%20realization%20does%20not%20check%20the%20number%20of%20h%20symbols.

PR #3370, opened 2022-02-01
Crash if pullups get enabled on the thermistor inputs

leptun:
WIth the hopes of tracking down why this happened

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -1754,6 +1754,10 @@ void stack_error() {
     crash_and_burn(dump_crash_reason::stack_error);
 }
 
+void pullup_error(bool fromTempISR) {
+    crash_and_burn(fromTempISR ? dump_crash_reason::bad_pullup_temp_isr : dump_crash_reason::bad_pullup_step_isr);
+}
+
 
 void trace();
 
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -500,6 +500,7 @@ void raise_z_above(float target, bool plan=true);
 
 extern "C" void softReset();
 void stack_error();
+void pullup_error(bool fromTempISR);
 
 extern uint32_t IP_address;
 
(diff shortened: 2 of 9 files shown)


mkbel (approved):
Looks OK for me.

I have no clue what branches do you use in this project and what are their purpose. So I can not tell if this is right target branch for this PR.

My comment on this is, that if firmware with this crash-on-error-feature is going to be released there should exist lower version numbered firmware which workarounds this potential problem without crash. So if user is getting crash with this firmware he can downgrade to firmware which is not crashing nor overheating nozzle.

Firmware/Marlin_main.cpp
```diff
@@ -1754,6 +1754,10 @@ void stack_error() {
     crash_and_burn(dump_crash_reason::stack_error);
 }
 
+void pullup_error(bool fromTempISR) {
+    crash_and_burn(fromTempISR ? dump_crash_reason::bad_pullup_temp_isr : dump_crash_reason::bad_pullup_step_isr);
```
DRracer, line 1758: please do the `#ifdef DEBUG_PULLUP_CRASH` in this function like this:

```
void pullup_error(bool fromTempISR) {
#ifdef DEBUG_PULLUP_CRASH
    crash_and_burn(fromTempISR ? dump_crash_reason.....
#else
    pull_up.disable();
#endif
}
```

This way you can avoid the #ifdefs at the calling places and you'd be ready for the production safety feature which turns off the pull-up if encountered.

DRracer (changes requested):
please improve the pullup_error as commented...

Firmware/Marlin_main.cpp
leptun: I don't like this approach. It would make no sense to constantly check if the pullups got enabled and then disable them. We can just disable them right away. Much quicker like this

DRracer: ok, then please add the "disable pull-up" into the #else branches at the correct spots in this PR

3d-gussner (approved):
Tested in MK404 and Test Pullup Crash works.

DRracer (approved):

PR #3377, opened 2022-02-05
Fix lang reserved area alignment

leptun:
I found something weird with the secondary language reserved area. It would appear as if the buffer is just randomly thrown in the PROGMEM. That wouldn't be a problem, except for the fact that we need to erase it. That part makes sense. We need to have it aligned to the page boundary. However, what makes no sense is how that is done:
```
#if (LANG_MODE != 0)
extern const char _SEC_LANG[LANG_SIZE_RESERVED];
extern const char* lang_get_translation(const char* s);
/** @def _SEC_LANG_TABLE
 *  @brief Align table to start of 256 byte page */
#define _SEC_LANG_TABLE ((((uint16_t)&_SEC_LANG) + 0x00ff) & 0xff00)
#endif //(LANG_MODE != 0)
```
I'm especially troubled by that define. Yes, it does align the pointer to the beginning of the next page... but what about the unused space at the beginning? And if that doesn't worry you, what about the missing space at the end??
Think of the following scenario: _SEC_LANG is misaligned and we fill a language to the limit. When we copy the language from xflash, where does the end of the table go? Does it just overwrite a bit of the following progmem?
I think in order to fix this, we just tell the compiler to align the page, as it should have done from the beginning:
```
const char _SEC_LANG[LANG_SIZE_RESERVED] __attribute__((aligned(SPM_PAGESIZE))) PROGMEM_I2 = "_SEC_LANG";
and
#define _SEC_LANG_TABLE ((uint16_t)&_SEC_LANG)
```
Before the changes:
```
&_SEC_LANG        = 0x051e
sizeof(_SEC_LANG) = 0x3000
&_lang_table0     = 0x0600                <--- aligned to next page by the define
```
After the changes:
```
&_SEC_LANG        = 0x0600
sizeof(_SEC_LANG) = 0x3000
&_lang_table0     = 0x0600                <--- already aligned by the compiler
```

This does however consume a bit more flash since the space before the aligned area is simply unused unfortunately. Maybe gcc will be smart enough one day to use that space efficiently.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -1242,9 +1242,9 @@ void setup()
 		printf_P(_n("_SEC_LANG_TABLE checksum = %04x\n"), sum);
 		sum = (sum >> 8) | ((sum & 0xff) << 8); //swap bytes
 		if (sum == header.checksum)
-			puts_P(_n("Checksum OK"), sum);
+			printf_P(_n("Checksum OK\n"), sum);
 		else
-			puts_P(_n("Checksum NG"), sum);
+			printf_P(_n("Checksum NG\n"), sum);
 	}
 	else
 		puts_P(_n("lang_get_header failed!"));
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -1242,9 +1242,9 @@ void setup()
 		printf_P(_n("_SEC_LANG_TABLE checksum = %04x\n"), sum);
 		sum = (sum >> 8) | ((sum & 0xff) << 8); //swap bytes
 		if (sum == header.checksum)
-			printf_P(_n("Checksum OK\n"), sum);
+			puts_P(_n("Checksum OK"));
 		else
-			printf_P(_n("Checksum NG\n"), sum);
+			puts_P(_n("Checksum NG"));
 	}
 	else
 		puts_P(_n("lang_get_header failed!"));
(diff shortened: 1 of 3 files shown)


leptun:
@gudnimg what do you think about this?

gudnimg:
I think this is a good idea.  I have not looked too deeply into the language code. I wonder if the start address/page of the table is always the same, does it depend on the program's size?

leptun:
It is not the same. It's whatever the linker decides to use. Other stuff in progmem will move this reserved space around

leptun:
I found out what happens if we fill up a language with a misaligned sec_lang space: update_lang.sh fails for the offending language. That script also aligns the sec lang space to the page size just like the define that I replaced. Now, this avoids a corrupted firmware, but doesn't solve the underlying problem: There is still wasted space that _could_ be used. @DRracer How should we proceed?

leptun:
I propose another approach to this issue which requires no changes in firmware (only linker script modifications):
https://github.com/3d-gussner/Arduino_Boards/pull/2
This change will come into the MK3 firmware via a new arduino boards version at a later time (but still into 3.12). I'll leave this PR open until the new arduino boards version is prepared.

leptun:
Update on this PR.
Both this and https://github.com/prusa3d/Arduino_Boards/pull/15 are needed simultaneously because the alignment must be done in firmware. Otherwise that small space after the vectors is wasted even with no _sec_lang reserved area.

leptun:
In combination with https://github.com/prusa3d/Prusa-Firmware/pull/3422, _SEC_LANG is finally page aligned and in the correct place.
image

@3d-gussner @DRracer fyi

3d-gussner (approved):
Tested with MK404 and reviewed the `lang-build.sh` results

Issue #1876, opened 2019-05-27
Feature Request: Software Disable MMU2


R0m3oF0x:
Please insert a **software setting** to **deactivate the MMU2**.

In my experience, the MMU2 cannot be used with abrasive filaments (stone and metal filaments) due to the long hoses. This only works if the filament is placed directly into the extruder (as it used to be), otherwise it gets jammed during printing.

Pulling the two plugs according to the MMU2 instructions is not a real solution in my opinion.

leptun:
Do you intend on having the ptfe tube disconnected or connected while having the mmu disabled?

I moddified the code to be able to disable it from menu, but did not do a pull request to the official firmware as I couldn’t test it and it would be kinda stupid just disabling the communication, but having the filament go through it.

If you want, I can port the changes to the latest version, but it will be english only.

R0m3oF0x:
Yes, I have the ptfe tube disconnected. Sounds just like I wished.

English shouldn't be a problem. Please port the changes.
Big THX!

leptun:
[FW371-Build2266-disable_MMU_from_lcd.zip](https://github.com/prusa3d/Prusa-Firmware/files/3227098/FW371-Build2266-disable_MMU_from_lcd.zip)

Here you go. Try it out an tell me how it works. The github project is https://github.com/leptun/Prusa-Firmware/tree/MK3_disable_MMU_from_lcd

One thing to note is that you need to reset the printer in order for it to have any effect.

mrcg:
I also would like to have feature to bypass the MMU2 for single printing.
It should also allow to leave the filament loaded in case of power off/on without the need to perform new load filament procedure when switching on. In about 95% of the cased i am printing with single filament and it's always a hassle to perform unload/load filament procedure when powering on the printer. Also in the forum lot of other user asking for single printing feature for the MMU2.

iamgnat:
I too would like to see this functionality be in the core firmware. As well as prints where it may be desirable to skip the MMU (due to filament or just avoiding extra complexity) it would also be an invaluable troubleshooting tool.

Sindarius:
This would be a great feature, I’d love to be able to do ninjaflex prints without going through the mmu.

leptun:
I might buy soon a MMU2S. If everything goes smoothly I plan to do some major changes to the communication layer between the printer and mmu that will also include this feature.

Zwirbel1:
Yes, please, this would be a great feature.

[...]

PR #3378, opened 2022-02-06
:construction:Extra optimizations for a future firmware

leptun:
This PR is composed of some optimizations which were left out of #3365. This PR is simply so that we don't lose track of these commits. There is no plan for now to merge this PR in 3.12.0

@gudnimg fyi


gudnimg:
Regarding the hypot commit. I will need to look at it from assembly level. It will take time. From first glance hypot seems to be slower, but I need to break down the subroutines to be sure.

Another note, I think we can convert the MarlinSerial ring buffer elements to 1 byte. (We are converting them to 2 bytes in places)

leptun:
We can profile those functions using the _micros as a timebase. I don't expect it to be _that_ much slower, but any speed penalty is not really welcome in the mbl code. It's already struggling right now when you push the stepper isr to the limit.
I also want to make sure the stepping is identical between implementations. If that is the case, then I see no issue with hypot.

I agree, we can change the ring_buffer to use uint8_t. We will also merge ring_buffer with rbuf at some point. One is used for UART0/1 and the other for UART2(MMU). We can use a common implementation, but for the purpose of this PR we should keep them split.
We could also optimize this further by making sure RX_BUFFER_SIZE is a multiple of 2 and that it's in the 1-256 range. This way we can use & instead of % for the index arithmetic

gudnimg:
Found this code is not used. `MMU_M600_SWITCH_EXTRUDER` is not defined anywhere and the function `choose_extruder_menu()` doesn't exit. Do you recognise this code @leptun ?
image

leptun:
It's dead SNMM (MMU1) code. It will be removed soon

gudnimg:
Took a look at `hypot()` commit. Measured the clock ticks with Timer 1 at 16MHz.

I start by first measuring the microseconds of the call `delayMicroseconds(1000)` just to see if my calculations are good. I then measure how many ticks it takes to calculate `sqrt(x * x + y * y)` where `x = 1000.2` and `y = 50.5`. Then calculate `hypot(x, y)` and check if the results are the same to 10 decimal places. Then I finally measure `delayMicroseconds(1000)` again to verify there hasn't been any drift.

Note that I assign the result from `sqrt(x * x + y * y)` and `hypot(x, y)` to a variable, that will have an effect on the tick measurement.

I see no loss in precision in my example and don't see much different in clock cycles. So I would say these two are basically the same.

Below are the results:

image

Arduino sketch with source code if anyone is interested:


<details>
<summary> Source Code </summary>
<br>

```cpp
float x;
float y;
volatile byte overflows = 0;

void setup()  
{ 
  Serial.begin(115200);
  while(!Serial);
  x = 1000.2;
  y = 50.5;
  float res = 0; // variable to force the program to store math result to memory

  Serial.println("|----------------- Verify 1000us Measurement -------------------|");
  Serial.flush(); // Make sure to flush so tick measurement is not affected by slow Serial stream
  
  TCCR0A = 0; // Disable Timer 0
  TCCR0B = 0; // Disable Timer 0
  TCCR1A = 0; // Set Timer 1 Control Register A to 0 
  TCCR1B = 0; // Set Timer 1 Control Register B to 0 
  TCCR1C = 0; // Set Timer 1 Control Register C to 0
  TIMSK1 = _BV(TOIE1); // Enable Time 1 overflow interrupt  

  TCNT1 = 0x0000; // Make sure tick count is 0
  TCCR1B = 1;  // Start Timer 1 with no prescaling (CLK = 16MHz)
  delayMicroseconds(1000); // Code being measured
  TCCR1B = 0;  // Stop Timer 1

  unsigned int temp_timer = TCNT1;  // Store ticks
  // Correct ticks in case there was overflow. -1 Here because ISR is called at start-up
  unsigned long ticks = ((((unsigned long)overflows - 1) << 16) | (unsigned long)temp_timer) - 4;

  Serial.print("Timer1 Ticks: ");
  Serial.println(temp_timer);
  Serial.print("Timer1 Overflows: ");
  Serial.println(overflows - 1);
  Serial.print("Timer1 Ticks after correcting for Overflows: ");
  Serial.println(ticks);

  Serial.print("Timer1 Ticks converted to microseconds: ");
  Serial.print((float)ticks/(F_CPU / 1000000), 4);
  Serial.println("us");

  Serial.println("|------------------ Measurement 1 -------------------|");
  Serial.flush();

  overflows = 0;       // Reset overflow counter
  TCNT1 = 0x0000;      // Make sure tick count is 0
  TCCR1B = 1;          // Start Timer 1 with no prescaling (CLK = 16MHz)
  res = sqrt(x*x+y*y); // Code being measured
  TCCR1B = 0;          // Stop Timer 1

  temp_timer = TCNT1;  // Store ticks
  // Correct ticks in case there was overflow
  ticks = ((((unsigned long)overflows) << 16) | (unsigned long)temp_timer) - 4;

  Serial.print("sqrt(x*x+y*y) = ");
  Serial.println(res, 10);
  Serial.print("Timer1 Ticks: ");
  Serial.println(temp_timer);
  Serial.print("Timer1 Overflows: ");
  Serial.println(overflows);
  Serial.print("Timer1 Ticks after correcting for Overflows: ");
  Serial.println(ticks);
  Serial.print("Timer1 Ticks converted to microseconds: ");
  Serial.print((float)ticks/(F_CPU / 1000000), 4);
  Serial.println("us");

  Serial.println("|------------------ Measurement 2 -------------------|");
  Serial.flush();

  overflows = 0;     // Reset overflow counter
  TCNT1 = 0x0000;    // Make sure tick count is 0
  TCCR1B = 1;        // Start Timer 1 with no prescaling (CLK = 16MHz)
  res = hypot(x, y); // Code being measured
  TCCR1B = 0;        // Stop Timer 1

  temp_timer = TCNT1; // Store ticks
  // Correct ticks in case there was overflow
  ticks = ((((unsigned long)overflows) << 16) | (unsigned long)temp_timer) - 4;

  Serial.print("hypot(x, y) = ");
  Serial.println(res, 10);
  Serial.print("Timer1 Ticks: ");
  Serial.println(temp_timer);
  Serial.print("Timer1 Overflows: ");
  Serial.println(overflows);
  Serial.print("Timer1 Ticks after correcting for Overflows: ");
  Serial.println(ticks);
  Serial.print("Timer1 Ticks converted to microseconds: ");
  Serial.print((float)ticks/(F_CPU / 1000000), 4);
  Serial.println("us");

  Serial.println("|----------------- Verify 1000us Measurement again -------------------| ");
  Serial.flush();

  overflows = 0;           // Reset overflow counter
  TCNT1 = 0x0000;          // Make sure tick count is 0
  TCCR1B = 1;              // Start Timer 1 with no prescaling (CLK = 16MHz)
  delayMicroseconds(1000); // Code being measured
  TCCR1B = 0;              // Stop the timer

  temp_timer = TCNT1;      // Store ticks
  // Correct ticks in case there was overflow
  ticks = ((((unsigned long)overflows) << 16) | (unsigned long)temp_timer) - 4;

  Serial.print("Timer1 Ticks: ");
  Serial.println(temp_timer);
  Serial.print("Timer1 Overflows: ");
  Serial.println(overflows);
  Serial.print("Timer1 Ticks after correcting for Overflows: ");
  Serial.println(ticks);
  Serial.print("Timer1 Ticks converted to microseconds: ");
  Serial.print((float)ticks/(F_CPU / 1000000), 4);
  Serial.println("us");
  Serial.flush();
  abort();
}

void loop() {}

ISR(TIMER1_OVF_vect)   
{  
  overflows++;  
}//end ISR 
```

</details>

leptun:
If you need to do measurement that requires clock precision in Prusa-Firmware (eg: measuring how long something takes in an ISR), you can do this very easily with TIM5. Beware that you need to use TIM3 on the miniRambo and that it has NO overflow handling.

At the end of setup:
```
    TIMSK5 = 0;
    TCCR5A = 0;
    TCCR5B = 1;
```
Where you want to measure:
```
    uint16_t initial = TCNT5;
    stepper_check_endstops(); //whatever function you want to measure
    printf_P(PSTR("time:%u\n"), TCNT5 - initial);
```
If a longer period is needed, increase TCCR5B (it increases the prescaler) and adjust your math accordingly. See datasheet for the prescaler values.


It is good that the code behaves correctly. I just hope it doesn't have any unforseen side effects.

gudnimg:
I have a feeling `buflen` in `cmdqueue.cpp` can be `uint8_t` instead of `int`. Even if all the commands in the buffer are 3 letters, Mxx or Gxx, that would still only fit 164 commands/lines. The command buffer only holds 493 bytes 🤔 

Highest I've seen so far for `buflen` is 16. I may look more into this next weekend as this can shave off 120 bytes of flash if possible.

leptun:
Please let's not touch the cmdqueue buffers for now. It works through misterious ways and I don't want to break anything there. A lot of the older lcd code uses it and I'm afraid we might break it quite easily accidentally.

gudnimg:
@leptun I noticed `endstops_hit_on_purpose()` function is not doing anything useful for us. We don't read it's return value anywhere we use it so it is just wasting clock cycles imo. Do you agree we should remove it? It saves 58 bytes of flash on my end.
https://github.com/prusa3d/Prusa-Firmware/blob/b654217a5b3d77816eaeeb5ea803cc41de5e9b74/Firmware/stepper.cpp#L228

Edit: Or instead of removing it, we remove the return value and just keep `endstop_hit = 0`. But then `checkHitEndstops()` can take its place I guess.

leptun:
I guess we can make endstops_hit_on_purpose() void and just clear endstop_hit. Please don't change it more than that

leptun:
@gudnimg Off topic question, do you have a MK3 or a MK2.5 with the PAT9125 filament sensor?

gudnimg:
I only have the MK3S+ printer

leptun:
@gudnimg make sure you don’t change something in the recently deleted snmm code. We should probably rebase this branch to the latest Mk3 to avoid this problem.

gudnimg:
@leptun Indeed, I'm trying to use the current MK3 as reference before switching to this branch. I'm really happy you removed the SNMM code 😄 

To add to my previous comment from 2 days ago... I got a MMU2S yesterday. So I will hopefully soon be able to test the MMU code directly 💪

leptun:
That’s great! Btw, see: https://github.com/prusa3d/Prusa-Firmware/issues/1876#issuecomment-1045965924

gudnimg:
@leptun I'm hesitant to run the rebase. Do you remember which git command you used before? I've usually used this guide https://medium.com/@topspinj/how-to-git-rebase-into-a-forked-repo-c9f05e821c8a to update my fork.

leptun:
@gudnimg I use gitkraken for everything git (unless that is not working, in which case I just use git directly). This is how I did the rebase above. I highly recommend you try this as well. There is a free version available that can be used for almost everything except private repos.

https://user-images.githubusercontent.com/17808203/154795545-9045ce25-6aef-45e9-a58a-2e79fe51abb0.mp4

As for why it's so zoomed-in, it's because I use it on a larger display normally and it looks correctly there. I moved it to my secondary screen for the screen recording so that the file size would be smaller, so I wouldn't have to compress the video

gudnimg:
Build fails. It looks like to me it is a problem in the current CMake script https://github.com/prusa3d/Prusa-Firmware/blob/MK3/CMakeLists.txt It doesn't know where to find `"Configuration_prusa.h"`

image

leptun:
Yes. You can't include "Configuration_prusa.h" in anything that has "unit tests" (loosely used word here). Please discard any changes to mmu related code (such is the case with the auto-deplete). This will be addressed later.

gudnimg:
@leptun Managed to somehow create 818 byte reduction in flash. This seems almost too good to be true. Do you see any issues I might be missing? https://github.com/prusa3d/Prusa-Firmware/pull/3378/commits/0b198bddf6c678f6adff8e043f08f52d8ebdf525

I tested this locally by changing the format of `myfloat`. I did not see any difference no matter how many digits/decimals there are or if the value is negative.

```c
  float myfloat = 123.6789;
  char buf[25] = {0};
  Serial.print("dtostrf() ->");
  sprintf_P(buf, PSTR("G92 E"));
  dtostrf(myfloat, 6, 3, buf + strlen(buf));
  Serial.println(buf);
  char buf2[25] = {0};
  Serial.print("sprintf() ->");
  sprintf_P(buf2,PSTR("G92 E%6.3f"),myfloat);
  Serial.println(buf2);
```

leptun:
@gudnimg I'd like to make a "PR3" of optimizations where we cherrypick changes from this one. I'd like to go for the smallest changes that have the largest flash impact (such as dtostrf).

Firmware/ultralcd.cpp
```diff
@@ -1031,11 +1031,11 @@ void lcd_commands()
...
 			setAllTargetHotends(0);  // reset all hotends temperature including the number displayed on the main screen
 			if (_Kp != 0 || _Ki != 0 || _Kd != 0) {
-			strcpy(cmd1, "M301 P");
+			strcpy_P(cmd1, PSTR("M301 P"));
 			strcat(cmd1, ftostr32(_Kp));
-			strcat(cmd1, " I");
+			strcat_P(cmd1, PSTR(" I"));
 			strcat(cmd1, ftostr32(_Ki));
-			strcat(cmd1, " D");
+			strcat_P(cmd1, PSTR(" D"));
```
leptun: Maybe this can be turned into a single sprintf_P. It's just building a command using the resulting PID values. Same for M303 above.

gudnimg: I thought about it but didn't have time to test it. `ftostr32` limits the string width, but `sprintf` wouldn't truncate the length.  

Edit: On second thought perhaps `snprintf_P` could work in that regard.

image

> Like [sprintf()](https://www.nongnu.org/avr-libc/user-manual/group__avr__stdio.html#ga6017094d9fd800fa02600d35399f2a2a), but instead of assuming s to be of infinite size, no more than n characters (including the trailing NUL character) will be converted to s.

> Returns the number of characters that would have been written to s if there were enough space.

gudnimg: Another note, slightly related to `ftostr32`. I've found that the code in general in `conv2str.cpp` can be optimised but 80% of the code is unused so perhaps we can just replace it by being a bit more smart with AVR's stdio library https://www.nongnu.org/avr-libc/user-manual/group__avr__stdio.html

gudnimg:
I replaced all `ftostr3` calls with `lcd_printf_P`. I didn't see memory savings until all `ftostr3` were replaced.

leptun:
Well, that makes sense. sprintf was also a function call, but by replacing all of them the ftostr3 was no longer included during linking.

Firmware/Marlin_main.cpp
```diff
@@ -8321,9 +8321,9 @@ SERIAL_PROTOCOLPGM("\n\n");
     */
 	case 861:
 		if (code_seen('?')) { // ? - Print out current EEPROM offset values
-			uint8_t cal_status = calibration_status_pinda();
 			int16_t usteps = 0;
-			cal_status ? SERIAL_PROTOCOLLN("PINDA cal status: 1") : SERIAL_PROTOCOLLN("PINDA cal status: 0");
+			SERIAL_PROTOCOL("PINDA cal status: ");
+			SERIAL_PROTOCOLLN(calibration_status_pinda());
```
leptun: ummm... can't this be the PGM variant?
So `SERIAL_PROTOCOLPGM("PINDA cal status: ");`

gudnimg: I think so. We can save a lot of SRAM (~50 bytes) but the flash usage will increase about 60 bytes if I remember correctly.

leptun: Flash increase... That sounds wrong. What if you use a printf_P (also for pinda status, ternary evaluation to 1:0)?

leptun: `printf_P(PSTR("PINDA cal status: %u\n"), calibration_status_pinda() ? 1 : 0);` sounds correct to me. I hope it's optimal. In any case, I'll always trade flash for ram. Issues caused by ram can be very deceiving, while a lack of flash means the build will fail (obviously trivial to debug).

leptun: Please don't commit to this. I have an awesome improvement. I'll squash this commit into the original Pinda status commit

leptun: See https://github.com/prusa3d/Prusa-Firmware/pull/3378/commits/aa7aefd6f844c470f1616d9a88632dd0827417ef

leptun:
In case of emergency, we could try to use `-fmerge-all-constants`:
> -fmerge-all-constants
Attempt to merge identical constants and identical variables.
This option implies -fmerge-constants. In addition to -fmerge-constants this considers e.g. even constant initialized arrays or initialized constant variables with integral or floating point types. Languages like C or C ++ require each variable, including multiple instances of the same variable in recursive calls, to have distinct locations, so using this option will result in non-conforming behavior.

Doing just a quick compile test resulted in 800B of savings. But as I said, only in case of emergency since it can have nasty side effects.

I wonder if this option merges duplicate PROGMEM strings. If it does, it would be amazing.

leptun:
@gudnimg Rebased onto latest MK3 and removed commits already merged

leptun:
Todo: Optimize `M115 U` and `M862.4`. There is a lot of duplicate code with these two

gudnimg:
Closing in favor of https://github.com/prusa3d/Prusa-Firmware/pull/4281

PR #3381, opened 2022-02-08
Pre-release updates and some fixes

3d-gussner:
- [x] `MSG_WIZARD_DONE c=20 r=8` to `c=20 r=3`
- [x] `MSG_YES c=3` change to `c=4`
- [x] Change `[um]` to `[μm]`
    - [x] Including `lang-im/export.sh` to convert HD44780 to UTF-8 and vice versa
- [x] Enable `lang-clean.sh` to clean up non active languages
- [x] Update German translation
- [x] `Bed level correct` is a submenu of `Calibration`. Changed the return from `Settings` to `Back`
- [x] Fix MSG in cz and hu as these have been different to other lang_en*.txt files
- [x] Update all po files

### DIFF
--- a/Firmware/messages.c
+++ b/Firmware/messages.c
@@ -120,7 +120,7 @@ const char MSG_WIZARD_HEATING[] PROGMEM_I1 = ISTR("Preheating nozzle. Please wai
 const char MSG_WIZARD_QUIT[] PROGMEM_I1 = ISTR("You can always resume the Wizard from Calibration -> Wizard."); ////c=20 r=8
 const char MSG_WIZARD_WELCOME[] PROGMEM_I1 = ISTR("Hi, I am your Original Prusa i3 printer. Would you like me to guide you through the setup process?"); //// c=20 r=7
 const char MSG_WIZARD_WELCOME_SHIPPING[] PROGMEM_I1 = ISTR("Hi, I am your Original Prusa i3 printer. I will guide you through a short setup process, in which the Z-axis will be calibrated. Then, you will be ready to print."); ////c=20 r=16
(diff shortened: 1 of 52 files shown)


DRracer (approved):
hopefully the "micro" sign doesn't cause issues on non-standard LCDs

leptun:
> hopefully the "micro" sign doesn't cause issues on non-standard LCDs

It better not cause problems. Those LCDs better have the A00 font identical to the hd44780 they are emulating

leptun (approved):
The changes look correct to me.
However, we must find an universal way to translate between TXT and PO files the translations. This will be problematic when instead of 1 custom character, we have 96.

Slight issue: In the error messages, \xe4 will be parsed in lang-check and the warnings/errors will show the wrong character. This should probably be addressed when we figure out the universal translation between encodings (utf8->lcd and back)

3d-gussner:
> > hopefully the "micro" sign doesn't cause issues on non-standard LCDs
> 
> It better not cause problems. Those LCDs better have the A00 font identical to the hd44780 they are emulating

I have somewhere an OLED display. So to final we should know.

3d-gussner:
> The changes look correct to me. However, we must find an universal way to translate between TXT and PO files the translations. This will be problematic when instead of 1 custom character, we have 96.
> 
> Slight issue: In the error messages, \xe4 will be parsed in lang-check and the warnings/errors will show the wrong character. This should probably be addressed when we figure out the universal translation between encodings (utf8->lcd and back)

Yes the lang-check.py shows now different char, but this is something I would like to get done until final.

PR #3383, opened 2022-02-09
LCD dynamic characters (+diacritics support)

leptun:
Post merge tasks:

|Language status | PO files | `TRANS_CHARS` |
| -------------------------- | ------------------------- |------------------------------ |
|cs :czech_republic: :heavy_check_mark: | :heavy_check_mark:          | :heavy_check_mark:   |
|de :de: :heavy_check_mark: | :heavy_check_mark:                        | :heavy_check_mark:  |
|es :es: :heavy_check_mark: | #4440 :heavy_check_mark:         | :heavy_check_mark:   |
|fr :fr: :heavy_check_mark: | #4439 :heavy_check_mark: , #4441 :heavy_check_mark:        | :heavy_check_mark:   |
|hr :croatia: |           |    |
|hu :hungary: |           | :heavy_check_mark:    |
|it :it: :heavy_check_mark: | #4442 :heavy_check_mark: | :heavy_check_mark:   |
|nl :netherlands: :heavy_check_mark: | :heavy_check_mark:           |:heavy_check_mark:    |
|no :norway: |           | :heavy_check_mark:   |
|pl :poland: :heavy_check_mark: | #4411 :heavy_check_mark: | :heavy_check_mark:   |
|ro :romania: :heavy_check_mark: | :heavy_check_mark:          | :heavy_check_mark:   |
|sk :slovakia: :heavy_check_mark: | :heavy_check_mark:          | :heavy_check_mark:   |
|sv :sweden:  |           |    |

- [x] Update font table based on the frequency of each diacritic in the PO files. To be done after all PO files are updated. #4456

This PR adds a system which allows using more than 8 custom characters in a simple way.
For reference, the hd44780 lcd controller that is used on the printer has a hardcoded font. Unfortunately, the variant of the chip uses the A00 font, which is the japanese font. But, regardless of hardcoded font, 8 custom characters can be created. Previously, the firmware had 2 banks of custom characters and manually switched between these two fonts. This allowed more than 8 custom characters, but the menu code would become more complicated because of this and the characters had to be carefully placed in the correct bank. 
To counter the manual switching limitation, I implemented a system where the two banks were switched automatically. This allowed up to 16 custom characters, with the condition that characters from different banks (but with the same ID) were never displayed at the same time. This is something that worked for the few custom characters that were used, but was not flexible enough for actual diacritics (since lots of custom characters are needed for this).

This PR is an improvement on the previous idea I had. Instead of having character banks, we treat each CGRAM slot as a custom character that can be allocated as needed. This means that at any given time, at most 8 distinct custom characters can be displayed, but there is no limitation to what those custom characters can be or in what order they appear. This approach allows much more custom characters to be implemented, but with some limitations.

The encoding used for the custom characters in strings is also improved. Instead of using ascii codes 0-7 as the LCD expects, we instead map the custom characters to extended ascii 0x80-0xDF, the code internally remapping the custom characters to the correct 0-7 character code based on what is loaded in CGRAM at the time. This approach allows up to 96 custom characters as long as no more than 8 distinct custom characters are on screen at the same time.

In case a screen uses more than 8 custom characters (which can happen with translations that have many diacritics in words), only the first 8 custom characters are rendered. All other custom characters will instead use the "alternate" form, that being the closest representation from the A00 font for that symbol (which is in most cases just the basic ascii variant).

In order to save flash, the font data is also compressed. This is possible since only 5 bits per row are used, so 3 bits per byte in the font data would be wasted. So instead of 8 bytes needed for the custom font data, only 5 bytes are needed per character. The compression is done by the language scripts generating the font table, the decompression is done by the firmware. The compression layout was chosen so that it would be the most efficient for the AVR to decompress. I wrote the decompression code by hand in assembly since that would result in the most flash savings. I tried my best to make the code readable, but there's only so much you can do about assembly readability.

Not all custom diacritics are going to be implemented however. There's too many diacritics to fit the 96 character limit. So some have to be dropped. The ones which are used the most infrequently have been chosen to be dropped. In most cases, these are the uppercase variants of some diacritics. For those, the lowercase version is used instead since it looks similar enough. Other symbols such as `¿¡` are simply going to be removed by the language preprocessing scripts.

most recent video:

https://github.com/prusa3d/Prusa-Firmware/assets/25530011/2fa6726b-d752-444c-a7bc-c8dd73350fe5

older videos:

https://github.com/prusa3d/Prusa-Firmware/assets/17808203/f7558791-6957-4c92-88ab-e8fe942970c6

https://github.com/prusa3d/Prusa-Firmware/assets/17808203/9914e607-7dab-46ed-ba26-33e62d0b88ef


leptun:
@gudnimg I found out why the status screen was breaking in e62bc4686f452f3222860146d28b7769b06200fd. Initially, I thought it was because of the custom characters since I've seen something really similar before (@TojikCZ can confirm). It might have been that, but what 100% is the problem with the status screen in that commit was changing from lcd_printf_P to lcd_puts_P. Turns out, puts_P does NOT return the number of characters that were printed, but rather a non-negative number for success. In our case, that would be 0. Changing that back to printf solved the issue, even though we are not providing any argument to the printf function besides the fmt. It might not be as efficient as lcd_puts_P, but since we require the return to mean the number of wrtitten characters, we have to sacrifice a few bytes of flash for this.

3d-gussner:
@leptun Can you please rebase it and I would like to present it. Milestone would be 3.14.0

3d-gussner:
Hello community translators @vintagepc @stelgenhof @Hauzman @AttilaSVK @Prime1910 @ingbrzy @shatter136 @Painkiller56 @OS-kar @trondkla,

We are preparing this Pull Request to show diacritics on the LCD screen.

- Please check the https://github.com/prusa3d/Prusa-Firmware/blob/MK3/lang/lib/charset.py if we have all characters in that list
  -  We would like to have all characters in that list EVEN if we don't use them at the moment.
    - To save some flash we may NOT show some them with the diacritics.
      - Example: `ÿ` is never used but we gonna keep it prepared in the source code BUT get into the firmware to save resources    
  -  Please point out which aren't used at all even exist in the alphabet of the language
     - Example: In Romanian the `â` can't be used at the beginning or the end of an word, this also means that the `Â` isn't used at all.
- Please review the current po files are translated correctly with diacritics.
  - If you can also try to shorten the translations to save additional space for future messages.  But that is only optional. Please focus on diacritics.
  - Side note: We are limited to 8 custom characters per 20x4 lines shown (page), so if you find a page with more than 8 custom chars please consider to change the translation.

arekm:
Polish looks good (all characters are there in charset.py).

leptun:
@arekm What about the rest of the questions in @3d-gussner's message? Any diacritics which aren't really used so we know that we can skip in the font table?

@Hauzman I know that romanian doesn't have the diacritics for older messages. Do you think you could add those please and make another PR with that?

arekm:
@leptun Polish translation is in a bad state for that - missing diacritics in almost all translations that should have them.

Started looking at that but it needs more work (https://github.com/arekm/Prusa-Firmware/commit/f86a6575e6689215a3ce5fa16668119a05568a16 ; commit also changes some weird translations). 

For now (my repository) at this moment "Ó", "Ć" and "Ł" are not used. The rest of PL diacritics is in use. 

Shorting messages will be problematic. Some are already too short (marked few with XXX to check if these can be a bit longer). Also terminology isn't unified and one thing is named using different words (which also should be checked against printer manual to use uniform ones).

ingbrzy:
Slovak looks good

Ro3Deee:
S-cedilla is not  in TRANS_CHARS in charset.py.

S-cedilla (majuscule: Ş, minuscule: ş) is a letter used in some of the [Turkic languages](https://en.wikipedia.org/wiki/Turkic_languages). It occurs in the [Azerbaijani](https://en.wikipedia.org/wiki/Azerbaijani_alphabet), [Gagauz](https://en.wikipedia.org/wiki/Gagauz_language), [Turkish](https://en.wikipedia.org/wiki/Turkish_alphabet), and [Turkmen](https://en.wikipedia.org/wiki/Turkmen_alphabet) alphabets. It is also planned to be in the Latin-based [Kazakh alphabet](https://en.wikipedia.org/wiki/Kazakh_alphabets).

S-comma (majuscule: Ș, minuscule: ș) is a letter which is part of the [Romanian alphabet](https://en.wikipedia.org/wiki/Romanian_alphabet), used to represent the sound /ʃ/. In early versions of [Unicode](https://en.wikipedia.org/wiki/Unicode), the [Romanian](https://en.wikipedia.org/wiki/Romanian_alphabet) letter [Ș](https://en.wikipedia.org/wiki/S-comma) (S-comma) was considered a glyph variant of Ş (S-cedilla), and therefore was not present in the Unicode Standard. It is also not present in the Windows 1250 (Central Europe) [code page](https://en.wikipedia.org/wiki/Code_page). The letter was only added to the standard in Unicode 3.0, and some texts in Romanian still use Ş instead.

Prime1910:
Looks good

3d-gussner:
> @leptun Polish translation is in a bad state for that - missing diacritics in almost all translations that should have them.
> 
> Started looking at that but it needs more work ([arekm@f86a657](https://github.com/arekm/Prusa-Firmware/commit/f86a6575e6689215a3ce5fa16668119a05568a16) ; commit also changes some weird translations).
> 
> For now (my repository) at this moment "Ó", "Ć" and "Ł" are not used. The rest of PL diacritics is in use.
> 
> Shorting messages will be problematic. Some are already too short (marked few with XXX to check if these can be a bit longer). Also terminology isn't unified and one thing is named using different words (which also should be checked against printer manual to use uniform ones).

@arekm Thanks for the feedback and your time. I used yesterday the VSCode Spell checkers and also found tons of missing diacritics especially in Polish. I know that shortening is hard but not the primary goal here.

3d-gussner:
> Looks good

Are you sure? I found lot of words with the VS Code Spell Checker Croatian
```
#. MSG_SELFTEST_FS_LEVEL c=20
#: ../../Firmware/ultralcd.cpp:6660
msgid "%s level expected"
msgstr "%s level ocekivan"
```
may should be `očekivan` or is the spell checker bad?

Prime1910:
> > Looks good
> 
> Are you sure? I found lot of words with the VS Code Spell Checker Croatian
> 
> ```
> #. MSG_SELFTEST_FS_LEVEL c=20
> #: ../../Firmware/ultralcd.cpp:6660
> msgid "%s level expected"
> msgstr "%s level ocekivan"
> ```
> 
> may should be `očekivan` or is the spell checker bad?

I will look once more on laptop

3d-gussner:
> Slovak looks good

VS Code spell check also show here lot misspelled words
Example:
```
#. MSG_IR_03_OR_OLDER c=18
#: ../../Firmware/Filament_sensor.cpp:280
#: ../../Firmware/Filament_sensor.cpp:357 ../../Firmware/messages.cpp:173
msgid " 0.3 or older"
msgstr " 0.3 a starsie"
```
may should be `staršie`

ingbrzy:
> > Slovak looks good
> 
> VS Code spell check also show here lot misspelled words Example:
> 
> ```
> #. MSG_IR_03_OR_OLDER c=18
> #: ../../Firmware/Filament_sensor.cpp:280
> #: ../../Firmware/Filament_sensor.cpp:357 ../../Firmware/messages.cpp:173
> msgid " 0.3 or older"
> msgstr " 0.3 a starsie"
> ```
> 
> may should be `staršie`

Ah sorry, missed second part of your message.. will review po file and fix diacritics..

trondkla:
All norwegian characters are in the file.

Only thing I find weird is that æ is translated into ä, which does not exist in norweigan.
We would use ae (a & e) but if this is a limitation, and you have to use one symbol, ä is most likely good enough, because most norwegians know the swedish alphabet.

3d-gussner:
> Only thing I find weird is that æ is translated into ä, which does not exist in norweigan.
> We would use ae (a & e) but if this is a limitation, and you have to use one symbol, ä is most likely good enough, because most norwegians know the swedish alphabet.

One of the reasons we would like to use the diacritics to avoid this.
In German ä|ö|ü|ß can be written as ae|oe|ue|ss but this auto replace 1 to 2 chars may "break" limited messages as it may would exceed these limits.

3d-gussner:
Again thanks guys for so fast responses! :hugs:

AttilaSVK:
The charset.py file is okay, but the Hungarian translation is partially without diacrirics, so I'll need to update that.

leptun:
> S-cedilla is not in TRANS_CHARS in charset.py.
> 
> S-cedilla (majuscule: Ş, minuscule: ş) is a letter used in some of the [Turkic languages](https://en.wikipedia.org/wiki/Turkic_languages). It occurs in the [Azerbaijani](https://en.wikipedia.org/wiki/Azerbaijani_alphabet), [Gagauz](https://en.wikipedia.org/wiki/Gagauz_language), [Turkish](https://en.wikipedia.org/wiki/Turkish_alphabet), and [Turkmen](https://en.wikipedia.org/wiki/Turkmen_alphabet) alphabets. It is also planned to be in the Latin-based [Kazakh alphabet](https://en.wikipedia.org/wiki/Kazakh_alphabets).
> 
> S-comma (majuscule: Ș, minuscule: ș) is a letter which is part of the [Romanian alphabet](https://en.wikipedia.org/wiki/Romanian_alphabet), used to represent the sound /ʃ/. In early versions of [Unicode](https://en.wikipedia.org/wiki/Unicode), the [Romanian](https://en.wikipedia.org/wiki/Romanian_alphabet) letter [Ș](https://en.wikipedia.org/wiki/S-comma) (S-comma) was considered a glyph variant of Ş (S-cedilla), and therefore was not present in the Unicode Standard. It is also not present in the Windows 1250 (Central Europe) [code page](https://en.wikipedia.org/wiki/Code_page). The letter was only added to the standard in Unicode 3.0, and some texts in Romanian still use Ş instead.

Thanks for the answer. We'll add them to TRANS_CHARS so that ş gets rewritten to ș

ingbrzy:
Updated SK po file with diacritics

https://github.com/leptun/Prusa-Firmware/pull/12

leptun:
@ingbrzy Thanks! I merged your PR. What about unused characters? Are there any of uppercase or lowercase diacritics that are used extremely infrequently or not at all?

ingbrzy:
> @ingbrzy Thanks! I merged your PR. What about unused characters? Are there any of uppercase or lowercase diacritics that are used extremely infrequently or not at all?

unused characters in translation are those uppercase letters:
```
Ä
É
Í (uppercase letter i)
Ĺ
Ň
Ó
Ŕ
Ú
Ý

```

Ro3Deee:
> Thanks for the answer. We'll add them to TRANS_CHARS so that ş gets rewritten to ș

There is also T-comma (majuscule: Ț, minuscule: ț)  which  was introduced only in Unicode 3.0.0, but also T-cedilla (majuscule: Ţ, minuscule: ţ). In early versions of Unicode, the Romanian letter [Ț](https://en.wikipedia.org/wiki/T-comma) (T-comma) was considered a glyph variant of Ţ (T-cedilla), and therefore was not present in the Unicode Standard. It is also not present in the [Windows-1250](https://en.wikipedia.org/wiki/Windows-1250) (Central Europe) [code page](https://en.wikipedia.org/wiki/Code_page). The letter was only added to the standard in Unicode 3.0 (1999), and some texts in Romanian still use Ţ instead.

@leptun , so Ţ and ţ should be added to TRANS_CHARS so that ţ gets rewritten to ț.

arekm:
> What about unused characters? Are there any of uppercase or lowercase diacritics that are used extremely infrequently or not at all?

Wouldn't be worth to automate such statistics? https://pypi.org/project/polib/

(Max length translation annotations could also be added to source code and then validated if translations do not cross such limit with xgettext --add-comments=XYZ and polib supports accessing such comments)

stelgenhof:
These seem to be the most frequently used diacritics in the Dutch language (Source in Dutch): https://nl.wikipedia.org/wiki/Accenttekens_in_de_Nederlandse_spelling#Frequentie

These diacritics are generally used with words of French origin, to put emphasis on the semantics or speech, or if words are from foreign origin. Considering this, we could safely ignore diacritics in this context as for firmware use we highly likely will not see words of French or other foreign origin. (Can totally ignore the emphasis rules).

leptun:
Just implemented á. The rest of the diacritics should now be easy to implement. The language scripts now properly handle the font table.

image

ingbrzy:
> Just implemented á. The rest of the diacritics should now be easy to implement. The language scripts now properly handle the font table.
> 
> image

Sorry but it looks ugly if all letters are not in the same level..

shatter136:
Exactly as @ingbrzy wrote. But perhaps all text could be pushed down to á level?

leptun:
Some diacritics can be made to be at the same level, for others this is not possible. The A00 font sadly has the leters one pixel higher than A02 font. The font I used for the diacritics is from the A02 font.

@shatter136 I can't edit the default font sadly. Only the diacritics can be "custom"

I could move the diacritic one pixel higher, but then there will be no space between the accent and the base letter.

image
image

leptun:
This is what it would look like if I push the base letter to the same level as other letters: 
image

And for others I really need that space between the base letter and the thing above:
image
image

leptun:
How it looks on screen:

image

shatter136:
@leptun I see. This is so much better. As you wrote though, question is how the other diacritics would look like if they are pushed up...

leptun:
Another thing to consider is diacritics that have things below them...
Take ț for example which is shifted one pixel higher than t
image

ingbrzy:
> How it looks on screen:
> 
> image

This is the way..

ingbrzy:
@leptun what about upper case letter like Č ? 
how it will looks like?

leptun:
What I was planning to use, with some minor modifications here and there. Not all of these would be implemented in the end due to being too many diacritics. Some are also duplicates between languages

image image image image image image image image image

trondkla:
@leptun is the norwegian/danish Ø (https://en.wikipedia.org/wiki/Close-mid_front_rounded_vowel) in A01 then? I only found the lower case ø on those pictures you posted :) 

EDIT: ah they are different hight, so you use either or, then using the swedish Ö or det lower case ø is most likely good enough :) Can't satisfy the whole wørld xD

leptun:
I didn't implement norwegian yet. I created that font before community languages were a thing. There isn't any A01 font. Only A00 (the one used on the printer) and A02 (extremely rare, only for reference).

trondkla:
Thanks for great answer and thanks for the work you are doing 💪 👍

leptun:
All diacritics that are used are implemented now. Here is SK for example:

https://github.com/prusa3d/Prusa-Firmware/assets/17808203/f7558791-6957-4c92-88ab-e8fe942970c6

ingbrzy:
> All diacritics that are used are implemented now. Here is SK for example:
> 
>  trim.EB8A9D32-BFF0-4B41-83BE-2FEDD12163E6.MOV

looks perfect! Good job!

ingbrzy:
> All diacritics that are used are implemented now. Here is SK for example:
> 
>  trim.EB8A9D32-BFF0-4B41-83BE-2FEDD12163E6.MOV

image

ď letter is still one pixel down..

leptun:
@ingbrzy Can't really avoid that. Some characters had to be one pixel off vertically. Most of them are not like that however.

3d-gussner:
@ingbrzy If you can draw a better `ď` please let us know.
Here a tool @leptun doesn't like but I do  https://maxpromer.github.io/LCD-Character-Creator/ set `Data type to hex`

ingbrzy:
> @ingbrzy If you can draw a better `ď` please let us know. Here a tool @leptun doesn't like but I do https://maxpromer.github.io/LCD-Character-Creator/ set `Data type to hex`

thanks for the tool!

here is better letter ď

image

```
#include <LiquidCrystal.h>

LiquidCrystal lcd(12, 11, 5, 4, 3, 2); // RS, E, D4, D5, D6, D7

byte customChar[] = {
  0x03,
  0x03,
  0x02,
  0x0E,
  0x12,
  0x12,
  0x12,
  0x0E
};

void setup() {
  lcd.begin(16, 2);
  lcd.createChar(0, customChar);
  lcd.home();
  lcd.write(0);
}

void loop() { }
```

leptun:
The one you drew is still shifted. The last row must be empty for it to not be shifted

shatter136:
> All diacritics that are used are implemented now. Here is SK for example:
> 
>  trim.EB8A9D32-BFF0-4B41-83BE-2FEDD12163E6.MOV

🤩 perfection.

ingbrzy:
> The one you drew is still shifted. The last row must be empty for it to not be shifted

ah ok..

fixed
image

```
#include <LiquidCrystal.h>

LiquidCrystal lcd(12, 11, 5, 4, 3, 2); // RS, E, D4, D5, D6, D7

byte customChar[] = {
  0x03,
  0x03,
  0x02,
  0x0E,
  0x12,
  0x12,
  0x0E,
  0x00
};

void setup() {
  lcd.begin(16, 2);
  lcd.createChar(0, customChar);
  lcd.home();
  lcd.write(0);
}

void loop() { }
```

leptun:
I asked internally what they think about this change. Personally, I still prefer how it is now.

leptun:
@3d-gussner @ingbrzy Also for future reference, I don't need the output code from that tool. I just need a screenshot of the character/grid so I can quickly replicate it in the tool I'm using (GLCD font creator, with windows xp SP3 compatibility layer enabled so that saving the .lcd file works correctly).

leptun:
@ingbrzy I am going to switch to your version of `ď`. But there are still issues with other letters. Here are the problematic ones:

image image image image image image image image image image image image

ingbrzy:
> @ingbrzy I am going to switch to your version of `ď`. But there are still issues with other letters. Here are the problematic ones:
> 
> image image image image image image image image image image image image

none of them are used in SK language.. :)

leptun:
Well yeah, but they are used in other languages. So they might need some adjustments still. With some of them it's doable, but with others I simply see no way to make it not shifted while still being recognizable as the correct letter/accent.

leptun:
I've started an internal vote for variants to the letters with issues. If anyone is interested in this, please vote as well. I'm adding pictures to the messages since I don't feel like retyping everything.

image
image
image

arekm:
PL for review: https://github.com/prusa3d/Prusa-Firmware/pull/4411

leptun:
@arekm Thanks for the PR. It is not for me to handle the official translations. I will let @3d-gussner deal with this.

trondkla:
@leptun do you have a sentence that hows how options 1 would look like in a full sentence?
Personally I would rather have it jump and be more recognicable. Since it's not the longest sentences :)

leptun:
@trondkla this is for norwegian:

image

leptun:
There is a slight issue at the moment with ț. It is incorectly being replaced with µ. I'll handle that issue as well soon

sarusani:
Probably a very stupid question, but why do the "standard" letters have the empty line at the bottom and not at the top? Wouldn't switching that solve the issue with the special characters without having to alter them?

leptun:
The standard A00 font has a line at the bottom. Since that font is hardcoded, we can't change it. The reason they placed the letters like that is because of the lcd's "cursor feature. One of the cursor modes underlines the current letter. The underline uses the last row of pixels. If the letters were to use that row, then the cursor would partly overwrite the letter. We don't use the cursor, but hitachi thought people might want to use that feature, hence why the A00 font has lowercase letters one pixel higher than the bottom row. As for why the space doesn't exist in the A02 font, it's because their diacritics are aligned at the bottom row. So it would make sense for the ascii to also be aligned like that. I hope this answers your question.

Firmware/lcd.cpp
```diff
@@ -348,22 +362,56 @@ static uint8_t __attribute__((noinline)) lcd_get_row_offset(uint8_t row)
 void lcd_set_cursor(uint8_t col, uint8_t row)
 {
 	lcd_set_current_row(row);
-	lcd_command(LCD_SETDDRAMADDR | (col + lcd_get_row_offset(lcd_currline)));
+    uint8_t addr = col + lcd_get_row_offset(lcd_currline);
+	lcd_ddram_address = addr;
+	lcd_command(LCD_SETDDRAMADDR | addr);
```
gudnimg, line 378: Do we need `uint8_t addr`?

I wonder if this saves any flash?
```c
        lcd_ddram_address = col + lcd_get_row_offset(lcd_currline);
	lcd_command(LCD_SETDDRAMADDR | lcd_ddram_address);
```

Firmware/lcd.cpp
```diff
@@ -348,22 +362,56 @@ static uint8_t __attribute__((noinline)) lcd_get_row_offset(uint8_t row)
...
 }
 
 void lcd_set_cursor_column(uint8_t col)
 {
-	lcd_command(LCD_SETDDRAMADDR | (col + lcd_get_row_offset(lcd_currline)));
+	uint8_t addr = col + lcd_get_row_offset(lcd_currline);
+	lcd_ddram_address = addr;
+	lcd_command(LCD_SETDDRAMADDR | addr);
```
gudnimg, line 385: Same here, do we need `uint8_t addr`?

Firmware/lcd.cpp
leptun: This accesses a global variable. So The code would end up storing to ram, then reading back from ram, which is wasteful. If you use the local variable, then only the store operation to ram will happen and then the cached values in registers will be passed to the functions.

Firmware/menu.cpp
```diff
@@ -73,6 +73,7 @@ void menu_start(void)
 		menu_top = lcd_encoder;
     menu_line = menu_top;
     menu_clicked = lcd_clicked(); // Consume click event
+	if (lcd_draw_update) lcd_frame_start();
```
gudnimg: Would it make sense to call `lcd_frame_start()` in `lcd_update()`? Then `lcd_frame_start()` is only called in one place. LCD updates are always enabled in the menus.

leptun: Hmmm, not really. Because we don't want to start a new frame if the entire screen isn't redrawn. I think if we did this in `lcd_update()`, then `lcd_frame_start()` gets called every 100ms, which is not what I want.

Firmware/lcd.cpp
gudnimg: Ok if current change is faster, then that's good 👍 I haven't looked at the assembly output.

leptun:
It is now possible to use diacritics in the sheet name. Of course, existing limitations apply (no more than 8 distinct custom characters on the screen at once).

image

Firmware/menu.cpp
leptun: Scrap my last comment. You were right. It is better to add the frame start in the lcd update function

ingbrzy:
> It is now possible to use diacritics in the sheet name. Of course, existing limitations apply (no more than 8 distinct custom characters on the screen at once).
> 
> image

and is it possible to translate that time message?

leptun:
It could be translated, but do note that if it used any extra diacritics, those would not be rendered because I already used all slots available. 5 are always used on the status screen, so this leaves only 3 slots for the sheet name and last row message. Should the status message translation exceed the 8 slots, the alternate form of those characters would be used instead.

As for this specific message, it seems to not be translated. I don't see why it couldn't be translated. In any case, that is beyond the scope of this PR.

leptun:
I think these are all the changes I intend for this PR. The plan is to have this PR reviewed and merged some time next week. The PO file translations for the other languages will come in follow-up PRs. The plan is to release 3.14 hopefully by the end of next month. Hopefully all PO files will be ready by then.

If anyone wants to play with the PR, here are the builds of the last commit I just pushed:
[FW3.13.1+PR3383+aee02c49d.zip](https://github.com/prusa3d/Prusa-Firmware/files/12774804/FW3.13.1%2BPR3383%2Baee02c49d.zip)

Firmware/lcd.h
```diff
@@ -171,21 +171,27 @@ class LcdUpdateDisabler
...
+#define LCD_STR_CONFIRM      "\x89"
 #define LCD_STR_SOLID_BLOCK  "\xFF"  //from the default character set
 
-extern void lcd_set_custom_characters(void);
-extern void lcd_set_custom_characters_nextpage(void);
+struct CustomCharacter {
+    uint8_t colByte;
+    uint8_t rowData[4];
+    char alternate;
+};
```
gudnimg: Nitpick: I feel this struct should be moved to lcd.cpp to limit the scope of it. `CustomCharacter` is only used in lcd.cpp

Firmware/ultralcd.cpp
```diff
@@ -627,8 +627,8 @@ void lcdui_print_status_line(void) {
 //! @endcode
 void lcdui_print_status_screen(void)
 {
-
-    lcd_set_cursor(0, 0); //line 0
+    lcd_frame_start();
```
gudnimg, line 630: Do we need `lcd_frame_start()` here? I think `lcd_update()` takes care of it now.

Firmware/Marlin_main.cpp
```diff
@@ -4093,8 +4092,7 @@ void process_commands()
 		}
 		else if (code_seen_P(PSTR("RESET"))) { // PRUSA RESET
 #if defined(XFLASH) && defined(BOOTAPP)
-            boot_app_magic = BOOT_APP_MAGIC;
-            boot_app_flags = BOOT_APP_FLG_RUN;
```
gudnimg, line 4097: Just wondering, is there any risk with this change?

Firmware/ultralcd.cpp
```diff
@@ -2814,7 +2813,7 @@ bool lcd_calibrate_z_end_stop_manual(bool only_z)
             }
         }
         // Let the user confirm, that the Z carriage is at the top end stoppers.
-        uint8_t result = lcd_show_fullscreen_message_yes_no_and_wait_P(_i("Are left and right Z~carriages all up?"), false);////MSG_CONFIRM_CARRIAGE_AT_THE_TOP c=20 r=3
+        uint8_t result = lcd_show_fullscreen_message_yes_no_and_wait_P(_i("Are left and right Z-carriages all up?"), false);////MSG_CONFIRM_CARRIAGE_AT_THE_TOP c=20 r=3
```
gudnimg, line 2816: `~` is usually parsed in the full screen code as "white space". We need to test this message just to be sure it still renders OK.

See in `lcd_display_message_fullscreen_nonBlocking_P`:

```
if (c == '~')
    c = ' ';
```

Firmware/ultralcd.cpp
```diff
@@ -5055,11 +5049,11 @@ static void lcd_rename_sheet_menu()
         lcd_encoder = menuData->name[0];
         menuData->initialized = true;
     }
-    if (lcd_encoder < '\x20') lcd_encoder = '\x20';
-    if (lcd_encoder > '\x7F') lcd_encoder = '\x7F';
+    if (lcd_encoder < 0x20) lcd_encoder = 0x20;
+    if (lcd_encoder > 0xFF) lcd_encoder = 0xFF;
```
gudnimg, line 5051: Just adding a note that this should be tested specifically. Can a sheet name now be a complete white block like we use in the MMU filament load test progress bar? :D

Firmware/ultralcd.cpp
```diff
@@ -7284,6 +7280,7 @@ void menu_lcd_lcdupdate_func(void)
 	{
 		if (lcd_draw_update) {
 			lcd_timeoutToStatus.start();
+            lcd_frame_start();
```
gudnimg, line 7281: @leptun I wonder, do we want to call this function also when the knob is rotated? I'm thinking if we can then remove the call to `lcd_frame_start()` in `lcd_display_message_fullscreen_nonBlocking_P`

`lcd_knob_update` is called even when LCD updates are disabled. I wonder if that will suffice to cover all fullscreen messages.

leptun: Idk, it feels weird tying the start of a frame with the knob update. Need to think more about this.

Firmware/ultralcd.cpp
leptun: Yup :P. The entire charset can be used now

Firmware/ultralcd.cpp
leptun: Oh damn. I didn't know that was the case. We should definitely check this message. If it renders fine, then I propose we remove this stupid exception.

Firmware/ultralcd.cpp
leptun: Hmmm. Maybe this will work. I need to test this.

Firmware/ultralcd.cpp
gudnimg: I was right. I think this is a feature not a bug 😄

Firmware/Marlin_main.cpp
leptun: There shouldn't be any risk. It's actually better now. You can look at the bootapp code in the bootloader to check it
https://github.com/prusa3d/stk500v2-prusa/blob/91fc4fa85c8fe001384a6d70359d113abd0325ec/stk500boot.c#L777

If we set the bootapp magic to anything besides 0x55aa55aa. then a watchdog reset will make the bootloader start the firmware immediately. Moreover, BOOT_APP_FLG_RUN is not a thing and it never was a thing.

Firmware/lcd.h
leptun: Will move it soon 👍

Firmware/ultralcd.cpp
gudnimg: Hmm I just noticed this is the only message that was using `~` like this? Seems like a low hanging fruit to remove it to simplify the code. I thought this was used in more messages.

gudnimg: Some translations may need to be updated. Like this one:

```
"Calibrating XYZ. Rotate the knob to move the Z carriage up to the end "
"stoppers. Click when done."
msgstr ""
"Kalibrace XYZ. Otacenim tlacitka posunte Z osu az k~hornimu dorazu. Potvrdte"
" tlacitkem."
```

leptun: @gudnimg I see what this is supposed to do. It's used to enforce that `Z` and `carriage` are not split between lines. This is stupid. A dash between the words is just as effective and it doesn't require this stupid hack. Do you know if any other message makes use of this feature?

gudnimg: It seems to be a bit inconsistent between translations. Some use `~` and some don't.
I found this one as well:

```
#. MSG_WIZARD_WELCOME_SHIPPING c=20 r=12
#: ../../Firmware/messages.cpp:126 ../../Firmware/ultralcd.cpp:3803
msgid ""
"Hi, I am your Original Prusa i3 printer. I will guide you through a short "
"setup process, in which the Z-axis will be calibrated. Then, you will be "
"ready to print."
msgstr ""
"Bok, ja sam vas Original Prusa i3 printer. Vodit cu vas kroz kratak proces "
"postavljanja, u kojem ce se kalibrirati Z~os. Tada cete biti spremni za "
"printanje."
```

Firmware/ultralcd.cpp
gudnimg: It seems to work as it is so I won't complain :) Tested using MMU error screen.

Firmware/ultralcd.cpp
leptun: I pushed a commit. I think I removed all tildes. We need to check how the text looks now. Worst case, we enforce a "page" break in the message so that words split between lines better.

Firmware/ultralcd.cpp
leptun: I thought about this more and I think I don't want to remove this. Thing is that for the status screen, the text on it can change even without interacting with the knob (take status message updates for example). So it makes sense that a full redraw might start even without interacting with the printer. This is in contrast to menus where only interacting with them will cause a redraw. So I'll keep this here

Firmware/lcd.cpp
```diff
@@ -809,151 +868,77 @@ void lcd_buttons_update(void)
...
+
+static void lcd_print_custom(uint8_t c) {
+	uint8_t charToSend = pgm_read_byte(&Font[c - 0x80].alternate); // in case no empty slot is found, use the alternate character.
+	int8_t slotToUse = -1;
+
+	for (uint8_t i = 0; i < 8; i++) {
+		// first check if we already have the character in the lcd memory
+		if ((lcd_custom_characters[i] & 0x7F) == (c & 0x7F)) {
+			lcd_custom_characters[i] = c; // mark the custom character as used
+			charToSend = i; // send the found custom character id
```
gudnimg, line 899: Would it make sense to add a constant or preprocessor symbol for `0x80` and `0x7F`?

This is my understanding:

* `0x7F` means custom character is unused
* `0x80` not sure what this one means

Firmware/ultralcd.cpp
gudnimg: `lcd_updatestatus()` sets `lcd_draw_update = 1;` This means `menu_lcd_lcdupdate_func` should be calling `lcd_frame_start()` before rendering the status screen:

```
if (lcd_draw_update) {
    lcd_timeoutToStatus.start();
    lcd_frame_start();
}
(*menu_menu)();
```

From what I can find in Vscode, it looks like `lcd_status_screen` is always called as a menu.

Firmware/lcd.cpp
leptun: Custom characters start at 0x80. 0x7F is the mask that is used for the custom character ID. Bit 7 has a different purpose. It is used to keep track of custom character usage. So if it's used on screen in the current frame, then bit 7 will be set. Frame start clears this bit for all characters. This bit is used to determine whether a character slot is available to be used (requires loading new character data to CGRAM, so slower) or potentially available (data already loaded, but it wasn't used yet on the current frame, will be discarded on the next frame start if unused this frame). A discarded character slot (or "empty" slot) is marked with the magic value 0x7F, not to be confused with the mask 0x7F used elsewhere.
I know this code is a bit messy because of these constants, but I really wanted to use every single bit of ram that I am allocating for this feature :)

Firmware/ultralcd.cpp
gudnimg: Only exception I can find is my hack in the MMU code, to render the solid blocks when LCD updates are disabled.

```c
void TryLoadUnloadReporter::Render(uint8_t col, bool sensorState) {
    // Set the cursor position each time in case some other
    // part of the firmware changes the cursor position
    lcd_insert_char_into_status(col, sensorState ? LCD_STR_SOLID_BLOCK[0] : '-');
    if (!lcd_update_enabled) lcdui_print_status_line();
}
```

Firmware/lcd.cpp
leptun: Hmmm, maybe I can use a bitset for this instead of masking. Not sure if the code will end up as efficient tho...

gudnimg: > A discarded character slot (or "empty" slot) is marked with the magic value 0x7F, not to be confused with the mask 0x7F used elsewhere.

Ok this is good to know :) I just see the same value everywhere and thought they had the same "meaning"

Firmware/ultralcd.cpp
leptun: That is not the only time the status screen changes. A change in displayed data can also happen based on the printing status (say SD printing or USB printing or idle, etc...). So I prefer to have an explicit frame start for the status screen

gudnimg: Marking this as resolved since there is no risk involved.

gudnimg (approved):
Note my testing is limited to only English language.

Tested the following on English only and Multilang build (with English language only)

* ✅ MMU error screen rendering
* ✅ MMU try-load-unload progress bar
* ✅ Changing sheet name
* ✅ MMU Load to Nozzle and Unload via LCD menu
* ✅ Running printer self-test

DRracer (approved):
nice work :+1:

PR #3389, opened 2022-02-11
Allow SN get only on the Einsy

leptun:
The miniRambo has no special 32u2 firmware, so don't get the SN there. Fixes an octoprint timeout on connect for the miniRambo variants

PFW-1284

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -1130,6 +1130,7 @@ void setup()
     }
 #endif //TMC2130
 
+#ifdef PRUSA_SN_SUPPORT
     //Check for valid SN in EEPROM. Try to retrieve it in case it's invalid.
     //SN is valid only if it is NULL terminated and starts with "CZPX".
     {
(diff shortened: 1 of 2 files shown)


leptun:
Quick test on the miniRambo:
```
;S
SN update failed
start
echo: 3.11.0-RC1-4842
echo: Last Updated: Feb 11 2022 17:36:10 | Author: (none, default config)
Compiled: Feb 11 2022
echo: Free Memory: 2183  PlannerBufferBytes: 1760
echo:Hardcoded Default Settings Loaded
adc_init
FSensor DISABLED
echo:SD init fail
LCD status changed
echo:SD init fail
```
Even though the 32u2 there has a SN, the verison of that code is so old that it doesn't recognize ;S (and not even the version command from usb). So I've disabled support there as it made no sense to ask for the SN when there is a 100% chance of failure.

DRracer (approved):
To be merged after 3.11.0-RC release

3d-gussner (approved):

PR #3390, opened 2022-02-11
Fix Octoprint connect because of SN GET on blank einsy boards

### LINKED ISSUE #3319: [BUG]Replacement non-genuine Einsy board timeouts connecting to OctoPrint from 3.10.1-rc onwards
**Printer type** - MK3S
**Printer firmware version** -  3.10.1-rc and newer

**SD card or USB/Octoprint**
Octoprint 1.7.2

**Describe the bug**
Replacement non-genuine Einsy board timeouts connecting to OctoPrint from 3.10.1-rc onwards. Simply this is because non-genuine board does not have CZPX...... serial number, and new code that harvests SN in CZPX format from AT90 prevents OctoPrint from connecting.
  
**To Reproduce**

  
**Expected behavior**
non-genuine boards are able to connect to OctoPrint.

leptun:
Depends on #3389
Fixes #3319 

PFW-1284

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -8974,7 +8974,7 @@ bool FarmOrUserECool(){
 
 #ifdef PRUSA_SN_SUPPORT
 void WorkaroundPrusaSN() {
-    const char *SN = PSTR("CZPX000000000000000");
+    const char *SN = PSTR("CZPXInvalidSerialNr");
     for (uint8_t i = 0; i < 20; i++) {
         eeprom_update_byte((uint8_t*)EEPROM_PRUSA_SN + i, pgm_read_byte(SN++));
     }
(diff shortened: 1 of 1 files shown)


Firmware/ultralcd.cpp
```diff
@@ -8972,6 +8972,15 @@ bool FarmOrUserECool(){
     return farm_mode || UserECoolEnabled();
 }
 
+#ifdef PRUSA_SN_SUPPORT
+void WorkaroundPrusaSN() {
+    const char *SN = PSTR("CZPX000000000000000");
```
DRracer: We decided not to assign a valid serial number to such printers, instead let's assign something human-readable:
`CZPXInvalidSerialNr`
should contain the same number of characters as the originally proposed:
`CZPX000000000000000`

DRracer (changes requested):
Please change the serial nr text as requested, then we should be good to go.

Firmware/ultralcd.cpp
leptun: done

DRracer (approved):
To be merged after 3.11.0-RC release

3d-gussner (approved):

PR #3395, opened 2022-02-12
crashDetection using sfilt

leptun:
The crash detection counters basically filtered the signal from the drivers so that 4 fullstep cycles of stall happen before a crash is issued. The funny thing is the drivers already have such a setting in hardware: sfilt.
In this proof of concept I've made it so that when we are doing crashDetection, sfilt is enabled (so the stallguard has additional filtering). When homing (or any other state where exact stall position is important), sfilt is disabled just like before. So the only affected behaviour is crashDetection while printing.
A quick test didn't reveal any issues with this approach: the printer did not crash by itself, but I was able to cause crashes to be detected if I blocked the axis with my fingers (just like before this PR).
Doing this in hardware instead of software saves 218B of flash and 16B of ram.
Something I found funny was that the MK2.5 variants had a define for crash detection counters... That was a mistake for sure as those printers don't have tmc2130 drivers.

PFW-1374
PFW-1393

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -2300,7 +2300,7 @@ bool calibrate_z_auto()
 #ifdef TMC2130
 static void check_Z_crash(void)
 {
-	if (READ(Z_TMC2130_DIAG) != 0) { //Z crash
+	if (!READ(Z_TMC2130_DIAG)) { //Z crash
 		FORCE_HIGH_POWER_END;
 		current_position[Z_AXIS] = 0;
 		plan_set_position_curposXYZE();
--- a/Firmware/mesh_bed_calibration.cpp
+++ b/Firmware/mesh_bed_calibration.cpp
@@ -974,7 +974,7 @@ bool find_bed_induction_sensor_point_z(float minimum_z, uint8_t n_iter, int
 		goto error;
 	}
 #ifdef TMC2130
-	if (READ(Z_TMC2130_DIAG) != 0)
+	if (!READ(Z_TMC2130_DIAG))
 	{
 		//printf_P(PSTR("crash detected 1, current_pos[Z]: %f \n"), current_position[Z_AXIS]);
 		goto error; //crash Z detected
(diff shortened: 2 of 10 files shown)


leptun:
Relevant section from the datasheet:
>14.2 StallGuard2 Update Rate and Filter
The StallGuard2 measurement value SG is updated with each full step of the motor. This is enough to safely detect a stall, because a stall always means the loss of four full steps. In a practical application, especially when using CoolStep, a more precise measurement might be more important than an update for each fullstep because the mechanical load never changes instantaneously from one step to the next. For these applications, the sfilt bit enables a filtering function over four load measurements. The filter should always be enabled when high-precision measurement is required. It compensates for variations in motor construction, for example due to misalignment of the phase A to phase B magnets. The filter should be disabled when rapid response to increasing load is required and for best results of sensorless homing using StallGuard.

leptun:
I've changed the DIAG pin mode from push-pull to open-collector. I have no idea why it was chosen to push-pull in the first place. It was a dangerous choice considering the pins are tied together:
image
There is also an external pullup, so the designers of the einsy clearly wanted to use open-collector in their implementation.

In any case, changing from push-pull to open-collector changed the polarity of the signal, so I also switched it accordingly in code. This also fixed a random crash detected on startup. I'm doing a longer test print right now. No false crashes so far (30 minutes)
I've also removed a duplicate initialization of the diag pins in st_init()

leptun:
No false crashes so far:
image
I tested and I could still trigger a crash if I pinched the axis by hand (tested in new print).
Now I'm testing how it performs with much higher velocities and acceleration.
image
image

leptun:
Finished the speedy benchy, no false crashes

leptun:
Rebased onto latest MK3 branch. Preparing feature test.

DRracer (approved):
Code changes look correct to me, hopefully these changes improve behavior of the TMC drivers at specific scenarios.
And of course, there is this obvious benefit of code and RAM saving...

PR #3403, opened 2022-02-13
Fix usb printing timer

leptun:
I hope this fixes the usb printing timer. Before this PR, the usb printing state would get enabled, but never time out. I also managed to optimize it more than before.
@gudnimg Please check this. I think we were the ones that broke it in the first place.

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -328,11 +328,10 @@ extern uint8_t host_keepalive_interval;
 
 extern unsigned long starttime;
 extern unsigned long stoptime;
+extern ShortTimer usb_timer;
 extern int bowden_length[4];
-extern bool is_usb_printing;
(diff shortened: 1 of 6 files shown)


gudnimg:
This commit may have broken it https://github.com/prusa3d/Prusa-Firmware/pull/3365/commits/66782e9c9dc21ced6ccfaec5c904a3ce09e76917 the `start()` statement is unreachable

leptun:
Yep. That's the one. Hopefully I managed to fix it here. It's too late for me now to continue checking. I'll verify this more during the week

wavexx (approved):

wavexx:
Took me a while to realize that ``timer::expired()`` also sets the "is_running" state. I'd love maybe a better name for this function.
But fixing it here would be wrong.

Everything looks good and much better!

PR #3405, opened 2022-02-14
Fixup suspicious indentation in lcd_show_sensors_state

wavexx:
Cleanup some odd indentation in the function lcd_print_state(), which makes the missing braces highly suspicious.

- Expand all tabs in the block.
- Add a relevant comment about the pat9125_update() call
- Align the columns to make it look nice :)
- Remove useless //// comments and non-existing translation IDs

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -3797,9 +3797,9 @@ static void lcd_print_state(uint8_t state)
 //! @code{.unparsed}
 //! |01234567890123456789|
 //! |PINDA N/A  FINDA N/A|  MSG_PINDA c=5 MSG_FINDA c=5
-//! |Fil. sensor      N/A|  MSG_FSENSOR 
-//! |Xd    000  Yd    000|  MSG_XD
-//! |Int   000  Shut  000|  
+//! |Fil. sensor      N/A|  MSG_FSENSOR
+//! | Xd   000    Yd  000|
+//! |Int   000  Shut  000|
 //! ----------------------
 //! @endcode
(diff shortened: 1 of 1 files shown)


Firmware/ultralcd.cpp
```diff
@@ -3843,26 +3843,18 @@ static void lcd_show_sensors_state()
...
-		//}
-	}
+    if (mmu_enabled == false)
+    {
+        // pat9125_update is already called while printing: only update manually when idling
+        if (!moves_planned() && !IS_SD_PRINTING && !is_usb_printing && (lcd_commands_type != LcdCommands::Layer1Cal))
+            pat9125_update();
+
+        lcd_set_cursor(0, 2);
+        lcd_printf_P(_N(" Xd: %4d    Yd: %3d\n"
```
leptun: What's with this change in the fmt? Wouldn't it make more sense to have it %6d as to handle the entire signed 16 bit number? I'd also change the spaces a bit if I did this.

Also please update the doxygen menu "render" if you change the layout a bit

wavexx: The formatting was badly aligned, it was looking out of place.
While printing this is unlikely to get anywhere higher than 100 as it's reset regularly.

But I was able to increase this by cheating, push filament in super-fast, then slowly pull it out (or just do the opposite), so that I could go higher.

I rearranged the layout so that we can show a full %6d. Looks a bit empty by comparison:

2022-02-14T233407

leptun (approved):
Looking good now

DRracer (approved):

PR #3407, opened 2022-02-15
Disable pullup crash test

leptun:
Disable the pullup test now. Only constantly clear the pullups for the adc pins with digital circuitry disabled.


3d-gussner (approved):
Approved

DRracer (approved):

PR #3409, opened 2022-02-16
MK3 3.10.2 crash on thermistor pullup

3d-gussner:
Re-merge to MK3_3.10.2-RC branch to prepare FW3.10.2-RC1 with debug and crash


DRracer (approved):
this is correct for 3.10.2-RC

PR #3410, opened 2022-02-16
Fix crashDet timer overflow handling

leptun:
I'm pretty confident that the crashDet timer was not expiring properly. Consider the following scenario:
- We get a crash detected at millis=1000
- The print recovers. At crashDet timer gets started at millis=1000
- Another crash happens at millis = 71000
- The crashDet timer is checked again for .expired(), but because it is a shortTimer, the time value is actually millis=5464 in the shortTimer code. In this scenario a shortTimer would say NOT expired (wrong), while a LongTimer would say that it expired and reset the crashDet_counter to 0 (correct). This happens because we are not constantly polling .expired(). So if the time between polls is more than shortTimer can handle (~65s), the expired event will be missed entirely.

I hope my scenario makes sense. If anyone can test the claims I'd be grateful.
Uses 8B of flash and 2B of ram (code size increase). So negligible.

I git blame @gudnimg :) . Can we check the optimizations PR to see if any similar case exists?

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -200,7 +200,7 @@ unsigned long pause_time = 0;
 unsigned long start_pause_print = _millis();
 unsigned long t_fan_rising_edge = _millis();
 LongTimer safetyTimer;
-static ShortTimer crashDetTimer;
+static LongTimer crashDetTimer;
 
 //unsigned long load_filament_time;
 
--- a/Firmware/cmdqueue.cpp
+++ b/Firmware/cmdqueue.cpp
@@ -24,7 +24,7 @@ int serial_count = 0;  //index of character read from serial line
 bool comment_mode = false;
 char *strchr_pointer; // just a pointer to find chars in the command string like X, Y, Z, E, etc
 
-ShortTimer farm_incomplete_command_timeout_timer;
+LongTimer farm_incomplete_command_timeout_timer;
 
 long gcode_N = 0;
 long gcode_LastN = 0;
--- a/Firmware/mmu.cpp
+++ b/Firmware/mmu.cpp
@@ -75,9 +75,9 @@ int16_t mmu_version = -1;
 
 int16_t mmu_buildnr = -1;
 
-ShortTimer mmu_last_request;
-ShortTimer mmu_last_response;
-ShortTimer mmu_last_finda_response;
+LongTimer mmu_last_request;
+LongTimer mmu_last_response;
+LongTimer mmu_last_finda_response;
 
 MmuCmd mmu_last_cmd = MmuCmd::None;
 uint16_t mmu_power_failures = 0;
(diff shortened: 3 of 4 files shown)


gudnimg:
Seems I broke a lot of timers :P When I check my commit message https://github.com/prusa3d/Prusa-Firmware/pull/3365/commits/ef8083a1485545cb6bd351939eb8213ee12b6da4 I seem to have thought at the time the expired value would fit into `uint16_t`.

leptun:
@gudnimg Well, the expired value would fit, but that assumes the .expired() is checked quicker than the shortTimer overflows (65536ms), so we must make sure that's the case at all times.

leptun:
@gudnimg I propose we search for all ShortTimer occurances in the code and update accordingly if the period between checks is not consistent, even if technically the expired amount that is checked fits into uint16_t.

leptun:
I also think I could close this PR and have a broader PR that fixes all the shortTimer timers, not only the crashDet one. What do you think? I can also use reverts instead of new commits if possible.

gudnimg:
I see, then this issue may be present in my other changes as well. I've only changed the timers in our optimisation PR which was recently merged https://github.com/prusa3d/Prusa-Firmware/pull/3365/commits

Yes I think a larger PR is good. We can have a checklist of all the timers to double-check. They aren't too many to be unmanageable :P

leptun:
👍 Ok.
I'm busy with uni stuff right now. If you have time, please prepare that PR and I'll close mine when you open yours.

gudnimg:
@leptun no problem. I will probably only start testing on friday. But I will create the PR today.

wavexx (approved):

wavexx:
The list of changed timers in #3365 seems to be:

```
tmc2130_overtemp_timer
autostart_atmillis
farm_incomplete_command_timeout_timer
tIRsensorCheckTimer
mmu_last_request
mmu_last_response
mmu_last_finda_response
```

Is there any other recent PR we should look at? (just to exclude existing short timers)

leptun:
Yes, I think only #3365 changed timers

gudnimg:
I’m thinking we should convert them all to `LongTimer` and then later we can optimise them one at a time for 3.12 or after. We’ve saved so much flash storage that this won’t be a problem memory wise 🤔

leptun:
I'm preparing now a PR with even more Short timers converted to long timers. Turns out there are even more that are vulnerable, but not all

DRracer (approved):

DRracer:
@leptun we can introduce some kind of detection how much bytes a timer needs just to get an idea what is safe and what is not. Then we can change the corresponding timers to more appropriate byte-length

leptun:
@DRracer Actually, we can't. The problem is not how big the .expired() value is, but rather how often .expired() is polled. If the polling in not frequent enough (~65s for ShortTimer), then an overflow of the timebase is not detected.

leptun:
Did not mean to break the pull request... Making a new one now 🤦

Issue #3321, opened 2021-12-17
[BUG]<Firmware upgrade impossible since 3.10.0>


LR-J:
**Printer type** -  MK3
**Printer firmware version** -  3.10.0 and 3.10.1
(files prusa3d_fw_MK3_3_10_1_4697.hex and prusa3d_fw_MK3_3_10_0_4481.hex)

**SD card or USB/Octoprint**
  upgrade via USB

**Describe the bug**
  Since firmware 3.10.0, I can't do the update on my machine. The logs do not show any errors. After the upgrade, the screen stays stuck on the welcome message, as if the boot sequence never ended. Downgrade to 3.9.3 is ok.
  
**To Reproduce**
  upgrade via usb using slic3r 2.3.3 on Ubuntu 18.04.
  
**Expected behavior**
  A normal boot sequence after upgrade...
  
**G-code**
  no relevant.

3d-gussner:
@LR-J Sorry to hear that.
Can you open a terminal to the USB after you have flashed the firmware with PrusaSlicer and report back the logfile?
- If there is a Serial number shown please remove some parts of it from the logfile before uploading.
A short video would also be helpful.
Is it a genuine Prusa MK3 or do you have a third party EINSY?

LR-J:
I have a 14 hours printing on going. I'll do this tomorrow.
It's a genuine Prusa MK3.

LR-J:
Here is the log :
[MK3_firmware update log file.md](https://github.com/prusa3d/Prusa-Firmware/files/7738955/MK3_firmware.update.log.file.md)

LR-J:
Here is the behaviour I get for the boot sequence...


https://user-images.githubusercontent.com/15336633/146635075-6baf842a-69be-40bd-98d8-144a2b80015d.mp4

gudnimg:
Could you try reading from the Serial port using [PuTTy](https://www.putty.org/) or similar tool? That will give us more info on how far the printer is booting.
 
The below settings work for me, the top image shows a successful boot on my end.
![image](https://user-images.githubusercontent.com/8218499/146636655-a9002543-a9b8-4ea4-92e6-ec89eaa3de9e.png)

![image](https://user-images.githubusercontent.com/8218499/146636664-f4ff737e-128d-44b7-a4da-4ea3612c2161.png)

The blue circled checkboxes need to be checked for the lines to appear right.
[...]

PR #3412, opened 2022-02-16
Hardware TWI timeout

leptun:
Prevents deadlock in the hardware I2C code if one of the data lines is shorted to ground (infinite wait for busy bus)

Fixess #3321 (intentional typo)


wavexx (approved):

leptun:
@wavexx Not good. Switching to c++ for pat9125.c and twi.c added quite a bit of ram back? Instead of 3B extra of ram I have 11B extra. This is not what I was expecting from switching to c++ in this example

DRracer:
@leptun that static timer's initialization is the cause. You don't need it static anyway...

DRracer (approved):

leptun:
PR uses 60B extra of flash. No ram difference

PR #3414, opened 2022-02-17
xflash partitioning

leptun:
Documentation about how the xflash space will be used

todo: update the build scripts to add the bin files to xflash and check for overlapping sections there


wavexx (approved):
Nice to have this documented!

PR #3420, opened 2022-02-20
Fix FINDA timer

gudnimg:
We want to set the FINDA state IF the timer has NOT expired.
It is helpful to see why by looking at the commit which broke
this https://github.com/prusa3d/Prusa-Firmware/pull/3365/commits/802b8860c862f1e8cca82146ab70c0a0cadf2629

Without this fix my FINDA sensor just shows N/A. But after this commit is works perfectly.


leptun (approved):

leptun:
Tested on MK3S. It does indeed fix the N/A issue

3d-gussner (approved):

PR #3423, opened 2022-02-21
PAT9125 new init sequence

leptun:
Enable the "new" PAT9125 init sequence and some code refactoring. It should make the sensor track plastic better.

PFW-1311

### DIFF
--- a/Firmware/config.h
+++ b/Firmware/config.h
@@ -42,7 +42,12 @@
 //#define PAT9125_I2C_ADDR  0x73  //ID=NC
 #define PAT9125_XRES      0
 #define PAT9125_YRES      240                   // maximum resolution (5*X cpi)
-#define PAT9124_YRES_MM   (5*PAT9125_YRES/25.4) // counts per mm
+#define PAT9125_YRES_MM   (5*PAT9125_YRES/25.4) // counts per mm
+#define PAT9125_INVERT_X  0 //1 means flipped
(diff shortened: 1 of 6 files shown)


leptun:
Rebased to MK3_MMU2 branch

leptun:
Added more code. Preparing for the fsensor refactoring.
Now we clock 10 dummy cycles on the i2c bus on init. This is so that if the printer was reset during a register read, the sensor might be left in a state which makes the bus "busy". In order to be able to talk to the sensor at that point, the sensor must get a NACK from the mcu, signaling to it that it should release the bus and go idle. If these cycles are not done, the mcu will do one of the following:
- TWI: The bus will be busy, 100ms timeout. fail to init until power cycle.
- SWI2C: There were no checks for busy bus. The sensor could have been fried by the mcu since it held the data line OUTPUT_HIGH, while the sensor was pulling it low (open collector). Maybe this is why early sensors were getting fried. Besides getting the sensor to idle, the SWI2C init sequence was also patched to avoid these dangerous gpio conditions on the bus.

I also changed how the pat9125 sensor is "probed". Instead of reading a register, we now just try to talk to the sensor without sending any data at all. Much quicker and the most barebones approach that could be done for checking if the sensor is there or not.

leptun:
Warning: This breaks MK404 for now. It will be fixed at a later time. To use MK404, disable the new init sequence.

DRracer (approved):

PR #3425, opened 2022-02-22
PFW-1296: Add Norwegian

3d-gussner:


DRracer (approved):
looking good to me (can't check the Norwegian texts themself though :wink: )

PR #3429, opened 2022-02-24
Filament sensor refactoring

leptun:
Still work in progress. It will probably break the current MMU2 implementation. I will set the branch base to a dev branch as soon as we have one so that the MK3 branch still works in the meantime.

Targets:
- Unify the filament sensor code without an ifdef hell (hopefully).
- Make the filament sensor implementation generic, so that the rest of the code has to change very little (or not at all) between filament sensor versions.
- Make the PAT9125 work with the new MMU2 implementation. This is needed for the MMU2(not S) variant. This will be accomplished by being able to detect filament presence without requiring motion, just like the IR sensor.

So far only the 3 MK2.5S/MK3S builds work. The PAT9125 implementation will come soon. As a result the build will most likely fail, but that is expected at this point.

PFW-1311
Todo:
- [ ] Implement some kind of action of fsensor error.
- [x] Implement PAT9125 SB and motion code.
- [x] Implement wrong sensor check on init.
- [ ] Translations


leptun:
Rebased to MK3_MMU2 branch

DRracer (approved):
approving to proceed with the new MMU implementation

PR #3431, opened 2022-02-25
Fill the doc for the EEPROM value at 0x0DA1 - active_sheet

TojikCZ:
This is my first one :tada: And I already had to do two amends just to get rid of spaces and put the stupid tabs there
Hopefully editing a doc doesn't break core functionality :sweat_smile:


leptun:
@TojikCZ The "Valid values" column is wrong for that eeprom item, but the formatting is somewhat inconsistent anyway. I'll let @3d-gussner handle this documentation as he wrote it.

@3d-gussner btw, there are some spurrious tabs at the end of the "Debug code" for the sheets. Should probably remove those.

TojikCZ:
@leptun had 00-07h there before, but I didn't see anything like that in the sheet i
temps or whatnot, so I made it look the same as those

wavexx:
@3d-gussner ping!

3d-gussner (approved):
Thanks for the PR

- [ ] Cherry-pick to MK3_3.12

PR #3432, opened 2022-02-26
memory and flash optimizations for 3.12 PR3

leptun:
Cherrypicked changes from #3378
More commits to come.

Saved 1588B of flash, 130B of ram

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -11956,7 +11956,7 @@ void M600_wait_for_user(float HotendTempBckp) {
 				else {
 					counterBeep = 20; //beeper will be inactive during waiting for nozzle preheat
 					lcd_set_cursor(1, 4);
-					lcd_print(ftostr3(degHotend(active_extruder)));
+					lcd_printf_P(PSTR("%3d"), (int16_t)degHotend(active_extruder));
 				}
 				break;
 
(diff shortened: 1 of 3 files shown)


leptun:
@gudnimg I think we can easily switch to printf and not use `ftostr3` anymore.

leptun:
@gudnimg Unless you have some other RAM->PGM improvements, this PR is final. Please let me know if you have any other plans, becase if you don't I'll ask for a review from a second reviewer.
For the future, it would be better if you made the PRs because this way I can approve and merge them. Can't really approve my own PRs :)

gudnimg:
@leptun I understand, I got no more changes for the time being :D

Firmware/Marlin_main.cpp
```diff
@@ -7786,14 +7786,14 @@ SERIAL_PROTOCOLPGM("\n\n");
 
         updatePID();
         SERIAL_PROTOCOLRPGM(MSG_OK);
-        SERIAL_PROTOCOL(" p:");
+        SERIAL_PROTOCOLPGM(" p:");
```
leptun, line 7789: I think flash usage increases because strings in ram are optimized much more than strings in progmem. So there will be duplicates in progmem, while in ram these are deduplicated. Sadly, I couldn't find a fix for this besides using PSTR only once and using the pointer on multiple PROTOCOLRPGM functions

DRracer (approved):

PR #3435, opened 2022-03-02
Use correct filament type in variant files

leptun:
I think this was incorrect.
- The MK2.5 and MK3 variants used the "MK25" and "MK3" filament sizes
- The MK2.5S miniRambo 1.0a variant used the "MK25S" filament size
- The MK2.5S miniRambo 1.3a variant and the MK3S used the "MK25" and "MK3" filament sizes.

Either the second or the third bullet point is wrong. This PR assumes the third bullet point is wrong and changes the names accordingly. This should only change the message in the support menu. And I actually quite like this considering the printer variant is not present in the support menu, but we have this "1_75mm_MK3" line instead which could confuse people who thought that was also the fw variant that was flashed.


wavexx (approved):

3d-gussner:
Is this a useful information at all?
Only used in `PRUSA Rev` and Support output.

leptun:
On hold until we confirm that "PRUSA Rev" is not broken because of this. @DRracer Please check this is the case and merge if everything is fine

wavexx (approved):

PR #3436, opened 2022-03-02
Reuse Gcode M123 in Prusa "FAN" command

gudnimg:
Changes save 90 bytes of flash

The `FAN` prusa command now not only shows RPM for both fans but also the fan speed (same as Gcode M123). I'm not sure if the format of the message breaks anything internally at Prusa. Please let me know. I mainly want to reduce the occurrences of `fan_speed[]` in the code.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -4489,7 +4489,7 @@ void process_commands()
         } else if( code_seen_P(PSTR("FANPINTST"))){
             gcode_PRUSA_BadRAMBoFanTest();
         }else if (code_seen_P(PSTR("FAN"))) { // PRUSA FAN
-			printf_P(_N("E0:%d RPM\nPRN0:%d RPM\n"), 60*fan_speed[0], 60*fan_speed[1]);
+            gcode_M123();
 		}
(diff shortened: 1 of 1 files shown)


leptun:
@TojikCZ can answer your question about compatibility

leptun:
As for the PSTR change, I already fixed it in 3.10.2 or in 3.11 (don't remember). Since that change already is there, this will interfere when we merge those branches to MK3 (soon™)

gudnimg:
> As for the PSTR change, I already fixed it in 3.10.2 or in 3.11 (don't remember). Since that change already is there, this will interfere when we merge those branches to MK3 (soon™)

Nice, I've removed the PSTR commit.

gudnimg:
Added another commit where we can reuse M123.

image

wavexx:
Huh, just touched that part in #3461. We can definitely replace the code in selftest (I will probably do the same already, since it would conflict anyway).

No idea about the "FAN" command though.. something else might rely on this

gudnimg:
I've updated the PR by removing the 2nd commit for fan selftest. Changes now only affect Prusa "FAN" command.

PR #3439, opened 2022-03-03
🚧 Bootloader rpi port support

leptun:
Using information from the bootloader, decide which action to take when running the second bootloader.
- ~Allows skipping the second bootloader if the first one is not used.~
- If the first bootloader used the primart uart port, then the second bootloader should do the same thing
- If the first bootloader used the secondary uart port, then the second bootloader should do the same thing
- If the bootloader is an older version which doesn't report anything, default to the primary uart port.

Used in combination with https://github.com/prusa3d/stk500v2-prusa/pull/3

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -77,17 +77,17 @@ extern FILE _uartout;
 
 #define SERIAL_PROTOCOL(x) (MYSERIAL.print(x))
 #define SERIAL_PROTOCOL_F(x,y) (MYSERIAL.print(x,y))
-#define SERIAL_PROTOCOLPGM(x) (serialprintPGM(PSTR(x)))
-#define SERIAL_PROTOCOLRPGM(x) (serialprintPGM((x)))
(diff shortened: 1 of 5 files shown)


ggppjj:
Hello!

I was personally interested in this PR, and was wondering if you wouldn't mind sharing the rationale behind its closure. Is there another PR that superseded it?

leptun:
There is no PR at the moment so supersede this one. I dropped the idea for the time being since I came to the conclusion that I have absolutely no way of updating the bootloader of the MK3 in production without a programmer. This is because the lock bits for the bootloader section are set to prevent accidental erasure of the existing bootloader. Should there be incentive again internally for this feature, I will open the PR again.

ggppjj:
Understood. I was hoping to see the day where I didn't have to plug my Pi4a that I have (at my own risk) in the rpi port to do an update, but it's understandable that it either isn't possible or isn't easily done. Would be nice internally for Prusa Connect through rpi for the neat sleek AiO package that it seems like the port was made for, but again understandable that this wasn't entirely designed for this purpose. 

Thanks again, here's hoping it can one day be done!

PR #3442, opened 2022-03-13
Add documentation for M701, M203, M201 parameters

gudnimg:
* Add `E` parameter to M701 documentation
* Remove missing `U` parameter in M702
* Document parameters for M203 and M201.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -7279,8 +7279,17 @@ Sigma_Exit:
     #endif
 
     /*!
-	### M203 - Set Max Feedrate <a href="https://reprap.org/wiki/G-code#M203:_Set_maximum_feedrate">M203: Set maximum feedrate</a>
+    ### M203 - Set Max Feedrate <a href="https://reprap.org/wiki/G-code#M203:_Set_maximum_feedrate">M203: Set maximum feedrate</a>
     For each axis individually.
+    ##### Usage
+
+    M203 [ X | Y | Z | E ]
+
+    ##### Parameters
+    - `X` - Maximum feedrate for X axis
+    - `Y` - Maximum feedrate for Y axis
+    - `Z` - Maximum feedrate for Z axis
+    - `E` - Maximum feedrate for extruder drives
     */
     case 203: // M203 max feedrate mm/sec
 		for (uint8_t i = 0; i < NUM_AXIS; i++)
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -8834,10 +8834,15 @@ Sigma_Exit:
     }
     break;
 
-  /*!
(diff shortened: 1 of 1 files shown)


wavexx:
@3d-gussner can you review this?

leptun (approved):

Firmware/Marlin_main.cpp
```diff
@@ -8834,10 +8852,15 @@ SERIAL_PROTOCOLPGM("\n\n");
...
 
-  /*!
-  ### M701 - Load filament <a href="https://reprap.org/wiki/G-code#M701:_Load_filament">M701: Load filament</a>
-  
-  */
+    /*!
+    ### M701 - Load filament <a href="https://reprap.org/wiki/G-code#M701:_Load_filament">M701: Load filament</a>
+    #### Usage
+
+        M701 [ E ]
```
3d-gussner: In RepRap gcode wiki it is https://reprap.org/wiki/G-code#M701:_Load_filament

Do we need to update the wiki for PrusaFimware using `E` instead of `T`?

Firmware/Marlin_main.cpp
```diff
@@ -8850,7 +8873,7 @@ SERIAL_PROTOCOLPGM("\n\n");
     ### M702 - Unload filament <a href="https://reprap.org/wiki/G-code#M702:_Unload_filament">G32: Undock Z Probe sled</a>
     #### Usage
     
-        M702 [ U | C ]
+        M702 [ C ]
```
3d-gussner, line 8877: In RepRap gcode wiki it is https://reprap.org/wiki/G-code#M702:_Unload_filament

Do we need to update the wiki for PrusaFimware using `C` for just current?

Firmware/Marlin_main.cpp
leptun: Hmmm. Now that's a stupid inconsistency from marlin.... Makes me wonder if we should duplicate the code_seen() so that T is also supported

Firmware/Marlin_main.cpp
leptun: Yes, we should do that. It's a really useful argument for end of print gcodes since you only want to unload the current filament when you have a mmu

Firmware/Marlin_main.cpp
leptun: I propose we merge this PR (documentation only) and then I add the T parameter alias as well in another PR. It shouldn't break any compatibility

Firmware/Marlin_main.cpp
leptun: Updated the wiki

Firmware/Marlin_main.cpp
leptun: Change of plans. I just added the commit here since the change is really small. Tested the change on the old mmu implementation

3d-gussner (approved):

gudnimg:
Kudos to you guys. The alias for M701 was a great idea 💪

Issue #2717, opened 2020-06-05
  [ENHANCEMENT]<Turn OFF print fan while changing the filament during a print>


majedation:
**MK3S**
**Firmware 3.9.0**
**PrusaSlcier 2.2.0**

When the filament runs out during a print, and while changing filament during a print, the print fan stays ON at full speed (for PLA). That pushes the new extruded/purged filament towards the printed part.

**Suggestion**
My suggestion is to turn OFF the print fan during the filament change, then turn it ON again when print resumes.

flyingmarvin:
I would like this feature as well, sometimes I even had my print being removed from the build plate because a full blowing fan at the edge of the print for a long time caused too much warping.

gudnimg:
This was fixed recently here https://github.com/prusa3d/Prusa-Firmware/pull/3746 (will be included in 3.13)

Prusa-Support:
Thank you for your collaboration.

This issue will be closed in the same way as issue https://github.com/prusa3d/Prusa-Firmware/issues/3670 that led to such Pull Request.

Michele Moramarco
Prusa Research

Issue #2518, opened 2020-03-02
  [ENHANCEMENT] Fan is on while waiting for nozzle heating


espr14:
i3 MK3S FW 3.8.1

1. Printer is printing
2. Filament runs out
3. Printer moves to the corner and turns nozzle heating off
4. I press button to start heating
5. Front fan is still running which slows down the heating.

Solution: turn fan off during heating and then turn it back on.

Photogad:
+1

ManuGithubSteam:
+1 
This would alsp speed up normal heating of the alu block not just at a filament change...

gudnimg:
For filament change this issue was fixed recently here https://github.com/prusa3d/Prusa-Firmware/pull/3746 (will be included in 3.13)

For resuming a print from RAM, this will be this will be fixed here https://github.com/prusa3d/Prusa-Firmware/pull/3443

Prusa-Support:
Hello, the new FW 3.13.0 is out, merging the fixes for this issue.
Please give the new firmware a try and possibly share feedback.

Michele Moramarco
Prusa Research

espr14:
I sold my i3 so no way how to check that.

PR #3443, opened 2022-03-13
Turn off print fan while resuming print from RAM

gudnimg:
When a print is restored from RAM the nozzle is heating up. While heating is taking place the print fan will be turned **OFF** to speed up the process.

Testing:
1.  Start a print
2. After the purge line is finished, go into the Tune menu and raise the speed of the fan
3. Wait and confirm the print fan is spinning.
4. Pause the print
5. Wait for the temperature to drop a bit
6. Resume the print
7. **Test 1:** While heating the nozzle, the print fan is forced off
8. **Test 2:** When heating is finished, the print is spinning with same speed as in Step 3

Change in memory:
Flash: +8 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -11260,12 +11260,17 @@ void restore_print_from_ram_and_continue(float e_move)
 	if ((fan_check_error != EFCE_OK) && (fan_check_error != EFCE_FIXED)) return;
     if (fan_check_error == EFCE_FIXED) fan_check_error = EFCE_OK; //reenable serial stream processing if printing from usb
 #endif
-	
+
+    // Make sure fan is turned off
+    fanSpeed = 0;
+
     // restore bed temperature (bed can be disabled during a thermal warning)
     if (degBed() != saved_bed_temperature)
         setTargetBed(saved_bed_temperature);
-    fanSpeed = saved_fan_speed;
     restore_extruder_temperature_from_ram();
+
+    // Restore saved fan speed
+    fanSpeed = saved_fan_speed;
     axis_relative_modes ^= (-saved_extruder_relative_mode ^ axis_relative_modes) & E_AXIS_MASK;
     float e = saved_pos[E_AXIS] - e_move;
     plan_set_e_position(e);


Panayiotis-git:
Nice touch!
Do you think that we should left the fan set on if the target temperature is lower than the current?

gudnimg:
> Nice touch! Do you think that we should left the fan set on if the target temperature is lower than the current?

@Panayiotis-git Yes I agree. If `M109` is called with a lower temperature than current temperature, it would speed up the cooldown process to enable the print fan. 

But when I was testing that functionality... M109 did not wait for the cooldown to complete. I will test that more next weekend.

Edit: Maybe we should add to `HeatingStatus`: `HeatingStatus::EXTRUDER_COOLING` and `HeatingStatus::EXTRUDER_COOLING_COMPLETE`

gudnimg:
I want to also look into the situation where filament runs out, but maybe that is a separate PR

tickets: 
* https://github.com/prusa3d/Prusa-Firmware/issues/2518
* https://github.com/prusa3d/Prusa-Firmware/issues/2717

gudnimg:
`M109` should now turn on the print fan if `R` parameter is used for cooldown. When the cooldown is complete, the print fan speed is set to whatever it was before `M109` was called.

gudnimg:
Marking as ready for review. 

I think the issue where filament runs out should be tackled in a separate PR.

wavexx:
I'm thorn by this. @mkbel might like this discussion, since he was able to convince me that the printer should be as dumb as reasonably possible.

The print fan status during print shouldn't be touched IMHO, since this is state the the user has directly control over.

However, I would have no objections to fix/change the behavior in restore_print_from_ram_and_continue(), since this is totally outside user's control and clearly has no influence on the print. We already do something similar in the selftest to speedup cooldown.

Thoughts?

Firmware/Marlin_main.cpp
```diff
@@ -11705,8 +11721,12 @@ void restore_print_from_ram_and_continue(float e_move)
 	{
 		setTargetHotendSafe(saved_extruder_temperature, saved_active_extruder);
 		heating_status = HeatingStatus::EXTRUDER_HEATING;
+		// The global fanSpeedBckp is intentionally not used here.
+		uint8_t fanSpeedBackup = fanSpeed;
+		fanSpeed = 0; // Turn off print fan while extruder is heating
 		wait_for_heater(_millis(), saved_active_extruder);
 		heating_status = HeatingStatus::EXTRUDER_HEATING_COMPLETE;
+		fanSpeed = fanSpeedBackup; // Restore previous speed setting for print fan
```
mkbel: I don't see any principal problem here.
It just needs to be tested if it doesn't induce big temperature dip in the moment the print fan is turned on.
If it would be found this is the problem, more complicated alternative approach would be to resume fan e.g. in the moment extruder heater temperature is closer than 10°C to target temperature.
In such case there would be enough time for integral part of regulation term to adapt for running fan.

wavexx: This part I'm not worried about, this is perfectly safe.

Just that M106/M109 S/R could set the fan speed as a result of this PR.
Technically, the same effect can be achieved by reordering the instructions already at the g-code level.

wavexx:
@gudnimg If we only fix up restore_print_from_ram_and_continue() I would be fine with this.

gudnimg:
> @gudnimg If we only fix up restore_print_from_ram_and_continue() I would be fine with this.

I'll try to cleanup this PR this weekend so we can close this PR :)

gudnimg:
@wavexx I've fixed the PR and updated the PR description with steps I used to test the change. Works well on my side.

wavexx (approved):

3d-gussner (approved):
Tested on MK404:
- `M155 S1 C3` to report temp and fan
- described steps to verify the PR. Fan set to `255`
- after pausing waited to cool down to 170°C fan at `0`
- resume print. Fan still at `0` 
- during heating nozzle the fan stayed off :heavy_check_mark: 
- after heating finished and printer continue printing fan again at set speed :heavy_check_mark:

Issue #3119, opened 2021-04-21
[ENHANCEMENT] Can you make the default/idle z-height just a little higher? 10mm or so?


rw3iss:
Prusa 2.3.0-64 linux
MK3S

This is a request to make the default/idle Z-height about 10mm higher, for multiple reasons:

If the nozzle is oozing, we generally want to clear it right before it starts printing... I do this with a quick wipe of the wire brush (see attached image). I can't fit the standard smallest width wire brushes under it... well just barely, but it's quite tight. Another 10-15mm would be perfect.
![1618503027468](https://user-images.githubusercontent.com/372120/115570014-c4275d00-a28b-11eb-8255-f4e91609211a.jpg)

The second argument is when inevitable blobs of failed prints get caught under the print head... we typically want to stop print and pull them off, but there's barely any more room to do that at the idle height, and going into settings to move the Z axis is tedious all the time.

Please! Or maybe someone knows a way to edit the Gcode to do this?

wavexx:
I feel this. I also modified the default "stop/pause height" to be minimum 50mm higher than the print, simply because I need more space to remove the steel sheet, and just like you say, I had to raise Z almost every time.

Anybody else having the same need?

rw3iss:
> I feel this. I also modified the default "stop/pause height" to be minimum 50mm higher than the print, simply because I need more space to remove the steel sheet, and just like you say, I had to raise Z almost every time.
> 
> Anybody else having the same need?

Nice @wavexx . Did you modify the default G-Code and have to re-flash the firmware? Or can you just modify it with PrusaSlicer or elsewhere?

Panayiotis-git:
This is one of the reasons I like the skelestruder: it provides plenty of room to access the nozzle.

wavexx:
> Nice @wavexx . Did you modify the default G-Code and have to re-flash the firmware? Or can you just modify it with PrusaSlicer or elsewhere?

It's done in multiple steps. I have this in the end-gcode:

```
{if layer_z < max_print_height}G1 Z{z_offset+min(layer_z+50, max_print_height)}{endif} ; Move print head up
```

For pause/stop, it's currently defined in the FW, so it needs a couple of simple changes (https://github.com/wavexx/Prusa-Firmware/tree/stop_raise_higher is one)

DaReaLDeviL:
I also whould prefer a higher standard value. Great suggestions.

rw3iss:
[...]

Issue #2830, opened 2020-09-11
  [ENHANCEMENT] Automatically raise extruder after the print is completed


Bill2462:
Printer : MK3S, 

Problem:
When the print is cancelled/finished, extruder parks very close to the bed surface. This can make removing the print surface a bit tricky and can lead to damage of the hotend if user is to lazy to raise it using axis movement option.

Solution:
Usually I go to settings->move axis and raise the extruder to around 80mm above print surface before I remove it.
I think this can happen automatically when print finishes or is cancelled.

Bill2462:
Edit:
I just noticed that that happens only when print is aborted. When print finishes normally, extruder moves all the way to the left (y=0) and that is no longer a problem. Not sure whether this is a bug or not.

leptun:
Doesn’t the nozzle raise 10mm before parking? Or is that not enough?

Bill2462:
It raises by 10mm but this is not enough to remove print surface freely if print is not very tall or if it fails on one of the first few layers.

It seems that when the print finishes normally, extruder moves to all the way to the left and I can remove the surface just fine.
But when the print is aborted, extruder does not move all the way to the left and it stays above the build platform.
Yes it raises by 10mm but this is often not enough.

I will test this tomorrow and provide some pictures but it also seems that there is a difference in y axis movement.

leptun:
The parking position defined in the gcode file for when the print finishes is not part of the firmware, but rather just 2 move instructions.
In firmware it is set up differently than it is in gcode.

Btw, there is a quick shortcut fo raising the nozzle more. Just press and hold the knob for 1s and the move Z menu will open

wavexx:
I agree with @Bill2462 here. I modified the end_gcode in the slicer and the stop code in the FW to raise at least 5cm above the current print position so that the steel sheet can be removed easily.

Changing the end_gcode is easy. On the FW I'm using this:

https://github.com/prusa3d/Prusa-Firmware/commit/3a3e65e057e0284ebd30072f180ba3bae67ca979

but for a PR maybe we should define the LIFT amount in the variant file instead of adding yet another magic value..

PR #3455, opened 2022-03-30
Stop raise higher

wavexx:
Check if Z axis has been homed and, if so, raise Z at least 5cm after
stopping so that the steel sheet can be comfortably removed. This also makes the distance configurable in the variant file (similar to the pause park feature).

I'm the camp that believes pause/stop, as well as parking after a complete print should lift higher to allow easier removal of the steel sheet. See #2830 #3119

For the final raising distance, the gcode is responsible so there's nothing we can do in the FW (we should move it to the prusaslicer profiles).

This only affects "stop print"

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3688,6 +3688,7 @@ static void gcode_M600(bool automatic, float x_position, float y_position, float
 
     //Lift Z
     current_position[Z_AXIS] += z_shift;
+    clamp_to_software_endstops(current_position);
     plan_buffer_line_curposXYZE(FILAMENTCHANGE_ZFEED);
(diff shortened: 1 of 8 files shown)


wavexx:
Related: https://github.com/prusa3d/PrusaSlicer-settings/issues/155

leptun (changes requested):
The stop move will crash the Z ends if the Z at the stop position is high. The target Z position is not clamped to the machine limits.

The Pause move is still only 20mm. Maybe that one should also be updated?

leptun (approved):
Fixed the issues in my last review. Also rebased on the latest MK3 branch.
The Pause raise amount is intentionally left as it is now

PR #3457, opened 2022-03-30
Cleanup: remove unused FILAMENT_RUNOUT_SUPPORT define

wavexx:
FILAMENT_RUNOUT_SUPPORT has never been supported in the MK3 branch, and
contains a ton of horrible and unmaintained code that doesn't even
compile anymore.

Since this code is currently polluting a sizable block inside the G0/G1
code block, simply get rid of it.


leptun:
In the MMU2 branch this code is already removed

wavexx:
Since this is self contained, maybe we can already merge this or cherrypick the commit from the MMU2 branch? This would reduce the diff size, which IMHO is a good thing during review

wavexx:
Looks like this is commit b0843aee56b4385e9427edbd3a05e1ed1a815ccb

leptun:
I can't find that commit. Can you please send a link to it?

wavexx:
The force-push changed the hash. Found it:

https://github.com/leptun/Prusa-Firmware/commit/1c3910efc053929e1d641e857e70e27056e9b1f7

gudnimg (approved):

leptun (approved):

PR #3459, opened 2022-03-30
Remove unused setting `PID_ADD_EXTRUSION_RATE`

### LINKED ISSUE #3458: [QUESTION] Global Kc is unused (see C parameter in M301) Should we remove it?
**What is your question?**

I noticed that the C parameter in M301 we assign the parameter value to a global variable Kc. But Kc is not used anywhere, it is only assigned. I am wondering if I should create a PR to remove it or if there are any current plans to use this variable?

After looking further into this. It seems we can sum this question into if we want to remove the `#ifdef PID_ADD_EXTRUSION_RATE`

gudnimg:
Even if the setting was turned on, it doesn't do anything useful. `Kc` variable is set but is never actually used anywhere. Closes #3458

### DIFF
--- a/Firmware/ConfigurationStore.cpp
+++ b/Firmware/ConfigurationStore.cpp
@@ -336,9 +336,6 @@ void Config_ResetDefault()
     
 #ifdef PIDTEMP
     updatePID();
-#ifdef PID_ADD_EXTRUSION_RATE
-    Kc = DEFAULT_Kc; //this is not stored by Config_StoreSettings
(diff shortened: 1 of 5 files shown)


wavexx:
PID_ADD_EXTRUSION_RATE was the precursor to PID_EXTRUSION_SCALING.

Seems like it's still disabled by default even in Marlin2, I guess because AUTOTEMP (which we also have) that replaces it with something that doesn't mess up normal PID regulation.

I'd be in favor of removing it.

wavexx (approved):

Issue #1520, opened 2019-02-10
3.5.2-RC1-3.7.1-RC1 - Print fan not detected during SelfTest


F0x06:
Hi ! 

On the latest 3.5.2-RC1 if i perform a Self-Test, the Print fan is not detected automatically and printer is asking if fans are rotating for a manual answer. Tested on previous version and auto detection worked before, probably related to recent PWM modifications ?

Dolnor:
I've also noticed the Support page on Extruder info is showing print fan RPM way off. At just 100/255, reported revolutions are over 4500, while the full fan speed per specification caps at 4200?

PavelSindler:
Hi and thank you for reporting this. We will re-add reliable automatic test for swapped fans detection soon.

F0x06:
Same issue on `3.7.1-RC1`

PR #3461, opened 2022-04-04
Fix/optimize fan selftest

wavexx:
The fan selftest code isn't properly measuring RPMs (by waiting until the measuring is terminated) and used wrong thresholds to detect for a swapped print/hotend fan.

I always assumed the question "Is front print fan spinning?" shown during selftest to be intentional, but it turns out they are probably the result of a faulty check. I've found issue #1520 mentioning this as a regression.

The delay waiting for the print fan was also nonsensical (10 seconds of spin-up followed by 5 of slow down?). The complete fan test now takes 7 seconds, and should no longer prompt on stock parts.

It might still occasionally trigger a "swapped fans" check if the hotend fan has been modified with a stronger fan, and/or if the print fan is starting to degrade, since in these cases the RPM thresholds can easily overlap. In such cases the "swapped fans" test still makes sense to be performed.

Due to the rewrite, the new code actually saves 134 bytes.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -7611,10 +7611,10 @@ static void lcd_selftest_measure_fans(uint8_t delay, uint8_t col, uint8_t row) {
 static FanCheck lcd_selftest_fan_auto(uint8_t _fan)
 {
     // speed threshold to differentiate between extruder and print fan
-    static const int printFanThr = 70; // >=4200 RPM
+    static const int printFanThr = FANCHECK_AUTO_PRINT_FAN_THRS; // >= FANCHECK_AUTO_PRINT_FAN_THRS RPS
 
     // speed threshold to mark a fan as failed
-    static const int failThr = 20; // <1200 RPM would mean either a faulty Noctua, Altfan or print fan
+    static const int failThr = FANCHECK_AUTO_FAIL_THRS; // < FANCHECK_AUTO_FAIL_THRS RPM would mean either a faulty Noctua, Altfan or print fan
 
 	switch (_fan) {
 	case 0:
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -7586,7 +7586,7 @@ static void lcd_selftest_setfan(uint8_t speed) {
 
 // Wait for the specified number of seconds while displaying some single-character indicator on the
 // screen coordinate col/row, then perform fan measurement
-static void lcd_selftest_measure_fans(uint8_t delay, uint8_t col, uint8_t row, int n) {
+static void lcd_selftest_measure_fans(uint8_t delay, uint8_t col, uint8_t row) {
     // spin-up delay
     static char symbols[] = {'-', '|'};
     static_assert(1000 / sizeof(symbols) * sizeof(symbols) == 1000);
@@ -7605,9 +7605,7 @@ static void lcd_selftest_measure_fans(uint8_t delay, uint8_t col, uint8_t row, i
         delay_keep_alive(100);
     }
 
-    printf_P(PSTR("Test %d:\n"), n);
-    printf_P(PSTR("Print fan speed: %d\n"), fan_speed[1]);
-    printf_P(PSTR("Extr fan speed: %d\n"), fan_speed[0]);
+    gcode_M123();
 }
 
 static FanCheck lcd_selftest_fan_auto(int _fan)
@@ -7622,7 +7620,7 @@ static FanCheck lcd_selftest_fan_auto(int _fan)
 	case 0:
         setExtruderAutoFanState(3); // extruder fan
         lcd_selftest_setfan(0); // print fan off
-        lcd_selftest_measure_fans(2, 18, 2, 1);
+        lcd_selftest_measure_fans(2, 18, 2);
         setExtruderAutoFanState(0); // extruder fan off
 		if (fan_speed[0] < failThr) {
 			return FanCheck::ExtruderFan;
@@ -7634,7 +7632,7 @@ static FanCheck lcd_selftest_fan_auto(int _fan)
 
 	case 1:
         lcd_selftest_setfan(255);
-        lcd_selftest_measure_fans(5, 18, 3, 2);
+        lcd_selftest_measure_fans(5, 18, 3);
         lcd_selftest_setfan(0);
         if (fan_speed[1] < failThr) {
             return FanCheck::PrintFan;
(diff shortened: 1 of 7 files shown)


wavexx:
After the suggestion from @gudnimg in #3436, this saves 148 bytes more!

Firmware/ultralcd.cpp
```diff
@@ -7574,90 +7574,72 @@ static bool lcd_selftest_manual_fan_check(int _fan, bool check_opposite,
...
-		setExtruderAutoFanState(3); //extruder fan
-#ifdef FAN_SOFT_PWM
-		extruder_autofan_last_check = _millis();
-		fan_measuring = true;
-#endif //FAN_SOFT_PWM
-		_delay(2000);
-		setExtruderAutoFanState(0); //extruder fan
-		manage_heater();			//count average fan speed from 2s delay and turn off fans
+    // speed threshold to differentiate between extruder and print fan
+    static const int printFanThr = 70; // >=4200 RPM
```
3d-gussner: Can we use variables/defines in variant files instead hard values?
There are some users modding their fans or for some reason the stock fan changes.

Firmware/ultralcd.cpp
```diff
@@ -7574,90 +7574,72 @@ static bool lcd_selftest_manual_fan_check(int _fan, bool check_opposite,
...
 
-		puts_P(PSTR("Test 1:"));
-		printf_P(PSTR("Print fan speed: %d\n"), fan_speed[1]);
-		printf_P(PSTR("Extr fan speed: %d\n"), fan_speed[0]);
+    // speed threshold to mark a fan as failed
+    static const int failThr = 20; // <1200 RPM would mean either a faulty Noctua, Altfan or print fan
```
3d-gussner: Can we use variables/defines in variant files instead hard values?
There are some users modding their fans or for some reason the stock fan changes.

Firmware/ultralcd.cpp
```diff
@@ -7574,90 +7574,72 @@ static bool lcd_selftest_manual_fan_check(int _fan, bool check_opposite,
...
 
-		if (fan_speed[0] < 20) { // < 1200 RPM would mean either a faulty Noctua or Altfan
+	switch (_fan) {
+	case 0:
+        setExtruderAutoFanState(3); // extruder fan
+        lcd_selftest_setfan(0); // print fan off
+        lcd_selftest_measure_fans(2, 18, 2);
```
3d-gussner, line 7623: Are these also hard coded values? If so I would prefer again variants variables.

Firmware/ultralcd.cpp
```diff
@@ -7574,90 +7574,72 @@ static bool lcd_selftest_manual_fan_check(int _fan, bool check_opposite,
...
-		// noctua speed is between 17 and 24, turbine more then 30
-		if (fan_speed[1] < 30) {
-			return FanCheck::SwappedFan;
-		}
-#else
-		// fan is spinning, but measured RPM are too low for print fan, it must
-		// be left extruder fan
-		else if (fan_speed[1] < 34) {
+        lcd_selftest_setfan(255);
+        lcd_selftest_measure_fans(5, 18, 3);
```
3d-gussner, line 7635: Are these also hard coded values? If so I would prefer again variants variables.

3d-gussner (commented):
As I don't have a MK25/S here I will have to somehow add some code to only trigger the fan selftest instead of a full run.
On the MK3S I can test it with
- Noctua coldend fan
- Prusa part fan
- Sunon coldend fan
- LDO part fan

Firmware/ultralcd.cpp
wavexx: These are spin up delay and screen coordinates, shouldn't require customization

Firmware/ultralcd.cpp
wavexx: I feel like the entire selftest is a giant hardcoded value.

We can do this, but I'd do it after we verify it's a good value since once it's in the variant file it's going to be 6 copies to maintain.

Firmware/ultralcd.cpp
wavexx: As above

Firmware/ultralcd.cpp
wavexx: Also doable, though this one doesn't likely require customization. Anything below this value should already be a broken/degraded fan.

Firmware/ultralcd.cpp
```diff
@@ -7574,90 +7574,72 @@ static bool lcd_selftest_manual_fan_check(int _fan, bool check_opposite,
...
+#endif
+    fan_measuring = true;
+    while(fan_measuring) {
+        delay_keep_alive(100);
+    }
+
+    gcode_M123();
+}
+
 static FanCheck lcd_selftest_fan_auto(int _fan)
```
gudnimg: Minor nitpick, but can the function parameter be `uint8_t`? The input is only `0` or `1`.

wavexx: Done

leptun:
I still get the swapped fans check (and I also thought you always get it). Here are my RPM readings during selftest:
```
E0:3840 RPM PRN1:0 RPM E0@:255 PRN1@:0
E0:0 RPM PRN1:3720 RPM E0@:0 PRN1@:255
```

wavexx:
@gudnimg could you also maybe report your values as well during selftest?
I wonder if I was just lucky

gudnimg:
Here are my results:
```
E0:4020 RPM PRN1:0 RPM E0@:255 PRN1@:0
E0:0 RPM PRN1:4320 RPM E0@:0 PRN1@:255
```
I get no errors. I also added `_swapped_fan` to the serial stream and it is set to `false`.

leptun:
Hmmm, so it appears I have underperforming fans. Disappointing

wavexx:
@3d-gussner reported also lower values. Can you post here your values as well?

3d-gussner:
Noctua + Prusa Part fan
```
E0:3780 RPM PRN1:0 RPM E0@:255 PRN1@:0
E0:0 RPM PRN1:3960 RPM E0@:0 PRN1@:255
```
Sunon MF40100V1-100C-G99 + LDO 5V DF5015H05S
The Sunon is detected as ALTFAN and spins at max 128 pwm = 50%
```
E0:2400 RPM PRN1:0 RPM E0@:128 PRN1@:0
E0:0 RPM PRN1:5760 RPM E0@:0 PRN1@:255
```

Firmware/ultralcd.cpp
3d-gussner: You are right, but moving it to variants would make custom mods easier (LDO part fan).

leptun (approved):

Firmware/ultralcd.cpp
leptun: Done

leptun:
Merging now since the code is ok. The threshold can be further adjusted as needed after we get more data.

PR #3466, opened 2022-04-07
 PFW-1303: Reset pullups on ADC

3d-gussner:
| Variant | before flash/mem | after flash/mem | diff flash/mem |
|---------|------------------|-----------------|----------------|
|         | FW3.11.0         | FW3.11.1-RC1    |                |
| MK3S    | 250550 / 6105    | 250566 / 6101   | +16 / -4       |
| MK3     | 251244 / 6133    | 251252 / 6129   | +8 / -4        |
| MK25S   | 222108 / 5968    | 222124 / 5964   | +16 / -4       |


3d-gussner:
@leptun Shouldn't we add `//#define DEBUG_PULLUP_CRASH //Test Pullup crash` also to the miniRAMBo variants?

Can this https://github.com/prusa3d/Prusa-Firmware/pull/3466/files#diff-16f11efcf1c662221c31bd948d195a160c12ee0b67f464d84308e9654ef71341R1692 cause issues on miniRAMbo, as the PK isn't used for thermistors?

leptun:
I mean, we could add it also to the miniRambo builds, but I don't know how the crash event will look like. Never tested it there.

PORTK shouldn't cause any issues on the miniRambo since I'm using the ADC mask directly. It would basically and the register value with 0xFF, which is more or less a nop.

leptun (approved):

PR #3467, opened 2022-04-09
Optimise `autostart_stilltocheck` variable

gudnimg:
Make the variable `static` within the `checkautostart()` function.
When the function is called for the first time `autostart_stilltocheck`
is set to `true` and will continue to live after the function is exited.
After it is set to `false` within `checkautostart()` it will continue to be
`false` forever.

Using `static` this way is more efficient than using a global variable.

Changes save 6 bytes of flash and 1 byte of SRAM

### DIFF
--- a/Firmware/cardreader.cpp
+++ b/Firmware/cardreader.cpp
@@ -30,7 +30,6 @@ CardReader::CardReader()
    memset(workDirParents, 0, sizeof(workDirParents));
    presort_flag = false;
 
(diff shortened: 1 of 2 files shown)


wavexx (approved):
Removes a global to booth!

gudnimg:
@wavexx we can also probably move `lastnr` in the same way and change it to one byte since it is only expected to range form 0 to 9.  I think we can in general optimise `checkautostart()` much more.

leptun (approved):
Tested the changes with an auto0.g. The file gets executed correctly on startup

PR #3471, opened 2022-04-11
New PO-based language translation support

wavexx:
The current internationalization support in the MK3 FW works, but leaves a lot to be desired  for the translators themselves.

In this PR we completely remove the old "txt-based" language files, and introduce a more traditional workflow that uses just the PO files and the gettext tooling.

See the README I wrote for an overview of how it works:

https://github.com/wavexx/Prusa-Firmware/blob/lang_po_workflow/lang/

This PR doesn't change the firmware itself aside from fixes in the message annotations themselves. It does however require a recheck of the translations, since I had to fix several long-standing inconsistencies between the txt and old po files.

This completely rewrites the tools that generate the multi-language image.

I have briefly updated ``build.sh`` and ``PF-build.sh``, but this needs to be tested more broadly. The main README also needs to be updated. "gawk" is no longer used, but python3 and several modules now take it's place.


DRracer (approved):
:pray: :+1: well done

esauvisky:
This PR introduced a regression when running `./build.sh` as it expects the python packages `pyelftools` and `polib` to be available in the current building environment:
```
$ ./build.sh 
Picked up JAVA_TOOL_OPTIONS: 
Loading configuration...
Initializing packages...
Preparing boards...
Verifying...
Sketch uses 250864 bytes (98%) of program storage space. Maximum is 253952 bytes.
Global variables use 6233 bytes of dynamic memory.
config.sh started
 Arduino main folder: OK
 AVR gcc tools:
   objcopy OK
 Output folder: OK
 Output elf file: OK
 Output hex file: OK
 Intl hex file prefix: OK
 Languages: cs de es fr it pl
 Community languages: nl ro hu hr sk sv no 
config.sh finished with success
fw-build.sh started
fw-build languages: cs de es fr it pl nl ro hu hr sk sv no 
generating firmware symbol map
Traceback (most recent call last):
  File "/home/emi/Coding/Prusa-Firmware/lang/./lang-map.py", line 4, in <module>
    import elftools.elf.elffile
ModuleNotFoundError: No module named 'elftools'
compiling language "cs" from po/Firmware_cs.po
Traceback (most recent call last):
  File "/home/emi/Coding/Prusa-Firmware/lang/./lang-check.py", line 36, in <module>
    import polib
ModuleNotFoundError: No module named 'polib'
po/Firmware_cs.po: NG! - translation contains warnings or errors
Traceback (most recent call last):
  File "/home/emi/Coding/Prusa-Firmware/lang/./lang-build.py", line 6, in <module>
    import polib
ModuleNotFoundError: No module named 'polib'
stat: cannot statx './tmp/lang_cs.bin': No such file or directory
compiling language "de" from po/Firmware_de.po
Traceback (most recent call last):
  File "/home/emi/Coding/Prusa-Firmware/lang/./lang-check.py", line 36, in <module>
    import polib
ModuleNotFoundError: No module named 'polib'
[...]
```

I can fix the documentation and submit a PR if desired.

wavexx:
These modules are mentioned inside lang/, but the main README is now completely outdated (gawk is still mentioned, but it's no longer used).

And looking at this, maybe we should do better error checking:  @3d-gussner a failure due to a language check is ok to ignore, but any error inside lang-build and lang-map should be fatal if a multi-lang FW is being built.

It would be nice to rewrite the main README completely.

@esauvisky where would you write the info? I think we could use some feedback on the documentation.

DJManas:
I am having issues with this.
```
Traceback (most recent call last):
  File "/home/djmanas/git/Prusa-Firmware/lang/lang-map.py", line 4, in <module>
    import elftools.elf.elffile
ModuleNotFoundError: No module named 'elftools'
ninja: build stopped: subcommand failed.
```

But when I launch python write import elftools.elf.elffile everything is working fine. Even tried to create venv and compile it under this environment. Just wondering is there some local venv for this firmware, where I am missing this or I do not know. I have python 3.9, 3.10 and 3.11 and in every one of them pyelftools polib, etc. are installed. Should I  raise ticket?

wavexx:
It's likely calling a different python executable/environment. Did you try to run `./lang/lang-map.py`?

DJManas:
Thanks. When I had prepared result for this thread I had noticed, that it launches python 3:11 and I did not had it in variables to modules were not compiled by system. I deleted 3.9 and left only 10/11, rebuild packages and its working now.

Thanks for the advice.

I was confused, because when I had launch the lang-map in terminal it only wanted elf file, but no error. I even tried to create prusa venv, which was python 3.10, polib/pyelftools installed by pip. Correct python interpreter was launched from cli, but when I launched ninja from same console it seems to miraculously used version 3.11 which did not had the package installed. Now its working like a charm.
Thanks!

PR #3472, opened 2022-04-13
PFW-1306 help.prusa3d.com

3d-gussner:
Use/show `help.prusa3d.com` instead of `howto.prusa3d.com`

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -1863,11 +1863,11 @@ void TestPullupCrash() {
 //! |  3.7.2.-2363       |	c=16
 //! | prusa3d.com        |	MSG_PRUSA3D
 //! | forum.prusa3d.com  |	MSG_PRUSA3D_FORUM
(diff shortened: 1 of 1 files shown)


Firmware/ultralcd.cpp
```diff
@@ -1971,7 +1971,7 @@ static void lcd_support_menu()
       
   MENU_ITEM_BACK_P(_n("prusa3d.com"));////MSG_PRUSA3D c=18
   MENU_ITEM_BACK_P(_n("forum.prusa3d.com"));////MSG_PRUSA3D_FORUM c=18
-  MENU_ITEM_BACK_P(_n("howto.prusa3d.com"));////MSG_PRUSA3D_HOWTO c=18
+  MENU_ITEM_BACK_P(_n("help.prusa3d.com"));////MSG_PRUSA3D_HELP c=18
```
leptun, line 1974: Just wondering, why do noloc messages get a MSG_ comment? It's not like any scripts touch these

3d-gussner: Added these for documentation.

PR #3473, opened 2022-04-14
Consolidate `menu_lcd_lcdupdate_func()` to create "Button Echo" and "Encoder Move" sounds.

gudnimg:
Marked as draft: Checkpoint, Need to test more, possibly more ways to optimize.

`menu_lcd_lcdupdate_func()` captures the event when a user:
1. Clicked the knob
2. Rotated the knob 

In this PR I propose we control the "Button Echo" and "Encoder Move" sounds from here. I believe this will be less bug prone and we always ensure that any button click or any knob rotation has sound feedback if that option is turned on.

Current changes save 140 bytes of flash memory.

Before: (01/10/2022)
```
Sketch uses 251302 bytes (98%) of program storage space. Maximum is 253952 bytes.
Global variables use 5684 bytes of dynamic memory.

```

After: (01/10/2022)
```
Sketch uses 251162 bytes (98%) of program storage space. Maximum is 253952 bytes.
Global variables use 5684 bytes of dynamic memory.
```

----

TODO:

- [x] Look into `lcd_quick_feedback()`
- [x] Remove `menu_back_if_clicked_fb()` it is now redundant because any click is guaranteed to have a sound feedback. `menu_back_if_clicked()` should be used instead.
- [x] Bug: (**Fixed by commit** e5ca3d0) When entering the Main Menu (click the knob on the Status Menu screen) then the menu will not render and the Button Echo feedback sound is endlessly beeping.
- [x] Bug: (**Fixed by commit** [4e33864](https://github.com/prusa3d/Prusa-Firmware/pull/3473/commits/4e338640e8effce634769875ec85530a89c8b023)) When folder is selected from "Print from SD" there is a double feedback and the user is returned to the main menu. I suspect we are not properly consuming the knob click here and the firmware thinks we want to back out of the menu until it the knob click is consumed somewhere else in the code. Function to look at: `lcd_sdcard_menu()`
- [x] Bug: (**Fixed by commit** [951dcb4](https://github.com/prusa3d/Prusa-Firmware/pull/3473/commits/951dcb419d7b04c8c9068669a9209e07f7e6c211)) When Toggling Brightness Mode, the LCD display does not update but I can clearly see visually that the firmware is applying the changes.
- [ ] Bug: Sometimes when clicking on folders, there are two beeps. This is due to how `lcd_sdcard_menu()` is written and `menu_lcd_lcdupdate_func()` is called twice until the click event is actually processed. I don't think this is a new issue. Potentially this can be fixed with `lcd_update_enable` flag?
- [x] Bug: (**Fixed by commit** [fd5dab1](https://github.com/prusa3d/Prusa-Firmware/pull/3473/commits/fd5dab17e819202dfb146bd49b366b40ab715e75)) Long-press menus are automatically exited as soon as user enters them.
- [ ] Bug: MMU2S - When I Load to Nozzle, the prompt asking if the extruded color is correct has no sound feedback when scrolling between Yes and No. Also clicking has no feedback.
- [x] Bug: (**Fixed by commit** [f822f91](https://github.com/prusa3d/Prusa-Firmware/pull/3473/commits/f822f9199f9c0f0feb937411fa2903087a37f01e)) Clicking `First layer cal.` -> `Continue` seems to skip a menu
- [ ] Clean up commits and add more description in commit messages.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -9578,12 +9578,12 @@ void manage_inactivity(bool ignore_stepper_queue/*=false*/) //default argument s
   // handle longpress
   if(lcd_longpress_trigger)
   {
-      // Consume the long-press click and give feedback
-      lcd_consume_click();
-      Sound_MakeSound(e_SOUND_TYPE_ButtonEcho);
       // long press is not possible in modal mode, wait until ready
       if (lcd_longpress_func && lcd_update_enabled)
       {
+      // Consume the long-press click and give feedback
+      lcd_consume_click();
+      Sound_MakeSound(e_SOUND_TYPE_ButtonEcho);
           lcd_longpress_func();
           lcd_longpress_trigger = 0;
       }
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -9578,6 +9578,9 @@ void manage_inactivity(bool ignore_stepper_queue/*=false*/) //default argument s
   // handle longpress
   if(lcd_longpress_trigger)
   {
+      // Consume the long-press click and give feedback
+      lcd_consume_click();
+      Sound_MakeSound(e_SOUND_TYPE_ButtonEcho);
       // long press is not possible in modal mode, wait until ready
       if (lcd_longpress_func && lcd_update_enabled)
       {
--- a/Firmware/menu.cpp
+++ b/Firmware/menu.cpp
@@ -117,15 +117,6 @@ void menu_back_if_clicked(void)
 		menu_back();
 }
 
-void menu_back_if_clicked_fb(void)
-{
-	if (lcd_clicked())
-	{
-        lcd_quick_feedback();
-		menu_back();
-	}
-}
-
 void menu_submenu(menu_func_t submenu)
 {
 	if (menu_depth < MENU_DEPTH_MAX)
--- a/Firmware/menu.cpp
+++ b/Firmware/menu.cpp
@@ -45,7 +45,7 @@ void menu_data_reset(void)
 	memset(&menu_data, 0, sizeof(menu_data));
 }
 
-void menu_goto(menu_func_t menu, const uint32_t encoder, const bool feedback, bool reset_menu_state)
+void menu_goto(menu_func_t menu, const uint32_t encoder, bool reset_menu_state)
 {
 	CRITICAL_SECTION_START;
 	if (menu_menu != menu)
@@ -56,8 +56,6 @@ void menu_goto(menu_func_t menu, const uint32_t encoder, const bool feedback, bo
 		CRITICAL_SECTION_END;
 		if (reset_menu_state)
 			menu_data_reset();
-
-		if (feedback) lcd_quick_feedback();
 	}
 	else
 		CRITICAL_SECTION_END;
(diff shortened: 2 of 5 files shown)


gudnimg:
Small update for future review. I am stumbling into the issue where the LCD knob click events are not consumed consistently in the firmware. I have seen this issue before but I did not intend to fix it in this PR but I might as well just do it since it is now affecting my changes negatively. I can use the sound "feedback" to debug it -- such that when there is a button echo feedback, then the knob click must be consumed to ensure correct behaviour.

leptun:
Just a heads up, there is a bug in gcc-avr 5.4.0 which causes random linker segmentation faults when changing these parts of the menu code. You might have encountered this already since you changed the the same code I was touching when I updated the print from sd menu. The issue might disappear if you clean the build folder after the error, and if that doesn't fix it, use arduino 1.8.13 (either by compiling with arduino, or by using PF-build.sh like this: `./PF-build.sh -o0 -i1.8.13`). If you want to find the code that triggered the issue, see those two lines in ultralcd.cpp (I couldn't return and call that function at the same time that always returned 1... something which works later in the same source file):
image


I'll review these changes next thursday since I'm busy with exams in the meantime, but it's looking good after a quick look over the changes.

gudnimg:
A few minor bugs I've found, working on fixes. I will commit changes when testing is done.

* When fullscreen message is displayed: `MMU needs user attention. Fix the issue and then press the button on MMU unit.` And I long-press --> the buzzer will beep endlessly. (**Fixed by** [9e9e338](https://github.com/prusa3d/Prusa-Firmware/pull/3473/commits/9e9e33886f64d445e91b62ae1487a4acaa918fb4))
* When long-pressing within a long-press menu, there is a double beep. (**Fixed by** [d3875ea](https://github.com/prusa3d/Prusa-Firmware/pull/3473/commits/d3875ea70b234421ff90548bac83b563205e031a) **and** [d7f7ef6](https://github.com/prusa3d/Prusa-Firmware/pull/3473/commits/d7f7ef604e9d214c0e85f959448fda1d6994e582))
* Pressing knob while Unloading Filament takes the screen --> results in endless beeping. The click is not being consumed. (**Fixed by** f5f411a)
* Long-pressing while Unloading Filament takes the screen (MMU) --> returns user to menu. I would expect the long-press to be ignored here. (**Fixed by** f5f411a)

leptun:
I recommend you wait a bit longer before changing mmu related stuff as it will be a nightmare to merge afterwards

gudnimg:
> I recommend you wait a bit longer before changing mmu related stuff as it will be a nightmare to merge afterwards

I think [f5f411a](https://github.com/prusa3d/Prusa-Firmware/pull/3473/commits/f5f411ac74c9733700e904b38db5ec4ec0658798) could be applied separately from this PR. The function `menu_lcd_lcdupdate_func()` needs to be disabled to suppress button feedback and also to disable long-presses. Currently in Master and release firmwares you can exit the Unload filament full-screen message with a long-press. The above commit should fix that.

gudnimg:
TODO for myself: `lcd_consume_click()` in `lcd_alright()` should not be needed after this PR as the firmware should guarantee there is no unconsumed click event waiting to be consumed.

gudnimg:
Rebased the PR now that we have the new MMU code.

gudnimg:
Removing the PR from Milestone. Probably won’t look into this until 3.13 is released.

gudnimg:
Notes for myself:

`Sound_MakeSound(e_SOUND_TYPE_ButtonEcho);` should be moved to `lcd_consume_click()` instead of `menu_lcd_lcdupdate_func()`. That way the sound is independent of whether or not LCD updates are enabled/disabled.

For the rotation event, I think we can create a similar function as `lcd_consume_click()` which will detect and consume the rotation event at global scope. We could also make it return a `enum` value to know if the rotation was clockwise or counter-clockwise.

```c
enum EncoderRotation {
    NONE = 0,
    CLOCKWISE = 1,
    COUNTER_CLOCKWISE = 2,
};
```
This function should also call `backlight_wake()` in case LCD updates are disabled.

PR #3475, opened 2022-04-17
Make M117 message persistent on LCD when used with M601 or G4

### LINKED ISSUE #1443: M117 fails
FW 3.5.1 for Prusa i3 Mk3

Entering the command `M117 Hello World ` does nothing to LCD display.

Trying to use the command in startup gcode and the FW parses lines above and below the M117 and ignores the display command.  

```
M109 S185 ; wait for extruder temp
M117 PINDA Soak Delay
G4 P15000 ; wait another 15 seconds to warm PINDA
```
In this case the display continues rolling the file name being printed from the SD card.

### LINKED ISSUE #2970: [FEATURE REQUEST] G4 "sleep..." message and M117
In custom scripts I often use the G4 command. When used alone there is no message on the printer display. But used as G4 P1000 or G4 S2 the display shows sleep... - which does not disappear after the G4 command.
Also if you want to display a message like: "M117 Waiting for 30 sec" this message is replaced by the "sleep..." when followed by a "G4 P30000" command.

My Request is to add a message string option to the G4 command, or to check wether there is something written by a previous M117 command in the display already, and then not to replace it with a not-anything-saying "sleep..."

Also, let the "sleep..." disapear when the pause is over.

Best regards...

### LINKED ISSUE #3316: [BUG]Custom Pause Print Message is overwritten
**Printer: MK3S**
**Firmware: 3.10**

**MMU: None**

**Print via SD Card**

**Describe the bug**
The PrusaSlicer's Custom Pause Message is quickly overwritten with 'Printer Paused' rendering the custom message ineffective.  The beeper does not sound according to the Sound Setting rules as does a Color Change.
  
  **To Reproduce**
* In PrusaSlicer, create a box shape.  
* In the Preview Mode, add a 'Pause Print' and custom message at any layer.  
* Monitor the LCD display as the paused layer begins to print.

**Expected Behavior**
* Printer pauses and persistently displays the Custom Message.
* Beeper sounds according to Sound Setting rules.

**G-code**
The relevant G-Code begins at Line 404 in the attached file.  Search for "MAGNETS".

**Video**
Attached iPhone video demonstates the bug
[Pause_Test.gcode.zip](https://github.com/prusa3d/Prusa-Firmware/files/7714837/Pause_Test.gcode.zip)

https://user-images.githubusercontent.com/45301979/146081423-7d7d8b7a-3ac9-4537-aaf3-00242f626cf1.MOV

.

gudnimg:
If M117 is called before M601 or G4 then we would like the custom message to
be visible on the LCD screen until the print is resumed.

Change in memory: +16 bytes of flash

Fixes #3316 (M601)
Fixes #1443 (G4)
Fixes #2970 (G4)

Video demonstration (with M601):

* Before (current Master branch): https://www.youtube.com/watch?v=8Mr-1lYwjsk
* After (this PR): https://www.youtube.com/watch?v=A7_uLBHg0qE

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -4272,7 +4272,7 @@ void process_commands()
         if (starpos != NULL)
             *(starpos) = '\0';
         lcd_setstatus(strchr_pointer + 5);
-        custom_message_type = CustomMsg::MsgUpdate;
+        custom_message_type = CustomMsg::M117;
     }
 
     /*!
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -4939,7 +4939,13 @@ if(eSoundMode!=e_SOUND_MODE_SILENT)
       codenum = 0;
       if(code_seen('P')) codenum = code_value(); // milliseconds to wait
       if(code_seen('S')) codenum = code_value() * 1000; // seconds to wait
-	  if(codenum != 0) LCD_MESSAGERPGM(_n("Sleep..."));////MSG_DWELL
+      if(codenum != 0)
+      {
+        if(custom_message_type != CustomMsg::M117)
+        {
+          LCD_MESSAGERPGM(_n("Sleep..."));////MSG_DWELL
+        }
+      }
       st_synchronize();
       codenum += _millis();  // keep track of when we started waiting
       previous_millis_cmd.start();
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -618,7 +618,7 @@ void lcdui_print_status_line(void)
         }
     } else { // Otherwise check for other special events
         switch (custom_message_type) {
-        case CustomMsg::MsgUpdate: //Short message even while printing from SD
+        case CustomMsg::M117:   // M117 Set the status line message on the LCD
         case CustomMsg::Status: // Nothing special, print status message normally
         case CustomMsg::M0Wait: // M0/M1 Wait command working even from SD
             lcd_print(lcd_status_message);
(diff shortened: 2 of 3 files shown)


Firmware/ultralcd.cpp
```diff
@@ -676,6 +676,9 @@ void lcdui_print_status_line(void)
         case CustomMsg::Resuming: //Resuming
             lcd_puts_at_P(0, 3, _T(MSG_RESUMING_PRINT));
             break;
+        case CustomMsg::M117:
```
leptun: Merge this case with these:
```
        case CustomMsg::MsgUpdate: //Short message even while printing from SD
        case CustomMsg::Status: // Nothing special, print status message normally
        case CustomMsg::M0Wait: // M0/M1 Wait command working even from SD
```
They are all identical and could probably be optimized better this way.

Firmware/Marlin_main.cpp
```diff
@@ -4272,7 +4272,7 @@ void process_commands()
         if (starpos != NULL)
             *(starpos) = '\0';
         lcd_setstatus(strchr_pointer + 5);
-        custom_message_type = CustomMsg::MsgUpdate;
+        custom_message_type = CustomMsg::M117;
```
leptun, line 4275: CustomMsg::MsgUpdate and CustomMsg::M117 can be one and the same.

gudnimg: Nice, didn't notice that!

gudnimg:
@leptun I renamed `CustomMsg::MsgUpdate` to `CustomMsg::M117` and reduced flash by 2 bytes.

Firmware/ultralcd.cpp
gudnimg: Done

gudnimg:
I added a commit which fixes the same issue when using `M117` with  `G4` command since it is supported by Marlin 2. Only adds 8 bytes more.

leptun (approved):
Seems to work nicely, except for one issue: The lcd_status_message gets cleared on resume. Probably not a serious issue, but I had to mention it.

PR #3477, opened 2022-04-18
Suggest features firmware management files

DrVallo:
I don't know if this is the right section, but I'd like to leave this message for a little suggestion that I think would be a great addition to include in the firmware of our Prusa printers.Quite often I find myself printing large projects split into many files. In order to remember what pieces I have printed I use a simple sheet of paper to note the files I have already printed. All this is ok, but sometimes I happened to lose the paper, and consequently I find myself to make confusion.It would be very nice to add a feature inside the Prusa firmware to allow to insert, for example, a tick symbol next to each file in the list of files contained in the sd card.

Maybe by holding down Knob on the LCD screen, you can insert or remove a tick next to each file name. In this way it will be very easy to keep track of the pieces already printed or to be printed.


3d-gussner:
@DrVallo  I think you want to report a feature request and not a pull request, please use https://github.com/prusa3d/Prusa-Firmware/issues/new/choose

Issue #3323, opened 2021-12-20
[BUG] PAUSE PRINT feature


rosauro3d:
3D printer brand / version + firmware version 
Prusa MK3S, firmware 3.10.0 and 3.10.1

Slicer
PrusaSlicer - Tested with 2.3.1, 2.3.2, 2.3.3

Operating system type + version
Windows 10 Pro, versioni 20H2

Behavior
In the Prusa slicer a PAUSE PRINT was added so that I could embed a nut to a project.

When the print job gets to that point of PAUSING and after adding the hardware I select resume print.
After selecting resume print the screen goes blank for 30 seconds.
The nozzle's temperature continues to drop as the print head continues with the job.
At the same time there is no filament being fed through the nozzle.

I've already reported this with Prusa support back in July 7, 2021.
I've done everything they've instructed me to do. (reflashing firmware, generate new gcode, redownload firmware files, full system refresh, tried slicers 2.3.1, 2.3.2 and 2.3.3, tried Pronterface, tried via SD, generate new gcode for a new project)
I thought the new firmware 3.10.1, that was release a few days ago, would have resolve this issue. Apparently not.
I was told by Prusa support that they will report the issue again and to also report it here as well.

For now I've gone back to firmware 3.9.3 where the PAUSE PRINT last worked.

There is a video I made of the issue that can be viewed here:
http://rosaurophotography.com/misc/videos/prusapause/

gudnimg:
I can try to reproduce this on my end (on MK3S+). But I’ve not used this feature before. Do you configure the pause in Prusa Slicer?

I think we should focus on fw v3.10.0

I’m currently printing a 2 day print which finishes tomorrow.

3d-gussner:
@rosauro3d have you searched the github for your issue? I think you have the same issue as https://github.com/prusa3d/Prusa-Firmware/issues/3148 describes.

Please read https://github.com/prusa3d/Prusa-Firmware/issues/3148#issuecomment-855201971
and https://github.com/prusa3d/Prusa-Firmware/issues/3148#issuecomment-966463702

Let us know if that helped.

rosauro3d:
@gudnimg Yes you configure the pause in Prusa Slicer.  
Check out this video:  https://www.youtube.com/watch?v=CtC5vAK94qA

[...]

Issue #3148, opened 2021-05-14
[BUG]3.10 (M601) doesn't heat after resume


redtif:
Printer type - MK3
Printer firmware version -  3.10

Octoprint

Describe the bug
Insert Pause via Prusa Slicer. Printer will pause at the desired Z. If you resume either at the printer or via Octoprint, it will start immediatily without heating up again. Also it will not extrude, because temperature are to low. It will finished the print without extruding. 
Going back to 3.9.3 will fix this.

If you need any more information, i am happy to provide them.

riedlingerm:
Identical Issue with my setup.
Printer type - MK3S
Printer firmware version - 3.10
Octoprint version - 1.6.1

I will stick with 3.9.3 as well since prints with insertion of parts (e.g. bearings, magnets) are practical impossible with this issue.

JohnnyDeer:
Hi @redtif, I can't reproduce this issue, work as expected. For how long you leave printer in pause? Can you please upload the Gcode where this is happening?

redtif:
Hi @JohnnyDeer, I tried the whole testcase again, 
the M601 ist by line 5344
first firmware 3.9.3 stops before 8.2, resume at printer, display shows again the printing temperature, all OK
Update Firmware
second firmware 3.10-4481, stops at 8.2, resume at printer, **No Display for a short time (5-10 seconds) but printing continues, the display normal but 0° temperature and printing continues........**
here the related gcode 

[TestInsert_0.2mm_PLA_MK3_11m_3.10.zip](https://github.com/prusa3d/Prusa-Firmware/files/6494429/TestInsert_0.2mm_PLA_MK3_11m_3.10.zip)

I revert to 3.9.3!

Rgds redtif

kg4iae:
I'm having the exact same issue.  Sliced the project with Prusa Slicer (V2.2).  Added a "Pause".  Printing using Repetier Server (1.04).  

In the previous firmware 3.9, the print would pause at the requested location.  The hotend would go to 0, and wait for me to put in the washer.  Then you have to do "Resume Print" on the printer.  The printer will heat back up and move into position.  Then you have to click "Continue" on Repetier Server.

[...]

PR #3481, opened 2022-04-22
Fix resuming USB print when it is paused via menu or M601

gudnimg:
This is my attempt at fixing: https://github.com/prusa3d/Prusa-Firmware/issues/3323 and https://github.com/prusa3d/Prusa-Firmware/issues/3148

Fix resuming USB print when it is paused via menu or M601
This fix behaves exactly like M602

I can't test this myself as I don't have Octoprint, so I will kindly ask for volunteers or feedback on the changes.

It is my understanding that when M601 calls `action:paused` we should resume with `action:resumed` since all control is handled via firmware. Such as extruder temperature and LCD visuals.

image

Current changes reduce flash use by 86 bytes.


gudnimg:
I noticed this screws up the state of the printer when you turn in on with SD card. The "pause print" and "resume print" menu item is available when it shouldn't :)

gudnimg:
Closing this PR for now, I might open it again once I setup Octoprint.

PR #3485, opened 2022-04-29
Fix AMBIENT_RAW_*_TEMP define names

wavexx:
The max/min temperature check were incorrectly using
HEATER_AMBIENT_RAW_*_TEMP (non-existing) instead of the correct
AMBIENT_RAW_*_TEMP (this is not a heater afterall).

This doesn't change the current behavior, since the check defaulted to
the correct path for the MK3+ configuration anyway.


leptun (approved):
Tested on MK3 with rebase. Not sure if this will cause conflicts with the new thermal regulation...

3d-gussner (approved):

PR #3486, opened 2022-04-30
Flash optimizations in loop() and status screen menu

gudnimg:
These are just small optimizations I have noticed. Thought I should commit them in case I see more or forget them.

The 2 commits save 22 bytes of flash. Removes one global variable

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -1940,7 +1940,7 @@ void loop()
 }
   //check heater every n milliseconds
   manage_heater();
-  isPrintPaused ? manage_inactivity(true) : manage_inactivity(false);
+  manage_inactivity(isPrintPaused);
   checkHitEndstops();
   lcd_update(0);
 #ifdef TMC2130
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -789,6 +789,11 @@ void lcd_status_screen()                          // NOT static due to using ins
 	else if (feedmultiply > 999)
 		feedmultiply = 999;
 
+	if (lcd_draw_update) {
+		// Update the status screen immediately
+		lcd_status_update_delay = 0;
+	}
+
 	if (lcd_status_update_delay)
 		lcd_status_update_delay--;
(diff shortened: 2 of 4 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -9839,15 +9839,7 @@ if(0)
 						lcd_update_enable(true);
 */
 						eFilamentAction=FilamentAction::AutoLoad;
-						bFilamentFirstRun=false;
-						if(target_temperature[0]>=EXTRUDE_MINTEMP){
-							bFilamentPreheatState=true;
-//							mFilamentItem(target_temperature[0],target_temperature_bed);
-							menu_submenu(mFilamentItemForce);
```
gudnimg: I have not tested this line. Need to verify we actually need this call to `menu_submenu`

gudnimg: I dropped this commit

Firmware/ultralcd.cpp
```diff
@@ -791,11 +792,9 @@ void lcd_status_screen()                          // NOT static due to using ins
 	if (lcd_status_update_delay)
 		lcd_status_update_delay--;
 	else
-		lcd_draw_update = 1;
```
leptun, line 794: I'm not so sure this change is needed. Before a lcd_draw_update event from elsewhere would re-render the status screen instantly. Now, a pending lcd update has to wait for the lcd_status_update_delay to expire, which will surely introduce some annoying delays. I'd keep it here how it was before

gudnimg: I see what you mean. Do you agree ca8d47a6da33e74b89a29b5c0a6bb1decac951bb is sufficient? I don't see a reason why `lcd_draw_update` needs to be assigned a value in `lcd_status_screen()`. But if it is set outside `lcd_status_screen()` then we can render the screen immediately. (Currently still saving 10 bytes overall :))

leptun: Yeah, that should work

DRracer (approved):

PR #3493, opened 2022-05-11
Add notices about Sheet::bed_temp and Sheet::pinda_temp being unused

wavexx:
Using the bed temp to compensate for different material expansion is a nice idea, but it's not currently done.

There is a potential for 2*8 bytes of extra sparse eeprom space, plus a
few bytes saved if we remove the update calls, so add a notice that
these values are currently only written-to, but never actually used.


leptun (approved):
Since it's only comments that are changed (and I also know the values aren't read anywhere), I'm approving the PR. @wavexx you can probably merge it immediately if you want

PR #3494, opened 2022-05-11
Jump to PLA temps in LCD temperature settings when starting from zero

### LINKED ISSUE #3085: [ENHANCEMENT] Manual temperature entry should have a non-zero initial value
MK3S - When there is no heating explicitly set and you go to Settings -> Temperature -> Bed | Nozzle and click the wheel the enter a temp, the initial value is zero, but I would suggest that the majority use-case is setting bed to >=60 and nozzle to >=200, which causes excess wear-and-tear on the rotary dial spinning it through (mostly) useless numbers.

wavexx:
When turning on one heater manually through the Settings -> Temperature menu, it's painfully slow to turn encoder to reach _any_ useful temperature (usually, a minimum of 180C is required).

This PR allows to set an initial "jump" value when starting from zero. It's used in the Settings -> Temperature menus for both the Nozzle and Bed, setting the jump point to 215/60C respectively.

That is, when the nozzle temp is zero, and the user is trying to increase the value, immediately set the value to 215 for further editing. The jump works only once after entering the menu, and the user is still allowed to decrease the value after the jump, so there's no functional restriction. For this reason, 215/60C seem a pretty good all-around value, since most other materials require higher temperatures.

This fixes #3085

It is somewhat expensive in terms of code size, increasing flash usage by 248 bytes (sadly, most of which is due to the ugly menu call interface).

### DIFF
--- a/Firmware/menu.h
+++ b/Firmware/menu.h
@@ -20,7 +20,8 @@ typedef struct
 {
     //Variables used when editing values.
     const char* editLabel;
-    void* editValue;
+    uint8_t editValueBits; // 8 or 16
+    void* editValuePtr;
     int16_t currentValue;
     int16_t minEditValue;
     int16_t maxEditValue;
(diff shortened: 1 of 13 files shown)


3d-gussner (commented):
I would suggest to jump to `EXTRUDE_MINTEMP` as this is close to the PLA temp but also to cold pull temp or MBL without oosing.

wavexx:
The temperature for a cold pull is not entirely defined though, depends on the material. MINTEMP is probably already too high for a cold pull.

I usually pull starting from cold, so any temp would do.

But I'm open to this. @leptun @gudnimg ?

leptun:
I'd jump to around 200C. This is usually the minimum temperature that is really used with any material. Jumping to EXTRUDE_MINTEMP is dangerous since due to temperature oscillations, the cold extrude check will kick in. The PLA temp I think is a bit too high and it might cause PLA to ooze, I'd much rather start at a safer temperature where the plastc oozes very little.

wavexx:
200 sounds quite reasonable to me.

wavexx:
The biggest problem here with the code size is that menu_item_edit_P is a template which gets expanded for uint8_t and int16_t (depending on the menu). This in turns generates an almost complete copy of _menu_edit_P.

IMHO these two functions could take a void* pointer and check for the type internally. And/or be rewritten so that most of the code just works with int16_t, and just does a final cast at the end.

The 8 and 16 bit types have different rendering. menu_draw_P have two completely different functions depending on type, which is also bad style to abuse template specialization for this.

wavexx:
I changed the default temperature to 200 and made it configurable in the variant file. I also added the same for the fan speed (where the default is simply 50%).

3d-gussner (approved):
Tested on MK404 and real printer
- Jump to min value :heavy_check_mark: 
  - Settings -> Temperature -> Nozzle :heavy_check_mark: 
  - Settings -> Temperature -> Bed :heavy_check_mark: 
  - Settings -> Temperature -> Fan :heavy_check_mark:

- Turning the knob clockwise jumps to set value
- Turning the knob counter clockwise disables the jump and following clockwise turns increase the temp as known.

3d-gussner:
Cost +250 bytes of flash

Firmware/menu.h
```diff
@@ -20,7 +20,8 @@ typedef struct
 {
     //Variables used when editing values.
     const char* editLabel;
-    void* editValue;
+    uint8_t editValueBits; // 8 or 16
+    void* editValuePtr;
```
gudnimg, line 24: I like this rename 👍 makes it more clear its a pointer to some other variable

wavexx:
Should hopefully be less wasteful with the latest changes.

3d-gussner (approved):
Retested on MK404

Save 92 bytes

gudnimg (approved):
Works great on my MK3S+, everything seems to work as expected.

PR #3495, opened 2022-05-11
Replace EXTRUDE_MINTEMP with the configurable extrude_min_temp

### LINKED ISSUE #2577: [BUG] The Move Extruder menu ignores the M302 (Allow cold extrude)
The M302 command allows for a cold extrusion. The following sequence of commands indeed moves the extruder, even with cold hotend:
M302
M83
G1 E4

But the Move Extruder by the Settings menu, responds with the "Preheat the nozzle!" error.

Printer type: MK3S
Printer firmware version: 

Expected behavior
  The Move Extruder by the Settings menu, should allow the movement of the extruder, if a M302 (or M302 S<temp>, where temp < current) has already been sent to the printer.

wavexx:
Everywhere MINTEMP is checked, use the configurable value set
by M302, not an hardcoded value. As mentioned in #2577, I hit this a couple of times.

EXTRUDE_MINTEMP is now used only as the initial default value.

Reduce the precision of extrude_min_temp to an integer to reduce the
generated code size (constant folding did in fact do the same previously
anyway). Having tenths of degrees is not necessary for this feature.

This is based on the previous version by @leptun in https://github.com/leptun/Prusa-Firmware/tree/MK3-cold_extrude (and the above is the reason why  it generated significantly larger code).

The new code though is still 83 bytes larger. I didn't check if this is due to the extra reads and/or some extra manual casts could help us get down further.

This fixes #2577

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3421,7 +3421,7 @@ static void gcode_G80()
     go_home_with_z_lift();
     //		SERIAL_ECHOLNPGM("Go home finished");
     //unretract (after PINDA preheat retraction)
-    if ((degHotend(active_extruder) > EXTRUDE_MINTEMP) && eeprom_read_byte((unsigned char *)EEPROM_TEMP_CAL_ACTIVE) && calibration_status_pinda() && (target_temperature_bed >= 50)) {
+    if ((degHotend(active_extruder) > extrude_min_temp) && eeprom_read_byte((unsigned char *)EEPROM_TEMP_CAL_ACTIVE) && calibration_status_pinda() && (target_temperature_bed >= 50)) {
         current_position[E_AXIS] += default_retraction;
         plan_buffer_line_curposXYZE(400);
     }
@@ -7916,8 +7916,8 @@ Sigma_Exit:
     */
     case 302:
     {
-	  float temp = .0;
-	  if (code_seen('S')) temp=code_value();
+	  int temp = 0;
+	  if (code_seen('S')) temp=code_value_short();
       set_extrude_min_temp(temp);
     }
     break;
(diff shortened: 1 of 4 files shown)


gudnimg:
>  I didn't check if this is due to the extra reads and/or some extra manual casts could help us get down further.

I think it is due to comparing a float to int. When I compare the output from godbolt.org I get a much larger assembly code with `if(float > int)` rather than `if (float > float)`. Very interesting.

`if(float > int)`:

```assembly
        lds r24,temp
        lds r25,temp+1
        mov __tmp_reg__,r25
        lsl r0
        sbc r26,r26
        sbc r27,r27
        mov r22,r24
        mov r23,r25
        mov r24,r26
        mov r25,r27
        rcall __floatsisf ; looks like the int is converted to float
        mov r14,r22
        mov r15,r23
        mov r16,r24
        mov r17,r25
        lds r24,my_current
        lds r25,my_current+1
        lds r26,my_current+2
        lds r27,my_current+3
        mov r18,r24
        mov r19,r25
        mov r20,r26
        mov r21,r27
        mov r25,r17
        mov r24,r16
        mov r23,r15
        mov r22,r14
        rcall __ltsf2
        tst r24
        brlt .L5
```

`if (float > float)`:
```assembly
        lds r24,my_current
        lds r25,my_current+1
        lds r26,my_current+2
        lds r27,my_current+3
        lds r18,temp
        lds r19,temp+1
        lds r20,temp+2
        lds r21,temp+3
        mov r22,r24
        mov r23,r25
        mov r24,r26
        mov r25,r27
        rcall __gtsf2
        cp __zero_reg__,r24
        brlt .L5
```

The code used to generate this:
```c
float my_current = 1.242352;
int temp = 230; // change int here to float to generate if (float > float)
int hello = 0;
void setup() {

}

void loop() {
    if (my_current > temp) {
        hello++;
    }
}
```

wavexx:
Yup, I suspected as much.

With the previous hard-coded define "185" would automatically take the type of either float (for degHotend) or int (for target_temperature).

wavexx:
By avoiding the float promotion we're now just 12 bytes extra. Much better.

leptun (approved):
The changes look good. Don't see any potential pitfalls.

PR #3504, opened 2022-05-22
Simplify `printer_smodel_check()`

gudnimg:
I was debugging this function on the MMU branch and the problem ended up not being due to `printer_smodel_check()`. But in the process I had simplified the function a bit.

**Changes saves 140 bytes of flash and 4 bytes of SRAM**

To test this I recommend using `M862.3 P "MK3SMMU2S"` directly in the serial stream. That way you don't actually need to start a print. Interchange `MK3SMMU2S` with anything you like, `MK3SMMU2S` is just one example.

The gcode `M862.3 P "string"` is found within any gcode file sliced by PrusaSlicer. And we need to check if the variable `sPrinterName` in the firmware matches the string parsed from the gcode file.

The comparison is also now case sensitive. Previous implementation ignored the case.

### DIFF
--- a/Firmware/util.cpp
+++ b/Firmware/util.cpp
@@ -444,35 +444,23 @@ pStrEnd=strchr(pStrBegin,GCODE_DELIMITER);
 if(!pStrEnd)
      return(NULL);
 *nLength=pStrEnd-pStrBegin;
-return(pStrBegin);
+pStrBegin[*nLength] = '\0';
+return pStrBegin;
 }
 
 void printer_smodel_check(char* pStrPos)
 {
 char* pResult;
 size_t nLength,nPrinterNameLength;
-bool bCheckOK;
-char sPrinterName[PRINTER_NAME_LENGTH+sizeof(ELLIPSIS)-1+1]="";
 
-nPrinterNameLength=strlen_P(::sPrinterName);
-pResult=code_string(pStrPos,&nLength);
-if(pResult!=NULL)
-     {
-     strlcpy(sPrinterName,pResult,min(nPrinterNameLength,nLength)+1);
-     if(nLength>nPrinterNameLength)
-          strcat(sPrinterName,ELLIPSIS);
-     bCheckOK=(nLength==nPrinterNameLength);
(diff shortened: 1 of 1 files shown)


Firmware/util.cpp
```diff
@@ -444,35 +444,23 @@ pStrEnd=strchr(pStrBegin,GCODE_DELIMITER);
 if(!pStrEnd)
      return(NULL);
 *nLength=pStrEnd-pStrBegin;
-return(pStrBegin);
+pStrBegin[*nLength] = '\0';
```
gudnimg, line 447: Note: This change make it so that the return string is `MK3SMMU2S` and not `MK3SMMU2S"`

Firmware/util.cpp
```diff
@@ -444,35 +444,23 @@ pStrEnd=strchr(pStrBegin,GCODE_DELIMITER);
...
-//SERIAL_ECHOLNPGM("\"");
-//SERIAL_ECHOPGM("expected: \"");
-////SERIAL_ECHO(sPrinterName);
-//SERIAL_ECHOLNPGM("\"");
+nPrinterNameLength = strlen_P(sPrinterName);
+pResult = code_string(pStrPos,&nLength);
+
+if(pResult != NULL && nLength == nPrinterNameLength) {
+     // Only compare them if the lengths match
+     if (strcmp_P(pResult, sPrinterName) == 0) return;
```
leptun, line 461: ~I think you can use strcasecmp here without increasing the code size since this one is also included by the sd sorting code~

nevermind. Just noticed we are using the _P variant. That one is indeed not included

leptun (approved):
I have a really hard time understanding the old code. Why are there so many copy and concat operations? And what is that elipsis supposed to even be? In any case, your new approach looks good to me. I tested this in MK404 with a MK3S setup. `M862.3 P "MK3SMMU2S"` gets refused, while `M862.3 P "MK3S"` gets accepted, which is correct. The MK3 variant also works properly.

While the new code is not 100% compatible with the old one (now it is case sensitive), I think it is perfectly fine and it should work properly with existing gcodes. The resources saved are also a nice thing to see.

leptun (approved):

PR #3519, opened 2022-06-29
Correctly show XFLASH initialization errors

wavexx:
Remove the first serial message, which cannot be displayed since serial
is not available yet.

Move both the serial message and display after
serial/temperature/planner is setup, so that we can use
lcd_show_fullscreen_message_and_wait_P properly.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -1048,8 +1048,8 @@ static void fw_crash_init()
 
 static void xflash_err_msg()
 {
-	lcd_clear();
-	lcd_puts_P(_n("External SPI flash\nXFLASH is not res-\nponding. Language\nswitch unavailable."));
+    puts_P(_n("XFLASH not responding."));
+    lcd_show_fullscreen_message_and_wait_P(_n("External SPI flash\nXFLASH is not res-\nponding. Language\nswitch unavailable."));
 }
 
 // "Setup" function is called by the Arduino framework on startup.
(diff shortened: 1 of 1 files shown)


leptun (approved):
The idea looks good, but I have some doubts about this claim:
> Remove the first serial message, which cannot be displayed since serial
is not available yet.

The serial should already be enabld at that point (see right above the xflash stuff). Below it is the farm mode exceptions which yes, do in fact reinitialize the serial port again, but I wouldn't consider that the end of the world. In any case, if everything works, then I don't have anything against this change.

wavexx:
@3d-gussner can you test this? We should be able to simulate this with MK404

3d-gussner:
I will do tomorrow

leptun:
Random question: where is the travis build?

3d-gussner:
> Random question: where is the travis build?

I had in the past similar issue that travis didn't run. I forced another change and reverted it, but that isn't ideal.
If I have PRs where travis fails, I create a fresh new worktree/branch and copy and commit the changes again. That worked for me at least few times.

3d-gussner (approved):
Tested on MK404 (with simulated broken xflash) and the LCD warning message now waits for acknowledge with knob click. Before this PR the message disappears quite fast.    
The serial out put `XFLASH not responding` has been there before.
Also the the serial output generates `echo:busy: paused for user` to keep connection to host systems like Octoprint alive.

PR #3521, opened 2022-07-01
Fix Mk2.5/S zip after change PO-based language translation support

3d-gussner:
Limit atmegaMK404 boards mem to 8,16,32


Firmware/language.h
```diff
@@ -29,6 +29,7 @@
 #define ISTR(s) s
 #define _i(s) _I(s)
 #define _T(s) s
+#define _O(s) s
```
wavexx, line 32: This is unrelated

wavexx:
Overall this looks ok, but I would drop [dc812da](https://github.com/prusa3d/Prusa-Firmware/pull/3521/commits/dc812dabb2769a7f31bfe5f0363a548b6d4500d4) which won't be needed in this relase.

Also, I would drop or merge the repeated commits changing the build numbers.

leptun:
@wavexx Is that really a problem? It's not like the _O() is used anywhere yet, so I don't see how it might cause an issue. But I see that the commit is a bit off topic in this PR.

wavexx:
Not a problem, but it just makes the PR busy. For 3.11.1 I'd expect only non-code fixes at this point

PR #3523, opened 2022-07-01
Fix mmu startup

vintagepc:
Fix the startup sequence where the MMU2 tries to printf() to the serial port before it's been initialized.


vintagepc:
Wrong destination

PR #3527, opened 2022-07-02
Fix a crash during file check when `CMDBUFFER_DEBUG` is defined

gudnimg:
There is so much serial output during the file check that it can slow the file check enough to timeout the watchdog. But this is only an issue if `CMDBUFFER_DEBUG` is defined

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -7784,6 +7784,11 @@ static bool check_file(const char* filename) {
 		card.sdprinting = true;
 		get_command();
 		result = check_commands();
+#ifdef CMDBUFFER_DEBUG
+		// Kick watchdog because the file check is very slow
+		// with the CMDBUFFER_DEBUG enabled
+		manage_heater();
+#endif // CMDBUFFER_DEBUG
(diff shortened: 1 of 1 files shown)


leptun (approved):
This appears to work, but I'm wondering if this will cause a conflict with the new thermal regulation. I think it won't, but we should check that just to be sure.
@wavexx what do you think of this? This manage_heater gets called very frequently during check_file. If it causes no issues, then I propose we merge this right away. In any case, it's debug code so it won't affect anything important

3d-gussner:
> This appears to work, but I'm wondering if this will cause a conflict with the new thermal regulation. I think it won't, but we should check that just to be sure. @wavexx what do you think of this? This manage_heater gets called very frequently during check_file. If it causes no issues, then I propose we merge this right away. In any case, it's debug code so it won't affect anything important

@wavexx Can you please comment/review this one?

wavexx (approved):

wavexx:
Looks good to me. manage_heater is self-limiting due to the ADC conversion, so it's not a problem to call it often.

PR #3530, opened 2022-07-04
tools: Do not use os.EX_DATAERR due to unavailability on windows

wavexx:
Just return a plain exit code...


leptun (approved):
Tested with git bash in windows 10

PR #3542, opened 2022-07-11
M300: do not set a default frequency if S is not set

wavexx:
Fallback to use the default HIGH value if a frequency has not been
specified.

This makes the tone of "M300" and "M300 P<x>" identical as a result.

Supersedes #3339


leptun:
I don’t see how the “HIGH” beep would be made when S is 0. To me it looks like just a delay and nothing more.

wavexx:
In Sound_MakeCustom this is checked explicitly.

wavexx:
I've removed the "else" branch entirely. IMHO it doesn't make sense anymore.

leptun (approved):
@wavexx Please see 3ed3f87. I've made it so M300 S... P0 doesn't beep at all. Otherwise the PR looks good to me

PR #3550, opened 2022-07-27
Fix crash due to recursion when printing a file is done

### LINKED ISSUE #3549: [BUG][High priority] Lack of M84 in g-code files causes firmware crash
**Printer type** - MK3S+
**Printer firmware version** -  3.11.0 and 3.11.1-RC1 using official hex releases

**SD card or USB/Octoprint**
Printing from SD card directly.

**Describe the bug**
Printing from g-code file (on sd card) that doesn't have M84 command causes firmware crash after processing
all existing g-code commands in file.

I've verified that all g-codes are processed by adding bunch of "M117 test" commands at the end of gcode
file to get "test "text on LCD and I got it on LCD.

**To Reproduce**
* decompress and run attached gcode file from sd card on the printer
* wait for it to finish and then observe crash 
 
**Expected behavior**
No crash.

**G-code**
This one reliably crashes for me:

[test-gcode-only-moves.gcode.gz](https://github.com/prusa3d/Prusa-Firmware/files/9198145/test-gcode-only-moves.gcode.gz)

Much less reliable but still sometimes even 3 line g-code file would also crash:
````
G90
G28 W
G1 Z50 F720 
````

**Crash dumps**
Nine crash dumps dumped by D21:
[crash-dumps.txt.gz](https://github.com/prusa3d/Prusa-Firmware/files/9198337/crash-dumps.txt.gz)

**Complete log**
[serial-crash-log.txt.gz](https://github.com/prusa3d/Prusa-Firmware/files/9200376/serial-crash-log.txt.gz)

**Video**
https://www.youtube.com/watch?v=FN9xwIY07h4

gudnimg:
The recursion could cause a crash

`manage_inactivity()` --> `get_command()` --> `card.printingHasFinished()` --> `st_synchronize()` --> `manage_inactivity()` (repeat)

Fixes #3549

### DIFF
--- a/Firmware/cardreader.cpp
+++ b/Firmware/cardreader.cpp
@@ -1007,7 +1007,6 @@ void CardReader::flush_presort() {
 
 void CardReader::printingHasFinished()
 {
-    st_synchronize();
     if(file_subcall_ctr>0) //heading up to a parent file that called current as a procedure.
     {
       file.close();


gudnimg:
**BEFORE:** Endless recursion until firmware crashes
image


**AFTER:**
image

leptun:
Can you pinpoint when this got broken? I’d like to see which PR made this change so that we confirm nothing gets broken by no st_sync.

gudnimg:
> Can you pinpoint when this got broken? I’d like to see which PR made this change so that we confirm nothing gets broken by no st_sync.

This call to `st_synchronize()` has been in the firmware for 6 years: https://github.com/prusa3d/Prusa-Firmware/commit/30f0528aba00e5949496f4fe7adcbef5b5a18b5c#diff-533294493df865ca3fdc626d3ed0ec2271c71c890cfc54f03e647d0a7417bd61R634

I haven't tried older versions of the firmware to see if this is a recent issue or not. 

I have a few logs from recent MMU prints and I don't see any recursion there. But as the reporter mentioned, the crash doesn't appear if M84 is present in the gcode file. `buflen` is `0` so I would have thought `st_synchronize()` wouldn't do anything here regardless if M84 is included or not.

wavexx:
We cannot really remove that st_synchronize without doing other changes.

Looks like this function was intended to handle gcode chaining? (never seen that so far TBH). However, the problem is that it calls quickStop() later, which discards the plan, so indeed this was intended to allow the plan to finish before calling quickStop().

Do we have instructions that refer to the current "card" object? I don't think so from memory. This is to know if we can close the file while finishing the moves. I would say it's safe to do.

The question is whether that quickStop() is needed at all there. In fact, seems like a dumb thing to do if the plan is already empty?

The second problem is the call to finishAndDisableSteppers(). This _will_ need an st_sync if we do it this way.

leptun:
[test-gcode-only-moves.zip](https://github.com/prusa3d/Prusa-Firmware/files/9202431/test-gcode-only-moves.zip)
Even with M84, you can still make the old code crash. If the M84 is moved one line higher, it stops crashing

gudnimg:
> Looks like this function was intended to handle gcode chaining? (never seen that so far TBH). However, the problem is that it calls quickStop() later, which discards the plan, so indeed this was intended to allow the plan to finish before calling quickStop().

In this crash scenario we don’t get to run `quickStop()` because the `st_synchronize()` crashes the FW first. 🤔

 I wonder if a solution to this could be to prevent `manage_inactivity` from running recursively? similar to `mmu_loop()` … I havent thought it through, just thinking out loud.

leptun:
@gudnimg we could try that. Something like a static variable acting as a “lock”. This might prevent the crash, but I think there is some other issue in the cmdqueue that is causing the eof section to trigger too early. It just doesn’t feel right that none of the protections we already have are not preventing the early eof event.
In the gcode I posted above, I don’t even think we get to execute the M84 because it’s already crashed.

wavexx:
Reading throughout the code, the main problem is similar to the current behavior of pause/stop, in the sense that it's trying to process the action immediately before actually going back to the main loop.

I have a fix for pause/stop coming in another PR which essentially removes the need of synchronizing. It does this by changing the lcd_command_type and handle this in the main loop instead. It would make sense to merge that first.

printingHasFinished is called in a few other places where it does need st_synchronize at the moment. I would fix this in another way. The major issue is that cmdqueue will still attempt process the SD card within any other st_synchronize call.

I think we should short-circuit this immediately in cmdqueue.cpp:639, by setting sdprinting to false there, before we call any other function. This will ensure any other manage_inactivity call will still process the serial, but skip the SD very early before triggering this again. This IMHO is needed even if we fix the "printing finished" correctly as stated above. I think setting sdprinting=false there is sufficient without introducing yet another global. Looks like printingHasFinished resets it anyway, and in the other path (chained gcode) it is reset to true again.

Another worrying thing is that I realized that this code is essentially running also during check_file right? So technically we could be overflowing the stack also quite early?

wavexx:
It's interesting that I couldn't reproduce this with the attached file. Probably your SD cards are faster than mine? By the time it took to read through the EOF comments the moves were completed already and it never crashed.

I can only do this if I truncate the file to the last G1, then it crashes as expected.

wavexx:
I created #3551. Seems to work (but please test it on your setup - I couldn't get it without modifying the file by hand), and I cannot spot any downside at the moment. It's not a perfect fix, but it's certainly something I would include in 3.11.1 before we have something proper.

gudnimg:
Closing this in favor of #3551 

We can reopen this if we want later.

PR #3551, opened 2022-07-27
Prevent re-entry in EOF command processing

### LINKED ISSUE #3549: [BUG][High priority] Lack of M84 in g-code files causes firmware crash

wavexx:
cmdqueue will run commands when EOF is reached without returning to the
main loop, which is already incorrect.

However, since it needs to ensure the queue is empty, an st_synchronize
call can result in a re-entrant call to get_command, which will
reprocess EOF again. Even if we removed st_synchronize, another command
could be picked by an unsuspecting manage_inactivity() somewhere else.

Short-circuit EOF processing by clearing card.sdprinting immediately.
This is safe to do, since it's already done by printingHasFinished()
called a few lines down. We're just doing this earlier.

This should fix #3549

### DIFF
--- a/Firmware/Dcodes.cpp
+++ b/Firmware/Dcodes.cpp
@@ -44,7 +44,7 @@ void print_hex_word(daddr_t val)
     print_hex_byte(val & 0xFF);
 }
 
-int parse_hex(char* hex, uint8_t* data, int count)
+int parse_hex(const char* hex, uint8_t* data, int count)
 {
 	int parsed = 0;
 	while (*hex)
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -4248,7 +4248,7 @@ void process_commands()
     */
 
     else if (code_seen_P(PSTR("M0")) || code_seen_P(PSTR("M1 "))) {// M0 and M1 - (Un)conditional stop - Wait for user button press on LCD
-        char *src = strchr_pointer + 2;
+        const char *src = strchr_pointer + 2;
         codenum = 0;
         bool hasP = false, hasS = false;
         if (code_seen('P')) {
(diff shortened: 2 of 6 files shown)


leptun:
Won’t this cause the lcd to switch to idle mode too quickly? Also, wouldn’t this cause a pause event after eof is reached to fail?

wavexx:
The LCD doesn't seem to be an issue, but the timers are checking IS_SD_PRINTING so it is a problem for that :(

I pushed another alternative approach on the same idea. We close the file early, and check for the file status (which is not a bad thing to do here anyway).

The file is correctly reopened when chaining, and I don't see relevant situation where the file state is checked besides in layer1cal

gudnimg:
Tested this quickly on my printer. No crash and the file is only closed once 👍

gudnimg (approved):

wavexx:
Rebased on the MK3 branch

DRracer (approved):

PR #3564, opened 2022-08-20
Filament loading state fixes

### LINKED ISSUE #2651: [BUG] Weird behavior if print is started during filament auto-load
**Printer type** - MK3S (kit, no hardware modifications except cosmetics)
**Printer firmware version**- 3.9.0-RC3

**MMU Upgrade** - none
  
**To Reproduce**

-   Printer on, no filament inserted, hot-end temperature ready for filament insert, filament autoload is on, SD-card with printable files are in.
-   Start inserting filament
-   Navigate to SD-card menu for any print and start print (press knob) while filament is being pulled in
  
  It results in short motor movement (including repeated ramming into end stop, as shown in video, have seen both X and Y movements, X in both directions) and after that sometimes printer restarts, sometimes it beeps (even if set to silence) and display "Inactivity Shutdown" and locks down.

**Expected behavior**

  Print is started, home screen is displayed. After filament is loaded - prompt about filament load and color is displayed, print is continued. Alternatively error is displayed that print can't be started during filament load or even SD-card menu is not accessible.

**Video**

https://www.youtube.com/watch?v=PC1uv23G0-E

### LINKED ISSUE #3297: [BUG] printer gets confused when starting a print while its waiting for fillament unload
MK3S, firmware 3.10.0

USB/octoprint

**Describe the bug**
If you start a print while the printer is already waiting for you to press the knob to initiate a fillament change, its getting confused.
  
**To Reproduce**

- Start manual fillament change while printer is cold.
- Wait for printer to heat up.
- Printer asks "press know to start fillament change."
- Instead of doing that, just start a print via ocotprint. (I forgot :) )
- Now the printer just starts printing while still staying in this screen.
- If you press the knob to initiate the change, it will do so.
- However, after this it will ask AGAIN to press the knob and remove the filament (you cant since its already removed)
- After fiddling around with reinserting etc, it continues printing, but at the wrong Z height. (since the initial change moved it up)
 
**Expected behavior**
I would have expected it to either ABORT/IGNORE the change, or do the change correctly.

### LINKED ISSUE #3595: [BUG]<Firmware crash at loading filament and print file Firmware 3.11>
**Printer type - MK3S
**Printer firmware version - 3.11

SD card 

Describe the bug
  During filament automatic loading, I have chosen a file to print from the SD card, and pressing print the printer suddenly restarted.
  
To Reproduce
Unload the filament and auto-load the filament. Go to the main menu and then choose a file to print, press print, and wahtch.

wavexx:
There are some issues in the management of "filament autoloading" when the printer is idle. Because the autoloading is performed asynchronously, we need to ensure incompatible actions are disabled while loading is taking place.

With this PR we make impossible to perform another autoload/unload/filament action when one is already taking place (the autoload/load/unload menus are now hidden while autoloading is taking place).

To do so we check for the loading_flag. This was previously done for loading only, but doing so for unload makes the behavior uniform (#2318 performs more than a single move and is also non-blocking - you shouldn't load while the old unload is still pending).

Using bFilamentState flag would have been a little bit cleaner here (removing the need for loading_flag), but it's currently used both as a flag for autoload and also for the autoload state...

We also now prevent to start an SD print while autoload is taking place (fixes https://github.com/prusa3d/Prusa-Firmware/issues/2651). You can still navigate the SD card when waiting, but clicking on the file will just beep (see my comments on b0561a0aa1ba757b6e442c17995f17fc4e969d2c)

As a side note, in the first commit, we change PRINTER_ACTIVE to consider any custom command to be non-idle, which is logical. All custom commands are incompatible with each other, and will return the printer to Idle when complete (LongPause included). Checking for Layer1Cal in the main menu is simply incorrect.

Change in memory (MK3S+ Multilang):
Flash: -8 bytes
SRAM: -1 byte

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -278,7 +278,6 @@ extern uint32_t start_pause_print; // milliseconds
 extern ShortTimer usb_timer;
 extern bool processing_tcode;
 extern bool homing_flag;
-extern bool loading_flag;
 extern uint32_t total_filament_used; // mm/100 or 10um
 
 /// @brief Save print statistics to EEPROM
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -179,7 +179,6 @@ bool mesh_bed_leveling_flag = false;
 
 uint32_t total_filament_used;
 HeatingStatus heating_status;
-bool loading_flag = false;
 int fan_edge_counter[2];
 int fan_speed[2];
 
@@ -3579,13 +3578,12 @@ void gcode_M701(float fastLoadLength, uint8_t mmuSlotIndex){
 
         Sound_MakeCustom(50, 500, false);
 
-        if (!farm_mode && loading_flag) {
+        if (!farm_mode && (eFilamentAction != FilamentAction::None)) {
             lcd_load_filament_color_check();
         }
         lcd_update_enable(true);
         lcd_update(2);
         lcd_setstatuspgm(MSG_WELCOME);
-        loading_flag = false;
         custom_message_type = CustomMsg::Status;
     }
 
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -513,9 +513,8 @@ bool __attribute__((noinline)) printJobOngoing() {
 bool __attribute__((noinline)) printer_active() {
     return printJobOngoing()
         || isPrintPaused
-        || (custom_message_type == CustomMsg::TempCal)
         || saved_printing
-        || (lcd_commands_type == LcdCommands::Layer1Cal)
+        || (lcd_commands_type != LcdCommands::Idle)
         || MMU2::mmu2.MMU_PRINT_SAVED()
         || homing_flag
         || mesh_bed_leveling_flag;
--- a/Firmware/Prusa_farm.cpp
+++ b/Firmware/Prusa_farm.cpp
@@ -242,7 +242,7 @@ void prusa_statistics(uint8_t _message) {
         else if (isPrintPaused) {
             prusa_statistics_case0(14);
         }
-        else if (IS_SD_PRINTING || loading_flag) {
+        else if (IS_SD_PRINTING || (eFilamentAction != FilamentAction::None)) {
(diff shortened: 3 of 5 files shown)


gudnimg:
Is `loading_flag` only applicable for autoloading? Should we consider renaming it to `auto_loading_flag` or something to that effect?

wavexx:
> Is loading_flag only applicable for autoloading? Should we consider renaming it to auto_loading_flag or something to that effect?

I would be in favor to rename this as auto[_]loading_flag, but we need to this after the MK3_MMU2 branch is merged.

DRracer (approved):

wavexx:
@gudnimg rereading the code it seems that now eFilamentAction does what we need, so I removed loading_flag completely.

The old code was resetting the "autoload" flag for some reason. I changed the meaning so that the "AutoLoad" state is the moment when autoload has been triggered, but loading is still not taking place. AutoLoad can be cancelled (if you're still in the preheat menu), while "Load" is already starting to feed the filament and cannot be undone.

IMHO this is a lot saner, and I hope I understood all existing scenarios.

Can you review the MMU part? I added one case in lcd_wizard_load which I think was missing.

@3d-gussner because eFilamentAction is now used everywhere this requires some retesting of #2651 but also retesting all scenarios with loading/unloading, autoloading+cancel, and using the menu load/load filament correctly clears the load flag, otherwise you won't see the "load/unload" menus anymore.

gudnimg (approved):
Looks good to me. I think you're right about `eFilamentAction = FilamentAction::MmuLoad` being missing before.

wavexx:
Should fix #3297

gudnimg:
@wavexx I rebased the PR to resolve the merge conflicts. Hopefully I didn't break anything. The PR was quite a bit behind MK3 branch :)

gudnimg:
Rebased again. For some reason I had to re-do the entire rebase (~600 commits)

3d-gussner (approved):
Approved and fixes issues like https://github.com/prusa3d/Prusa-Firmware/issues/3595
Other tests with MK404 show that this PR is an improvement to previous version.

PR #3566, opened 2022-08-21
Fix multi segment pause-resume

### LINKED ISSUE #3490: [BUG] Fix resume/pause for arc (G2/G3) commands
This relates to all printer types and all firmware up to the latest release for both SD and serial printing.

Pause/resume is broken for arc commands.  To reproduce, send a long arc command, pause in the middle of the arc, and resume.
  
Test gcode for the issue can be found [here - arcs(1).zip](https://github.com/prusa3d/Prusa-Firmware/files/8621142/arcs.1.zip)

Video of the issue, from @leptun: https://photos.app.goo.gl/duduGsUVGoHSdT7k8

leptun:
This PR fixes #3490.

~This PR is based on top of #3552. As a result, there are a lot of unrelated commits. If you want to see a clean diff, see: https://github.com/wavexx/Prusa-Firmware/compare/temp_model_check...leptun:Prusa-Firmware:fix_multi_segment_pause_resume?expand=1~
This PR will get merged to the MK3 branch after #3552 is merged some time next week.

Previously, the pause-resume worked like this: For each G0/1 command, along with each planner segment, the target destination of that command was storred so that in the event the move was interrupted, it could be resumes. This patch was needed since executing the same gcode from the paused position when relative moves were enabled (as it is the case for the extruder), the relative distance would be applied to the current position (pause position), not to the start position of the move (the position before the current command even began executing.
This worked fine for G0/1 since all you needed to know in order to resume was where the pause position was (current position) and what the target was the first time the command got executed (to account for relative moves). Unfortunately, this approach could not be used for G2/3 command. This is because the Arc commands also need the center position of the arc, not just the target XYZE position. This center position was always relative, so on resume the center would be recalculated incorrectly, the offsets being applied to the pause position. In theory, the same apprach as for the G1 could be used (storing the initially computed I/J offests in the planner), but that would have required at least 128B of extra memory for the planner and some more math in order to correctly apply the offsets on resume. This approach would have been wasteful on resources, especially since G2/3 usage isn't mainstream YET (hopefully it will be soon).

The new approach:
Instead of storing the initially computed target position, we can instead store the start position of the gcode command. This way, when we resume, we can apply the relative offsets directly on top of the start position all over again. This leaves another problem though: How do you continue from the pause position if you only know the start position and target position (and I/J offsets)? The answer is that you don't. In order to make this work, you also need to know which segment got interrupted in order to be able to resume. As a result, along with the start position, we also store a uint16_t segment index in each block. This index is the one used by the for loop in the gcode that generates the multi-segment move.
As a result, we now know on resume the start position, the target position (and I/J offsets), the current position (where we paused) and also the index of the segment that was interrupted.
In the case of G0/1, we just initialize the start index of the for loop with the saved index. This means that only the needed segments are queued. This is possible since the current segment doesn't rely on information computed on the previous segment.
In the case of G2/3, we can't initialize the for loop to the saved segment since inside the for loop, the current segment relies on data computed in the previous segment. The only problem this causes is performance loss on resume since all positions up to the saved segment index need to be recomputed and discarded (move not planned). The worst this can do is create a small pause before resuming.

Another advantage of the new approach besides fixing the issue in #3490 is the consistency of the resumed segments was improved. I tried to explain this problem here: https://github.com/prusa3d/Prusa-Firmware/pull/2657#issuecomment-1114581657
In the context of G2/3, the problem was that the distance between the current position and the center was not constant (and it really couldn't have been since we are interpolating the arc using segments, so it's not a perfect arc). The issue would show itself on resume when the pause position did not happen exactly at the end/beginning of a segment. This would mean that the resumed segments would not align exactly with what might have been a previous or next layer and result in a surface artifact. This would have been especially apparent with arcs since those are used usually on the perimeters and any inconsistency in the facets could be seen with the correct lighting. With the new approach we requeue the multi-segment move ***exactly*** as before the pause, so there can be no inconsistency between layers caused by the pause/resume.
This fix also improves the G0/1 moves to some extent. While with G2/3 the inconsistency would show itself on the XYZ motion, with G0/1 it would only happen on the Z axis since that's the one that splits an otherwise straight move (for mesh bed leveling). In this case, the inconsistency would show itself as a variation of the Z height compared to the previously printed layers since the 30mm splits would happen on different XY coordinates on the resumed move compared to the previous layer. Just like with G2/3, this issue is now gone since we are requeuing the exact same segments on resume, so the 30mm splits happen at the exact same XY coordinates, so the Z position gets interpolated by MBL to the same Z coordinates as before the pause.

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -236,8 +236,6 @@ void FlushSerialRequestResend();
 void ClearToSend();
 void update_currents();
 
-void get_coordinates();
-void prepare_move();
 void kill(const char *full_screen_message = NULL, unsigned char id = 0);
 void finishAndDisableSteppers();
 
@@ -252,7 +250,6 @@ bool IsStopped();                           // Returns true if the print has bee
 //put an ASCII command at the begin of the current buffer, read from flash
 #define enquecommand_front_P(cmd) enquecommand_front(cmd, true)
 
-void prepare_arc_move(bool isclockwise);
 void clamp_to_software_endstops(float target[3]);
 void refresh_cmd_timeout(void);
 
@@ -288,6 +285,10 @@ extern int fanSpeed;
 extern uint8_t newFanSpeed;
 extern int8_t lcd_change_fil_state;
 extern float default_retraction;
+
+void get_coordinates();
+void prepare_move(uint16_t start_segment_idx = 0);
+void prepare_arc_move(bool isclockwise, uint16_t start_segment_idx = 0);
 
 #ifdef TMC2130
 void homeaxis(uint8_t axis, uint8_t cnt = 1, uint8_t* pstep = 0);
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -289,6 +289,7 @@ extern float default_retraction;
 void get_coordinates();
 void prepare_move(uint16_t start_segment_idx = 0);
 void prepare_arc_move(bool isclockwise, uint16_t start_segment_idx = 0);
+uint16_t restore_interrupted_gcode();
 
 #ifdef TMC2130
 void homeaxis(uint8_t axis, uint8_t cnt = 1, uint8_t* pstep = 0);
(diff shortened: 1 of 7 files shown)


leptun:
To aid testing this, `RESUME_DEBUG` can be defined at the beginning of ultralcd.cpp. This will enable a menu item in the main menu which pauses and resumes the print, but without moving away and stopping the heaters.

For testing the G2/3 handling, this file can be used: https://github.com/prusa3d/Prusa-Firmware/files/8621142/arcs.1.zip
For testing the G0/1 handling, just print a large first layer so that it has multi-segment moves.

DRracer (commented):
I checked the relevant changes and I haven't found anything obviously wrong with this approach.
So let's wait for #3552 getting merged and then we can merge this PR as well.

leptun:
Rebased onto MK3.

DRracer (approved):

PR #3569, opened 2022-08-24
Fix "Card removed" on startup

leptun:
Allow the SDCARDDETECT line to be raised by the pullups before sampling it for the first time. The chosen delay is arbitrary and worked on a problematic printer I had around.


DRracer (approved):
interesting :wink:

PR #3570, opened 2022-08-25
Report correct fan speeds in M155 during calibration

wavexx:
We didn't set fanSpeed during calibration, resulting in M155 not showing the current set fan speed correctly.


leptun (approved):
Looks good to me

PR #3571, opened 2022-08-26
Optimisation: `workDirDepth` can be one byte

gudnimg:
We set the max working directory depth to 6 (see `MAX_DIR_DEPTH`)

Changes save 1 byte of SRAM and 50 bytes of flash


DRracer (approved):
nice :+1:

PR #3573, opened 2022-08-26
Don't include M851 in build if Auto Bed Leveling is not enabled

gudnimg:
The variable `cs.zprobe_zoffset` is not used unless
auto bed leveling is enabled

This reduces flash usage by 270 bytes. No change in SRAM.


leptun (approved):
Good catch

PR #3574, opened 2022-08-26
More optimizations for 3.12 to fit

leptun:
Yet another PR for optimizations. This shall be merged soon and it is based on the most recent MK3 branch.

flash: -322
RAM: -7

### DIFF
--- a/Firmware/conv2str.cpp
+++ b/Firmware/conv2str.cpp
@@ -104,7 +104,7 @@ char *ftostr43(const float &x, uint8_t offset)
 //Float to string with 1.23 format
 char *ftostr12ns(const float &x)
 {
-  long xx = x * 100;
+  int xx = x * 100;
 
   xx = abs(xx);
   conv[0] = (xx / 100) % 10 + '0';
--- a/Firmware/mesh_bed_calibration.cpp
+++ b/Firmware/mesh_bed_calibration.cpp
@@ -418,9 +418,9 @@ BedSkewOffsetDetectionResultType calculate_machine_skew_and_offset_LS(
     for (uint8_t i = 0; i < npts; ++i) {
         float x = vec_x[0] * measured_pts[i * 2] + vec_y[0] * measured_pts[i * 2 + 1] + cntr[0];
         float y = vec_x[1] * measured_pts[i * 2] + vec_y[1] * measured_pts[i * 2 + 1] + cntr[1];
-        float errX = sqr(pgm_read_float(true_pts + i * 2) - x);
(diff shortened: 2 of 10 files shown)


Firmware/mesh_bed_calibration.cpp
```diff
@@ -418,9 +418,9 @@ BedSkewOffsetDetectionResultType calculate_machine_skew_and_offset_LS(
     for (uint8_t i = 0; i < npts; ++i) {
         float x = vec_x[0] * measured_pts[i * 2] + vec_y[0] * measured_pts[i * 2 + 1] + cntr[0];
         float y = vec_x[1] * measured_pts[i * 2] + vec_y[1] * measured_pts[i * 2 + 1] + cntr[1];
-        float errX = sqr(pgm_read_float(true_pts + i * 2) - x);
-        float errY = sqr(pgm_read_float(true_pts + i * 2 + 1) - y);
-        float err = sqrt(errX + errY);
+        float errX = pgm_read_float(true_pts + i * 2) - x;
+        float errY = pgm_read_float(true_pts + i * 2 + 1) - y;
+        float err = hypot(errX, errY);
```
DRracer, line 423: Introducing the `hypot` function will take some space if it hasn't been used anywhere else yet.
But I can see it is being used at multiple spots later on...

Firmware/conv2str.cpp
```diff
@@ -104,7 +104,7 @@ char *ftostr43(const float &x, uint8_t offset)
 //Float to string with 1.23 format
 char *ftostr12ns(const float &x)
 {
-  long xx = x * 100;
+  int xx = x * 100;
```
DRracer, line 107: are you sure the value can never exceed 32K?

DRracer (approved):
this PR looks ok so far

Firmware/mesh_bed_calibration.cpp
leptun: Well, it saved flash anyway so I wouldn't worry too much about it

Firmware/conv2str.cpp
leptun: It could, but then even the old implementation would be broken. The largest number this can render is 9.99 (unsigned). So when you make it *100, it will be a maximum of 999, which is fine for int16_t.

3d-gussner (approved):
Tested on MK404 
- MBL with `SUPPORT_VERBOSITY` enabled and `M45 V20` :heavy_check_mark:
- `G28 W` and  `G28` :heavy_check_mark:

PR #3575, opened 2022-08-28
Add ability to stop print *fast* remotely, just like via LCD.

arekm:
Add ability to stop print fast remotely, just like via LCD.
Currently when sitting next to printer it is possible
to stop print immediately by choosing "Stop print"
option on LCD.

When monitoring printer remotely and having a need to stop
print immediately there is no such possibility at all.
M603 which stops the print is pushed into planner and executed
way too late to be useful.

So far the only methods to stop print were 1) to disconnect
from serial and reconnect causing reset, 2) reset via command
;C2560_RES, 3) issue emergency M112 (which is processed early)

**This change adds M603 (stop the print) command to be processed
early, too.**

For this to be useful with OctoPrint (without hacks) EMERGENCY_PARSER
capability needs to be implemented which is also done.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -8133,6 +8133,7 @@ Sigma_Exit:
 
     /*!
     ### M603 - Stop print <a href="https://reprap.org/wiki/G-code#M603:_Stop_print">M603: Stop print</a>
+    It is processed much earlier as to bypass the cmdqueue.
     */
     case 603: {
         lcd_print_stop();
--- a/Firmware/cmdqueue.cpp
+++ b/Firmware/cmdqueue.cpp
@@ -476,10 +476,10 @@ void get_command()
             return;
         }
         // Handle KILL early, even when Stopped
-        if(strcmp(cmdbuffer+bufindw+CMDHDRSIZE, "M112") == 0)
+        if(strcmp_P(cmdbuffer+bufindw+CMDHDRSIZE, PSTR("M112")) == 0)
           kill(MSG_M112_KILL, 2);
         // Stop print
-        if (strcmp(cmdbuffer+bufindw+CMDHDRSIZE, "M603") == 0)
+        else if (strcmp_P(cmdbuffer+bufindw+CMDHDRSIZE, PSTR("M603")) == 0)
           lcd_print_stop();
         // Handle the USB timer
         if ((strchr_pointer = strchr(cmdbuffer+bufindw+CMDHDRSIZE, 'G')) != NULL) {
--- a/Firmware/cmdqueue.cpp
+++ b/Firmware/cmdqueue.cpp
@@ -478,6 +478,9 @@ void get_command()
         // Handle KILL early, even when Stopped
         if(strcmp(cmdbuffer+bufindw+CMDHDRSIZE, "M112") == 0)
           kill(MSG_M112_KILL, 2);
+        // Stop print
+        if (strcmp(cmdbuffer+bufindw+CMDHDRSIZE, "M603") == 0)
+          lcd_print_stop();
         // Handle the USB timer
         if ((strchr_pointer = strchr(cmdbuffer+bufindw+CMDHDRSIZE, 'G')) != NULL) {
             if (!IS_SD_PRINTING) {


Firmware/Marlin_main.cpp
```diff
@@ -4027,6 +4027,8 @@ static void extended_capabilities_report()
 #endif //FANCHECK and TACH_0 or TACH_1
     // AUTOREPORT_POSITION (M114)
     cap_line(PSTR("AUTOREPORT_POSITION"), ENABLED(AUTO_REPORT));
+    // M108, M112, M410 and Prusa specific M603.
+    cap_line(PSTR("EMERGENCY_PARSER"), 1);
```
wavexx: @leptun would this be sufficient to expose EMERGENCY_PARSER?

leptun: Hmm... I'm not sure. I don't think it's enough since afaik we don't constantly run get_command()

arekm: Expose where / for what?

```
2022-08-28 15:35:13,239 - Recv: FIRMWARE_NAME:Prusa-Firmware 3.11.1 based on Marlin FIRMWARE_URL:https://github.com/prusa3d/Prusa-Firmware PROTOCOL_VERSION:1.0 MACHINE_TYPE:Prusa i3 MK3S EXTRUDER_COUNT:1 UUID:00000000-0000-0000-0000-000000000000
2022-08-28 15:35:13,243 - Recv: Cap:AUTOREPORT_TEMP:1
2022-08-28 15:35:13,270 - Recv: Cap:AUTOREPORT_FANS:1
2022-08-28 15:35:13,362 - Recv: Cap:AUTOREPORT_POSITION:1
2022-08-28 15:35:13,451 - Recv: Cap:EMERGENCY_PARSER:1
2022-08-28 15:35:13,565 - Recv: Cap:EXTENDED_M20:1
2022-08-28 15:35:13,586 - Recv: Cap:PRUSA_MMU2:1
```

OctoPrint picks that up.

```
2022-08-28 00:59:42,383 - octoprint.util.comm - INFO - Printer reports firmware name "Prusa-Firmware 3.11.1 based on Marlin"
2022-08-28 00:59:42,383 - octoprint.util.comm - INFO - Detected Prusa firmware, enabling relevant features for issue free communication
2022-08-28 00:59:42,413 - octoprint.util.comm - INFO - Firmware states that it supports temperature autoreporting
2022-08-28 00:59:42,446 - octoprint.util.comm - INFO - Firmware states that it supports emergency GCODEs to be sent without waiting for an acknowledgement first
```
(last line)

leptun: At least I interpreted the question as "are these changes enough to make the emergency parser feature to work?". Because, we can report that we support it, but that doesn't mean the commands get processed early enough. I hope they do, but I think they don't

arekm: Ahh, ok.

From my observation there are some delays like when "G28 W" is being made for example. When actual object printing object it was reacting quite fast (but obviously have no idea about possible corner cases).

Also this means current MK3 branch M112 handling has the same issues.

Question is can it be made more responsive (without some huge rewrite) or are such delays acceptable for this capability.

arekm: Marlin did that via state machine in MarlinSerial.cpp store_char() and MarlinSerial.h checkRx():
https://github.com/MarlinFirmware/Marlin/commit/a129078927c8968022569175cec7fea35c920fd2

leptun: The delays are not acceptable in my opinion (that’s the whole point of this, absolutely no delays). In theory we _could_ implement the real emergency parser, but it might not work very well. This is because it would make the UART ISR longer, taking up valuable clock cycles. Also, it’s important that this feature uses as little flash as possible since we are seriously running out of.

arekm:
I've dropped emergency part from this pull request since emergency parser
was there only to make OctoPrint happy.

I would like to get first patch merged though.

OctoPrint will be persuaded to send M601 immediately via plugin.

Firmware/cmdqueue.cpp
```diff
@@ -478,6 +478,9 @@ void get_command()
         // Handle KILL early, even when Stopped
         if(strcmp(cmdbuffer+bufindw+CMDHDRSIZE, "M112") == 0)
           kill(MSG_M112_KILL, 2);
+        // Stop print
+        if (strcmp(cmdbuffer+bufindw+CMDHDRSIZE, "M603") == 0)
```
leptun: Please change this to strcmp_P and wrap the string with PSTR(). Do this also to the M112 above and maybe also change this to a `else if` since it might save a few bytes of flash

Firmware/Marlin_main.cpp
arekm: btw I wonder what is taking so much space in prusa firmware? For example on my other printer (DIY corexy, with arduino mega + ramps 1.4) current Marlin 2 with tons of features (close to prusa ones but there are differences) turned on takes:
```
RAM:   [=======   ]  69.1% (used 5658 bytes from 8192 bytes)
Flash: [======    ]  62.7% (used 159308 bytes from 253952 bytes)
```

(if these PlatformIO stats are accurate)

leptun: I think the UI is what kills most of the flash. The xyz calibration also takes a few percents. Translations are 10% of flash.
I think @wavexx knows more about this since he’s disabled significant parts of the firmware during testing before.

wavexx: Hard to say since not all parts can be easily disabled, but the whole ui/menu structure is dating back to marlin 1 and is very inefficient code-wise but would take some serious effort to improve around. xyzcal and selftest are relatively big. The current mmu code is scattered but it also takes up space, as well as multi-language support. Plus several small things that summed up I'm sure are eating space.. for example the la10->12 conversion doesn't take up much, but it's still extra stuff marlin2 doesn't need.

alexiri:
What's the status of this? I was looking for this functionality too. I'd like to have something like `M112`, but that doesn't shutdown the firmware completely and require a power cycle (and that doesn't have such an annoying alarm). I thought `M410` would be something like that, but it's not implemented for Prusa firmware.

3d-gussner:
@arekm I assigned this PR to myself and will discuss this with the development team.

PR #3576, opened 2022-08-28
Save target bed temperature, not current temperature

leptun:
Save the target bed temperature instead of current bed temperature when pausing a print


wavexx (approved):
Looks like set_temp_error() already did the right thing.
Approving

PR #3577, opened 2022-08-28
PSU_DELTA alternative

leptun:
This PR is an alternative proposal to the infamous PSU_DELTA code.

The idea behind PSU_DELTA was to automatically switch the Z motors to stealthchop when these motors were supposedly unused. This was because the PSU_DELTA has more noise on the 24V rail, causing any active stepper to create high pitched noise.

In practice, that had quite a few issues:
- not only the Z motor got switched to stealthchop when idle/"disabled", but also all the other steppers. As a result, the conservative motion profile got applied automatically to all steppers as soon as the Z went silent. Even more annoyingly, the only way to get the other steppers to the correct mode was to move the Z motor by any distance (effectively "enabling" the stepper, switching all the other ones to the correct mode as well).
- As you might have noticed in my point above, I put quotation marks around enabled and disabled. That's because the PSU_DELTA code would hijack the enable and disable macros in marlin to do the mode switching. This was a poor choice in my opinion since these sometimes get called from stupid places such as the babystepping in the temperature isr (if that is still there). @wavexx knows what I'm talking about (from the mystery crashing printer). The problem is that these functions access the SPI bus without regarding that the bus could be busy from the main context.
- The PSU_DELTA code is a mess. Trying to read it made my head hurt.

My approach:
Instead of manually toggling the mode (in stupid ways), why not just let the Z driver do that on it's own? This is what I attempt to do in this PR. In stealth mode, the Z motor is always in stealthchop mode. However, in normal mode the driver is configured so that if the velocity is above standstill, it switches to spreadcycle, otherwise stay in stealthchop.

flash: -202
RAM: -1

PFW-1455

### DIFF
--- a/Firmware/tmc2130.cpp
+++ b/Firmware/tmc2130.cpp
@@ -12,6 +12,7 @@
 
 #define TMC2130_GCONF_NORMAL 0x00000000 // spreadCycle
 #define TMC2130_GCONF_SGSENS 0x00000180 // spreadCycle with stallguard (stall activates DIAG0 and DIAG1 [open collector])
+#define TMC2130_GCONF_DYNAMIC_SGSENS 0x00000184 // stealthChop/spreadCycle (dynamic) with stallguard (stall activates DIAG0 and DIAG1 [open collector])
 #define TMC2130_GCONF_SILENT 0x00000004 // stealthChop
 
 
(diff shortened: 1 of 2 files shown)


wavexx:
Should we completely remove the PSU_DELTA code as well? We know it's buggy, and we know it can actually interrupt SPI communication, so I wouldn't even re-enable that.

leptun:
If my proposal will get approved, then yes. Until then I’ll just leave it there.

wavexx (approved):
Can't test this at the moment, but looks good

DRracer:
rebased onto MK3

gudnimg:
Moving this from 3.14 to 3.13 since this is already on the 3.13 branch

PR #3579, opened 2022-08-30
Allow to keep model checking enabled during autotuning

wavexx:
Allow running the model checking while doing model autotune. The C/R0 values should be stable enough to provide a further safety mechanism for printers of the same variant right from the factory.

This updates the default C/R0 estimates for MK3/MK3S variants, as well as a new parameter `M310 A F1` to force self-check during autotune.

With this PR, the following:

```
M502
M310 A F1
```

will autotune the model while still checking the hotend during the heating steps, aborting if the hotend is too far from the factory defaults. The warning threshold still generate beeps during autotune, although short beeps during autotune are considered to be expected in this scenario.

### DIFF
--- a/Firmware/temperature.cpp
+++ b/Firmware/temperature.cpp
@@ -2486,7 +2486,7 @@ void log_isr()
     if(!log_buf.enabled) return;
 
     uint32_t stamp = _millis();
-    uint8_t delta_ms = stamp - log_buf.entry.stamp - (TEMP_MGR_INTV * 1000);
+    uint8_t delta_ms = stamp - log_buf.entry.stamp - (uint32_t)(TEMP_MGR_INTV * 1000);
     log_buf.entry.stamp = stamp;
 
     ++log_buf.entry.counter;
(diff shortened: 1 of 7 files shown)


Firmware/temperature.cpp
```diff
@@ -2316,26 +2315,31 @@ static void check_temp_raw()
 #ifdef TEMP_MODEL
 namespace temp_model {
 
-void model_data::reset(uint8_t heater_pwm, uint8_t fan_pwm, float heater_temp, float ambient_temp)
+void model_data::reset(uint8_t heater_pwm _UNUSED, uint8_t fan_pwm _UNUSED,
```
DRracer, line 2318: Do we need to keep the parameters if they are unused now?

wavexx: Yes. I already checked that there's no flash cost if they are unused. I want to keep the code providing the initial values if we want to change this in the future so that the model can start a few milliseconds sooner.

DRracer: ok

DRracer (approved):

DRracer (approved):

DRracer (approved):

3d-gussner (approved):
Working great!
Tests with EINSY simulating broken thermistor catched `sim-broken-thermistor` every time. :1st_place_medal:

PR #3585, opened 2022-09-10
Fix lost changes during rebase

gudnimg:
See changes already merged into 3.12 here: https://github.com/prusa3d/Prusa-Firmware/pull/3551/files
This fixes an issue where the cmdbuffer can be corrupted


gudnimg:
Examples of the corruption fixed in this PR

image

image

DRracer (approved):
I remember choosing that strange `pStrBegin[*nLength] = '\0';` while rebasing, so yes - this was my fault while rebasing, sorry.
Thanks for fixing this.

PR #3586, opened 2022-09-10
PFW-1362 Implement SpoolJoin + FINDA filament runout

gudnimg:
The pull request rewrites SpoolJoin implementation (no drastic changes to old behaviour though) and now the firmware uses the FINDA sensor for filament runout detection when the MMU2S is used. Works pretty well in my testing so far.

* Commit ca67fa9cbedd6d2cfa80eb014d9f1d6f7e77f16c fixes an important issue where FINDA values was not being written to findaPressed variable.


Firmware/SpoolJoin.h
```diff
@@ -0,0 +1,59 @@
+/// @file
```
DRracer, line 1: nice, clean code and well documented :+1:

Firmware/SpoolJoin.cpp
```diff
@@ -0,0 +1,80 @@
...
+        updateSpoolJoinStatus(EEPROM::Disabled);
+    } else {
+        updateSpoolJoinStatus(currentStatus);
+    }
+
+    // Useful information to see during bootup
+    SERIAL_ECHOPGM("SpoolJoin is ");
+    if (isSpoolJoinEnabled())
+    {
+        SERIAL_ECHOLNPGM("enabled");
```
DRracer: :thinking: don't we have some already existing message (messages.h) which could be used instead of these "enabled/disabled" texts?

gudnimg: We have "Off" and "On" those can serve the same purpose.

Firmware/SpoolJoin.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+void SpoolJoin::updateSpoolJoinStatus(EEPROM newStatus)
+{
+    status = newStatus;
+    eeprom_write_byte((uint8_t*)EEPROM_AUTO_DEPLETE, (uint8_t)status);
+}
+
+void SpoolJoin::initSpoolJoinStatus()
+{
+    EEPROM currentStatus = (EEPROM)eeprom_read_byte((uint8_t*)EEPROM_AUTO_DEPLETE);
```
DRracer, line 25: another topic for potential optimization - many variables read from the eeprom are checked for unitialized values. We could try implementing a non-inline function:
`uint8_t eeprom_get_byte(uint8_t * address, uint8_t defaultValue)`

We could save some space doing the check for defaults at one spot and having always a valid value as a result. This function could be updating the eeprom cell with the default as well if necessary.

Firmware/SpoolJoin.cpp
```diff
@@ -0,0 +1,80 @@
...
+    {
+        SERIAL_ECHOLNPGM("enabled");
+    } else {
+        SERIAL_ECHOLNPGM("disabled");
+    }
+}
+
+void SpoolJoin::toggleSpoolJoin()
+{
+    if (eeprom_read_byte((uint8_t*)EEPROM_AUTO_DEPLETE) == (uint8_t)EEPROM::Disabled)
```
DRracer, line 46: potential optimization - eeprom_toggle()? Do we have similar implementation in the code base multiple times?

Firmware/SpoolJoin.cpp
```diff
@@ -0,0 +1,80 @@
...
+    {
+        eeprom_write_byte((uint8_t*)EEPROM_AUTO_DEPLETE, (uint8_t)EEPROM::Enabled);
+    } else {
+        eeprom_write_byte((uint8_t*)EEPROM_AUTO_DEPLETE, (uint8_t)EEPROM::Disabled);
+    }
+}
+
+uint8_t SpoolJoin::isSpoolJoinEnabled()
```
DRracer: Can't it be `bool`?

Firmware/SpoolJoin.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+void SpoolJoin::setSlot(uint8_t slot)
+{
+    currentMMUSlot = slot;
+}
+
+uint8_t SpoolJoin::nextSlot()
+{
+    SERIAL_ECHOPGM("SpoolJoin: Changing slot from ");
```
DRracer: let's optimize this message :wink: 
`SpoolJoin: x -> y`

Firmware/Tcodes.cpp
```diff
@@ -57,14 +57,14 @@ void TCodes(char *const strchr_pointer, uint8_t codeValue) {
         }
     } else {
         SChooseFromMenu selectedSlot;
-//        if (strchr_pointer[index] == '?')
-//            selectedSlot = TCodeChooseFromMenu();
-//        else {
+        if (strchr_pointer[index] == '?') {
+            selectedSlot = TCodeChooseFromMenu();
+        /*} else if (MMU2::mmu2.Enabled() && SpoolJoin::spooljoin.isSpoolJoinEnabled()) {
+            // TODO: What if the next slot has no filament?
```
DRracer, line 63: I think the original idea was to select next slot again of there was no filament, but we can skip that for now.

Firmware/SpoolJoin.cpp
gudnimg: Done. Saves 16 bytes. I used `_O()` on the strings since we don't care about translations in this case.

Firmware/SpoolJoin.cpp
gudnimg: Done, saved 20 bytes!

Firmware/mmu2_protocol_logic.cpp
```diff
@@ -316,6 +316,7 @@ StepStatus ProtocolLogic::CommandStep() {
         SendFINDAQuery();
         return Processing;
     case ScopeState::FINDAReqSent:
+        findaPressed = rsp.paramValue;
```
DRracer, line 319: :man_facepalming: I forgot about this one ... moreover - there are 2 idential pieces of code (in CommandStep and in IdleStep). May be we could make them one function if it saves some bytes.

Firmware/SpoolJoin.cpp
gudnimg: Done. You're right, I'm just used to using `uint8_t` for everything in C

Firmware/mmu2_protocol_logic.cpp
```diff
@@ -316,6 +316,7 @@ StepStatus ProtocolLogic::CommandStep() {
...
         SendReadRegister(4, ScopeState::StatisticsSent);
         scopeState = ScopeState::StatisticsSent;
```
DRracer: even this line can be removed, I think SendReadRegister does that already

DRracer (approved):
well done :+1: , just think about the optimization ideas, may be we can shave off some bytes.

Firmware/SpoolJoin.h
```diff
@@ -31,6 +31,7 @@ class SpoolJoin {
     static void toggleSpoolJoin();
 
     /// @brief Check if SpoolJoin is enabled
+    /// @returns 1 if enabled, 0 if disabled
```
DRracer: probably true/false now (not 1/0)

gudnimg: Done

Firmware/SpoolJoin.cpp
gudnimg: I'm not sure we do. Probably a good idea though.

gudnimg:
Rebased the PR with respect to `MK3_MMU2` branch

gudnimg:
If the MMU is disabled. The firmware will still trigger a FINDA runout (even though there is no FINDA sensor). I was kind of expecting this :)

EDIT: Fixed here [1524aae](https://github.com/prusa3d/Prusa-Firmware/pull/3586/commits/1524aae2bffa69f1e6e14ceb9d4bf6a890e7dccc)

Firmware/mmu2_protocol_logic.cpp
gudnimg: Done, this saved 2 bytes of flash. See 279aa8e

Firmware/mmu2.cpp
```diff
@@ -731,6 +734,9 @@ StepStatus MMU2::LogicStep() {
     StepStatus ss = logic.Step();
     switch (ss) {
     case Finished:
+        // At this point it is safe to trigger a runout and not interrupt the MMU protocol
```
DRracer, line 737: this is actually a good point :+1:

DRracer (approved):

PR #3587, opened 2022-09-10
Allow sending uint16_t values to registers

leptun:
Registers can also be 16 bit. Read the command arguments correctly.


DRracer (approved):
yes, this is correct :+1: , good catch

gudnimg (approved):

DRracer (approved):

PR #3588, opened 2022-09-11
Fix inconsistencies in MMU buttons

gudnimg:
- Unload Manually error did not send any buttons to the MMU (even though it does have an error screen, so clicking the button may seem like the MK3S does nothing or the MMU does nothing)
- Fsensor errors only have the Retry button. The Done button was removed several weeks ago.
- Filament Already Loaded error only has two buttons defined. I removed the one which does not appear on the error screen.

Changes save 2 bytes of flash.


gudnimg:
This should fix https://dev.prusa3d.com/browse/PFW-1356

I'm not sure whether we want to add the "Done" button in this PR since this PR contains only fixes. We should consider to re-test whether having only "Retry" button is sufficient and then decide if having an additional "Done" button makes sense?

gudnimg:
Closing in favor of https://github.com/prusa3d/Prusa-Firmware/pull/3601

PR #3589, opened 2022-09-11
Fix issue where physical MMU buttons do not dismiss error screen

gudnimg:
Steps to reproduce issue:

1. Boot up printer
2. Send T0 over serial
3. Wait for MMU to raise an error screen
4. Try clicking any of the physical MMU buttons
5. Observe that the error screen is not dismissed.

---
Solution here is to write down the error source into SRAM as soon as the error is reported (and only once). Then there can not be any confusion whether the error originated from the Printer or the MMU.

Change in memory:
Flash: -2 bytes
SRAM: +1 byte


gudnimg:
Marking this as a draft for now, because we need to test more than just the problem I am trying to fix. To make sure this doesn't break anything else.

gudnimg:
Closing for now, need to create a new PR later

PR #3592, opened 2022-09-13
Extract PRINTER_ACTIVE into a noinline function

DRracer:
Having the original PRINTER_ACTIVE macro copied at multiple spots doesn't make sense. Refactoring it into a non-inline function saved ~400 bytes of code. It should be safe in terms of performance, all occurrences are at non-time critical spots.


gudnimg (approved):
Looks good 💪

DRracer:
closing as this change has been already applied by rebasing onto MK3 (3.12)

PR #3593, opened 2022-09-13
Extract PRINTER_ACTIVE into a noinline function

DRracer:
Having the original PRINTER_ACTIVE macro copied at multiple spots doesn't make sense. Refactoring it into a non-inline function saved ~400 bytes of code. It should be safe in terms of performance, all occurrences are at non-time critical spots.

Cherry picked from #3592

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -355,7 +355,6 @@ extern uint16_t gcode_in_progress;
 extern LongTimer safetyTimer;
 
 #define PRINT_PERCENT_DONE_INIT 0xff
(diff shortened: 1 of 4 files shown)


Firmware/Marlin.h
```diff
@@ -355,7 +355,6 @@ extern uint16_t gcode_in_progress;
 extern LongTimer safetyTimer;
 
 #define PRINT_PERCENT_DONE_INIT 0xff
-#define PRINTER_ACTIVE (IS_SD_PRINTING || usb_timer.running() || isPrintPaused || (custom_message_type == CustomMsg::TempCal) || saved_printing || (lcd_commands_type == LcdCommands::Layer1Cal) || mmu_print_saved || homing_flag || mesh_bed_leveling_flag)
```
leptun, line 358: I propose you make this “#define PRINTER_ACTIVE printer_active()” so that you don’t have to change all occurances of this macro in the code.

DRracer: no, I intentionally want to get rid of these non-debuggable macros

leptun (approved):
Looks good to me

gudnimg (approved):

PR #3600, opened 2022-09-17
Fix initialization of LCD status line message

### LINKED ISSUE #3581: [BUG] 3.12 - During G28 with 7x7 mesh, Nozzle icon disappears while calibration points left are 10 or more
**Printer type** - All
**Printer firmware version** -  3.12 (unreleased)

**MMU upgrade** - N/A - MMU can be left disconnected

**SD card or USB/Octoprint**
 N/A - SD card can be left disconnected

**Describe the bug**

While the calibration points left are double digits (larger than 9) the nozzle icon is overwritten with a space.

This for-loop seems to be responsible. It is called during mesh bed leveling and at various other scenarios such as at printer's boot-up. https://github.com/prusa3d/Prusa-Firmware/blob/66b97683bbb838661e5e8a2a3deda7cc7e8eb9d8/Firmware/ultralcd.cpp#L674

I cannot explain why this wasn't an issue before. Looking at the Git history, the issue first starts here in my fork: https://github.com/gudnimg/Prusa-Firmware/commit/b64377b9de17103ad970369fed8e967d872602d2 but this does not make sense since even if I take these changes out, the issue is still there. We likely need to look more deeply into this.

image

  
**To Reproduce**
Make sure the Mesh Bed Level settings are set at 7x7 mesh. (Problem does not appear with 3x3 mesh)

1. Open Serial stream
2. Send G28 command to the printer
  
**Expected behavior**
Nozzle icon should be present on the LCD at all times.
  
**G-code**
N/A

**Video**
N/A

gudnimg:
Fixes #3581 

Change in memory:
Flash: -48 bytes
SRAM: -1 byte

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -7518,6 +7518,9 @@ void ultralcd_init()
   lcd_oldcardstatus = IS_SD_INSERTED;
 #endif//(SDCARDDETECT > 0)
   lcd_encoder_diff = 0;
+
+  // Initialise status line
+  lcd_setstatuspgm(MSG_WELCOME);
 }
 
 void lcd_ignore_click(bool b)
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -94,7 +94,7 @@ static float manual_feedrate[] = MANUAL_FEEDRATE;
 /* LCD message status */
 static LongTimer lcd_status_message_timeout;
 static uint8_t lcd_status_message_level;
-static char lcd_status_message[LCD_WIDTH + 1] = WELCOME_MSG;
+static char lcd_status_message[LCD_WIDTH + 1];
 
 /* !Configuration settings */
 
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -669,11 +669,6 @@ void lcdui_print_status_line(void)
             break;
         }
     }
-
-    // Fill the rest of line to have nice and clean output
-    for(uint8_t fillspace = 0; fillspace < LCD_WIDTH; fillspace++)
-        if ((lcd_status_message[fillspace] <= 31 ))
-            lcd_print(' ');
 }
 
 //! @brief Show Status Screen


leptun:
@gudnimg Do we even need this snippet of code anymore now that we always set the status line correctly? I think this code was used only for this welcome message and for old code which didn't set the status line with the specialized function.
```
   // Fill the rest of line to have nice and clean output
    for (uint8_t fillspace = 0; fillspace < LCD_WIDTH; fillspace++)
        if ((lcd_status_message[fillspace] <= 31))
            lcd_print(' ');
```

gudnimg:
> @gudnimg Do we even need this snippet of code anymore now that we always set the status line correctly? I think this code was used only for this welcome message and for old code which didn't set the status line with the specialized function.
> 
> ```
>    // Fill the rest of line to have nice and clean output
>     for (uint8_t fillspace = 0; fillspace < LCD_WIDTH; fillspace++)
>         if ((lcd_status_message[fillspace] <= 31))
>             lcd_print(' ');
> ```

I think you're right. Now `lcd_status_message` is only set in `lcd_updatestatus()` which always calls `lcd_finishstatus()`.

Firmware/ultralcd.cpp
```diff
@@ -94,7 +94,7 @@ static float manual_feedrate[] = MANUAL_FEEDRATE;
 /* LCD message status */
 static LongTimer lcd_status_message_timeout;
 static uint8_t lcd_status_message_level;
-static char lcd_status_message[LCD_WIDTH + 1] = WELCOME_MSG;
+static char lcd_status_message[LCD_WIDTH + 1];
```
leptun, line 97: I don't really understand how this saved 1B of RAM

gudnimg: Perhaps a clue:
If we initialize the array with all elements as 1: it takes 1 extra byte of SRAM
`static char lcd_status_message[LCD_WIDTH + 1] = {1};`

If we initialize the array with all elements as 0: it takes no extra SRAM. 
`static char lcd_status_message[LCD_WIDTH + 1] = {0};`

DRracer (approved):

Firmware/ultralcd.cpp
leptun: I think I know why this saved 1B of ram. It's because when you initialize the array with {1} it goes into the .data section, whch is in turn stored in flash. When you initialize the variable with {0} or don't initialize at all (in a global context), the variable gets placed in the .bss section which is initialized to 0x00 before jumping to main.
I see that the .data section is aligned to 2B while .bss is not. I think that's where a byte gets lost.

gudnimg: That explanation sounds reasonable 👍

PR #3601, opened 2022-09-17
Fix Unload Manually MMU error being impossible to resolve

gudnimg:
- Unload Manually error did not send any buttons to the MMU (even though it does have an error screen, so clicking the button may seem like the MK3S does nothing or the MMU does nothing)
- Fsensor errors only have the Retry button. The Done button was removed several weeks ago.
- Filament Already Loaded error only has two buttons defined. I removed the one which does not appear on the error screen.

Changes save 2 bytes of flash.

PFW-1356


DRracer (approved):
:+1:

PR #3602, opened 2022-09-17
PFW-1369 Use same ramming sequence for MMU unload as default settings in Slicer

gudnimg:
The current ramming sequence is known to have some stringing issues. Let's see if we use the same sequence as the default PrusaSlicer settings, does the situation improve?

Change in memory:
Flash: +56 bytes
SRAM: 0 bytes


DRracer (approved):
let's see if it improves the unload

PR #3603, opened 2022-09-18
Optimisation: Remove duplicate code in `cmdqueue_could_enqueue_back()`

gudnimg:
PR removes duplicated code in `cmdqueue_could_enqueue_back()`. The code under `if (serial_count > 0)` and its respective `else` is exactly the same. Removing the if statement saves both flash memory and about ~8 clock cycles.

Code is 28 lines shorter :)

Change in memory:
Flash: -64 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/cmdqueue.cpp
+++ b/Firmware/cmdqueue.cpp
@@ -1,3 +1,4 @@
+#include <util/atomic.h>
 #include "cmdqueue.h"
 #include "cardreader.h"
 #include "ultralcd.h"
@@ -155,7 +156,7 @@ static bool cmdqueue_could_enqueue_front(size_t len_asked)
 // len_asked does not contain the zero terminator size.
 // This function may update bufindw, therefore for the power panic to work, this function must be called
 // with the interrupts disabled!
-static bool cmdqueue_could_enqueue_back(size_t len_asked, bool atomic_update = false)
+static bool cmdqueue_could_enqueue_back(size_t len_asked)
 {
     // MAX_CMD_SIZE has to accommodate the zero terminator.
     if (len_asked >= MAX_CMD_SIZE)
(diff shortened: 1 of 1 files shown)


gudnimg:
I noticed also if `atomic_update` can be `true` by default that alone would save an additional 22 bytes of flash. I did not include that change here though.

Firmware/cmdqueue.cpp
```diff
@@ -165,61 +165,33 @@ static bool cmdqueue_could_enqueue_back(size_t len_asked, bool atomic_update = f
...
+        (endw <= sizeof(cmdbuffer) && CMDBUFFER_RESERVE_FRONT <= bufindr))
+        return true;
+    // Could one fit both to the start?
+    if (len_asked + (1 + CMDHDRSIZE) + CMDBUFFER_RESERVE_FRONT <= bufindr) {
+        // Mark the rest of the buffer as used.
+        memset(cmdbuffer+bufindw, 0, sizeof(cmdbuffer)-bufindw);
+        // and point to the start.
+        // Be careful! The bufindw needs to be changed atomically for the power panic & filament panic to work.
+        if (atomic_update)
+            cli();
```
DRracer: Can't we use an ATOMIC_BLOCK in all cases? Do we really need to check for the `atomic_update` variable (twice!)?

DRracer (approved):

Firmware/cmdqueue.cpp
gudnimg: Should we consider removing `atomic_update` parameter? @wavexx what do you think?

The parameter is always `true` except for in `enqueue_command()` but its not set explicitly.

leptun: I don’t see why this couldn’t be always atomic. The clock cycles gained by the lack of cli/sei is lost by the if statements on the argument. I’d just make it always atomic and get rid of the argument. I’d also switch to an atomic block.
I think doing like this will save ~20B of flash and some clock cycles

wavexx: Yes, let's ditch this and change it to an atomic block (this, and the second case following). Fully in favor of removing the argument as well.

gudnimg:
I'm surprised just removing the parameter increases code size 🤔 but keeping it and setting the default value to true, saves flash.

gudnimg:
Turns out after I removed the parameter, the compiler inlines the function (or so I think). Now when I forcefully disable inlining the function I get a similar flash size as if I set the default value of `atomic_update` to `true`.

DRracer (approved):

PR #3605, opened 2022-09-18
MMU: Improve purge line during first layer calibration

gudnimg:
Sync the purge line gcode with PrusaSlicer output when using Tcodes.

These 2 lines:
```
G1 X55.0 E32.0 F1073.0
G1 X5.0 E32.0 F1800.0
```

Should be like this
```
G1 X55.0 E29.0 F1073.0
G1 X5.0 E29.0 F1800.0
```

Change in memory:
Flash: 0 bytes
SRAM: 0 bytes


gudnimg:
To fix this for multimaterial prints, the slicer settings need to be fixed as well.

**Before:**
----

image

image


**After:**
-----

image

PR #3608, opened 2022-09-19
Update FW version for 3.12 branch

wavexx:


Firmware/Configuration.h
```diff
@@ -17,8 +17,8 @@ extern PGM_P sPrinterName;
 
 // Firmware version
 #define FW_MAJOR 3
-#define FW_MINOR 11
-#define FW_REVISION 1
+#define FW_MINOR 12
+#define FW_REVISION 0
 //#define FW_FLAVOR RC      //uncomment if DEBUG, DEVEL, ALPHA, BETA or RC
```
leptun, line 22: Should this be set to BETA?

wavexx: IMHO there's no point to set anything besides major/minor in the sources, which is helpful to track branches.

The actual release comes from the tag, not from the sources. `git describe` will list the closest tag once done, which is why I'm pushing for it instead of fiddling with defines which mean very little when building by hand.

leptun (approved):

DRracer (approved):

PR #3609, opened 2022-09-19
Update current FW version

wavexx:
Update the FW_MINOR/REVISION version for the next upcoming release


leptun:
Why is this based on the MK3? What is the point of incrementing to 3.12.0 if we increment again to 3.13.0 right after?

DRracer:
@leptun the idea is correct, branch MK3 is heading towards 3.13

DRracer (approved):

PR #3615, opened 2022-09-21
Do not reset line on serial commands without N

wavexx:
Fix regression introduced in fc10ca31466e1c58bcb5b9b90976a489239a6fde.

Accept incoming serial commands without line numbers (assumed to be injected by the host), but do not reset the last line count when doing so.


leptun (approved):
This looks fine to me. Just wondering if we should worry about line number overflow. Probably not

wavexx:
I only found a couple of references that require line numbers to be positive...

DRracer (approved):
looks ok to me

PR #3618, opened 2022-09-21
Do not hang during startup

wavexx:
In ultralcd_init() do not use lcd_setstatuspgm() to initialize the welcome message.

The internal call to lcd_finishstatus() requires the serial to be already available. Split the function into lcd_padstatus() to pre-pad the string and save some space.


DRracer (commented):
nice :+1: ... and its even shorter :tada:

DRracer (approved):

3d-gussner (approved):
Tested on MK404 and it fixes the issue :medal_sports:

PR #3620, opened 2022-09-22
Improve warning behavior during thermal anomaly

wavexx:
The current code forces any warning to return the user to the status screen in order to show the message.

Thermal anomaly warnings can repeat at very short intervals, making menu navigation (to pause/tune the print) impossible.

We now check if the message to be displayed is the same and only force a kickback for new messages.

This partially reverts https://github.com/prusa3d/Prusa-Firmware/pull/3600 since we need the string to be null terminated for ease of comparison.

We pad the status line at display time instead using the new lcd_print_pad() function which achieves the same effect.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -242,9 +242,6 @@ static void lcd_cutter_enabled();
 #endif
 static void lcd_babystep_z();
 
-//! Beware: has side effects - forces lcd_draw_update to 2, which means clear the display
-void lcd_finishstatus();
-
 static void lcd_sdcard_menu();
 static void lcd_sheet_menu();
 
@@ -7422,7 +7419,6 @@ static bool check_file(const char* filename) {
 	card.printingHasFinished();
 
 	lcd_setstatuspgm(MSG_WELCOME);
-	lcd_finishstatus();
 	return result;
 }
 
@@ -7524,11 +7520,6 @@ void lcd_ignore_click(bool b)
   wait_for_unclick = false;
 }
 
-void lcd_finishstatus() {
-  SERIAL_PROTOCOLLNRPGM(MSG_LCD_STATUS_CHANGED);
-  lcd_draw_update = 2;
-}
-
 static bool lcd_message_check(uint8_t priority)
 {
     // regular priority check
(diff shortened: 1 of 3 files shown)


wavexx:
@gudnimg sadly this eats up most of the savings of the previous PR. Do you see any way to optimize the code size?

gudnimg (commented):
I wonder if `lcd_print_pad()` is eating more flash or if it is the `strcmp` functions. 

Any chance the compiler is inlining the functions? I can’t test this as I don’t have access to my computer until Monday.

Firmware/ultralcd.cpp
```diff
@@ -7537,7 +7526,6 @@ void lcd_ignore_click(bool b)
 
 void lcd_finishstatus() {
   SERIAL_PROTOCOLLNRPGM(MSG_LCD_STATUS_CHANGED);
-  lcd_padstatus();
```
gudnimg, line 7540: Doesn’t this make the `lcd_finishstatus()` function a bit useless? It used to pad the status line if the message was not 20 characters. 

I think its better to do it here than in `lcd_ui_print_status_line()`, because then we aren’t padding the message multiple times with each print on the LCD. Before we only needed to pad the mesage once when `lcd_status_message` was set.

wavexx: Indeed. I moved the only relevant part inside `lcd_updatestatus()` itself.

The only difference to before is that in `check_file()` we were forcing a full display clear, but by testing this it doesn't look like it's needed at all (@leptun was this really necessary?). I get a clean screen after the file check, even if I have the additional warning messages due to incompatible gcode.

The issue is that we need to compare if the message we're going to display is newer or not and compare it to the current. The new message is not padded, so we would need a custom comparison function to ignore (but check) the extra padding or extra memory to pre-pad the new string and compare it. I didn't test this, but I suspect neither would save space.

I don't think lcd_print_pad is much slower than before (it's not a critical path anyway). Using lcd_print_pad() also fixes two other cases (PID autotune) that were displaying a custom status message without padding so could show something in the background.

leptun: I don't remember why that was needed. If it works now, then it can be removed. It might have been a relic of the past that I just kept anyway

wavexx: @3d-gussner when you test this, also check if after the "checking file" step the screen is fine. It looks like it always work correctly to me.

3d-gussner (approved):
Tested on real hardware
To simulate it I set the warning value from 1.20 to 0.10 with `M310 W0.10`
As soon the Thermal anomaly is reported on screen and printer beeps constantly I was able to access the menu and go to Preheat -> Cooldown.
It keeps beeping for a while.

gudnimg (approved):
Approving as I don’t see any issues. 

We need to remember to apply this to the MMU branch as well because I added a new call to `lcd_print()` to print the progress codes.

PR #3627, opened 2022-09-26
Add Menu `Calibration -> Temp. model cal`

3d-gussner:
Runs `M310 A F0`
Enables with `M310 S1`
Saves settings `M500`

Tested on MK404 with Octoprint serial logging.

ToDos:

- [ ] cherry-pick to MK3_3.12
  - [ ] Update pot and po files
- [ ] Update pot and po files in MK3 branch

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -3712,6 +3712,16 @@ void lcd_mesh_calibration_z()
   lcd_return_to_status();
 }
 
+#ifdef TEMP_MODEL
+void lcd_temp_model_cal()
+{
+  enquecommand_P(PSTR("M310 A F0"));
+  enquecommand_P(PSTR("M310 S1"));
+  enquecommand_P(PSTR("M500"));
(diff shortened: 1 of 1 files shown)


Firmware/ultralcd.cpp
```diff
@@ -3712,6 +3712,16 @@ void lcd_mesh_calibration_z()
   lcd_return_to_status();
 }
 
+#ifdef TEMP_MODEL
+void lcd_temp_model_cal()
+{
+  enquecommand_P(PSTR("M310 A F0"));
+  enquecommand_P(PSTR("M310 S1"));
```
wavexx, line 3719: In case calibration fails, M310 S1 will refuse to enable the model (which is fine), however M500 will still trigger an eeprom save. If the previous calibration succeeded, this will result in the model being permanently disabled.

3d-gussner: Okay then I need another approach with lcd_command_step or so.

PR #3638, opened 2022-09-30
TM: Consume LCD click after calibration

wavexx:
Consume any pending LCD click after the temperature model calibration. Since we're already in the status menu, the usual consume action is not done automatically here.

### DIFF
--- a/Firmware/lcd.h
+++ b/Firmware/lcd.h
@@ -204,11 +204,12 @@ private:
 extern void lcd_set_custom_characters(void);
 extern void lcd_set_custom_characters_nextpage(void);
 
-//! @brief Consume click event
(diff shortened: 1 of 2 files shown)


leptun (approved):
Looks correct to me

wavexx:
@3d-gussner can you do a test on this before I merge it also on 3.12?

3d-gussner (approved):
On the MK404 I had not really results if this PR fixed it or not.
But on real hardware I can confirm that the long button press just after starting `M310 A F0` is not buffered.
Tested of course the previous version and the long button press was buffered. So after the TM autotune finished I saw the Z +0.2 screen.

PR #3639, opened 2022-09-30
Cleanup warnings

wavexx:
Minor changes to keep the build clean.

### DIFF
--- a/Firmware/Filament_sensor.cpp
+++ b/Firmware/Filament_sensor.cpp
@@ -278,7 +278,9 @@ void IR_sensor_analog::voltUpdate(uint16_t raw) { // to be called from the ADC I
 }
 
 uint16_t IR_sensor_analog::getVoltRaw() {
-    ATOMIC_BLOCK(ATOMIC_RESTORESTATE) { return voltRaw; }
+    uint16_t ret;
+    ATOMIC_BLOCK(ATOMIC_RESTORESTATE) { ret = voltRaw; }
+    return ret;
 }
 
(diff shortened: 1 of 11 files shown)


Firmware/Filament_sensor.cpp
```diff
@@ -278,7 +278,9 @@ void IR_sensor_analog::voltUpdate(uint16_t raw) { // to be called from the ADC I
 }
 
 uint16_t IR_sensor_analog::getVoltRaw() {
-    ATOMIC_BLOCK(ATOMIC_RESTORESTATE) { return voltRaw; }
+    uint16_t ret;
+    ATOMIC_BLOCK(ATOMIC_RESTORESTATE) { ret = voltRaw; }
```
leptun, line 282: I need David's approval on these. I think this is how I did them initially.

wavexx: I checked the generated asm is identical. The alternative would be to add a dummy "return", which is a lot uglier IMHO.

leptun: If the generated code is identical, then I have 0 problems with this change. Still, I'd let David approve this part since he changed it after me

leptun (approved):
Very nice. No more warnings. I checked that the MMU still communicates and it does show error messages

Firmware/Filament_sensor.cpp
DRracer: I agree. Too bad the compiler is stupid and doesn't understand the intent in this case

DRracer (approved):

PR #3644, opened 2022-10-01
Optimisation: Don't attempt to set LCD to 8-bit mode if its not supported

gudnimg:
`lcd_displayfunction` is used to set the LCD to 8-bit or 4-bit mode. See: `lcd_command(LCD_FUNCTIONSET | lcd_displayfunction)`
4-bit mode is set by default with zero-initialization.

Change in memory:
Flash: -64 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/lcd.cpp
+++ b/Firmware/lcd.cpp
@@ -64,15 +64,13 @@
 
 // bitmasks for flag argument settings
 #define LCD_RS_FLAG 0x01
-#define LCD_HALF_FLAG 0x02
 
 FILE _lcdout; // = {0}; Global variable is always zero initialized, no need to explicitly state that.
 
-uint8_t lcd_displayfunction = 0;
-uint8_t lcd_displaycontrol = 0;
-uint8_t lcd_displaymode = 0;
-
-uint8_t lcd_currline;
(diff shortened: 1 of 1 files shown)


leptun:
NO! We need to set the LCD to 8 bit mode. That is the init sequence and it is very important for nibble desync recovery :)

gudnimg:
I didn't see any configuration which uses the pins `LCD_PINS_D0`, `LCD_PINS_D1`, `LCD_PINS_D2`, `LCD_PINS_D3`. I didn't see any obvious issues on my end.

leptun:
Yes. We don't use the D0-D3 pins in any of the config (it's from an old config, maybe we can remove the ifdeffed out code). The problem you removed the actual init sequence of the LCD. It must switch to 8 bit mode and then back to 4 bit every time the lcd initializes. Otherwise you can't recover from an out of sync nibble. And you have to do this regardless if you use 4 bits or 8 bits since you can never assume what is the current bus mode while initializing

leptun:
This is the exact init sequence that we must follow:
image

gudnimg:
I was using the wrong datasheet. Read the one you referenced above and am convinced the current implementation is correct. 👍

PR #3646, opened 2022-10-02
PFW-1403 Fix issue where physical MMU buttons do not dismiss error screen

gudnimg:
When the physical MMU buttons are used instead of the LCD buttons to resolve error screen, the screen will not be dismissed. Even if the error itself is resolved by pressing the button, the error screen will stay indefinitely on the LCD.

**Steps to reproduce issue:**

1. Boot up printer
2. Send T0 over serial
3. Wait for MMU to raise an error screen
4. Try clicking any of the physical MMU buttons
5. Observe that the error screen is not dismissed.

----
Solution here is to write down the error source into SRAM as soon as the error is reported (and only once). Then there can not be any confusion whether the error originated from the Printer or the MMU.

### DIFF
--- a/Firmware/mmu2.cpp
+++ b/Firmware/mmu2.cpp
@@ -216,7 +216,7 @@ void MMU2::mmu_loop() {
     if (is_mmu_error_monitor_active){
         // Call this every iteration to keep the knob rotation responsive
         // This includes when mmu_loop is called within manage_response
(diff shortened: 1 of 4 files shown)


ursius2020:
Like to mention, issue since FW 3.11.0

gudnimg:
> Like to mention, issue since FW 3.11.0

This fix is only relevant to the new, rewritten MMU code. It hasn’t been released yet.

The MMU code included in FW 3.11.0 is bundled with MMU FW 1.0.6. Both will be replaced in FW 3.13.0.

gudnimg:
Rebased branch to sync with MK3 branch

DRracer (approved):
looks correct to me :+1:

PR #3648, opened 2022-10-02
cleanup: Expand tabs to fixup suspicious indentation warnings

wavexx:


vintagepc (approved):

leptun (approved):
Looks good. We should really consider hard reformatting the entire code now that the big PRs are merged. Maybe after 3.12.0 final gets released

wavexx:
The interesting thing is that I thought I fixed all of those already for gcc 7.3.0, but those popped up as "new" when using cmake.
:rainbow: magic :rainbow:

PR #3652, opened 2022-10-03
cmake build improvements v2

wavexx:
Further progress on the cmake build infrastructure

- Reformat CMakeList.txt through cmake-format due to mixed formatting
- Define, but not build single-language targets by default (these are for debugging only). The target ALL_ENGLISH and ALL_FIRMWARE still drag those in.
- Fix multi-language builds: correctly copy/patch the base binary as produced by `lang-build.py`.
- Uniform binary structure in dual-language builds.
- Fix rules not to rebuild binaries if nothing changed

The multi-language builds are still slightly larger than expected, but this will be investigated later.

### DIFF
--- a/.cmake-format.py
+++ b/.cmake-format.py
@@ -1,0 +1,21 @@
+# If a statement is wrapped to more than one line, than dangle the closing
+# parenthesis on it's own line.
+dangle_parens = True
+dangle_align = 'child'
+
+# If true, the parsers may infer whether or not an argument list is sortable
+# (without annotation).
+autosort = True
+
+# How wide to allow formatted cmake files
+line_width = 100
+
+additional_commands = {
+    "target_sources": {
+        "kwargs": {
+            "PUBLIC": "*",
+            "PRIVATE": "*",
+            "INTERFACE": "*",
+        }
+    },
+}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -323,7 +323,6 @@ function(add_base_binary variant_name)
     ${variant_name}
     PRIVATE ${PRUSA_BOARDS_DIR}/cores/prusa_einsy_rambo/
             ${PRUSA_BOARDS_DIR}/variants/prusa_einsy_rambo/
-            ${VARIANT_CFG_DIR} # Include the header for this variant.
             ${CMAKE_SOURCE_DIR}/Firmware
     )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -329,7 +329,7 @@ function(add_base_binary variant_name)
 
   target_link_libraries(${variant_name} avr_core)
 
-  # # configure linker script
+  # configure linker script
   set(LINKER_SCRIPT ${PRUSA_BOARDS_DIR}/ldscripts/avr6.xn)
   target_link_options(${variant_name} PUBLIC -Wl,-T,${LINKER_SCRIPT})
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -358,24 +358,14 @@ function(add_base_binary variant_name)
     ${variant_name} PUBLIC -Wl,-Map=${CMAKE_CURRENT_BINARY_DIR}/${variant_name}.map
     )
 
-  target_compile_definitions(${variant_name} PRIVATE CMAKE_LANG_CONTROL)
+  target_compile_definitions(${variant_name} PRIVATE CMAKE_CONTROL)
 endfunction()
 
 function(fw_add_variant variant_name)
   # Set FW_SOURCES to be reproducible in this variant as it's set in a separate project
   set_reproducible_sources(FW_SOURCES "Firmware/")
 
-  # Create the Configuration_Prusa.h for this variant so it can be #included.
-  set(VARIANT_CFG_DIR ${CMAKE_CURRENT_BINARY_DIR}/include)
-  set(VARIANT_CFG_DST ${VARIANT_CFG_DIR}/Configuration_prusa.h)
-  set(VARIANT_CFG_SRC ${CMAKE_SOURCE_DIR}/Firmware/variants/${variant_name}.h)
-  add_custom_command(
-    OUTPUT ${VARIANT_CFG_DST}
-    COMMAND ${CMAKE_COMMAND} -E copy ${VARIANT_CFG_SRC} ${VARIANT_CFG_DST}
-    COMMENT "Generating Configuration_prusa.h for ${variant_name}"
-    BYPRODUCTS ${VARIANT_CFG_DIR}
-    DEPENDS ${VARIANT_CFG_SRC}
-    )
+  set(variant_header "variants/${variant_name}.h")
   string(REPLACE "1_75mm_" "" variant_name "${variant_name}")
   string(REPLACE "-E3Dv6full" "" variant_name "${variant_name}")
 
@@ -384,7 +374,7 @@ function(fw_add_variant variant_name)
   set(FW_HEX ${CMAKE_BINARY_DIR}/${FN_PREFIX}-${FW_EN}.hex)
 
   add_base_binary(${FW_EN})
-  target_compile_definitions(${FW_EN} PUBLIC LANG_MODE=0)
+  target_compile_definitions(${FW_EN} PUBLIC LANG_MODE=0 FW_VARIANT="${variant_header}")
   add_custom_command(
     TARGET ${FW_EN}
     POST_BUILD
@@ -399,7 +389,7 @@ function(fw_add_variant variant_name)
   set(FW_LANG_BASE "${variant_name}_Multilang_base")
   set(FW_LANG_PATCH "${variant_name}_Multilang_patch")
   add_base_binary(${FW_LANG_BASE})
-  target_compile_definitions(${FW_LANG_BASE} PUBLIC LANG_MODE=1)
+  target_compile_definitions(${FW_LANG_BASE} PUBLIC LANG_MODE=1 FW_VARIANT="${variant_header}")
 
   # Construct language map
   set(LANG_TMP_DIR lang)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -459,34 +459,35 @@ function(fw_add_variant variant_name)
       BYPRODUCTS ${LANG_HEX}
       DEPENDS ${FW_LANG_PATCH}.hex ${LANG_CATHEX}
       COMMENT "Generating final ${FW_LANG_FINAL}.hex")
+
     add_custom_target(${FW_LANG_FINAL} DEPENDS ${FW_LANG_FINAL}.hex)
     add_dependencies(ALL_MULTILANG ${FW_LANG_FINAL})
   else()
     set(ALL_VARIANT_HEXES "")
     # Non-xflash, e.g. MK2.5
     foreach(LANG IN LISTS SELECTED_LANGUAGES)
-      set(LANG_HEX_FN ${variant_name}-en_${LANG})
-      set(LANG_HEX ${CMAKE_BINARY_DIR}/${FN_PREFIX}-${LANG_HEX_FN}.hex)
+      set(FW_LANG_FINAL ${variant_name}-en_${LANG})
+      set(LANG_HEX ${CMAKE_BINARY_DIR}/${FN_PREFIX}-${FW_LANG_FINAL}.hex)
       set(LANG_BIN ${LANG_TMP_DIR}/${variant_name}_${LANG}.bin)
-      set(LANG_FWBIN_TMP ${LANG_TMP_DIR}/${variant_name}-en_${LANG}.bin)
 
       # Intermediate 2-lang bin
       add_custom_command(
-        OUTPUT ${LANG_FWBIN_TMP}
-        COMMAND ${CMAKE_COMMAND} -E copy ${LANG_FWBIN} ${LANG_FWBIN_TMP}
-        COMMAND ${CMAKE_SOURCE_DIR}/lang/lang-patchsec.py ${FW_MULTI}
-                ${LANG_BIN} ${LANG_FWBIN_TMP}
-        DEPENDS ${LANG_FWBIN} ${LANG_BIN}
-        COMMENT "Generating ${variant_name}-en_${LANG}.bin")
+        OUTPUT ${FW_LANG_FINAL}.bin
+        COMMAND ${CMAKE_OBJCOPY} -O binary ${FW_LANG_BASE} ${FW_LANG_FINAL}.bin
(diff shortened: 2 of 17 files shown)


Firmware/config.h
```diff
@@ -62,7 +62,7 @@
 
 // This is set by the cmake build to be able to take control of
 // the language flag, without breaking existing build mechanisms.
-#ifndef CMAKE_LANG_CONTROL
```
vintagepc, line 65: The reason I did this was to make it blatantly obvious where the language control comes from (cmake). With this change I can see users getting frustrated/confused why their setting is not taking effect if they change it here and try to do a cmake build.

CMakeLists.txt
```diff
@@ -2,92 +2,114 @@ cmake_minimum_required(VERSION 3.15)
...
+option(SECONDARY_LANGUAGES "Secondary language support in the firmware" ON)
 
-SET(MAIN_LANGUAGES cs de es fr it pl CACHE STRING "The list of 'main' languages to be included, in the correct order")
-SET(COMMUNITY_LANGUAGES nl ro hu hr sk sv no CACHE STRING "The list of community languages to be included, in the correct order")
-SET(SELECTED_LANGUAGES ${MAIN_LANGUAGES} ${COMMUNITY_LANGUAGES})
+# Language configuration
+set(MAIN_LANGUAGES
+    cs de es fr it pl
+    CACHE STRING "The list of 'main' languages to be included, in the correct order"
+    )
```
vintagepc, line 66: Can we agree the auto-formatter is garbage? why is one separate lines and one not?

CMakeLists.txt
```diff
@@ -2,92 +2,114 @@ cmake_minimum_required(VERSION 3.15)
...
+file(STRINGS ${CMAKE_CURRENT_SOURCE_DIR}/Firmware/config.h MAX_SIZE_LINE
+     REGEX "^#define \+LANG_SIZE_RESERVED \+"
+     )
+string(REGEX MATCH "0x[0-9]+" MAX_SIZE_HEX "${MAX_SIZE_LINE}")
 math(EXPR LANG_MAX_SIZE "${MAX_SIZE_HEX}" OUTPUT_FORMAT DECIMAL)
 message("Language maximum size (from config.h): ${LANG_MAX_SIZE} bytes")
 
-set (LANG_BIN_MAX 249856) # Ditto, this in xflash_layout.h but needs invocation of the preprocessor... :-/
+set(LANG_BIN_MAX 249856) # Ditto, this in xflash_layout.h but needs invocation of the
+                         # preprocessor... :-/
```
vintagepc: this is also garbage reformat,, why split the comment? We have large displays now, column limits are arbitrary anachronisms that need to die

CMakeLists.txt
```diff
@@ -208,298 +229,321 @@ set(AVR_SOURCES
...
-        TARGET ${FW_EN}
-        POST_BUILD
-        COMMAND ${CMAKE_OBJCOPY} -O ihex ${CMAKE_CURRENT_BINARY_DIR}/${FW_EN} ${CMAKE_BINARY_DIR}/${FN_PREFIX}-${FW_EN}.hex
-        BYPRODUCTS ${CMAKE_BINARY_DIR}/${FN_PREFIX}-${FW_EN}.hex
-        COMMENT "Generating ${variant_name} hex"
+      OUTPUT ${LANG_BIN}
+      #[[
+      # Check po file:
+      #COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/lang/lang-check.py --no-warning --map ${LANG_MAP} ${PO_FILE}
+      #]]
```
vintagepc, line 418: This should probably be re-enabled for the language builds... no?
I turned it off because it wasn't working for me but it is used by PF-build.

CMakeLists.txt
```diff
@@ -208,298 +229,321 @@ set(AVR_SOURCES
...
+    set(LANG_CATHEX ${LANG_TMP_DIR}/${variant_name}_cat.hex)
+
+    add_custom_command(
+      OUTPUT ${LANG_CATBIN}
+      COMMAND ${CMAKE_COMMAND} -E cat ${LANG_BINS} > ${LANG_CATBIN}
+      DEPENDS ${LANG_BINS}
+      COMMENT "Merging language catalogs"
+      )
+    #[[
+    #add_custom_command(OUTPUT ${LANG_FINAL_BIN}
```
vintagepc, line 444: if this is vestigial code, remove it.

CMakeLists.txt
wavexx: cmake-format is bad, or at least the defaults we are using are pretty bad. I've set the same values as the Buddy FW for the moment and just accepted the garbage...

vintagepc: Not sure I agree with that mindset, but :shrug:

vintagepc (changes requested):
Needs work. Doing a build followed by a "clean all" does not clean the MK2.5 hex files from the build/ dir. 

There may also be a pathing issue with one of the targets because it leaves crap like this in the build dir too:

image

CMakeLists.txt
wavexx: Yup. But this (and the following) were commented before and remained commented as I didn't have time to check them.

In this case, lang-check will emit huge amounts of output with the current build layout due to the number of targets. IMHO it needs some sort of special handling.

CMakeLists.txt
wavexx: As above. This was just a pass-through in this PR.

Firmware/config.h
wavexx: I dropped the comment.

CMakeLists.txt
wavexx: Yeah, the formatter here is just using the argument _count_ as a threshold...

CMakeLists.txt
wavexx: Following 4 projects with where this is being used, I'm just trying to make it uniform. I hate all autoformatters in general BTW.

If you can find some better settings for cmake-format (0.6.13, not the outdated 0.6.0 as used by buddy) I'm all ears. I'll try to push it though-out. I tried a bit in the past, but it always looked like garbage no matter what I tried.

CMakeLists.txt
vintagepc: that's why I don't agree with the principle of proliferating garbage ;) the more it spreads, the more widely accepted it becomes.

.cmake-format.py
```diff
@@ -0,0 +1,21 @@
+# If a statement is wrapped to more than one line, than dangle the closing
+# parenthesis on it's own line.
+dangle_parens = True
+dangle_align = 'child'
+
+# If true, the parsers may infer whether or not an argument list is sortable
+# (without annotation).
+autosort = True
```
vintagepc, line 8: can this be dangerous for the language lists?

wavexx: @3d-gussner probably knows.

I think the language list was always hand-sorted. The important part here is that we keep the two groups sorted separately, which is already the case.

wavexx:
I fixed the cleanup.

The path issue seems pre-existing, probably a side-effect of the nested project/s?
__ stands for dir-up here. If you look at the generated rules, all the path names are constructed as project-relative, so they include __ and these are just the paths holding the final objects.

Firmware/config.h
wavexx: I meant commit here.

wavexx:
BTW I wouldn't mind a different cmake formatter entirely, but I couldn't find alternatives.

CMakeLists.txt
vintagepc: skimmed the docs and didn't see anything particularly useful configuration wise :(

vintagepc:
> I fixed the cleanup.
> 
> The path issue seems pre-existing, probably a side-effect of the nested project/s? __ stands for dir-up here. If you look at the generated rules, all the path names are constructed as project-relative, so they include __ and these are just the paths holding the final objects.

OK, they must be part of the generated changes then - I can't recall seeing them before but it has been a while....

Do hard-link work on windows out of the box? I can't find any clear documentation on how that works on non posix platforms...
If not that change won't work. Last I checked you needed some weird developer mode in :window: s to be able to create symlinks - it's not a stock feature.

CMakeLists.txt
wavexx: Looking at buddy, where there isn't much discipline, I came to appreciate autoformatters. But I'm with you here, cmake-format is just subpar in everything it does. Some flags help with one case and destroy 20 more. I'm just trying to jump a little bit ahead and clone what we do on MMU here.

I wish I had alternatives.

vintagepc: change it back to tabs and I'll be happy /s :rofl:

wavexx:
Reading the docs, yes it does, and I guess it will lead to less issues than people copying the final binary compared to a symlink.
The only limit is that of course both paths need to be on the same filesystem, which I think is always going to be the case here unless some crazy folk decides to split the tree by hand.

wavexx:
@leptun can you confirm this works as expected on windows?

vintagepc (approved):
Cleans properly now - Approving provisionally once we confirm the linking behaviour.

CMakeLists.txt
arekm: "ColumnLimit" setting for that one thing (?).

OctoPrint uses pre-commit hook (and tool with the same name) for checking and reformatting (and this is also wired into github actions for all commits and pull requests). 

https://pre-commit.com/
https://github.com/OctoPrint/OctoPrint/blob/maintenance/.pre-commit-config.yaml
https://github.com/OctoPrint/OctoPrint/blob/master/.github/workflows/build.yml

There are various reformatting hooks supported, including simple things like trailing-whitespace or end-of-file-fixer while also more complicated things. 

~Full list:
https://pre-commit.com/hooks.html

c++ ones are also there but can't tell if these will fit your needs.

Example, manual run, for python project:
```
(venv) pi@prusa-octopi:~/devel/OctoPrint-Sdwire $ pre-commit run --hook-stage manual --all-files
fix end of files.........................................................Passed
trim trailing whitespace.................................................Failed
- hook id: trailing-whitespace
- exit code: 1
- files were modified by this hook

Fixing octoprint_sdwire/__init__.py

check for case conflicts.................................................Passed
check json...........................................(no files to check)Skipped
check yaml...............................................................Passed
check toml...........................................(no files to check)Skipped
check for merge conflicts................................................Passed
fix python encoding pragma...............................................Passed
OctoPrint codemod: batch.................................................Passed
isort....................................................................Passed
black....................................................................Passed
flake8...................................................................Passed
prettier.................................................................Passed
eslint...................................................................Passed
```

wavexx: @arekm See what we do here https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/master/.pre-commit-config.yaml

Historically though this project was always buildable with just Arduino which makes is very approachable, but is also why many things are clunky.

3d-gussner:
I have found a "bug" if `Configuration_prusa.h` is present:
- Cmake MK3S-EINSy10a_Multilang--Utility
- PF-build.sh -p1 MK3S
  - Doesn't delete `Configuration_prusa.h `
- Cmake MK3-EINSy10a_Multilang--Utility
Then both are identical.

Minor ToDos:
- The build time isn't shown on LCD Support screen just `1970-01-01`
- Project version also doesn't read the FW_FLAVOR and FW_FLAVERSION
- Also the `FW_COMMIT_NR` isn't updated
- NOT cmake related.
  - The `Update language ...` during language change flickers with FW version.

CMakeLists.txt
```diff
@@ -315,9 +315,18 @@ add_custom_target(ALL_MULTILANG)
...
+
+add_custom_target(
+  test_wrong_config COMMAND ${CMAKE_COMMAND} -DPROJECT_SOURCE_DIR=${PROJECT_SOURCE_DIR} -P
+                            ${PROJECT_CMAKE_DIR}/Check_wrong_config.cmake
+  )
+
 function(add_base_binary variant_name)
   add_executable(${variant_name} ${FW_SOURCES} ${FW_HEADERS} ${VARIANT_CFG_DST})
   set_target_properties(${variant_name} PROPERTIES EXCLUDE_FROM_ALL TRUE)
+  # add_dependencies(${variant_name} test_wrong_config) # Uncomment to check at build time
```
vintagepc: We should probably enable this - otherwise someone could configure with cmake, (passing the configure-time check) and then run pf build and break the build without knowing why.

wavexx:
The date/time/commit will be dealt with in another PR as it needs extra code to fetch the commit information.

I now also check if a stale Configuration_prusa.h is present in the source tree at configuration time (that is, when you run cmake) and refuse to continue until it's removed.

This cannot check if the file is added _later_ after configuration. For this purpose, I also added (but left commented) a build-time check that would catch such a case. This requires one rule to run at every make/ninja invocation even for a no-op build, which is something I deeply dislike.

CMakeLists.txt
wavexx: I commented below, but this results in a rule that runs even if the build should be a no-op. IMHO this should _never_ happen and sucks big time.

I couldn't find a way to do better than that via cmake though.

vintagepc: There's probably a way to prevent this by specifying a dummy output/byproducts file for cmake to look at so that it won't re-run unless other dependencies have changed. Maybe there's even a way to key it off the presence of the configuration_prusa.h file :thinking:

vintagepc: Thought: a little clunky but perhaps it could be e.g. a post-build command for the elf - sure, it'll still build the incorrect binary, and throw an error after that. But this way if the elf hasn't changed, it won't rerun the check either leaving it a true no-op.

wavexx: It also doesn't work the way I would like, as it runs late. This target would need to be a dependency on every object.

wavexx: Using a target with a dummy would be better. If you attach a post-build command it's going to run for all the variants. At least with a single global target you get just one run.

wavexx:
I wonder if we're just trying to fix the wrong thing. We could actually specify the file to include at build time with something hacky like this:

```
#ifndef VARIANT
#define VARIANT Configuration_prusa.h
#endif
#define INCLUDE(v) <v>
#include INCLUDE(VARIANT)
```

then pass `-DVARIANT=variant/path`, which would also avoid the copy. We could change current references that include "Configuration_prusa.h" directly to a small wrapper. In that case, Configuration_prusa.h would still be used as a fallback, but only when the variant header is not supplied. If that still exists in the build tree it would become irrelevant for the cmake build.

Thoughts?

wavexx:
I went with the above approach. Not only it avoids the need to copy/check Configuration_prusa, but you have the added advantage that you get compiler errors pointing _correct file_ when changing the variant header.

3d-gussner (approved):
Tested on MK404
Compiling with `PF-build.sh` and `cmake` works even with `Configuration_prusa.h` existing.

wavexx:
Merging then

PR #3653, opened 2022-10-03
cleanup: Fix more spourious missing return warnings

wavexx:
Rewrite to use a temporary and get rid of the warning. The generated asm is *unchanged*.


leptun (approved):
You probably didn't get these until now since they are not used in the MK3S build

PR #3656, opened 2022-10-04
Change fan RPM display to 5 digits

WarrenSchultz:
Using an altfan that has a lot more power in a small package (30mm, 10K RPM), the display can't handle the additional character, and pushes it to the third line. I've changed this in my local firmware build, and it works great, but thought it might be useful down the road for others, for a cleaner look for high-RPM fans.


leptun:
@3d-gussner The translations for this need to be checked. I strongly suspect that some languages might not fit with c=15.

WarrenSchultz:
Character tallies by language: Extruder fan/Print fan, French and Italian for Extruder Fan seem to be the only issues. (Those can probably also be compressed easily to match others abbreviations in the translations.)

```
CS:
11
14

DA:
0
0

DE:
15
12

ES:
14
11

FR:
16
11

HR:
15
12

HU:
15
10

IT:
16
10

LB:
0
0

LT:
0
0

NL:
13
10

NO:
15
11

PL:
11
11

RO:
15
12

SK:
11
14

SL:
0
0

SV:
14
15
```

3d-gussner (approved):
Thanks for the PR and we change the `Extruder fan` to `Hotend fan`.

3d-gussner (changes requested):
We need this PR to be based on MK3 branch and at this moment a rebase causes too many conflicts.
@WarrenSchultz Can you please make these changes to MK3 branch, or should I do these?

I have to create another PR to change from `Extruder fan` to `Hotend fan` anyway.

WarrenSchultz:
@3d-gussner Probably better for you to do it, if you have the time. I try to touch as few files as possible as an outside contributor :)

One thing I was wondering, would it make more sense to move the RPM values to the next line for each, so that the translations have more headroom? Or are there memory constraints that prevent that?

3d-gussner:
> @3d-gussner Probably better for you to do it, if you have the time. I try to touch as few files as possible as an outside contributor :)
> 
> One thing I was wondering, would it make more sense to move the RPM values to the next line for each, so that the translations have more headroom? Or are there memory constraints that prevent that?

I want to keep the messages as short as possible and also the used space on LCD. There may come other requests to be shown in this screen, so keep we it short and simple.

I will commit your the 5 digits to my https://github.com/prusa3d/Prusa-Firmware/pull/3694

Thanks again for the PR and contribution to the firmware development, very appreciated.

Please close the PR.

WarrenSchultz:
Thanks! Much appreciated!

3d-gussner:
@WarrenSchultz see https://github.com/prusa3d/Prusa-Firmware/pull/3694/commits/e83728f72c1921daf1f2dff8b17f763ece87c321

PR #3661, opened 2022-10-05
CI: Update travis to build/test with cmake

wavexx:
Include the new cmake-based build in travis.
Split the build into 3 stages: cmake, legacy and tests.

cmake: new cmake-based build (*all* variants, including english only)
legacy: old build.sh based build (using arduino-builder)
tests: cmake tests (not terribly useful at the moment)


3d-gussner (approved):
Finally it works with Travis

PR #3663, opened 2022-10-06
Add padded size for individual languages

3d-gussner:
Detailed output of used space.


wavexx:
Do we want to show the padded size also the mk3/mk3s? In that case the padding is done for the final catalog (ie, not for each language).

3d-gussner:
> Do we want to show the padded size also the mk3/mk3s? In that case the padding is done for the final catalog (ie, not for each language).
This is just bit more detailed output, which can help us to determine how much space we have per each languages to reach the padded size.

The final padding is already there https://github.com/prusa3d/Prusa-Firmware/blob/MK3/cmake/Check_final_lang_bin_size.cmake

wavexx (approved):

PR #3665, opened 2022-10-06
:construction: WIP: PFW-1307 build: Set all version data based on git tags

wavexx:
Drop the ALPHA/BETA/custom version handling. Set the version string and firmware date based on the current git tree which has all the required details.

- Use the short HEAD commit hash as the default suffix. This also overrides the old "build number" which is essentially useless for reproducibility purposes. A truncated 16bit version of the hash is now also used in its place in the eeprom.
- Use the output of "describe" as the full version string.
- Use the HEAD commit timestamp as SOURCE_DATE_EPOCH.

With the new structure it's possible to checkout an old tree and get a reproducible artifact. The Configuration.h header provides fallback values whenever git is not available (for example, for zip archives).

In order to get a meaningful description we need to create dummy annotated tags when the version number is bumped. After this is done (for example, tagging 9f1f396c18ecc91230959c2b47f6271824398fed with t3.13), the resulting version would look something like:
```
-- Project version ...........: 3.13.0
-- Project version suffix ....: 49bc78425
-- Project version description: t3.13-431-g49bc78425
```

A final release would have a much shorter description as a result:
```
-- Project version ...........: 3.13.0
-- Project version suffix ....: 49bc78425
-- Project version description: v3.13.0
```


vintagepc (approved):

Firmware/Configuration.h
```diff
@@ -27,42 +27,20 @@ extern PGM_P sPrinterName;
     #define FW_VERSION STR(FW_MAJOR) "." STR(FW_MINOR) "." STR(FW_REVISION) "-" STR(FW_FLAVOR) "" STR(FW_FLAVERSION)
 #endif
 
-#define FW_COMMIT_NR 4987
+#ifndef FW_FLAVERSION
```
3d-gussner: Please keep `#define FW_COMMIT_NR 4987`

3d-gussner (changes requested):
We have to keep the Build number

Firmware/Configuration.h
wavexx: I will add the ability to include a build number via cmake, but not in the sources

3d-gussner:
https://github.com/prusa3d/Prusa-Firmware/pull/4044 is based on this PR setting it to draft and we may close it in favor of the PR 4044

3d-gussner:
Obsolete by https://github.com/prusa3d/Prusa-Firmware/pull/4366

PR #3674, opened 2022-10-09
MMU: Sync first layer purge line with PrusaSlicer generated output

gudnimg:
If one generates a multicolor gcode file with PrusaSlicer, the purge line gcode will look like this.

This reduces clicking sounds from the extruder, filament extrusion is reduced by 6mm.


3d-gussner (approved):
Same values as in PS2.5.0 and latest PS profiles for the MK3S + MMU2S

PR #3675, opened 2022-10-09
Fix issue where MMU status messages are not padded

gudnimg:
Fix issue where MMU status messages were not padded. Also combined common cases to reduce code size (saves 10 bytes of flash)


leptun (approved):
Looks good to me. It should fix the lcd padding for the mmu messages

PR #3683, opened 2022-10-12
cmake: Toolchain tweaks

wavexx:
Allow two toolchains to be used:

- AvrGcc: use avr-gcc from dependencies
- AnyAvrGcc: use system's avr-gcc for testing

Switch avr-gcc to Microchip's version, which has less overhead.

Update travis and vscode to use AvrGcc.

### DIFF
--- a/.github/travis/cmake-build.sh
+++ b/.github/travis/cmake-build.sh
@@ -4,7 +4,7 @@ rm -rf build
 mkdir build
 cd build
 cmake .. \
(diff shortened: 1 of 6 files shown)


leptun (approved):
Works fine in my setup. This is the first time I build with cmake and it worked right away (after calling bootstrap.py)

3d-gussner:
Gonna test on Windows and Linux asap.

3d-gussner (approved):
Works on Linux and Windows

PR #3685, opened 2022-10-12
Optimisation: Reduce code size in rendering SD file names and folders

gudnimg:
No change in UI or behaviour just more efficient code size. Tested on MK3S+ with SD card.

Change in memory:
Flash: -88 bytes
SRAM: 0 bytes


leptun (approved):
Tested with files and dirs of W-1, W and W+1 sizes, All displayed properly

PR #3686, opened 2022-10-13
Add targets for language checking

wavexx:
Add the following macro targets to check translations:

- check_lang: check all languages for all variants
- check_lang_[variant]: check all languages for [variant]
- check_lang_[lang]: check all variants against [lang]
- check_lang_[variant]_[lang]: check a single variant/language

For example:

- ninja check_lang: check *everything*
- ninja check_lang_de: check German in all variants
- ninja check_lang_MK3S-EINSy10a: check all languages in the MK3S
- ninja check_lang_MK3S-EINSy10a_de: check German in the MK3S

The build now also checks for language *errors*. The old output is now split into a new travis stage due to the amount of output it produces.  See "Show all checks" -> Details -> "Stage 2: lang"

### DIFF
--- a/.github/travis/cmake-lang.sh
+++ b/.github/travis/cmake-lang.sh
@@ -1,0 +1,10 @@
+#!/bin/sh
+set -xe
+rm -rf build
+mkdir build
+cd build
+cmake .. \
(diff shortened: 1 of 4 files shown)


3d-gussner (approved):
Great addition to for translators.

PR #3687, opened 2022-10-13
lang: Add git diff filter instructions for po files

wavexx:
Add some gitattributes for po/pot files and relevant instructions for translators.


3d-gussner (approved):
Nice addition, works well

PR #3691, opened 2022-10-15
Cleanup: remove `automatic` parameter from `unload_filament()`

gudnimg:
In M600 and M702, the Z-lift is taken care of outside this function (via gcode) so it makes no sense to have a Z-lift within `unload_filament()`

Only exception to this is in the wizard, there I moved the Z-lift to happen before we start heating the nozzle.

Change in memory:
Flash: -12 bytes
SRAM: 0 bytes


gudnimg:
Note: I've been sitting on this change since we recently changed the gcodes for the MMU. I didn't want to touch the wizard code in that PR.

wavexx (approved):

gudnimg:
rebased PR to sync with MK3 branch :)

DRracer (approved):

PR #3692, opened 2022-10-16
:construction: WIP MMU Jam detection, the Redux

### LINKED ISSUE #2478: [ENHANCEMENT] MMU jam/grinding detection.
**Printer type** - MK3S
**MMU Upgrade** - MMU2S

For the -S variant with the IR sensor, the sensor state is ignored when printing. If there's a jam, the filament grinds and the IR sensor will change to zero. Seems useful to detect this and likely save the print (or at least the user's time). This should work for both heat-creep jams or feed-tube/buffer jams and binding. 

I haven't quite finished poking around in the mmu handling routines yet but I suspect the end result will probably be a jam notice that the mmu needs attention. Depending on the nature of the jam it may just make the situation more difficult if automatic action is attempted - I think it's best to let the user determine what's up. 

-> This will only work for the 3S variant; I hear the Mk3 sensor's reliability has improved significantly, but I suspect it's rarely used for the MMU anymore owing to the improved loading from using the IR lever.

It's been requested before in #2214 which was closed and noted it was in the backlog. I'm noting this as a new open issue because I'm going to take a stab at this; some of my filament seems more prone to this problem than others. Please feel free to add comments/concerns.

vintagepc:
This is a redux of #2499 for the new MMU infrastructure (closes #2478)

Obsoletes #2499 

Please note it's only been tested in that it compiles; I would like to gather feedback on the proposed implementation/theory before spending the time to polish and test it.


3d-gussner:
@vintagepc Are you working on this?

Phetdanai:
`hello`

3d-gussner:
@Phetdanai Please stop write useless comments. https://github.com/prusa3d/Prusa-Firmware/pull/3155#issuecomment-1890833114

PR #3694, opened 2022-10-17
PFW-1415: Rename `Extruder fan` to `Hotend fan`

3d-gussner:
@JakoobCZ FYI
Translated also `Hotend fan` in all languages with the existing `Hotend` and `fan` translations.

- [X] cherry-pick to MK3_3.12


Firmware/Marlin_main.cpp
```diff
@@ -1247,7 +1247,7 @@ void setup()
     temp_mgr_init();
 
 #ifdef EXTRUDER_ALTFAN_DETECT
-	SERIAL_ECHORPGM(_n("Extruder fan type: "));
+	SERIAL_ECHORPGM(_n("Hotend fan type: "));
```
leptun, line 1250: I wouldn't touch this one. It's not visible anyway and prusa link might be parsing it

If it is not parsed, then I see no issue with this change

3d-gussner: As "Hotend fan" is the only term we should use from now on I will create an issue for PrusaLink and add a Release Note. There might be some OctoPrint Plugins that depend on it.

3d-gussner: @leptun Thanks pointing this out and I have found something related in PrusaLink.

DRracer (approved):

PR #3698, opened 2022-10-17
Farm mode: cache filenames in the presort function

leptun:
This should fix the first filename in the SD menu with farm mode enabled.
Even though we are not sorting, we still want to benefit from the speedup of filename caching, so we allow presort() to cache the file positions in the FAT.

PFW-1414


3d-gussner (approved):
Tested on real hardware.

- [X] SD card files are shown correctly
- [X] SD card sort is disabled in Settings menu in FARM Mode

Issue #3270, opened 2021-08-21
[BUG] Exiting `Support` submenu like `Extruder Info` and others does not reset `lcd_timeoutToStatus` timer


gudnimg:
**Printer type** - MK3S
**Printer firmware version**-  3.10.1-RC1 and before

**MMU Upgrade** - N/A

**SD card or USB/Octoprint**
 SD card

**Describe the bug**

Minor bug which shouldn't be a huge concern except maybe to increase consistency.

When a user enters any of the submenus under `Support` like `Extruder Info` and then when the user exits the submenu the LCD will wait indefinitely in `Support` without user input. The expected behaviour would be after 30 seconds the printer returns to the main menu.

The bug here is `lcd_timeoutToStatus.stop()` is called and when `menu_back_if_clicked()` is called or `menu_back()`... the timer is not reset. This is another issue of when the LCD enters a menu, the timer is not set.

Currently this affects any code that uses: `lcd_timeoutToStatus.stop()` except `Sensor Info` menu.
  
**To Reproduce**
1. Turn on printer
2. Click knob and scroll down and select the `Support` menu.
3. Scroll down and select `Extruder info`
4. Now click the knob to exit, be careful not to rotate the knob!
5. Wait for 30 seconds...
6. Observe that there is no change on the LCD.
  
**Expected behavior**
In step 6 above I would expect the LCD to return to the main menu.
 
**Note**

This is not an issue for `Sensor Info` but I think we should change the code

```cpp
void lcd_menu_show_sensors_state()                // NOT static due to using inside "Marlin_main" module ("manage_inactivity()")
{
	lcd_timeoutToStatus.stop();
	lcd_show_sensors_state();
	if(LCD_CLICKED)
	{
		lcd_timeoutToStatus.start();
		menu_back();
	}
}
```

to this below for consistency, but then this will be an issue too:

```cpp
void lcd_menu_show_sensors_state()                // NOT static due to using inside "Marlin_main" module ("manage_inactivity()")
{
	lcd_timeoutToStatus.stop();
	lcd_show_sensors_state();
	menu_back_if_clicked();
}
```

gudnimg:
Another note… even though clicking/pushing the knob to exit the submenu _does_ restart the timer, the submenu is executed once more afterwards which will stop the timer again.

leptun:
> Another note… even though clicking/pushing the knob to exit the submenu does restart the timer, the submenu is executed once more afterwards which will stop the timer again.

[...]

PR #3713, opened 2022-10-22
PFW-1397 Add Toolchange counter

gudnimg:
This PR implements a new statistic for MMU2S users. Every tool change is now counted over the printer's lifetime. This statistic can be useful to determine when it is time to replace the "small" PTFE tube which is inside the extruder. 

The following will increment the counter:
1.  Toolchanges via T-code: `T0`, `T1`, `T2`, `T3`, `T4`, `T5`
2. Single material toolchange via T-code: `Tx` 
    - **NOTE:** `Tc` does not increment the counter. Because `Tx` and `Tc` are always used together, only `Tx` increments the counter to prevent double counts.
4. Any calls to `MMU2::load_filament_to_nozzle()`, this includes:
    - T-code `T?`
    - Loading to the nozzle via LCD menu
    - G-codes: `M600` and `M701` 

⚠️ **NOTE:**
> Currently it has not been decided how to show this statistic to users, so for the time being I've placed it inside the **Fail stats MMU** menu for convenience only. It had one unused menu item row.

----

Change in memory:
Flash: +262 bytes
SRAM: +1 bytes

NOTE: The EEPROM code is eating up a lot of this.

### DIFF
--- a/Firmware/menu.cpp
+++ b/Firmware/menu.cpp
@@ -29,7 +29,7 @@ uint8_t menu_line = 0;
 uint8_t menu_item = 0;
 uint8_t menu_row = 0;
 uint8_t menu_top = 0;
-static bool menu_changed; // flag to indicate a new menu was entered
+bool menu_changed; // flag to indicate a new menu was entered
 
 uint8_t menu_clicked = 0;
 
@@ -347,11 +347,6 @@ uint8_t menu_item_back_P(const char* str)
 
 bool __attribute__((noinline)) menu_item_leave(){
     return ((menu_item == menu_line) && menu_clicked && (lcd_encoder == menu_item)) || menu_leaving;
-}
-
-bool menu_item_enter() {
-    menu_changed = false;
-    return !menu_changed;
 }
 
 uint8_t menu_item_function_P(const char* str, menu_func_t func)
--- a/Firmware/menu.cpp
+++ b/Firmware/menu.cpp
@@ -29,6 +29,7 @@ uint8_t menu_line = 0;
 uint8_t menu_item = 0;
 uint8_t menu_row = 0;
 uint8_t menu_top = 0;
+static bool menu_changed; // flag to indicate a new menu was entered
 
 uint8_t menu_clicked = 0;
 
@@ -53,6 +54,7 @@ void menu_goto(menu_func_t menu, const uint32_t encoder, const bool feedback, bo
 		menu_menu = menu;
 		lcd_encoder = encoder;
 		menu_top = 0; //reset menu view. Needed if menu_back() is called from deep inside a menu, such as Support
+		menu_changed = true;
 		CRITICAL_SECTION_END;
 		if (reset_menu_state)
 			menu_data_reset();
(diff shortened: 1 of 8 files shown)


leptun:
I'm not sure it's a good idea to update the eeprom for every toolchange. This would wear out the eeprom of the atmega2560 after 100000 toolchanges. This exact thing happened in the MMU firmware where there is a load balancing algorithm for the eeprom cells. It might be better if this statistic is only saved to eeprom at the end of the print. What do you think about this proposal?

gudnimg:
@leptun I like that idea, but that means we need to sacrifice 2 bytes of SRAM which I am fine with. 2 bytes because the print can easily go over 255 toolchanges, but I doubt any print goes over 65535 toolchanges. 

I also want to limit the EEPROM read within the menu. With the way I have it now, we are reading from the EEPROM address in a loop(). Might be better to either add a timer or find a way to have the value constant when the menu is opened.

leptun:
> find a way to have the value constant when the menu is opened.

You may be able to use the menu_data scratch buffer for this. Look at the SD menu and how the EEPROM_SD_SORT is cached

gudnimg:
Note: Issue https://github.com/prusa3d/Prusa-Firmware/issues/3270 can now be fixed easily if we enclose all relevant `lcd_timeoutToStatus.stop()` calls with `ON_MENU_ENTER()`.

DRracer:
@gudnimg we can save the toolchanges per print after the print finishes just like other stats. We'd lose the stats only if the user resets the printer mid print (i.e. we can save the value after print finished or stopped).

Those guaranteed 100K EEPROM writes can be achieved pretty fast assuming one toolchange per minute:
100K/60/24 = ~69 days, which is something like 3 bigger prints :wink: 
On the MMU side the same amount of toolchanges is split across a wider EEPROM memory range.

gudnimg:
For what its worth I did print two prints this weekend to test the counter. And seems to work great. Counter is updated when the print finishes. 

58882960-FB52-4B57-B0CB-58C7F0A83755

Firmware/menu.cpp
```diff
@@ -347,6 +349,11 @@ bool __attribute__((noinline)) menu_item_leave(){
     return ((menu_item == menu_line) && menu_clicked && (lcd_encoder == menu_item)) || menu_leaving;
 }
 
+bool menu_item_enter() {
+    menu_changed = false;
+    return !menu_changed;
```
gudnimg: I suspect this is not correct. This always returns true. Will look at this when I have time.

Firmware/mmu2.h
```diff
@@ -286,6 +300,7 @@ class MMU2 {
     
     bool inAutoRetry;
     uint8_t retryAttempts;
+    uint16_t toolchange_counter;
```
DRracer, line 303: Total toolchange count is 32bit in eeprom while this variable is only 16bit. That assumes there will be no more than 64K toolchanges in a single print ... on MK3 this is probably a realistic assumption...

DRracer (approved):

PR #3714, opened 2022-10-23
Optimisation: use `failstats_reset_print()` in `factory_reset_stats()`

gudnimg:
Change in memory: 
Flash: -56 bytes
SRAM: 0 bytes


leptun (approved):
Looks like a good optimization

PR #3720, opened 2022-10-28
MK3_3.12: Add Menu Calibration -> Temp Model cal. 

3d-gussner:
Runs M310 A F0
Enables with M310 S1
Saves settings M500

Tested on MK404 with Octoprint serial logging.

### DIFF
--- a/Firmware/temp_model.h
+++ b/Firmware/temp_model.h
@@ -59,6 +59,7 @@ struct model_data
 };
 
 static bool enabled;          // model check enabled
+static bool valid = false;    // model is valid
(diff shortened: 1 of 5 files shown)


Firmware/temp_model.h
```diff
@@ -59,6 +59,7 @@ struct model_data
 };
 
 static bool enabled;          // model check enabled
+static bool valid = false;    // model is valid
```
wavexx, line 62: Why is this needed? temp_model::calibrated() should do the same and avoid having to set/reset another state variable?

DRracer (approved):

PR #3731, opened 2022-11-04
Bump up Revision to ALPHA1 6051

3d-gussner:
Temporary fix for MSG_BTN_MORE
pre-release pot and po update

- [ ] Create tag t3.13.0-ALPHA1 after merge


DRracer (approved):
looks ok to me :+1:

PR #3737, opened 2022-11-06
Fix long-press LCD inactivity timeout

### LINKED ISSUE #3715: [BUG]3.12.0-BETA1 Inconsistent long-press for Live-Z behavior
I'm testing 3.12.0-BETA1 on my MK3S+. Please note I'm also using PrusaLink 0.7RC1 so this behavior could be related to that (due to the changes on how LCD bottom line messages are being handled that were introduced in 3.12).

Printer's behavior back on 3.11:
- After encoder button long-press, the Live-Z menu was consistently popping up
- After some delay, where no Live-Z changes were being dialed-in for some time, the Live-Z menu would disappear.

Current behavior on 3.12.0.BETA1, when I print via PrusaConnect and the LCD bottom line scrolls the name of the file being printed:
- First encoder button long-press gets ignored, only resulting in what looks like an LCD redraw
- Second long press brings up the Live-Z menu, but then:
- Usually it doesn't ever time out, or:
- Sometimes times-out prematurely.

gudnimg:
Fixes #3715

Issue was likely introduced during 3.12 development here: https://github.com/prusa3d/Prusa-Firmware/pull/3268

When a long-press is triggered, the LCD inactivity timer needs to be restarted. This is done with `lcd_timeoutToStatus.start()`. The function `menu_lcd_lcdupdate_func()` does not reset the timer in this scenario so we need to do it in `menu_lcd_longpress_func()`

This also fixes unpredictable case where a long-press is triggered, and the timer times out immediately (because it was not reset properly)

Change in memory:
Flash: +8 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -7582,7 +7582,10 @@ uint8_t get_message_level()
 
 void menu_lcd_longpress_func(void)
 {
-	backlight_wake();
+    // Wake up the LCD backlight and,
+    // start LCD inactivity timer
+    lcd_timeoutToStatus.start();
+    backlight_wake();
     if (homing_flag || mesh_bed_leveling_flag || menu_menu == lcd_babystep_z || menu_menu == lcd_move_z || menu_block_mask != MENU_BLOCK_NONE)
     {
         // disable longpress during re-entry, while homing, calibration or if a serious error


DRracer (approved):
makes sense, but needs to be retested if this changes doesn't break something else

3d-gussner (approved):
Test confirm that OctoPrint print and long-press do time out.

PR #3746, opened 2022-11-10
M600: Fix bug where fan is not turned off

### LINKED ISSUE #3670: [ENHANCEMENT] 3.11. Cooling Fan Stays on while colorchange M600
--> MK3S, 3.11, no MMU, Sd card, 

I printed a larger model and somehow Prusa Slicer 2.5.0 added a color change....  So 

1. after M600, he parked the printer head above the I3 sign of the prusa metal sheet. 
4. the printing cooling fan was still active. 
5. i was somewhere else, so i detected that message an hour later. 
6. While the printer wanted me to unload filament, the front fan was all the time engaged-> BUG. Thus, it was cooling down my printed element, which was, unfortunatley on that corner lifted up. :/
7. i unloaded filament, reloaded it, continued printing, still happy. 

Wanted Behaviour: If the printer detects that the filament stucks or is empty, ther printer does not print anymore, so 

- please turn off the front cooling fan, as it is not needet anylonger (cause we cant currently print) until we reload and restart printing. Thats all :)

I enclosed the model, and the g code. you will find it in line 130939. 

Thanks a lot!

failure

Cheers George
[Model and GCode.zip](https://github.com/prusa3d/Prusa-Firmware/files/9741045/Model.and.GCode.zip)

gudnimg:
The firmware saves the current fanspeed before parking, and the fanspeed is restored after unparking. The problem is the fan was never actually paused.

Turning the fan off just before the retraction is consistent with Marlin 2.

Fixes #3670

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3538,6 +3538,9 @@ static void gcode_M600(bool automatic, float x_position, float y_position, float
 
     memcpy(lastpos, current_position, sizeof(lastpos));
 
+    // Turn off the fan
+    fanSpeed = 0;
+
     // Retract E
     current_position[E_AXIS] += e_shift;
     plan_buffer_line_curposXYZE(FILAMENTCHANGE_RFEED);


gudnimg:
I see the same issue applies to `feedmultiply` (feed rate). We save the current value but we then never do anything with it. Do we need to backup this value?

leptun (approved):
I tested the fan change and it works properly.

About the `feedmultiplyBckp`, I think there might be something wrong going on with a menu that messes up the `feedmultiply`. I had the `feedmultiply` set to 100%. Executed M600. Unloaded the filament, switched to another filament, had an error, recovered, selected filament not loaded, reloaded the same filament, selected color not clear, selected resume. I noticed that while the M600 was restoring the print, the `feedmultiply` on the screen was set to a higher value. This leads me to believe that some menu isn't clearing the lcd_encoder_diff or something like that, resulting in unintentional rotation on the status screen once M600 finishes. What hid this issue was the save and restore of `feedmultiplyBckp` since after M600 finished executing, the `feedmultiply` was set back to 100%, but it was indeed broken for a brief moment. If this other issue is solved, then I have no issue with removing `feedmultiplyBckp`. I like the idea of getting rid of sprintf for gcode generation.

gudnimg:
@leptun interesting findings with `feedmultiply`, I have not noticed this before.

leptun:
Something which maybe wasn't too clear in my message above, I was using the MMU without spooljoin for unloading and reloading the filament. That's what caused the error (FSENSOR_TOO_EARLY since I think it didn't try unloading at all. I was just prompted to select the new filament since the FINDA wasn't sensing a filament). The way I had the filament loaded was with a previous firmware (3.12) bypassing the MMU. After I updated to 3.13, the MMU connected, but the ptfe wasn't connected initially.

DRracer (approved):

PR #3748, opened 2022-11-11
Fix PFW-1364 & possible comms timeout during reheat

vintagepc:
Per title. 

Underlying issue was recursion guarding preventing the MMU protocol from continuing to query while in the reheat wait loop. 

This should also fix issues with MMU communications timing out after a reheat if it took too long.

### DIFF
--- a/Firmware/mmu2.cpp
+++ b/Firmware/mmu2.cpp
@@ -221,14 +221,14 @@ void MMU2::mmu_loop() {
         return;
     avoidRecursion = true;
 
-    mmu_loop_inner();
+    mmu_loop_inner(true);
 
     avoidRecursion = false;
 }
 
-void MMU2::mmu_loop_inner()
+void __attribute__((noinline)) MMU2::mmu_loop_inner(bool reportErrors)
 {
-    logicStepLastStatus = LogicStep(); // it looks like the mmu_loop doesn't need to be a blocking call
+    logicStepLastStatus = LogicStep(reportErrors); // it looks like the mmu_loop doesn't need to be a blocking call
 
     if (is_mmu_error_monitor_active){
         // Call this every iteration to keep the knob rotation responsive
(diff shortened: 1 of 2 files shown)


DRracer (approved):

gudnimg:
@vintagepc is this branch based on latest MK3? I'm seeing flash usage increase by 510 bytes.

MK3S Multilang build
Before:
```
[build] Program:  252058 bytes (96.2% Full)
[build] 
[build] Data:       5683 bytes (69.4% Full)

```
After:
```
Program:  252616 bytes (96.4% Full)
[build] 
[build] Data:       5683 bytes (69.4% Full)
```

vintagepc:
Fixed, now it's only 12b more

gudnimg:
Minor issue: I can confirm te MMU communications are no longer blocked. But when I hit "Retry" on error screen, the error screen will overwrite `"MMU Retry: Restoring temperature..."` screen.

vintagepc:
we can probably move the error handler if out of the mmu_loop_inner() then. Though I didn't see that  in the sim :thinking:

Firmware/mmu2.cpp
```diff
@@ -635,9 +635,9 @@ void MMU2::ResumeHotendTemp() {
         lcd_display_message_fullscreen_P(_i("MMU Retry: Restoring temperature...")); ////MSG_MMU_RESTORE_TEMP c=20 r=4
         //@todo better report the event and let the GUI do its work somewhere else
         ReportErrorHookSensorLineRender();
-        waitForHotendTargetTemp(100, [this]{
+        waitForHotendTargetTemp(100, []{
             manage_inactivity(true);
-            this->mmu_loop_inner(); // This keeps the comms alive, the call in manage_inactivity is blocked by recursion guard.
+            mmu2.mmu_loop_inner();
```
DRracer: :scream: is the compiler really so conservative (stupid)?

vintagepc: I think it's more a case of it pulling in additional function overhead for a pointer reference to the class... (which is otherwise a global static singleton)

Firmware/mmu2.cpp
```diff
@@ -779,30 +779,36 @@ StepStatus MMU2::LogicStep() {
...
     default:
-        break;
+        if(reportErrors) {
+            switch (ss)
+            {
+            case CommandError:
+                ReportError(logic.Error(), ErrorSourceMMU);
+                break;
+            case CommunicationTimeout:
+                state = xState::Connecting;
```
DRracer, line 795: I don't feel comfortable with these CommunicationTimeout and ProtocolError states, but let's keep them for later.

DRracer (approved):

PR #3749, opened 2022-11-12
Optimisation: reduce code size in showing file name being printed

gudnimg:
The PR optimizes how the SD card file name is printed on the screen.

`lcd_print_pad()` now returns the last character it's pointer pointed to which is `'\0'` (NULL) when the end of the string was reached, or a non-zero value which indicates the opposite. We can use this directly to make the file name scroll to the left on the LCD by shifting the starting pointer by one index each time `lcd_print_pad()` didn't reach the end of the string.

Affected UI:
* SD card filename on the status screen
* SD card filenames in the "Print from SD" menu.

Change in memory:
Flash: -140 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/lcd.cpp
+++ b/Firmware/lcd.cpp
@@ -528,13 +528,14 @@ void lcd_print(const char* s)
 	while (*s) lcd_write(*(s++));
 }
 
-void lcd_print_pad(const char* s, uint8_t len)
+char lcd_print_pad(const char* s, uint8_t len)
 {
     while (len && *s) {
         lcd_write(*(s++));
         --len;
     }
     lcd_space(len);
+    return *s;
 }
 
 void lcd_print(char c, int base)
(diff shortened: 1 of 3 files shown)


gudnimg:
Note: I have not tested this with a SD file which has a name length less than `LCD_WIDTH`. Will hopefully test that later today.

leptun (commented):
I feel like we can optimize this much further. We just need a way of knowing if lcd_print_pad got to the end of the string or not. I propose that lcd_print_pad `return *s;` so that we can check if we reached the end of the string or not. That would simplify this whole filename printing on the status screen. The same optimization could be applied also to the SD menu (maybe)

gudnimg:
Marking as ready for review since I tested both short filenames and long filenames. Both work without issues.

leptun (approved):
This looks perfect. Can't test it though, so I take your word that it works.
The next step might be to make the SD menu printing similar to this. It should be quite similar to this in the way it's rendered (- the folder icon and the cursor).

leptun (approved):
Tested on MK404. Reapproving

DRracer (approved):

PR #3750, opened 2022-11-12
Optimisation: remove global variable

gudnimg:
Remove `progressbar_block_count`

Change in memory:
Flash: -6 bytes
SRAM: -1 byte

### DIFF
--- a/Firmware/menu.cpp
+++ b/Firmware/menu.cpp
@@ -556,18 +556,18 @@ void menu_progressbar_init(uint16_t total, const char* title)
 	progressbar_total = total;
 	
 	lcd_set_cursor(0, 1);
-	lcd_printf_P(PSTR("%-20.20S\n"), title);
+	lcd_printf_P(PSTR("%-20.20S"), title);
 }
 
 void menu_progressbar_update(uint16_t newVal)
 {
-	uint8_t newCnt = (newVal * LCD_WIDTH) / progressbar_total;
-	if (newCnt > LCD_WIDTH)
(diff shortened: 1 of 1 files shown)


Firmware/menu.cpp
```diff
@@ -566,10 +564,9 @@ void menu_progressbar_update(uint16_t newVal)
 	uint8_t newCnt = (newVal * LCD_WIDTH) / progressbar_total;
 	if (newCnt > LCD_WIDTH)
 		newCnt = LCD_WIDTH;
-	while (newCnt > progressbar_block_count)
+	while (newCnt--)
```
gudnimg: Main thing to test is if this should be `--newCnt` or `newCnt--`

leptun:
The reason for having progressbar_block_count was so that I would only render the "remaining" number of blocks that need to be added to the progress bar. So if 2 blocks are rendered and we need to display 3 blocks, I don't draw all three blocks, but only the newest block. This was an optimization so that we don't waste time rendering the display during sorting / checking file since those algorithms update the progress bar quite often. Without progressbar_block_count, every time that we call menu_progressbar_update, the entire progress bar is redrawn. Most of the time no redrawing actually is needed since most of the time progressbar_block_count will be equal to newCnt. If the two are equal, the no LCD changes are needed during that update cycle. As such, I am against this code size optimization.

leptun (changes requested):
... see comment above

gudnimg:
In the end only 6 bytes are saved. I agree here we prefer speed over size. Closing the PR.

PR #3756, opened 2022-11-16
Display Welcome message after MMU progress report ends

DRracer:
Should make the status line look cleaner when the printer is not printing.

PFW-1426


leptun (commented):
Can this change overwrite the text message of a critical error?

DRracer:
I don't think so, but I need to test this change properly

DRracer:
It seems to have fixed the issue when operating the MMU while not printing.

3d-gussner (approved):

PR #3758, opened 2022-11-16
Optimize EEPROM usage/functions

DRracer:
It looks like the amount of code surrounding eeprom_read/update_* is significant.
By introducing 4 functions which do the repeated stuff I've been able to save ~270B. The proposed principle can be extended further.

### DIFF
--- a/Firmware/ConfigurationStore.cpp
+++ b/Firmware/ConfigurationStore.cpp
@@ -18,69 +18,16 @@
 
 M500_conf cs;
 
-//! @brief Write data to EEPROM
-//! @param pos destination in EEPROM, 0 is start
-//! @param value value to be written
-//! @param size size of type pointed by value
-//! @param name name of variable written, used only for debug input if DEBUG_EEPROM_WRITE defined
-//! @retval true success
-//! @retval false failed
-#ifdef DEBUG_EEPROM_WRITE
-static bool EEPROM_writeData(uint8_t* pos, uint8_t* value, uint8_t size, const char* name)
-#else //DEBUG_EEPROM_WRITE
-static bool EEPROM_writeData(uint8_t* pos, uint8_t* value, uint8_t size, const char*)
-#endif //DEBUG_EEPROM_WRITE
-{
-#ifdef DEBUG_EEPROM_WRITE
-	printf_P(PSTR("EEPROM_WRITE_VAR addr=0x%04x size=0x%02x name=%s\n"), pos, size, name);
-#endif //DEBUG_EEPROM_WRITE
-  while (size--)
-  {
-
-        eeprom_update_byte(pos, *value);
-        if (eeprom_read_byte(pos) != *value) {
-            SERIAL_ECHOLNPGM("EEPROM Error");
-            return false;
-        }
-
-    pos++;
-    value++;
-  }
-    return true;
-}
-
-#ifdef DEBUG_EEPROM_READ
-static void EEPROM_readData(uint8_t* pos, uint8_t* value, uint8_t size, const char* name)
-#else //DEBUG_EEPROM_READ
-static void EEPROM_readData(uint8_t* pos, uint8_t* value, uint8_t size, const char*)
-#endif //DEBUG_EEPROM_READ
-{
-#ifdef DEBUG_EEPROM_READ
-	printf_P(PSTR("EEPROM_READ_VAR addr=0x%04x size=0x%02x name=%s\n"), pos, size, name);
-#endif //DEBUG_EEPROM_READ
-    while(size--)
-    {
-        *value = eeprom_read_byte(pos);
-        pos++;
-        value++;
-    }
-}
-
 #define EEPROM_VERSION "V2"
 
 #ifdef EEPROM_SETTINGS
 void Config_StoreSettings()
 {
-  strcpy(cs.version,"000"); //!< invalidate data first @TODO use erase to save one erase cycle
-  
-  if (EEPROM_writeData(reinterpret_cast<uint8_t*>(EEPROM_M500_base),reinterpret_cast<uint8_t*>(&cs),sizeof(cs),0), "cs, invalid version")
-  {
+  strcpy(cs.version, EEPROM_VERSION);
+  eeprom_update_block(reinterpret_cast<uint8_t*>(&cs), reinterpret_cast<uint8_t*>(EEPROM_M500_base), sizeof(cs));
 #ifdef TEMP_MODEL
-      temp_model_save_settings();
+  temp_model_save_settings();
 #endif
-      strcpy(cs.version,EEPROM_VERSION); //!< validate data if write succeed
-      EEPROM_writeData(reinterpret_cast<uint8_t*>(EEPROM_M500_base->version), reinterpret_cast<uint8_t*>(cs.version), sizeof(cs.version), "cs.version valid");
-  }
 
   SERIAL_ECHO_START;
   SERIAL_ECHOLNPGM("Settings Stored");
@@ -269,12 +216,11 @@ bool Config_RetrieveSettings()
 {
   bool previous_settings_retrieved = true;
     char ver[4]=EEPROM_VERSION;
-    EEPROM_readData(reinterpret_cast<uint8_t*>(EEPROM_M500_base->version), reinterpret_cast<uint8_t*>(cs.version), sizeof(cs.version), "cs.version"); //read stored version
+    eeprom_read_block(reinterpret_cast<uint8_t*>(cs.version), reinterpret_cast<uint8_t*>(EEPROM_M500_base->version), sizeof(cs.version));
     //  SERIAL_ECHOLN("Version: [" << ver << "] Stored version: [" << cs.version << "]");
     if (strncmp(ver,cs.version,3) == 0)  // version number match
     {
-
-        EEPROM_readData(reinterpret_cast<uint8_t*>(EEPROM_M500_base), reinterpret_cast<uint8_t*>(&cs), sizeof(cs), "cs");
+        eeprom_read_block(reinterpret_cast<uint8_t*>(&cs), reinterpret_cast<uint8_t*>(EEPROM_M500_base), sizeof(cs));
         calculate_extruder_multipliers();
 
     //if max_feedrate_silent and max_acceleration_units_per_sq_second_silent were never stored to eeprom, use default values:
--- a/Firmware/ConfigurationStore.cpp
+++ b/Firmware/ConfigurationStore.cpp
@@ -198,17 +198,6 @@ static const M500_conf default_conf PROGMEM =
 };
 
 
-static bool is_uninitialized(void* addr, uint8_t len)
-{
-    while(len--)
-    {
-        if(reinterpret_cast<uint8_t*>(addr)[len] != 0xff)
-            return false;
-    }
-    return true;
-}
-
-
 //! @brief Read M500 configuration
 //! @retval true Succeeded. Stored settings retrieved or default settings retrieved in case EEPROM cs was empty.
 //! @retval false Failed. Default settings has been retrieved, because of version mismatch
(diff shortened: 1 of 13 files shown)


gudnimg (approved):
I see no issue with this PR so I'm approving 👍

Firmware/eeprom.cpp
```diff
@@ -102,12 +102,12 @@ if (eeprom_read_byte((uint8_t*)EEPROM_PINDA_TEMP_COMPENSATION) == 0xff) eeprom_u
 	if (eeprom_read_dword((uint32_t*)EEPROM_JOB_ID) == EEPROM_EMPTY_VALUE32)
```
leptun: Would doing this also for DWORD help?

Firmware/mmu2/error_codes.h
```diff
@@ -46,7 +46,10 @@ enum class ErrorCode : uint_fast16_t {
     ///< - a piece of filament was left inside - pushed in front of the loaded filament causing the fsensor trigger too early
     ///< - fsensor is faulty producing bogus triggers
 
-    MOVE_FAILED = 0x800a, ///< generic move failed error - always reported with the corresponding axis bit set (Idler or Selector) as follows:
+    FINDA_FLICKERS = 0x800a, ///< FINDA flickers - seems to be badly calibrated and happens to be pressed at spots where it used to be not pressed before.
```
leptun: I feel like these changes are unintentional in this PR

leptun (changes requested):
- [x] There is one extra commit from #3757. I believe that commit should not be part of this PR.
- [x] change EEPROM_writeData to eeprom_update_block.
- [x] I think there are also some init_default which were missed (byte/word).
- [x] It might also be beneficial to add the DWORD equivalent for increment and init_default. We might be able to squeeze a few more bytes from that.
- [x] I'm not sure how I feel about the name "increment". When I think of increment I think of `++`, not of `+=`. I don't have a better name in mind, so it could stay like this in the end. -> Split it into increment (++) and add(+=).

DRracer:
@leptun I completely agree with you, this PR was meant as a proof of concept, that we can really save considerable amount of code by refactoring the EEPROM-related code.
Therefore I really don't care about the extra commit and/missing dword support - on the contrary - I hope someone will pick up the idea and polish it into a clean PR :wink:

Firmware/ConfigurationStore.cpp
```diff
@@ -18,69 +18,16 @@
...
 #define EEPROM_VERSION "V2"
 
 #ifdef EEPROM_SETTINGS
 void Config_StoreSettings()
 {
-  strcpy(cs.version,"000"); //!< invalidate data first @TODO use erase to save one erase cycle
-  
-  if (EEPROM_writeData(reinterpret_cast<uint8_t*>(EEPROM_M500_base),reinterpret_cast<uint8_t*>(&cs),sizeof(cs),0), "cs, invalid version")
-  {
+  strcpy(cs.version, EEPROM_VERSION);
```
DRracer: May be assigning individual bytes results in a shorter code than `strcpy`.

Firmware/ConfigurationStore.cpp
```diff
@@ -269,12 +216,11 @@ bool Config_RetrieveSettings()
 {
   bool previous_settings_retrieved = true;
     char ver[4]=EEPROM_VERSION;
-    EEPROM_readData(reinterpret_cast<uint8_t*>(EEPROM_M500_base->version), reinterpret_cast<uint8_t*>(cs.version), sizeof(cs.version), "cs.version"); //read stored version
+    eeprom_read_block(reinterpret_cast<uint8_t*>(cs.version), reinterpret_cast<uint8_t*>(EEPROM_M500_base->version), sizeof(cs.version));
```
DRracer, line 204: we could read the version as a dword if that saves something ... it is just 4B long anyway.

Firmware/eeprom.cpp
```diff
@@ -182,6 +136,20 @@ void eeprom_switch_to_next_sheet()
     if (sheet >= 0) eeprom_update_byte(&(EEPROM_Sheets_base->active_sheet), sheet);
 }
 
+bool __attribute__((noinline)) eeprom_is_sheet_initialized(uint8_t sheet_num) {
+    return (eeprom_read_word(reinterpret_cast<uint16_t*>(&(EEPROM_Sheets_base->s[sheet_num].z_offset))) != EEPROM_EMPTY_VALUE16);
+}
+
+bool __attribute__((noinline)) eeprom_is_initialized_block(const void *__p, size_t __n) {
+    const uint8_t *top = (const uint8_t*)__p + __n;
+    for (const uint8_t *p = (const uint8_t *)__p; p < top; p++) {
```
DRracer: iterating 16bit variables will take more space (a pointer is a 16bit variable)

Firmware/ConfigurationStore.cpp
leptun: Thank you for pointing this out. Turns out there was another EEPROM_VERSION in RAM. I also found out the strncmp_P parameters were reversed. I'm now comparing with the version string in progmem from default_config

Firmware/ConfigurationStore.cpp
leptun: I don't think that's worth it. I prefer not messing too much with the dumpster fire that is the CS. In any case, a lot of stuff needs to be retested after these changes

Firmware/eeprom.cpp
leptun: I've changed this in the meantime. I'm now using __n as a downcounter and incrementing __p

Firmware/ConfigurationStore.cpp
DRracer: ok

leptun (approved):
I'm approving my own changes so that the PR can move forward. It needs serious testing before it can be merged

Firmware/ConfigurationStore.cpp
```diff
@@ -283,16 +282,13 @@ bool Config_RetrieveSettings()
...
-    //In case that storing to eeprom was not used yet, do not inform user that hardcoded settings are used.
-    if (eeprom_read_byte(reinterpret_cast<uint8_t*>(&(EEPROM_M500_base->version[0]))) != 0xFF ||
-      eeprom_read_byte(reinterpret_cast<uint8_t*>(&(EEPROM_M500_base->version[1]))) != 0xFF ||
-      eeprom_read_byte(reinterpret_cast<uint8_t*>(&(EEPROM_M500_base->version[2]))) != 0xFF)
-    {
-      previous_settings_retrieved = false;
-    }
+        //Return false to inform user that eeprom version was changed and firmware is using default hardcoded settings now.
+        //In case that storing to eeprom was not used yet, do not inform user that hardcoded settings are used.
+        if (eeprom_is_initialized_block(EEPROM_M500_base->version, sizeof(EEPROM_M500_base->version))) {
```
DRracer, line 266: `sizeof(EEPROM_M500_base->version) != 3` but 4, I think this doesn't match the previous implementation. Not sure though if it matters.

DRracer: May be: if the EEPROM chip is empty, version at the memory cells will be just 4 bytes od 0xff, so the check will probably work ok.

Firmware/eeprom.cpp
```diff
@@ -140,15 +140,23 @@ bool __attribute__((noinline)) eeprom_is_sheet_initialized(uint8_t sheet_num) {
     return (eeprom_read_word(reinterpret_cast<uint16_t*>(&(EEPROM_Sheets_base->s[sheet_num].z_offset))) != EEPROM_EMPTY_VALUE16);
 }
 
+
 bool __attribute__((noinline)) eeprom_is_initialized_block(const void *__p, size_t __n) {
-    const uint8_t *top = (const uint8_t*)__p + __n;
-    for (const uint8_t *p = (const uint8_t *)__p; p < top; p++) {
-        if (eeprom_read_byte(p) != EEPROM_EMPTY_VALUE)
+    const uint8_t *p = (const uint8_t*)__p;
```
DRracer, line 140: Are we using a block of data bigger than 255 bytes? If not, we could limit `size_t` down to `uint8_t`. But a good check would have been necessary (I don't know how to do that nicely).

Firmware/eeprom.cpp
```diff
@@ -191,7 +199,17 @@ void __attribute__((noinline)) eeprom_init_default_dword(uint32_t *__p, uint32_t
         eeprom_write_dword(__p, def);
 }
 
+void __attribute__((noinline)) eeprom_init_default_float(float *__p, float def) {
+    if (eeprom_read_dword((uint32_t*)__p) == EEPROM_EMPTY_VALUE32)
+        eeprom_write_float(__p, def);
```
DRracer, line 215: Do we need eeprom_write_float at all? Or was the AVR libc smart enough to implement it the same way like eeprom_write_dword?

Firmware/Marlin_main.cpp
```diff
@@ -9773,14 +9753,8 @@ bool setTargetedHotend(int code, uint8_t &extruder)
...
 {
-	if (eeprom_read_byte((uint8_t *)EEPROM_TOTALTIME) == 255 && eeprom_read_byte((uint8_t *)EEPROM_TOTALTIME + 1) == 255 && eeprom_read_byte((uint8_t *)EEPROM_TOTALTIME + 2) == 255 && eeprom_read_byte((uint8_t *)EEPROM_TOTALTIME + 3) == 255)
-	{
-		eeprom_update_dword((uint32_t *)EEPROM_TOTALTIME, 0);
-		eeprom_update_dword((uint32_t *)EEPROM_FILAMENTUSED, 0);
-	}
-
-	unsigned long _previous_filament = eeprom_read_dword((uint32_t *)EEPROM_FILAMENTUSED); //_previous_filament unit: cm
-	unsigned long _previous_time = eeprom_read_dword((uint32_t *)EEPROM_TOTALTIME); //_previous_time unit: min
+	unsigned long _previous_filament = eeprom_init_default_dword((uint32_t *)EEPROM_FILAMENTUSED, 0); //_previous_filament unit: cm
```
DRracer: please prefer `uint32_t` over `unsigned long` (I know it has been in the original code)

Firmware/Prusa_farm.cpp
```diff
@@ -390,12 +390,8 @@ void prusa_statistics_update_from_lcd_update() {
 }
 
 void farm_mode_init() {
-    farm_mode = eeprom_read_byte((uint8_t*)EEPROM_FARM_MODE); 
-    if (farm_mode == 0xFF) {
-        farm_mode = false; //if farm_mode has not been stored to eeprom yet and farm number is set to zero or EEPROM is fresh, deactivate farm mode
-        eeprom_update_byte((uint8_t*)EEPROM_FARM_MODE, farm_mode);
-    }
-    else if (farm_mode) {
+    farm_mode = eeprom_init_default_byte((uint8_t*)EEPROM_FARM_MODE, 0);
```
DRracer, line 393: :+1:

Firmware/mesh_bed_calibration.cpp
```diff
@@ -3125,11 +3125,7 @@ void mbl_settings_init() {
 //magnet elimination: use aaproximate Z-coordinate instead of measured values for points which are near magnets
 	eeprom_init_default_byte((uint8_t*)EEPROM_MBL_MAGNET_ELIMINATION, 1);
 	eeprom_init_default_byte((uint8_t*)EEPROM_MBL_POINTS_NR, 3);
-	mbl_z_probe_nr = eeprom_read_byte((uint8_t*)EEPROM_MBL_PROBE_NR);
-	if (mbl_z_probe_nr == 0xFF) {
-		mbl_z_probe_nr = 3;
-		eeprom_update_byte((uint8_t*)EEPROM_MBL_PROBE_NR, mbl_z_probe_nr);
-	}
+	mbl_z_probe_nr = eeprom_init_default_byte((uint8_t*)EEPROM_MBL_PROBE_NR, 3);
```
DRracer, line 3128: :thinking: let's name `eeprom_init_default_byte` something like `eeprom_read_byte_or_default`. It's semantics has evolved from a simple init if unset.

DRracer (commented):
@leptun most of my comments are nitpicks, but there is one important - working with the eeprom version. We must be careful about the string operation - i.e. 3 bytes of characters + the '\0' at the end. That makes 4 bytes. I'm not sure if the init and check eeprom blocks are completely correct after refactoring.

Firmware/eeprom.cpp
leptun: CS is 209B. It's quite close. I think that's the largest one. In any case, I don't think this will save that much. The thing is with the avr, all function parameters are expanded to at least 2 bytes, so even with uint8_t it still takes 2 registers

Firmware/eeprom.cpp
leptun: I do not know if it got implemented more efficiently. In theory the code is a duplicate of the above, but I hope the compiler doesn't treat uint32_t and float differently in this case

Firmware/mesh_bed_calibration.cpp
leptun: I don't really see the problem with eeprom_init_default_byte. Init here could be taken as both initializing the eeprom cell and as initializing the cached value (assignment). I find byte_or_default more confusing tbh.

leptun:
I'll get to the rest of the issues tomorrow

Firmware/eeprom.cpp
leptun: I just checked the avr libc source and it does indeed look like the DWORD and FLOAT implementations in ASM are identical. Both symbols are attached to the same piece of code. As such, even the rest of the function should be identical. I expect the LTO to combine our two functions into one.

https://github.com/avrdudes/avr-libc/blob/55e8cac69935657bcd3e4d938750960c757844c3/libc/misc/eewr_dword.S#L45-L46

Firmware/eeprom.cpp
leptun: This only results in 22B of savings

Firmware/ultralcd.cpp
```diff
@@ -7694,16 +7685,6 @@ void lcd_crash_detect_disable()
...
-void lcd_experimental_toggle()
-{
-    uint8_t oldVal = eeprom_read_byte((uint8_t *)EEPROM_EXPERIMENTAL_VISIBILITY);
-    if (oldVal == EEPROM_EMPTY_VALUE)
-        oldVal = 0;
-    else
-        oldVal = !oldVal;
-    eeprom_update_byte((uint8_t *)EEPROM_EXPERIMENTAL_VISIBILITY, oldVal);
-}
-
```
gudnimg, line 7706: There is a similar function also here: `lcd_pinda_temp_compensation_toggle()`

Firmware/ConfigurationStore.cpp
leptun: That was my thought as well. An eeprom that is empty for real will have all cells 0xFF. If any of those cells are not 0xFF, then it means something was stored there. If the data is garbage, then the Version string will mismatch and the default settings will be loaded

Firmware/ConfigurationStore.cpp
```diff
@@ -251,50 +183,45 @@ static const M500_conf default_conf PROGMEM =
...
-static bool is_uninitialized(void* addr, uint8_t len)
+void Config_StoreSettings()
 {
-    while(len--)
-    {
-        if(reinterpret_cast<uint8_t*>(addr)[len] != 0xff)
-            return false;
-    }
-    return true;
+  strcpy_P(cs.version, default_conf.version);
```
leptun, line 188: I feel like this line might be redundant, but I'll leave it here anyway since it doesn't hurt too much

Firmware/ConfigurationStore.cpp
```diff
@@ -251,50 +183,45 @@ static const M500_conf default_conf PROGMEM =
...
+//! @retval false Failed. Default settings has been retrieved, because of version mismatch
 bool Config_RetrieveSettings()
 {
-  bool previous_settings_retrieved = true;
-    char ver[4]=EEPROM_VERSION;
-    EEPROM_readData(reinterpret_cast<uint8_t*>(EEPROM_M500_base->version), reinterpret_cast<uint8_t*>(cs.version), sizeof(cs.version), "cs.version"); //read stored version
+    eeprom_read_block(reinterpret_cast<uint8_t*>(cs.version), reinterpret_cast<uint8_t*>(EEPROM_M500_base->version), sizeof(cs.version));
     //  SERIAL_ECHOLN("Version: [" << ver << "] Stored version: [" << cs.version << "]");
-    if (strncmp(ver,cs.version,3) == 0)  // version number match
+    if (strncmp_P(cs.version, default_conf.version, sizeof(EEPROM_VERSION)) == 0)  // version number match
```
leptun, line 206: I always found this thing a bit stupid. Why not limit this to 4 characters? `sizeof(cs.version)`

Firmware/mesh_bed_calibration.cpp
DRracer: ok

PR #3759, opened 2022-11-16
Remove dead code (G-code level check in menus)

gudnimg:
The code was added in v3.8.0 but the menu was never implemented according to the Git history. The menu code has always been commented out. I propose we remove this code so it doesn't need to be maintained.

Translation list is one item shorter :)

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -4337,22 +4337,6 @@ void process_commands()
                 plan_set_z_position(current_position[Z_AXIS]);
             }
         }
-
-//-//
-/*
-    } else if(code_seen("rrr")) {
-MYSERIAL.println("=== checking ===");
-MYSERIAL.println(eeprom_read_byte((uint8_t*)EEPROM_CHECK_MODE),DEC);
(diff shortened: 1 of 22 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -4369,52 +4353,8 @@ eeprom_update_word((uint16_t*)EEPROM_NOZZLE_DIAMETER_uM,0xFFFF);
...
-
-  } 
-  // This prevents reading files with "^" in their names.
-  // Since it is unclear, if there is some usage of this construct,
-  // it will be deprecated in 3.9 alpha a possibly completely removed in the future:
-  // else if (code_seen('^')) {
-  //  // nothing, this is a version line
-  // }
+    }
+  }
```
leptun, line 4357: ouf. This is some horrible indentation above. Can't wait for the day we hard reformat the repo

leptun (approved):

Firmware/Marlin_main.cpp
gudnimg: It was very tempting to fix some the surrounding code 🤣

DRracer (approved):

PR #3761, opened 2022-11-17
Recover from MMU restart while performing a command

DRracer:
This PR will bring smooth recovery from accidental MMU restarts while performing a command. 
Both printer and MMU are capable of recovering cleanly, it just hasn't been addressed yet to full extent.

Moreover, the infrastructure will be possibly leveraged to recovery of failed loads into extruder PTFE tube.

PFW-1427
PFW-1407
PFW-1432
PFW-1355

### DIFF
--- a/Firmware/mmu2.cpp
+++ b/Firmware/mmu2.cpp
@@ -44,8 +44,8 @@ static constexpr float MMU2_LOAD_TO_NOZZLE_LENGTH = 87.0F + 5.0F;
 // The printer intercepts such a call and sets its extra load distance to match the new value as well.
 static constexpr uint8_t MMU2_TOOL_CHANGE_LOAD_LENGTH = 5; // mm
 
-static constexpr uint8_t MMU2_EXTRUDER_PTFE_LENGTH = 42.3f; // mm
-static constexpr uint8_t MMU2_EXTRUDER_HEATBREAK_LENGTH  = 17.7f; // mm
+static constexpr float MMU2_EXTRUDER_PTFE_LENGTH = 42.3f; // mm
+static constexpr float MMU2_EXTRUDER_HEATBREAK_LENGTH  = 17.7f; // mm
 
 static constexpr float MMU2_LOAD_TO_NOZZLE_FEED_RATE = 20.0F; // mm/s
 static constexpr float MMU2_UNLOAD_TO_FINDA_FEED_RATE = 120.0F; // mm/s
--- a/Firmware/mmu2.cpp
+++ b/Firmware/mmu2.cpp
@@ -340,9 +340,9 @@ bool MMU2::VerifyFilamentEnteredPTFE()
 }
 
 void MMU2::ToolChangeCommon(uint8_t slot){
-    tool_change_extruder = slot;
     for(;;) { // while not successfully fed into extruder's PTFE tube
         for(;;) {
+            tool_change_extruder = slot;
             logic.ToolChange(slot); // let the MMU pull the filament out and push a new one in
             if( manage_response(true, true) )
                 break;
--- a/Firmware/mmu2.cpp
+++ b/Firmware/mmu2.cpp
@@ -306,7 +306,7 @@ void MMU2::DecrementRetryAttempts() {
     }
 }
 
-bool MMU2::FSensorCalibrationCheck()
+bool MMU2::VerifyFilamentEnteredPTFE()
 {
     st_synchronize();
 
@@ -327,6 +327,7 @@ bool MMU2::FSensorCalibrationCheck()
 
     if (fsensorState)
     {
+        IncrementLoadFails();
         return false;
     } else {
         // else, happy printing! :)
@@ -339,10 +340,8 @@ bool MMU2::FSensorCalibrationCheck()
 }
 
 void MMU2::ToolChangeCommon(uint8_t slot){
-    bool calibrationCheckResult = true;
     tool_change_extruder = slot;
-    do {
-        if (!calibrationCheckResult) unload(); // TODO cut filament
+    for(;;) { // while not successfully fed into extruder's PTFE tube
         for(;;) {
             logic.ToolChange(slot); // let the MMU pull the filament out and push a new one in
             if( manage_response(true, true) )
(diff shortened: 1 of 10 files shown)


DRracer:
rebased onto upsteam/MK3

Firmware/mmu2.cpp
```diff
@@ -306,9 +306,43 @@ void MMU2::DecrementRetryAttempts() {
...
+    while(blocks_queued())
+    {
+        // Wait for move to finish and monitor the fsensor the entire time
+        // a single 0 (1) reading is enough to fail the test
+        fsensorState |= !fsensor.getFilamentPresent();
+    }
+
+    if (fsensorState)
+    {
+        return false;
```
DRracer, line 338: IncrementMMULoadFails() :wink:

Firmware/mmu2.h
```diff
@@ -283,6 +280,17 @@ class MMU2 {
     /// @returns false if the MMU is not ready to perform the command (for whatever reason)
     bool WaitForMMUReady();
 
+    /// Redundancy test. After MMU completes a tool-change command
+    /// the printer will retract the filament by a distance set by the
+    //  Extra Loading Distance MMU register. If the Fsensor untriggers
```
vintagepc: This comment isn't right. The test load should feed forward and then retract a differential amount. 
i.e. from "Toochange done position" -> forward some amount `X` to just shy of the melt zone, followed by a retract of the same amount (since we are already at gears +5mm)

Firmware/mmu2.cpp
```diff
@@ -860,6 +913,16 @@ void MMU2::ReportError(ErrorCode ec, ErrorSource res) {
         lastErrorCode = ec;
         lastErrorSource = res;
         LogErrorEvent_P( _O(PrusaErrorTitle(PrusaErrorCodeIndex((uint16_t)ec))) );
+
+        if( ec != ErrorCode::OK ){
+            IncrementMMUFails();
+
+            // check if it is a "power" failure - we consider TMC-related errors as power failures
+            if( (uint16_t)ec & 0x7e00 ){ // @@TODO can be optimized to uint8_t operation
```
vintagepc: Magic numbers are bad. Static constexpr please.

Firmware/mmu2.h
gudnimg: Added a fix here 1f957d0 and used a similar constants as in the old MMU implementation.

Firmware/mmu2.cpp
```diff
@@ -44,6 +44,9 @@ static constexpr float MMU2_LOAD_TO_NOZZLE_LENGTH = 87.0F + 5.0F;
 // The printer intercepts such a call and sets its extra load distance to match the new value as well.
 static constexpr uint8_t MMU2_TOOL_CHANGE_LOAD_LENGTH = 5; // mm
 
+static constexpr uint8_t MMU2_EXTRUDER_PTFE_LENGTH = 42.3f; // mm
+static constexpr uint8_t MMU2_EXTRUDER_HEATBREAK_LENGTH  = 17.7f; // mm
```
vintagepc: ummm... what's the point of specifying `.f` if it's getting downgraded to uint8_t? Should it be float?

Firmware/mmu2.cpp
```diff
@@ -313,15 +316,14 @@ bool MMU2::VerifyFilamentEnteredPTFE()
     if (!fsensor.getFilamentPresent()) return false;
 
     uint8_t fsensorState = 0;
-    // MMU has finished its load, move filament back by ExtraLoadDistance
-    // If Fsensor reads 0 at any moment, then report FAILURE
-    current_position[E_AXIS] -= logic.ExtraLoadDistance();
+    // MMU has finished its load, push the filament further by some defined constant length
+    // If the filament sensor reads 0 at any moment, then report FAILURE
+    current_position[E_AXIS] += MMU2_EXTRUDER_PTFE_LENGTH + MMU2_EXTRUDER_HEATBREAK_LENGTH;
```
vintagepc: This can't be 60 anymore, you need to subtract the ExtraLoadDistance from this value or you'll make blobs on the tower.

Firmware/mmu2.cpp
```diff
@@ -332,7 +334,7 @@ bool MMU2::VerifyFilamentEnteredPTFE()
     } else {
         // else, happy printing! :)
         // Revert the movements
-        current_position[E_AXIS] += logic.ExtraLoadDistance();
+        current_position[E_AXIS] -= MMU2_EXTRUDER_PTFE_LENGTH + MMU2_EXTRUDER_HEATBREAK_LENGTH;
```
vintagepc: Same, needs to be adjusted by ExtraLoadDistance

gudnimg: Added here 031d00f. Also added brackets around the subtracted constant so the signs are correct.

Firmware/mmu2.cpp
gudnimg: Omg, this was stupid of me.

gudnimg: Fixed [df67c13](https://github.com/prusa3d/Prusa-Firmware/pull/3761/commits/df67c1384ca27243f9fe3eacadb256cff599dfe5)

Firmware/mmu2.cpp
```diff
@@ -299,52 +302,91 @@ void MMU2::ResetRetryAttempts(){
...
+        IncrementLoadFails();
+        return false;
+    } else {
+        // else, happy printing! :)
+        // Revert the movements
+        current_position[E_AXIS] -= (MMU2_EXTRUDER_PTFE_LENGTH + MMU2_EXTRUDER_HEATBREAK_LENGTH - logic.ExtraLoadDistance());
+        plan_buffer_line_curposXYZE(MMU2_LOAD_TO_NOZZLE_FEED_RATE);
+        st_synchronize();
+        return true;
+    }
```
vintagepc, line 342: You need to check the fsensor during and *after* the unload for this check to be effective. Checking it once after the push isn't really going to help anything.

gudnimg: Done: [6ccd375](https://github.com/prusa3d/Prusa-Firmware/pull/3761/commits/6ccd375f006f31137813de471eadb267edd1290e) the filament sensor is checked continously. I am using the binary operator `|=` to set the variable if 1 is seen. I'm assuming in this test that the Fsensor reading is debounced enough elsewhere in our filament sensor code so should reduce risk of false positives.

gudnimg (approved):
Approving @DRracer 's changes.

DRracer:
Approving @gudnimg 's changes

PR #3767, opened 2022-11-19
Improve `lcdui_print_extruder` to cover two more scenarios

gudnimg:
Simplify the rendering to cover more scenarios:
* `"[nr.]>?"`
* `"?>?"`
Both of these new scenarios indicate there is a bug in the firmware. Currently these scenarios were not rendered correctly and would result in the status screen being corrupted. This is because the firmware was trying to write a 3 digit number where there is no space on the LCD for it.

| BEFORE  |  AFTER  |
|---|---|
|  IMG_2676 |  IMG_2677 |

Change in memory:
Flash: -52 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -455,12 +455,12 @@ uint8_t lcdui_print_extruder(void) {
     lcd_space(1);
     if (MMU2::mmu2.get_current_tool() == MMU2::mmu2.get_tool_change_tool()) {
         lcd_putc('F');
-        lcd_putc(MMU2::mmu2.get_current_tool() == (uint8_t)MMU2::FILAMENT_UNKNOWN ? '?' : MMU2::mmu2.get_current_tool() + 1);
+        lcd_putc(MMU2::mmu2.get_current_tool() == (uint8_t)MMU2::FILAMENT_UNKNOWN ? '?' : MMU2::mmu2.get_current_tool() + '1');
         chars += 2;
     } else {
-        lcd_putc(MMU2::mmu2.get_current_tool() == (uint8_t)MMU2::FILAMENT_UNKNOWN ? '?' : MMU2::mmu2.get_current_tool() + 1);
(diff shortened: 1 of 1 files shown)


Firmware/ultralcd.cpp
```diff
@@ -441,24 +441,27 @@ void lcdui_print_percent_done(void)
...
+    uint8_t chars = 1;
+    lcd_space(1);
     if (MMU2::mmu2.get_current_tool() == MMU2::mmu2.get_tool_change_tool()) {
-        if (MMU2::mmu2.get_current_tool() == (uint8_t)MMU2::FILAMENT_UNKNOWN) {
-            chars = lcd_printf_P(_N(" F?"));
-        } else {
-            chars = lcd_printf_P(_N(" F%u"), MMU2::mmu2.get_current_tool() + 1);
-        }
+        lcd_putc('F');
+        lcd_putc(MMU2::mmu2.get_current_tool() == (uint8_t)MMU2::FILAMENT_UNKNOWN ? '?' : MMU2::mmu2.get_current_tool() + 1);
```
leptun: I have a feeling these should be + '1' instead of + 1.

gudnimg: You're correct. Fix added here [7e341bf](https://github.com/prusa3d/Prusa-Firmware/pull/3767/commits/7e341bf96c939f09988b0bf6fc012a1d84180598)

gudnimg:
Only thing left to test before this PR can be merged is to see Scenario 5 work:

```
// Scenario 5: "[nr1.]>[nr2.]"
//              [nr1.] ranges from 1 to 5.
//              [nr2.] ranges from 1 to 5.
//              Filament [nr1.] was loaded, but [nr2.] is currently being loaded via tool change
```
It only works when printing.

Firmware/ultralcd.cpp
gudnimg: Will rebase the PR later so the fixup will be a part of the first commit.

gudnimg:
Can confirm Scenario 5 works without issues :) So I'm opening the PR for review/merge.

DRracer (approved):
nice :+1:

PR #3772, opened 2022-11-20
Unload first before retrying

DRracer:
Should free the filament if accidentally held by the fsensor.

Also the unloading distance and speed have been greatly increased.

PFW-1406


Firmware/mmu2.cpp
```diff
@@ -735,6 +735,9 @@ void MMU2::CheckUserInput(){
         SERIAL_ECHOLN(btn);
         ResumeHotendTemp(); // Recover the hotend temp before we attempt to do anything else...
         Button(btn);
+        // at this spot it is necessary to move the E-motor if fsensor is pressed (or every time)
+        // Just plan the move, let the MMU take over when it is ready
+        HelpUnloadToFinda();
```
gudnimg: I don't agree with this. Specifically, I don't think this applies to FINDA didn't trigger error. Even on Selector or Idler couldn't home errors -> E-motor should not move.

gudnimg (approved):

PR #3775, opened 2022-11-23
Fix & optimize FW version parsing and checking

DRracer:
PFW-1433


3d-gussner (approved):
Tested with MK404 and works as expected.
The commented out serial output works as well and might help in the future.

PR #3779, opened 2022-11-23
Don't show full screen message meant for single color setup

gudnimg:
`load_filament_to_nozzle()` will show a more precise full screen message when an MMU is used. `loading_flag` is also not used with MMU in M701 (autoloading).


vintagepc (approved):

wavexx (approved):

wavexx:
@gudnimg can we add a comment on the mmu "enabled" branch to say that the "insert" message will be performed by M701 in that case? Or the other way around, that M701 doesn't show instructions by default. Either is fine.

Just so that it's clear for the next time.

gudnimg:
Yes I agree,  I’ll add it tomorrow 👍

gudnimg:
Added a comment

PR #3780, opened 2022-11-24
optimisation: Improve Stop Print menu

gudnimg:
Optimise the "Stop print" menu by using the menu API methods. **UI is visually unchanged.**

Before:

```
Program:  234486 bytes (89.4% Full)
(.text + .data + .bootloader)
 
Data:       5680 bytes (69.3% Full)
(.data + .bss + .noinit)
```

After:
```
Program:  234370 bytes (89.4% Full)
(.text + .data + .bootloader)
 
Data:       5680 bytes (69.3% Full)
(.data + .bss + .noinit)
```

Change in memory:
Flash: -116 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -5819,31 +5819,18 @@ void lcd_temp_model_cal()
 
 void lcd_sdcard_stop()
 {
+    // Show static message
+    lcd_puts_at_P(0, 0, _T(MSG_STOP_PRINT));
+    lcd_putc_at(0, 1, '\n');
 
-	lcd_puts_at_P(0, 0, _T(MSG_STOP_PRINT));
-	lcd_puts_at_P(2, 2, _T(MSG_NO));
-	lcd_puts_at_P(2, 3, _T(MSG_YES));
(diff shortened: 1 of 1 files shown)


leptun:
I have not tested this, but I don't think this is ok. To me it looks as if the menu is blocking. The old menu would allow the rest of the firmware to work while the prompt is displayed. Is it the case with this new one or does it prevent the printer from printing while it is displayed?

gudnimg:
> I have not tested this, but I don't think this is ok. To me it looks as if the menu is blocking. The old menu would allow the rest of the firmware to work while the prompt is displayed. Is it the case with this new one or does it prevent the printer from printing while it is displayed?

Just double-checked and you're right. If I wait long enough, the current move will finish and then the printer halts until the menu goes away.

gudnimg:
@leptun I've fixed the issue you mentioned while still reducing the flash memory consumption. I removed my changes to "Resume print" menu items as they are not really relevant here.

wavexx (approved):

3d-gussner (approved):
Tested on MK404 and works as expected

PR #3781, opened 2022-11-24
Fix an issue on multilang builds where first two characters are not rendered

gudnimg:


3d-gussner (approved):

DRracer (changes requested):
Good point, but we should make the text a MSG_TOOLCHANGE_COUNT and have it translated like the rest of our messages.

leptun:
I agree with David. The string should be placed in messages.c with ISTR and then be retrieved here with _T. I'm not sure I understand, but why was _O used here anyway? It looks like a string you might want translated, so I don't see the reason to have it with _O.

gudnimg:
> I agree with David. The string should be placed in messages.c with ISTR and then be retrieved here with _T. I'm not sure I understand, but why was _O used here anyway? It looks like a string you might want translated, so I don't see the reason to have it with _O.

The _O was a mistake on my part as I didnt test on multilang build. 

The menu item as I implemented was only supposed to be temporary/placeholder so I didnt really bother with translations.

DRracer:
@gudnimg since your temporary placeholder became the production solution, let's translate that string as well :wink:

DRracer (approved):

PR #3784, opened 2022-11-25
Optimize rendering MMU Error sensor line

DRracer:
CPUFLASH: -32B


gudnimg (approved):
Tested both a normal error screen and after letting the cool-down start during an error screen (and re-heat). In both cases the error screen renders correctly.

PR #3786, opened 2022-11-27
PFW-1423 UNLOAD MANUALLY: error persistency 2

gudnimg:
- Drop `Retry` button and add `Restart MMU` button.


vintagepc (commented):
I am not sure how this is supposed to fix the "FINDA vs EEPROM discrepancy" as reported in the JIRA ticket. 

Also, restarting the MMU feels like a crude hack. There is probably a better way to address the problem.

gudnimg:
> I am not sure how this is supposed to fix the "FINDA vs EEPROM discrepancy" as reported in the JIRA ticket.
> 
> Also, restarting the MMU feels like a crude hack. There is probably a better way to address the problem.

It is a hack. The MMU FW would need to be changed to make the “Retry” work. I’m not sure how easy that is to do.

gudnimg:
Converting this PR to draft for now, so it cant be merged. Im going to try to tackle the problem on the MMU FW side.

**Update:** I've got a working solution in the MMU FW by adding handling for Retry. Now just to add unit tests...

PR #3787, opened 2022-11-27
PFW-1448 Fix underextrusion + compensate load to nozzle extruder sequence for Extra Loading Distance

gudnimg:
The Extra Loading Distance parameter is configurable by the user. We need to compensate the hard-coded extruder sequence
such that it does not extrude too much or too little. Currently the firmware extrudes too little.

The problem is split in two:
* The current sequence is too short
* The sequence was made even shorter when the Extra Loading Distance was changed from 30mm to 5mm in development.

Where is this sequence used?

* M600: Loading filament
* M701: Loading filament
* `T?` gcode
* `Tc` gcode (single color print with MMU)
* LCD menu "Load to Nozzle"

### DIFF
--- a/Firmware/mmu2.cpp
+++ b/Firmware/mmu2.cpp
@@ -249,9 +249,9 @@ bool MMU2::VerifyFilamentEnteredPTFE()
     uint8_t fsensorState = 0;
     // MMU has finished its load, push the filament further by some defined constant length
     // If the filament sensor reads 0 at any moment, then report FAILURE
-    current_position[E_AXIS] += MMU2_EXTRUDER_PTFE_LENGTH + MMU2_EXTRUDER_HEATBREAK_LENGTH - MMU2_LOAD_DISTANCE_PAST_GEARS;
+    current_position[E_AXIS] += MMU2_EXTRUDER_PTFE_LENGTH + MMU2_EXTRUDER_HEATBREAK_LENGTH - (logic.ExtraLoadDistance() - MMU2_FILAMENT_SENSOR_POSITION);
     plan_buffer_line_curposXYZE(MMU2_LOAD_TO_NOZZLE_FEED_RATE);
-    current_position[E_AXIS] -= (MMU2_EXTRUDER_PTFE_LENGTH + MMU2_EXTRUDER_HEATBREAK_LENGTH - MMU2_LOAD_DISTANCE_PAST_GEARS);
+    current_position[E_AXIS] -= (MMU2_EXTRUDER_PTFE_LENGTH + MMU2_EXTRUDER_HEATBREAK_LENGTH - (logic.ExtraLoadDistance() - MMU2_FILAMENT_SENSOR_POSITION));
     plan_buffer_line_curposXYZE(MMU2_LOAD_TO_NOZZLE_FEED_RATE);
 
     while(blocks_queued())
@@ -840,7 +840,7 @@ void MMU2::execute_extruder_sequence(const E_Step *sequence, uint8_t steps) {
 void MMU2::execute_load_to_nozzle_sequence() {
     st_synchronize();
     // Compensate for configurable Extra Loading Distance
-    current_position[E_AXIS] -= MMU2_LOAD_DISTANCE_PAST_GEARS;
+    current_position[E_AXIS] -= (logic.ExtraLoadDistance() - MMU2_FILAMENT_SENSOR_POSITION);
     execute_extruder_sequence((const E_Step *)load_to_nozzle_sequence, sizeof(load_to_nozzle_sequence) / sizeof (load_to_nozzle_sequence[0]));
 }
 
--- a/Firmware/mmu2.cpp
+++ b/Firmware/mmu2.cpp
@@ -355,7 +355,7 @@ bool MMU2::tool_change(char code, uint8_t slot) {
 
     case 'c': {
         waitForHotendTargetTemp(100, []{});
-        execute_load_to_nozzle_sequence((const E_Step *)load_to_nozzle_sequence, sizeof(load_to_nozzle_sequence) / sizeof (load_to_nozzle_sequence[0]));
+        execute_load_to_nozzle_sequence();
     } break;
     }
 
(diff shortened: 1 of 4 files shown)


Firmware/mmu2.cpp
```diff
@@ -83,15 +83,13 @@ static constexpr E_Step ramming_sequence[] PROGMEM = {
     { -35.0F,   2000.0F / 60.F},
 };
 
-static constexpr E_Step load_to_nozzle_sequence[] PROGMEM = { 
-    { 10.0F,  810.0F / 60.F}, // feed rate = 13.5mm/s - Load fast until filament reach end of nozzle
-    { 25.0F,  198.0F / 60.F}, // feed rate = 3.3mm/s  - Load slower once filament is out of the nozzle
+static constexpr E_Step load_to_nozzle_sequence[] PROGMEM = {
+    { MMU2_EXTRUDER_PTFE_LENGTH,       810.0F / 60.F}, // feed rate = 13.5mm/s - Load fast while not at heatbreak
+    { MMU2_EXTRUDER_HEATBREAK_LENGTH,  198.0F / 60.F}, // feed rate = 3.3mm/s  - Load slower once filament reaches heatbreak
```
vintagepc: Is this correct? Seems like the magnitude of the two parameters has changed here, i.e. we've gone from a shorter value first to a longer value in the first entry...

gudnimg: To be honest I’m not sure if this is correct. It is my understand that “Load to nozzle” should leave the filament at the very end of the nozzle (just far enough to not extrude out of the nozzle). But if this is incorrect, then I need some definition for the lengths. 

The Prusa testing team reported that Load to Nozzle was extruding too little. So I am assuming here thats because 35mm extruder sequence is too short. 

In the loading test we added last week, we use the full distance as I’m proposing in this PR. Maybe thats not correct either.

gudnimg:
A nice way to test the load length is running a single color MMU print.

I'm seeing the current sequence (currently in `MK3` branch) is wayyy too short. Close to half of the purge line is missing.

```
static constexpr E_Step load_to_nozzle_sequence[] PROGMEM = {
    { 10.0F,       810.0F / 60.F}, // feed rate = 13.5mm/s
    { 25.0F,  198.0F / 60.F}, // feed rate = 3.3mm/s
};
```
----
With the changes I propose in this PR, the purge line is missing ~1/4 of the first line.
```
static constexpr E_Step load_to_nozzle_sequence[] PROGMEM = {
    { 42.3F,       810.0F / 60.F}, // feed rate = 13.5mm/s - Load fast while not at heatbreak
    { 17.7F,  198.0F / 60.F}, // feed rate = 3.3mm/s  - Load slower once filament reaches heatbreak
};
```

gudnimg:
So far the result looks good.

IMG_2742

gudnimg:
Only thing left to check is to manually change the extra loading distance and check that the purge line looks exactly the same.

To test:  `M708 A0x0b X30` to set Extra Loading Distance to 30mm (current default is 5)

gudnimg:
Testing this more (after getting ☕ ), I realised the First Layer Calibration doesn't use this sequence, so the result I posted earlier today is not relevant here; I deleted the comment. 

But meanwhile I noticed the First Layer Calibration purge line is not compensated for the extra loading distance. Another thing to add to our TODO list :)

**Running a single material print, this is what the purge line looks like currently:**


The Gcode (Default PrusaSlicer 2.5)
```
Tc
; purge line
G1 X55 F2000
G1 Z0.3 F1000
G92 E0
G1 X240 E25 F2200
G1 Y-2 F1000
G1 X55 E25 F1400
G1 Z0.2 F1000
M73 Q0 S402
G1 X5 E4 F1000
```

As can be seen in the Gcode, we move the X-axis by 55mm, and during these 55mm there is no extrusion happening. Because after `Tc` is finished running, the firmware is not pushing any filament.

IMG_2746


Compare this with a multicolor purge line
```
T0
; initial load
G1 X55 E29 F1073
G1 X5 E29 F1800
G1 X55 E8 F2000
G1 Z0.3 F1000
G92 E0
G1 X240 E25 F2200
G1 Y-2 F1000
G1 X55 E25 F1400
G1 Z0.2 F1000
G1 X5 E4 F1000
G92 E0
```

gudnimg:
I think this PR is now ready for review for merge.

Not sure what we can do about the missing first bits of the purge line. I think that issue is a problem on the Slicer side.

leptun (approved):
Tested on MK2.5_MMU2. The load to nozzle distance is correct regardless of `M708 A0x0b Xxx` setting. I tried up to `M708 A0x0b X60` and it still worked correctly when I loaded to nozzle from the LCD.

gudnimg:
@leptun I just tested this PR again on my MK3S+ and can confirm the purge line looks as expected 👍

Firmware/mmu2.cpp
```diff
@@ -839,6 +837,13 @@ void MMU2::execute_extruder_sequence(const E_Step *sequence, uint8_t steps) {
     }
 }
 
+void MMU2::execute_load_to_nozzle_sequence() {
+    st_synchronize();
+    // Compensate for configurable Extra Loading Distance
+    current_position[E_AXIS] -= (logic.ExtraLoadDistance() - MMU2_FILAMENT_SENSOR_POSITION);
```
DRracer, line 843: /nitpick: if extra load distance is set to some insane value, this subtraction can cause the extruder to actually unload the filament :wink: .
But I assume when someone is tweaking the extra load distance he probably knows what he is doing.

leptun: Yeah, I know. When I picked the value for testing I had to make sure not to exceed that. I don't think it's necessary to add an extra check for that since only knowlageable users (or print profile makers) should use the register anyway.

DRracer: I agree, that register should be primarily set with a hard-coded per-printer-type value (or may be some mods may need to tweak this value). I don't see a reason why a user would want to tweak extra loading distance per G-code or even at runtime.

DRracer (approved):

PR #3790, opened 2022-11-30
Fix yes no for Nozzle change

3d-gussner:


leptun (approved):
checked with mk404. Seems to be working correctly

PR #3793, opened 2022-12-03
PFW-1453 Stop Print: Don't unload if FINDA is not triggered

gudnimg:
If FINDA does not detect filament, then U0 command doesn't do anything.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -5863,14 +5863,14 @@ void print_stop()
         fanSpeed = 0;
     }
 
-    if (MMU2::mmu2.Enabled())
+    if (MMU2::mmu2.Enabled() && MMU2::mmu2.FindaDetectsFilament())
     {
         if (isPrintPaused)
         {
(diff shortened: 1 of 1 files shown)


Firmware/ultralcd.cpp
```diff
@@ -5863,14 +5863,14 @@ void print_stop()
...
 
-    if (MMU2::mmu2.Enabled())
+    if (MMU2::mmu2.Enabled() && MMU2::mmu2.FindaDetectsFilament())
     {
         if (isPrintPaused)
         {
             // Restore temperature saved in ram after pausing print
             restore_extruder_temperature_from_ram();
         }
-        MMU2::mmu2.unload(); //M702 C
```
gudnimg, line 5873: Note, I removed the C here because it is not longer a parameter in M702

vintagepc (approved):

wavexx (approved):

leptun (approved):

gudnimg:
3 Approvals 💪 merging

PR #3794, opened 2022-12-03
Fix undefined pointer in `M862.2 Q` and `M862.3 Q`

### LINKED ISSUE #3788: 3.13.0 returns 0 to "M862.3 Q" - Get printer type
MK3S MMU2S on 3.13.0-Alpha1 
Send "M862.3 Q" over serial. It should not return 0, but that's what seems to be happening

gudnimg:
`fSetMmuMode()` needs to be called
in order for the commands to work:
* M862.2 Q
* M862.3 Q

Fixes #3788 

Change in memory:
Flash: -70 bytes
SRAM: -4 bytes

### DIFF
--- a/Firmware/Configuration.cpp
+++ b/Firmware/Configuration.cpp
@@ -5,6 +5,3 @@ const uint16_t _nPrinterType PROGMEM=PRINTER_TYPE;
 const char _sPrinterName[] PROGMEM=PRINTER_NAME;
 const uint16_t _nPrinterMmuType PROGMEM=PRINTER_MMU_TYPE;
 const char _sPrinterMmuName[] PROGMEM=PRINTER_MMU_NAME;
-
-uint16_t nPrinterType;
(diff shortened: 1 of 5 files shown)


vintagepc:
Feels like there's an opportunity to make the code more resilient in general by adding e.g. a null pointer and other checks?

leptun:
@gudnimg I pushed a change which gets rid of the global nPrinterType and sPrinterName. This avoids the null pointer issue highlighted by @vintagepc and also saves quite a bit of resources (better optimization when inlined in this case).

leptun (approved):
Tested my changes in MK404. I'd like one more approval besides mine since I did significant changes before we merge this.

3d-gussner (approved):
Tested on MK404

- [X] MK3S
- [X] MK3S + MMU2
- [X] MK3
- [X] MK3 + MMU2
- [X] MK2.5/S

PR #3795, opened 2022-12-03
Optimisation: Simplify parsing M862

gudnimg:
This gets rid of float conversion.
We just need to check the 6th character when M862 is detected.

Change in memory:
Flash: -20 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -8011,10 +8011,9 @@ Sigma_Exit:
 	
     */
     case 862: // M862: print checking
-          float nDummy;
-          uint8_t nCommand;
-          nCommand=(uint8_t)(modff(code_value(),&nDummy)*10.0+0.5);
-          switch((ClPrintChecking)nCommand)
+    {
+        ClPrintChecking nCommand = static_cast<ClPrintChecking>(strtol(strchr_pointer+5, NULL, 10));
+        switch(nCommand)
                {
                case ClPrintChecking::_Nozzle:     // ~ .1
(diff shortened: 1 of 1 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -8034,10 +8034,9 @@ SERIAL_PROTOCOLPGM("\n\n");
 	
     */
     case 862: // M862: print checking
-          float nDummy;
-          uint8_t nCommand;
-          nCommand=(uint8_t)(modff(code_value_float(),&nDummy)*10.0+0.5);
-          switch((ClPrintChecking)nCommand)
+    {
+        ClPrintChecking nCommand = static_cast<ClPrintChecking>(strtol(strchr_pointer+5, NULL, 10));
```
wavexx: We need to check the string length here before we do strchr_pointer+5. We only know the string is at least 4 chars long including null.

wavexx (changes requested):

Firmware/Marlin_main.cpp
gudnimg: @wavexx here is my attempt to address this issue: [0bd85cc](https://github.com/prusa3d/Prusa-Firmware/pull/3795/commits/0bd85cc23ebb3b34ea2f985fc06957da546e903e) It looks like checking the string length or looking for a null byte consumes much more flash than we are trying to save. But with this commit we still save 20 bytes. 👍

DRracer (approved):

wavexx (approved):

PR #3797, opened 2022-12-04
PFW-1454 Send button to MMU on MMU SELFTEST FAILED screen

gudnimg:
Fixes an issue where the button on the MMU error screen doesn't do anything when MMU SELFTEST FAILED is rendered


vintagepc (approved):
Ah, this probably broke when we turned the single error back into one specific to the failed channel

PR #3805, opened 2022-12-06
MK3: Fix `M862.4` with [strict] mode

3d-gussner:
Max 8 flavor versions
Minor changes: easier to un-comment serial debug

Tested with MK404
Settings -> HW Setup -> Checks -> Firmware set to `[warn]`
- [X] Send `M862.4 P3.11.0` nothing happens as FW is newer
- [X] Send `M862.4 P3.13.0-RC1` warning shown on LCD to continue with print
- [X] Send `M862.4 P3.13.0` warning shown on LCD to continue with print

Settings -> HW Setup -> Checks -> Firmware set to `[strict]`
- [X] Send `M862.4 P3.11.0` nothing happens as FW is newer
- [X] Send `M862.4 P3.13.0-RC1` warning shown on LCD and print canceled 
- [X] Send `M862.4 P3.13.0` warning shown on LCD and print canceled

ToDo's:
- [ ] Cherry-pick to MK3_3.12

### DIFF
--- a/Firmware/util.cpp
+++ b/Firmware/util.cpp
@@ -366,20 +366,19 @@ uint8_t mCompareValue(uint16_t nX, uint16_t nY) {
 }
 
 void fw_version_check(const char *pVersion) {
-    uint16_t aVersion[4];
-    uint8_t nCompareValueResult;
-
     if (oCheckVersion == ClCheckVersion::_None)
         return;
+
+    uint16_t aVersion[4];
+    uint8_t nCompareValueResult;
     parse_version(pVersion, aVersion);
     nCompareValueResult = mCompareValue(aVersion[0], eeprom_read_word((uint16_t *)EEPROM_FIRMWARE_VERSION_MAJOR)) << 6;
     nCompareValueResult += mCompareValue(aVersion[1], eeprom_read_word((uint16_t *)EEPROM_FIRMWARE_VERSION_MINOR)) << 4;
     nCompareValueResult += mCompareValue(aVersion[2], eeprom_read_word((uint16_t *)EEPROM_FIRMWARE_VERSION_REVISION)) << 2;
     nCompareValueResult += mCompareValue(aVersion[3], eeprom_read_word((uint16_t *)EEPROM_FIRMWARE_VERSION_FLAVOR));
-    if (nCompareValueResult == COMPARE_VALUE_EQUAL)
+    if (nCompareValueResult <= COMPARE_VALUE_EQUAL)
         return;
-    if ((nCompareValueResult < COMPARE_VALUE_EQUAL) && (oCheckVersion == ClCheckVersion::_Warn || oCheckVersion == ClCheckVersion::_Strict))
-        return;
+
 /*
     SERIAL_ECHO_START;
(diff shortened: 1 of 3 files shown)


Firmware/util.cpp
```diff
@@ -378,27 +378,28 @@ void fw_version_check(const char *pVersion) {
...
+    SERIAL_ECHO_START;
+    SERIAL_ECHOLNPGM("Printer FW version differs from the G-code ...");
+    SERIAL_ECHOPGM("actual  : ");
+    SERIAL_ECHO(eeprom_read_word((uint16_t *)EEPROM_FIRMWARE_VERSION_MAJOR));
+    SERIAL_ECHO('.');
+    SERIAL_ECHO(eeprom_read_word((uint16_t *)EEPROM_FIRMWARE_VERSION_MINOR));
+    SERIAL_ECHO('.');
+    SERIAL_ECHO(eeprom_read_word((uint16_t *)EEPROM_FIRMWARE_VERSION_REVISION));
+    SERIAL_ECHO('.');
+    SERIAL_ECHO(eeprom_read_word((uint16_t *)EEPROM_FIRMWARE_VERSION_FLAVOR));
```
vintagepc, line 392: ... why are we reading the version from EEPROM instead of PROGMEM? Doesn't this have a risk of being wrong?

Firmware/util.cpp
```diff
@@ -378,27 +378,28 @@ void fw_version_check(const char *pVersion) {
     nCompareValueResult += mCompareValue(aVersion[3], eeprom_read_word((uint16_t *)EEPROM_FIRMWARE_VERSION_FLAVOR));
     if (nCompareValueResult == COMPARE_VALUE_EQUAL)
         return;
-    if ((nCompareValueResult < COMPARE_VALUE_EQUAL) && oCheckVersion == ClCheckVersion::_Warn)
+    if ((nCompareValueResult < COMPARE_VALUE_EQUAL) && (oCheckVersion == ClCheckVersion::_Warn || oCheckVersion == ClCheckVersion::_Strict))
```
wavexx: It looks like to me the original idea of "strict" was "exactly the reversion", at least to the code.
I think we can change this to:

```
if (nCompareValueResult <= COMPARE_VALUE_EQUAL)
```

since we handle _None above, and everything else is handled later.

wavexx: We also have exactly the same

```
if (nGcodeLevel == (uint16_t)GCODE_LEVEL)
```

that we could change to:

if (nGcodeLevel <= (uint16_t)GCODE_LEVEL)

at line 424 (can't attach a comment)

wavexx (changes requested):

wavexx:
Was this re-tested?

3d-gussner:
> Was this re-tested?

Retested on MK404:
All conditions working correctly.

- [X] Strict Only triggers when gcode sends newer firmware version
- [X] Warn Only triggers when gcode sends newer firmware version
- [X] None ignores any `M862.4 Px.y.z-abc` gcode

wavexx (approved):

PR #3814, opened 2022-12-09
TM: reset fan measuring state when changing speed

### LINKED ISSUE #3791: [BUG]   3.12.0-RC1-5576 PRINT FAN ERROR
**Printer type**  MK3S
**Printer firmware version** -3.12.0-RC1-5576
No modifications to hotend. Have added PrusaLink.

**SD card or USB/Octoprint**
 N/A

**Describe the bug**
  "PRINT FAN ERROR" message when running temperature calibration.
  
**To Reproduce**
Run calibration from menu.
  
**Expected behavior**
 Complete calibration
  
**G-code**
 n/a

**Video**
No video but it is a consistent error in an 18-19 deg C room.

wavexx:
Ensure that fan checks are reset and use the new speed at each step of the calibration.

This also gives extra time to the fan to ramp-up from a cold start, when a fancheck could previously start right *after* the speed change.

Should fix #3791

- [x] cherry-pick to MK3_3.12 branch https://github.com/prusa3d/Prusa-Firmware/pull/3825

### DIFF
--- a/Firmware/temperature.cpp
+++ b/Firmware/temperature.cpp
@@ -2630,6 +2630,11 @@ namespace temp_model_cal {
 // set current fan speed for both front/backend
 static __attribute__((noinline)) void set_fan_speed(uint8_t fan_speed)
 {
+#if (defined(EXTRUDER_0_AUTO_FAN_PIN) && EXTRUDER_0_AUTO_FAN_PIN > -1)
+    // reset the fan measuring state due to missing hysteresis handling on the checking side
+    fan_measuring = false;
+    extruder_autofan_last_check = _millis();
+#endif
     fanSpeed = fan_speed;
 #ifdef FAN_SOFT_PWM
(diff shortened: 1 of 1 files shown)


wavexx:
Fan hysteresis strikes again. The whole code surrounding the fan speed and checks seems super-sketchy.

Not only we don't account for rpm hysteresis, but every part that sets the fan speed is currently working around the fan checks...

3d-gussner:
Is there any way to test this issue and then the fix?

wavexx:
I think it's hard to reproduce this, since we never had this. You need a fan that's slow to start in addition to good timing.

3d-gussner (approved):
I had few FAN ERRORs at 18-19°C ambient temperatures during testing the temp model cal on real printer. With this PR I had until now zero issues.

DRracer (approved):

PR #3816, opened 2022-12-10
Optimisation: simplify some of the LCD fullscreen code

gudnimg:
Removed only lines of code that shouldn't have any effect on the UI for the user. Mostly just redundant code.

Change in memory:
Flash: -62 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -2990,7 +2990,6 @@ static inline bool pgm_is_interpunction(const char *c_addr)
  */
 static const char* lcd_display_message_fullscreen_nonBlocking_P(const char *msg)
 {
-    lcd_set_cursor(0, 0);
     const char *msgend = msg;
     bool multi_screen = false;
     for (uint8_t row = 0; row < LCD_HEIGHT; ++ row) {
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -3190,9 +3190,6 @@ uint8_t lcd_show_multiscreen_message_with_choices_and_wait_P(
             if (allow_timeouting && _millis() - previous_millis_cmd > LCD_TIMEOUT_TO_STATUS) {
                 return LCD_BUTTON_TIMEOUT;
             }
-            manage_heater();
-            manage_inactivity(true);
-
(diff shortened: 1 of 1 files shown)


Firmware/ultralcd.cpp
```diff
@@ -2992,23 +2992,15 @@ static const char* lcd_display_message_fullscreen_nonBlocking_P(const char *msg)
 {
     lcd_set_cursor(0, 0);
     const char *msgend = msg;
-    uint8_t row = 0;
     bool multi_screen = false;
-    for (; row < 4; ++ row) {
-        while (pgm_is_whitespace(msg))
```
leptun, line 2998: I'm not sure it's ok to remove this. What if the word on the previous line finishes with the last character on the last column of the screen? Wouldn't that cause the next line to have a useless space at the beginning?

gudnimg: You're correct. See images below:

`lcd_display_message_fullscreen_P(_i("Averyveryverlongword Press the knob to pr eheat nozzle and continue."));`

MK3 branch test:
IMG_2731


PR branch test:
IMG_2729

gudnimg: Fixed in aca9e79

Firmware/ultralcd.cpp
```diff
@@ -3198,9 +3192,6 @@ uint8_t lcd_show_multiscreen_message_with_choices_and_wait_P(
             if (allow_timeouting && _millis() - previous_millis_cmd > LCD_TIMEOUT_TO_STATUS) {
                 return LCD_BUTTON_TIMEOUT;
             }
-            manage_heater();
-            manage_inactivity(true);
```
DRracer, line 3202: Why was this removed? Is lcd_show_multiscreen_message_with_choices_and_wait_P always accompanied with some other ways of managing heaters and inactivity?

gudnimg: There is a call to `delay_keep_alive(50)` above. It calls these functions. There is no need to do it twice, that just increases code size.

DRracer: :+1:

DRracer (approved):

PR #3818, opened 2022-12-12
TM: Provide full defaults for model data and enable it by default

wavexx:
This PR provides average values for all fan levels for the MK3/MK3S, allowing the model **defaulting to enabled**. Since the model cannot be disabled by the UI, this greatly reduces all corner cases that need to be handled for a safe default.

The wizard and the model calibration menu have been switched to force self-check, to ensure the hotend is always tested at every step of the setup process. A failed calibration will now always restore the full previous state, returning the printer to the default (model enabled).

Thermal errors during the hotend checks are now handled by interrupting the selftest early.

Default values are likely suboptimal for fan values, as aging can be wild on the blowers. The wizard will complain about the thermal model calibration not being performed to avoid prints failing prematurely at the second layer when the fan is switched on for the first time.

If the wizard is interrupted due to a failed calibration, using the "calibration -> Temp Model Cal." will now clear the "temperature calibration required" message correctly.

The thermal model can thus now only be disabled via gcode:

```
D3 Ax0ff7 X 01 ; set the printer as fully calibrated if the wizard can't be completed
M310 S0 ; model off
M500 ; save
```

- [ ] Create cherry-pick PR to MK3_3.12 branch

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -8657,16 +8657,6 @@ Sigma_Exit:
     }
     break;
 
-    /*!
-    ### M999 - Restart after being stopped <a href="https://reprap.org/wiki/G-code#M999:_Restart_after_being_stopped_by_error">M999: Restart after being stopped by error</a>
-    @todo Usually doesn't work. Should be fixed or removed. Most of the time, if `Stopped` it set, the print fails and is unrecoverable.
-    */
-    case 999:
-      Stopped = false;
-      lcd_reset_alert_level();
-//@@TODO      gcode_LastN = Stopped_gcode_LastN;
-      FlushSerialRequestResend();
-    break;
 	/*!
 	#### End of M-Commands
     */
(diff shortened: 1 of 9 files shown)


3d-gussner (approved):
Tested on real MK3S with/without MMU2 hardware and it works.
During print at some point (fan kicked in) the default values reported thermal anomalies, which is okay.
After running the `Calibration -> Temp model cal. ` and/or the `Wizard` (including this calibration) the print went well with out any warnings.

I have been monitoring both `default` and `calibrated` using `M310 B0 W0.1` (no beeper and very low warning  value) to report the temp model values during these tests. And as expected the default weren't that good as the calibrated one. 
 

On MK404 (not updated yet for temp model) it blocked the access to the printer as from start up the temp model reported a TM ERROR. This can be solved by https://github.com/prusa3d/Prusa-Firmware/pull/3822 + one minor change via serial gcode commands.

This PR will work on stock Prusa MK3/MK3S but may fail modded printers, that's why it is important to merge the  https://github.com/prusa3d/Prusa-Firmware/pull/3822 + one minor change as well

DRracer (approved):

PR #3819, opened 2022-12-12
If an old firmware is detected, make sure the protocol is actually stopped

gudnimg:
If `StopKeepPowered()` has been called, the Printer's MMU protocol should shut up until it is started again :) `CommunicationTimeout` was triggered in this state and would restart the MMU protocol so you would repeatedly get more S0 queries.

Fixes:
* Fix `ok\n` detection from old MMU FW. The `\n` is actually never seen, probably consumed by the MMU protocol parsing. This was accidently creating `ProtocolError` (which will force send S0 again).
* If `ok` is detected from the old MMU FW, report `VersionMismatch` immediately to let the user know of the problem.
* Don't report `CommunicationTimeout` error if the protocol was stopped with `StopKeepPowered()`. 

----

Change in memory:
Flash: -108 bytes
SRAM: 0 bytes

----

Steps I used to test this:

1. Flash old FW 1.0.6 on the MMU2S
2. Open Serial port to see the logs
3. S0 is sent to MMU
4. MMU responds with "ok"
5. => Expected behavior is a VersionMismatch error and MMU serial port is disabled
6. => Actual behavior is Printer sends S0 endlessly, and MMU responds with multiple "ok". Essentially floods the serial port.

### DIFF
--- a/Firmware/mmu2_protocol_logic.cpp
+++ b/Firmware/mmu2_protocol_logic.cpp
@@ -124,8 +124,6 @@ struct OldMMUFWDetector {
             return State::MatchingPart;
         } else if(ok == 1 && c == 'k'){
             ++ok;
-            return State::MatchingPart;
-        } else if(ok == 2 && c == '\n'){
             return State::Matched;
         }
         return State::SomethingElse;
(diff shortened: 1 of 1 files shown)


Firmware/mmu2_protocol_logic.cpp
```diff
@@ -796,7 +793,6 @@ StepStatus ProtocolLogic::Step() {
         break;
     case VersionMismatch:
         LogError(PSTR("Version mismatch"));
-        Stop(); // cannot continue
```
gudnimg, line 799: **Note:** Removed this call because it's redundant. `Stop()` is called via `StopKeepPowered()` when the `VersionMismatch` error is reported one level higher.

Firmware/mmu2_protocol_logic.cpp
```diff
@@ -124,8 +124,6 @@ struct OldMMUFWDetector {
             return State::MatchingPart;
         } else if(ok == 1 && c == 'k'){
             ++ok;
-            return State::MatchingPart;
-        } else if(ok == 2 && c == '\n'){
```
DRracer, line 128: removing this part is a bit speculative, but since we don't use any lowercase letters in the protocol yet, it seems to be safe

DRracer (commented):
looks correct to me

leptun (approved):
Works nicely with an older v2 firmware on the MK2.5

PR #3821, opened 2022-12-13
Add 'Yes' and 'No' choices to messages like: "G-code sliced for a different printer type. Continue?"

gudnimg:
Fixes a minor thing that annoys me personally. If I get the this screen accidentally, then clicking the knob can't cancel the print.

This PR adds 'Yes' and 'No' choices for all the warning screens and we save flash too :)

Change in memory:
Flash: -98 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -3212,7 +3212,8 @@ uint8_t lcd_show_multiscreen_message_with_choices_and_wait_P(
         for (uint8_t i = 0; i < 100; ++i) {
             delay_keep_alive(50);
             if (allow_timeouting && _millis() - previous_millis_cmd > LCD_TIMEOUT_TO_STATUS) {
-                return LCD_BUTTON_TIMEOUT;
+                current_selection = LCD_BUTTON_TIMEOUT;
+                goto exit;
             }
             manage_heater();
             manage_inactivity(true);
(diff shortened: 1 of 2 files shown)


wavexx (approved):
I hated the check so I disabled it. This looks a much better solution :))

DRracer (changes requested):
The message timeout option is missing from the new implementation. It should be kept.

gudnimg:
> The message timeout option is missing from the new implementation. It should be kept.

`extern uint8_t lcd_show_fullscreen_message_yes_no_and_wait_P(const char *msg, bool allow_timeouting = true, uint8_t default_selection = LCD_MIDDLE_BUTTON_CHOICE);`

The timeout is on by default. See `allow_timeouting`. The timeout is currently set to 30 seconds which the same value as before.

```
#define LCD_TIMEOUT_TO_STATUS 30000ul //!< Generic timeout to status screen in ms, when no user action.
if (allow_timeouting && _millis() - previous_millis_cmd > LCD_TIMEOUT_TO_STATUS) {
    return LCD_BUTTON_TIMEOUT;
}
```

gudnimg:
Timeout seems to work 👍 I would approve if this wasn't my PR :)

leptun (approved):

Firmware/ultralcd.cpp
```diff
@@ -3265,6 +3263,11 @@ uint8_t lcd_show_multiscreen_message_with_choices_and_wait_P(
             lcd_show_choices_prompt_P(current_selection, first_choice, second_choice, second_col, third_choice);
         }
     }
+exit:
```
DRracer, line 3266: it would be better to use some RAII pattern, but the compiler makes the jmp instructions anyway :wink: , so I don't complain

DRracer (approved):
approving now

PR #3822, opened 2022-12-14
MK3: TM: Prevent lockout on invalid model values

### LINKED ISSUE #3630: [BUG] TM errors without a running print cannot be resumed
**Printer type** - n/a
**Printer firmware version** -  3.12+

**MMU upgrade** - n/a

**SD card or USB/Octoprint**
n/a

**Describe the bug**
If a Temperature Model error is triggered when there is no current print, the printer enters the "Stopped" state (as it would while printing). But since there's no print, it's impossible to "resume".
  
**To Reproduce**
`M310 S1 E0.01` will immediately enter this state.

**Expected behavior**
It's correct that a thermal error should still prevent the serial to work, just like while printing.

This can be actually _worse_ than a normal thermal error, since it means that the heater managed to trigger a problem not only while being preheated, but also potentially while being off.

In such case getting stuck in a "THERMAL ERROR - PLEASE RESET" like other thermal features might actually be good thing to do.

wavexx:
If the model is changed to invalid values (by manually running ``M310 C0.01`` for example) the printer will enter a continuous state of thermal error where 1) the LCD is unusable and 2) the serial is disabled. This cannot be undone except for a factory reset.

Allow again to use the menus after a thermal error. Since the thermal model is managing the heaters, the user can't worsen the situation. This can be used to run gcode from the SD.

Allow M310 to bypass the serial lock. M310 doesn't control heaters, and can't be used to bypass the stopped state, so it's safe to run in an emergency situation to reset the model.

~This is still incomplete though: if the thermal anomaly is constantly triggering (as happens with wrong values), even though the menus are unlocked, the error will continuously force redisplay, preventing you from doing anything.~

- [X] Create cherry-pick PR to [MK3_3.12 branch](https://github.com/prusa3d/Prusa-Firmware/pull/3845)

### DIFF
--- a/Firmware/cmdqueue.cpp
+++ b/Firmware/cmdqueue.cpp
@@ -421,10 +421,10 @@ void get_command()
 			  }
 
 			  // Don't parse N again with code_seen('N')
-			  cmdbuffer[bufindw + CMDHDRSIZE] = '$';
+			  *cmd_head = '$';
 		}
         // if we don't receive 'N' but still see '*'
-        if ((*cmd_head != 'N') && (cmdbuffer[bufindw + CMDHDRSIZE] != '$') && (strchr(cmd_start, '*') != NULL))
+        if ((*cmd_head != 'N') && (*cmd_head != '$') && (strchr(cmd_start, '*') != NULL))
         {
             SERIAL_ERROR_START;
             SERIAL_ERRORRPGM(_n("No Line Number with checksum, Last Line: "));////MSG_ERR_NO_LINENUMBER_WITH_CHECKSUM
--- a/Firmware/cmdqueue.cpp
+++ b/Firmware/cmdqueue.cpp
@@ -366,22 +366,24 @@ void get_command()
         comment_mode = false; //for new command
         return;
       }
-      cmdbuffer[bufindw+serial_count+CMDHDRSIZE] = 0; //terminate string
+      cmdbuffer[bufindw+serial_count+CMDHDRSIZE] = 0; // terminate string
+      char* cmd_head = cmdbuffer+bufindw+CMDHDRSIZE; // current command pointer
+      char* cmd_start = cmd_head; // pointer past the line number (if any)
+
       if(!comment_mode){
-		  
-		  long gcode_N = -1;
+		  long gcode_N = -1; // seen line number
 
 		  // Line numbers must be first in buffer
-
 		  if ((strstr_P(cmdbuffer+bufindw+CMDHDRSIZE, PSTR("PRUSA")) == NULL) &&
-			  (cmdbuffer[bufindw+CMDHDRSIZE] == 'N')) {
+			  (*cmd_head == 'N')) {
 
-			  // Line number met. When sending a G-code over a serial line, each line may be stamped with its index,
-			  // and Marlin tests, whether the successive lines are stamped with an increasing line number ID
-			  gcode_N = (strtol(cmdbuffer+bufindw+CMDHDRSIZE+1, NULL, 10));
-			  if(gcode_N != gcode_LastN+1 && (strstr_P(cmdbuffer+bufindw+CMDHDRSIZE, PSTR("M110")) == NULL) ) {
-				  // M110 - set current line number.
-				  // Line numbers not sent in succession.
+			  // Line number met: decode the number, then move cmd_start past all spaces.
+			  gcode_N = (strtol(cmd_head+1, &cmd_start, 10));
+			  while (*cmd_start == ' ') ++cmd_start;
+
+			  // Test whether the successive lines are stamped with an increasing line number ID.
+			  if(gcode_N != gcode_LastN+1 && strncmp_P(cmd_start, PSTR("M110"), 4)) {
+				  // Line numbers not sent in succession and M110 not seen.
 				  SERIAL_ERROR_START;
 				  SERIAL_ERRORRPGM(_n("Line Number is not Last Line Number+1, Last Line: "));////MSG_ERR_LINE_NO
 				  SERIAL_ERRORLN(gcode_LastN);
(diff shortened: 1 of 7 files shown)


3d-gussner (changes requested):
Please review and add https://github.com/wavexx/Prusa-Firmware/pull/10

Tested this on MK404 which does directly into TM Error.
- Sending `M310 S0` :heavy_check_mark: 
  - But after reset, which is the only option printer is still locked
    - With the PR above I am able to disable the temp model and save it, so next time it reboots it stays disabled.

Firmware/cmdqueue.cpp
```diff
@@ -422,25 +424,24 @@ void get_command()
 			  cmdbuffer[bufindw + CMDHDRSIZE] = '$';
 		}
         // if we don't receive 'N' but still see '*'
-        if ((cmdbuffer[bufindw + CMDHDRSIZE] != 'N') && (cmdbuffer[bufindw + CMDHDRSIZE] != '$') && (strchr(cmdbuffer+bufindw+CMDHDRSIZE, '*') != NULL))
+        if ((*cmd_head != 'N') && (cmdbuffer[bufindw + CMDHDRSIZE] != '$') && (strchr(cmd_start, '*') != NULL))
```
DRracer: Why wasn't `cmdbuffer[bufindw + CMDHDRSIZE]` also changed to `*cmd_head`? I don't see a difference yet...

Firmware/cmdqueue.cpp
```diff
@@ -422,25 +424,24 @@ void get_command()
...
+        if(strcmp_P(cmd_start, PSTR("M112")) == 0)
           kill(MSG_M112_KILL, 2);
+
         // Handle the USB timer
-        if ((strchr_pointer = strchr(cmdbuffer+bufindw+CMDHDRSIZE, 'G')) != NULL) {
-            if (!IS_SD_PRINTING) {
-                usb_timer.start();
-            }
-        }
+        if ((*cmd_start == 'G') && !IS_SD_PRINTING)
```
DRracer: suggest parentheses around `IS_SD_PRINTING` as it is a macro and unintentional change of it may cause hard-to-debug issues in the future

DRracer (approved):
Looks correct to me, just 2 minor stylistic questions. Otherwise I don't see any further issues with these changes

Firmware/cmdqueue.cpp
wavexx: Probably because I removed it entirely on #3823. I will fix this

wavexx:
Regarding https://github.com/wavexx/Prusa-Firmware/pull/10 I would handle it differently. We need to allow the user to unlock the printer from the LCD.

Besides adding code in a hot path, M500 can save garbage state, and allowing the command there will make it available also while breaking the paused protocol, something I'm really not keen on.

wavexx:
We the latest commits we do the following:

- Repeated thermal model warnings/errors now do not prevent menu access
- Only partial menu access is available: preheat/filament/calibrate actions are disabled
- The user can clear a thermal error by using "Acknowledge error". This works for a one-time error, and will resume USB serial processing as expected in the same way aborting/resuming a thermal error pause does.
- "Print from SD" is now available during a thermal error, and a new print will implicitly *clear* the error state. This is allowed as an escape hatch to run g-code containing "M310 S0" to disable the model. If the model is still running, the SD print will still be interrupted pretty fast, so it should pose no threat.

Firmware/messages.cpp
```diff
@@ -169,6 +169,7 @@ extern const char MSG_PAUSED_THERMAL_ERROR[] PROGMEM_I1 = ISTR("PAUSED THERMAL E
 #ifdef TEMP_MODEL
 extern const char MSG_THERMAL_ANOMALY[] PROGMEM_I1 = ISTR("THERMAL ANOMALY");////MSG_THERMAL_ANOMALY c=20
 extern const char MSG_TM_NOT_CAL[] PROGMEM_I1 = ISTR("Temp model not calibrated yet.");////MSG_TM_NOT_CAL c=20 r=4
+extern const char MSG_ACK_ERROR[] PROGMEM_I1 = ISTR("Acknowledge error");////MSG_ACK_ERROR c=17
```
3d-gussner: A menu can be `c=18`
The message may not be that clear to the users.
```
01234567890123456789
>Clear TM error    ➔
>Confirm TM error  ➔
>Unlock TM error   ➔

**Comment by wavexx** [comment]
This fixes #3630

File: Firmware/messages.cpp [hunk 4]
```diff
@@ -169,6 +169,7 @@ extern const char MSG_PAUSED_THERMAL_ERROR[] PROGMEM_I1 = ISTR("PAUSED THERMAL E
 #ifdef TEMP_MODEL
 extern const char MSG_THERMAL_ANOMALY[] PROGMEM_I1 = ISTR("THERMAL ANOMALY");////MSG_THERMAL_ANOMALY c=20
 extern const char MSG_TM_NOT_CAL[] PROGMEM_I1 = ISTR("Temp model not calibrated yet.");////MSG_TM_NOT_CAL c=20 r=4
+extern const char MSG_TM_ACK_ERROR[] PROGMEM_I1 = ISTR("Clear TM error");////MSG_TM_ACK_ERROR c=18
```
3d-gussner, line 172: Adding new message needs an update of the pot and po files.
Also the new message should be translated.
Do want to do that in a separate PR?
Splitting "unblocks" this PR as we don't need to wait for the translations, which may take a while.

wavexx: Why would it block it?

3d-gussner (approved):
Works well on MK404 and real printer.
On MK404 after factory reset I am able to recover from TM error and "unlock" the printer. This solves the worst case scenario of locking out users.

Firmware/messages.cpp
3d-gussner: Because we have to wait for the translations

PR #3824, opened 2022-12-14
MK3: French shorten translation

3d-gussner:
Fix `[μm]` in translations

- [X] Create cherry-pick to MK3_3.12

### DIFF
--- a/lang/lib/charset.py
+++ b/lang/lib/charset.py
@@ -2,10 +2,9 @@
 CUSTOM_CHARS = {
     '\x06': '⏬',
     '\x04': '🔃',
-    '\xe4': 'µ',
     '\xdf': '°',
     '\xe1': 'ä',
-    '\xe4': 'μ',
+    '\xe4': 'µ', #on keyboard AltGr+m it is \xC2\xB5
     '\xef': 'ö',
     '\xf5': 'ü',
 }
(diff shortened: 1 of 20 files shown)


lang/lib/charset.py
```diff
@@ -2,10 +2,10 @@
 CUSTOM_CHARS = {
     '\x06': '⏬',
     '\x04': '🔃',
-    '\xe4': 'µ',
+    '\xe4': 'µ', #\xC2\xB5
     '\xdf': '°',
     '\xe1': 'ä',
-    '\xe4': 'μ',
+    #'\xe4': 'μ', #xCE\xBC
```
wavexx: What are we doing here? You cannot have multiple entries in this map. Remove the comment entirely.

wavexx (changes requested):

lang/lib/charset.py
wavexx: This is better, but I would still remove completely this line.

3d-gussner: I only kept is as some reminder. Looks like some tools used by translators use odd/unconventional chars
I will move the comment above as so we keep an eye on it.

wavexx: I see but this is not the place to put it anyway because the map can't contain more than a single entry.

Doesn't U+03bc (the wrong one) fail to build anyway?

3d-gussner:
@leptun The french translation has been provided by the internal translator and approved by content team.

leptun (approved):

3d-gussner:
@wavexx Can I merge this one? Now it compiles and shows the micron correctly on LCD in all translations?

wavexx (approved):
Everything looks good now

DRracer (approved):

PR #3833, opened 2022-12-16
MK3: fix MK2.5/S getting stuck in the bootloader after watchdog reset

### LINKED ISSUE #3832: [BUG]MK2.5S Factory Reset makes Z-motor steppers sing, and hangs!
**Printer type** - [MK2.5S]
**Printer firmware version** -  [prusa3d_fw_3_11_0_MK25S]

**MMU upgrade** - [NO]
**MMU upgrade firmware version** - [NO]

**SD card or USB/Octoprint**
  Please let us know if you print via SD card or USB/Octoprint

**Describe the bug**
  Factory Reset makes Z-stepper motors sing, and HANGS!
  
**To Reproduce**
  Push and hold selector knob. Push Reset while continuing to hold selector knob. When you see 'Factory Reset',
  select ALL DATA.
  
**Expected behavior**
  Expect memory to be reset, and command complete
  
**G-code**
  NOT INVOLVED

**Video**

leptun:
The miniRambo has a broken bootloader which doesn't properly handle the watchdog timer. The expected behaviour is for the bootloader to check the reset source and jump quickly to the application if the reason for reset was a watchdog rest. However, that is not done of the old bootloader, instead waiting the full timeout period for the host to upload firmware. To work around this issue, when soft-resetting, use a watchdog reset period that covers the entire bootloader stage so that the watchdog timer can then be disabled by the main firmware after the bootloader jumps to the main app. In this PR I used a 2s timeout.

The watchdog is now also disabled early in the app in case a stupid bootloader forgot to disable watchdog timer.

Should fix #3832

PFW-1466

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -697,8 +697,7 @@ void __attribute__ ((naked)) __attribute__ ((section (".init3"))) stopWatchdogOn
 }
 }
 
-void softReset()
-{
+void softReset(void) {
     cli();
 #ifdef WATCHDOG
     // If the watchdog support is enabled, use that for resetting. The timeout value is customized
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -688,7 +688,24 @@ void failstats_reset_print()
 void softReset()
 {
     cli();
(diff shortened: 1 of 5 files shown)


DRracer (approved):
looks correct to me :+1:

PR #3835, opened 2022-12-17
PF-build: improve storage and caching of dependencies

wavexx:
A few fixes for PF-build to improve caching of the dependencies:

- check for the correct python version (we only use python3)
- do not re-download dependencies if already extracted
- use an unversioned root so we can cache dependencies across versions
- add that root to the travis cache

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,10 @@ language: minimal
 
 cache:
   directories:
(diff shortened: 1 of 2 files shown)


3d-gussner (approved):
Thanks working well

PR #3837, opened 2022-12-18
PFW-1457 Do not unload at start of First Layer Cal

gudnimg:
Fixes issue where First Layer Cal. triggers an unload before any filament is loaded

The purpose of this `if()` statement is to handle the edge case where a user or developer is sending T-codes to the printer directly via Serial. Such as when one first sends `T0` and then `T4`. And unload must be triggered in-between.

My first solution was to simply check if a print is on-going. But this doesn't take into account first layer calibration. We can generalize the check to only apply when FINDA detects filament.

Change in memory:
Flash: -10 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -22,7 +22,6 @@
 #include "pins.h"
 #include "Timer.h"
 #include "mmu2.h"
-#include "cardreader.h" // for IS_SD_PRINTING
 extern uint8_t mbl_z_probe_nr;
 
 #ifndef AT90USB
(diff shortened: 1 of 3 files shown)


Firmware/Marlin.h
```diff
@@ -22,6 +22,7 @@
 #include "pins.h"
 #include "Timer.h"
 #include "mmu2.h"
+#include "cardreader.h" // for IS_SD_PRINTING
```
wavexx: Do we need to move this here? I know it's more convenient, but I'd try to keep includes in Marlin.h as low as possible as this is included everywhere, resulting in total rebuilds if you touch any of the dependencies.

gudnimg: Yes, I would otherwise get a compiler error.

gudnimg: image

gudnimg: Found an alternative. Making CHECK_FSENSOR into a function. 👍

gudnimg: Costs no additional memory :D

wavexx (approved):

PR #3839, opened 2022-12-18
MMU2 split variant config

leptun:
Separate the MMU2 config from the MMU2S config. This is due to the different extruder requiring different lengths of filament to be pushed/pulled.
This PR also changes the load to nozzle sequence so that it pushes enough filament.


gudnimg (approved):
Looks good to me

DRracer (approved):
nice and clean :+1:

PR #3841, opened 2022-12-18
build: Use the cached path for the build environment

wavexx:
This should allow the obsolete "build.sh" to cache the build environment as well.


3d-gussner (approved):
Tested local build without any issues.

PR #3830, opened 2022-12-15
MK3: XYZ cal change message and add LCD ouput

3d-gussner:
@Hauzman @vintagepc @AttilaSVK @Prime1910 @ingbrzy @shatter136 @Painkiller56 @OS-kar

Please check my changes to the community translations.
:+1:  correct
:-1:  needs change and please comment here

- [x] Create cherry-pick PR to MK3_3.12 branch

PR #3846, opened 2022-12-20
MK3_3.12: XYZ cal change message and add LCD ouput

3d-gussner:
Cherry-pick from https://github.com/prusa3d/Prusa-Firmware/pull/3830

### DIFF
--- a/Firmware/xyzcal.cpp
+++ b/Firmware/xyzcal.cpp
@@ -135,10 +135,10 @@ pos_mm_t pos_2_mm(float pos){
 	return pos * 0.01f;
 }
 
-void xyzcal_meassure_enter(void)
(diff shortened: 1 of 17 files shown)


DRracer (approved):

Firmware/xyzcal.cpp
```diff
@@ -135,10 +135,10 @@ pos_mm_t pos_2_mm(float pos){
 	return pos * 0.01f;
 }
 
-void xyzcal_meassure_enter(void)
+void xyzcal_meassure_center(void)
```
wavexx, line 138: I think this is fixing the wrong typo. We had enter -> leave. Now "center" doesn't make much sense.
I would instead have fixed "meassure" to "measure"

PR #3851, opened 2022-12-24
Minor header cleanup in Marlin_main.cpp

gudnimg:
* One step closer to delete conv2str.cpp/h. 🎅 
    * In this case `fan_speed_rec` is `uint8_t` so it cannot exceed 3 characters. `itostr3()` is therefore redundant here.
* `ultralcd.h` is included 3 times in `Marlin_main.cpp`, I removed the extra ones, only one include is needed.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -66,7 +66,6 @@
 
 #include "menu.h"
 #include "ultralcd.h"
-#include "conv2str.h"
 #include "backlight.h"
 
 #include "planner.h"
@@ -11043,8 +11042,7 @@ void restore_print_from_eeprom(bool mbl_was_active) {
 	sprintf_P(cmd, PSTR("M220 S%d"), feedmultiply_rec);
 	enquecommand(cmd);
   // Set the fan speed saved at the power panic.
-	strcpy_P(cmd, PSTR("M106 S"));
-	strcat(cmd, itostr3(int(fan_speed_rec)));
+	sprintf_P(cmd, PSTR("M106 S%u"), fan_speed_rec);
 	enquecommand(cmd);
 
   // Set a position in the file.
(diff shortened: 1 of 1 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -76,7 +76,6 @@
 #include "cardreader.h"
 #include "ConfigurationStore.h"
 #include "language.h"
-#include "pins_arduino.h"
```
leptun, line 80: this file does in fact exist. It's part of the arduino board variant (Einsy rambo in this case). See https://github.com/prusa3d/Arduino_Boards/blob/master/IDE_Board_Manager/prusa3dboards/variants/prusa_einsy_rambo/pins_arduino.h

Firmware/Marlin_main.cpp
```diff
@@ -11043,8 +11042,7 @@ void restore_print_from_eeprom(bool mbl_was_active) {
 	sprintf_P(cmd, PSTR("M220 S%d"), feedmultiply_rec);
 	enquecommand(cmd);
   // Set the fan speed saved at the power panic.
-	strcpy_P(cmd, PSTR("M106 S"));
-	strcat(cmd, itostr3(int(fan_speed_rec)));
+	sprintf_P(cmd, PSTR("M106 S%c"), fan_speed_rec);
```
leptun: This looks wrong. %c is used for printing ascii characters, which is not the intended case here. Use %u instead. I know it's unfortunate, but on the AVR, all function parameters get blown up to at least 2B anyway, so it's not like there is a more efficient option. %hhu also doesn't work unfortunately.

gudnimg: Fixed 👍

Firmware/Marlin_main.cpp
gudnimg: I see it is included with `Arduino.h` in `Marlin.h` so I believe we are safe in removing this particular include in `Marlin_main.cpp`

I'll reword the commit message "doesn't exist" is inaccurate as you point out 👍 

image

leptun (approved):
lgtm

PR #3853, opened 2022-12-24
Remove T parameter from M221

### LINKED ISSUE #3852: [BUG] M221 doesn't change extrusion multiplier if T parameter is used
**Printer type** -  MK3S+
**Printer firmware version** -  3.11 and before

**MMU upgrade** - N/A

**SD card or USB/Octoprint** - N/A

**Describe the bug**
In the M221 Gcode the variable `extruder_multiply[]` is assigned but it is never used anywhere. I suspect `extruder_multiplier[]` or `extrudemultiply` should be used here.

There are some references to this in the planner code:

```
#ifdef LA_FLOWADJ
        // M221/FLOW should change uniformly the extrusion thickness
        float delta_e = (e - position_float[E_AXIS]) / extruder_multiplier[extruder];
#else
        // M221/FLOW only adjusts for an incorrect source diameter
        float delta_e = (e - position_float[E_AXIS]);
```

I propose the `T` parameter be removed since the firmware doesn't really handle multiple extruders. The T parameter is only used by Prusa,
  
**To Reproduce**
Inspect by code review
  
**Expected behavior**
M221 should have an effect on the extrusion
  
**G-code**
M221 S100 T1

**Video**

gudnimg:
Fixes #3852

Just a proposal. The T parameter doesn't work anyway from what I can see.

We decided a few months ago the firmware should only support one extruder, so I doubt we will fix the T parameter to work.

Change in memory:
Flash: -28 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -272,7 +272,6 @@ extern uint8_t axis_relative_modes;
 extern float feedrate;
 extern int feedmultiply;
 extern int extrudemultiply; // Sets extrude multiply factor (in percent) for all extruders
-extern int extruder_multiply[EXTRUDERS]; // sets extrude multiply factor (in percent) for each extruder individually
 extern float extruder_multiplier[EXTRUDERS]; // reciprocal of cross-sectional area of filament (in square millimeters), stored this way to reduce computational burden in planner
 extern float current_position[NUM_AXIS] ;
 extern float destination[NUM_AXIS] ;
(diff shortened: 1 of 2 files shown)


leptun (approved):
lgtm

gudnimg:
@DRracer Are you OK with this change?

DRracer:
@gudnimg not sure ... wasn't the T parameter intended to compensate different flow rate for individual slots of the MMU? I mean - compensate for filament which need to have different flow rate? I don't see a reason why this cannot be computed by the slicer, but who knows?

DRracer:
After internal discussion, this parameter seems to be a remainder of the old SNMM code which has been removed a while ago.

PR #3854, opened 2022-12-27
Speed optimisation: Replace `labs` with `fabs` in `wait_for_heater()`

gudnimg:
Both `degHotend` and `degTargetHotend` return a `float`
In this case it is better to use `fabs()`

Change in memory:
Flash: -14 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -9887,7 +9887,7 @@ static void wait_for_heater(long codenum, uint8_t extruder) {
 			or when current temp falls outside the hysteresis after target temp was reached */
 			if ((residencyStart == -1 && target_direction && (degHotend(extruder) >= (degTargetHotend(extruder) - TEMP_WINDOW))) ||
 				(residencyStart == -1 && !target_direction && (degHotend(extruder) <= (degTargetHotend(extruder) + TEMP_WINDOW))) ||
-				(residencyStart > -1 && labs(degHotend(extruder) - degTargetHotend(extruder)) > TEMP_HYSTERESIS))
+				(residencyStart > -1 && fabs(degHotend(extruder) - degTargetHotend(extruder)) > TEMP_HYSTERESIS))
 			{
 				residencyStart = _millis();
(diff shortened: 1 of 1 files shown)


wavexx (approved):

gudnimg:
To verify this "speed" improvement via assembly I tested this code in Godbolt with AVR GCC 5.4.0 (7.3.0 is not available):

The assignment below to `my_serial` is only there so the labs and fabs calls are not optimised away.

```
#include <stdlib.h>
#include <math.h>

float current = 1.2345;
float my_serial = 0;

void setup(){
  
}
void loop(){

    my_serial = labs(current);
    my_serial = fabs(current);
}
```

The size difference between the two outputs below demonstrate why we save 14 bytes of flash:

my_serial = fabs(current); output
=========

```asm
        lds r24,current
        lds r25,current+1
        lds r26,current+2
        lds r27,current+3
        andi r27,0x7f
        sts my_serial,r24
        sts my_serial+1,r25
        sts my_serial+2,r26
        sts my_serial+3,r27
```

my_serial = labs(current); output
=========
```asm
        lds r24,current
        lds r25,current+1
        lds r26,current+2
        lds r27,current+3
        mov r22,r24
        mov r23,r25
        mov r24,r26
        mov r25,r27
        rcall __fixsfsi
        mov r27,r25
        mov r26,r24
        mov r25,r23
        mov r24,r22
        tst r27
        brge .L3
        com r27
        com r26
        com r25
        neg r24
        sbci r25,lo8(-1)
        sbci r26,lo8(-1)
        sbci r27,lo8(-1)
.L3:
        mov r22,r24
        mov r23,r25
        mov r24,r26
        mov r25,r27
        rcall __floatsisf
        mov r27,r25
        mov r26,r24
        mov r25,r23
        mov r24,r22
        sts my_serial,r24
        sts my_serial+1,r25
        sts my_serial+2,r26
        sts my_serial+3,r27
```

DRracer (approved):
makes sense to me, approving

PR #3855, opened 2022-12-27
Revised firmware calibration and upgrade checks

### LINKED ISSUE #3836: [BUG] Change CALIBRATION_STATUS to a bitfield
**Printer type** - any
**Printer firmware version** -  3.13+

**MMU upgrade** - n/a

**SD card or USB/Octoprint**
n/a

**Describe the bug**
`calibration_status()` is currently a *sequence* of actions to be performed before `CALIBRATION_STATUS_CALIBRATED` is reached. This means that we cannot easily skip a single calibration step (XYZ, model or 1st layer) or run them in any other other.

If the wizard is skipped, incomplete calibration can result if you don't follow the on-screen prompts.
I consider this a bug.
  
**To Reproduce**
Stop wizard immediately. Run thermal model calibration -> printer is now fully calibrated, but XYZ has never been run.
  
**Expected behavior**
Change the eeprom value to a bitfield so that we can know the status of each step, run them in any order without risking to miss anything.

wavexx:
In order to *properly* handle new optional calibration steps such as temperature model calibration the entire wizard/calibration workflow had to be revised. A new eeprom variable has been introduced that handles the calibration stages as a bitfield (fixing #3836) and cleans up most of the weird calibration handling of the previous versions. The variable `EEPROM_CALIBRATION_STATUS_V2` properly handles upgrade workflows and new calibration features transparently. The old variable (now `EEPROM_CALIBRATION_STATUS_V1`) is kept unused to allow downgrading to a previous version.

With this change we can now explicitly perform temperature model calibration after a firmware upgrade, so that the default model values are not used on existing printers which already completed the full wizard workflow. We currently do not ask: we just warn that the temperature model will be performed. This is similar in behavior to FW 3.2 where selftest was enforced to perform belt calibration.

Due to this change there are a lot of things that needs to be retested surrounding the wizard handling:

- ALTFAN override initialization has been cleaned up. Ensure ALTFAN is still detected and handled properly when installed after a FACTORY RESET.
- Ensure ALTFAN behavior is not changed if FW is upgraded *without* a FW reset.
- Ensure the entire wizard works in the same order as before.
- Ensure the wizard still asks about calibrating more sheets as before.
- Ensure the wizard after a factory prep reset works as before.
- When upgrading from FW 3.2.0.3 (or older) self-test needs to be enforced as before.
- When upgrading from FW 3.11 (or older) there is now a new message warning that the model calibration will start.
- If the full wizard is skipped, the usual "printer is not calibrated, follow calibration flow" should be shown.
- If the wizard is aborted due to selftest failure, hotend or xyz or temperature model calibration failure there should be a warning about incomplete wizard. This wasn't handled properly before.
- Ensure the wizard still terminates properly on failure (due to selftest, hotend, xyz or model failure) and doesn't attempt to continue indefinitely.
- Even if the wizard is skipped due to a later failure, a warning about a missing temperature model calibration should be shown. The warning is NOT shown if the wizard is skipped intentionally at the first question.
- The warning should NOT be shown if the model is disabled afterwards.
- You should be able to RESET the printer to abort temp model calibration (the next boot should show that model calibration is required, but it not done automatically again).
- Even if the wizard is skipped due to failure, a warning about a missing Z offset should be shown (as before).
- The same warning should be shown when switching to a new (unused) bed sheet.
- The warning should go away if we switch back to a known sheet and reset.
- Ensure we can downgrade to a previous version without failure.

Doing a "SHIPPING PREP" reset should require (and do) _only_ Z calibration after power-up. This assumes model calibration will be done in the factory. We can adjust this if needed and it's easy to do so now.

There is a new string and an updated string regarding the temperature model calibration to be updated.

PFW-1469

### DIFF
--- a/Firmware/util.cpp
+++ b/Firmware/util.cpp
@@ -181,19 +181,19 @@ inline int8_t is_provided_version_newer(const char *version_string)
     return 0;
 }
 
-bool eeprom_fw_version_older_than(const uint16_t (&ver_req)[4])
+bool eeprom_fw_version_older_than_p(const uint16_t (&ver_req)[4])
 {
     uint16_t ver_eeprom[4];
-
     ver_eeprom[0] = eeprom_read_word((uint16_t*)EEPROM_FIRMWARE_VERSION_MAJOR);
     ver_eeprom[1] = eeprom_read_word((uint16_t*)EEPROM_FIRMWARE_VERSION_MINOR);
     ver_eeprom[2] = eeprom_read_word((uint16_t*)EEPROM_FIRMWARE_VERSION_REVISION);
     ver_eeprom[3] = eeprom_read_word((uint16_t*)EEPROM_FIRMWARE_VERSION_FLAVOR);
 
     for (uint8_t i = 0; i < 4; ++i) {
-        if (ver_req[i] > ver_eeprom[i])
+        uint16_t v = pgm_read_word(&ver_req[i]);
+        if (v > ver_eeprom[i])
             return true;
-        else if (ver_req[i] < ver_eeprom[i])
+        else if (v < ver_eeprom[i])
             break;
     }
 
--- a/Firmware/util.cpp
+++ b/Firmware/util.cpp
@@ -513,3 +513,24 @@ void ip4_to_str(char* dest, uint8_t* IP)
 {
     sprintf_P(dest, PSTR("%u.%u.%u.%u"), IP[0], IP[1], IP[2], IP[3]);
 }
+
+
+bool calibration_status_get(CalibrationStatus components)
+{
+    CalibrationStatus status = eeprom_read_byte((uint8_t*)EEPROM_CALIBRATION_STATUS_V2);
+    return ((status & components) == components);
+}
+
+void calibration_status_set(CalibrationStatus components)
+{
+    CalibrationStatus status = eeprom_read_byte((uint8_t*)EEPROM_CALIBRATION_STATUS_V2);
+    status |= components;
+    eeprom_update_byte((uint8_t*)EEPROM_CALIBRATION_STATUS_V2, status);
+}
+
+void calibration_status_clear(CalibrationStatus components)
+{
+    CalibrationStatus status = eeprom_read_byte((uint8_t*)EEPROM_CALIBRATION_STATUS_V2);
+    status &= ~components;
+    eeprom_update_byte((uint8_t*)EEPROM_CALIBRATION_STATUS_V2, status);
+}
--- a/Firmware/util.cpp
+++ b/Firmware/util.cpp
@@ -181,32 +181,23 @@ inline int8_t is_provided_version_newer(const char *version_string)
     return 0;
 }
 
-bool force_selftest_if_fw_version()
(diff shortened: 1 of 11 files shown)


wavexx:
There is currently no way to set model as "calibrated". For example, when setting all values via M310: the printer will still moan that model calibration has not been performed.

I do not want to abuse existing flags in M310 for this.

D10 allows to set the XYZ calibration status, I wonder if we need the same for M310.

wavexx:
Also note: when applying this to 3.12, we should still keep the same eeprom offset even though EEPROM_HEAT_BED_ON_LOAD_FILAMENT is not available.

3d-gussner:
> Also note: when applying this to 3.12, we should still keep the same eeprom offset even though EEPROM_HEAT_BED_ON_LOAD_FILAMENT is not available.

Easiest is then also cherry-pick that PR to MK3_3.12 so we have the that in sync, or not?

3d-gussner:
First tests are working well. I want to run some additional tests with serial debug of the calibration status.

Firmware/util.cpp
```diff
@@ -181,32 +181,23 @@ inline int8_t is_provided_version_newer(const char *version_string)
     return 0;
 }
 
-bool force_selftest_if_fw_version()
+bool eeprom_fw_version_older_than(const uint16_t (&ver_req)[4])
```
DRracer: we can probably optimize the parameter as PROGMEM

Firmware/util.cpp
```diff
@@ -522,3 +513,24 @@ void ip4_to_str(char* dest, uint8_t* IP)
 {
     sprintf_P(dest, PSTR("%u.%u.%u.%u"), IP[0], IP[1], IP[2], IP[3]);
 }
+
+
+bool calibration_status_get(CalibrationStatus components)
```
DRracer, line 518: Are these functions inlined? Does it make sense making them __attribute__((noiniline))?

DRracer (approved):
very nice, generally I approve this PR :+1: 

there are a few minor comments which can be addressed later

Firmware/util.cpp
wavexx: I was /about/ to say they _cannot_ be, but with -flto you can't say right away.
They're not inlined. As they're all external, they should never be.

Firmware/util.cpp
wavexx: Done

3d-gussner:
Tested on MK404 and it works well, just one or two issue I have found.
- Factory reset ALL :heavy_check_mark: 
- Factory reset Shipping/Service :heavy_check_mark:
- LCD Selftest :heavy_check_mark:
- LCD TM cal :heavy_check_mark:
- XYZ cal. :question: As it resets the cal status from 31 to 15 and not back to 31 after finished
- Same for MMU2SR :heavy_check_mark:
- No thermal anomaly during hotend selftest :heavy_check_mark:
- Reset the printer during Wizard it continues (if selected with yes) at the missed wizard step :heavy_check_mark: 
- Minor issue `calibration_status_clear` and maybe `calibration_status_set` update the eeprom value even it didn't change.

DRracer (approved):

PR #3857, opened 2022-12-28
:construction: Optimisation: re-use `delay_keep_alive()` in a few places

gudnimg:
The optimisation is based on seeing these three lines together:

```c
manage_heater();
manage_inactivity(true);
lcd_update(0);
```

To reduce code size we can call one function instead: `delay_keep_alive()` which will then call the above three functions.

Note that `delay_keep_alive()` assumes `manage_inactivity()` takes `true` for it's argument. A few places don't specify this value, I believe using `true` is safe in most cases.

Change in memory on **Multi-lang** build (MK3S+):
Flash: -360 bytes
SRAM: 0 bytes

Change in memory on **English only** build (MK3S+):
Flash: -316 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -9672,18 +9672,18 @@ static void wait_for_heater(long codenum, uint8_t extruder) {
 			SERIAL_PROTOCOL((int)extruder);
 
 #ifdef TEMP_RESIDENCY_TIME
-				SERIAL_PROTOCOLPGM(" W:");
-				if (residencyStart > -1)
-				{
-					codenum = ((TEMP_RESIDENCY_TIME * 1000UL) - (_millis() - residencyStart)) / 1000UL;
-					SERIAL_PROTOCOLLN(codenum);
-				}
-				else
-				{
-					SERIAL_PROTOCOLLN('?');
-				}
+			SERIAL_PROTOCOLPGM(" W:");
+			if (residencyStart > -1)
+			{
+				codenum = ((TEMP_RESIDENCY_TIME * 1000UL) - (_millis() - residencyStart)) / 1000UL;
+				SERIAL_PROTOCOLLN(codenum);
+			}
+			else
+			{
+				SERIAL_PROTOCOLLN('?');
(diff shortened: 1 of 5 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -4485,13 +4483,7 @@ void process_commands()
         }
       }
       st_synchronize();
-      codenum += _millis();  // keep track of when we started waiting
-      previous_millis_cmd.start();
```
gudnimg, line 4382: I shouldn't have removed previous_millis_cmd.start(); here... adding it back

Firmware/Marlin_main.cpp
```diff
@@ -4484,6 +4484,7 @@ void process_commands()
       }
       st_synchronize();
       delay_keep_alive(codenum);
+      previous_millis_cmd.start();
```
gudnimg, line 4381: Removing or adding `previous_millis_cmd.start();` here:
Interesting this commit didn't have effect on code size 🤔 is this being optimised away?

gudnimg:
Rebased with MK3 to resolve merge conflict

Firmware/planner.cpp
```diff
@@ -714,10 +714,7 @@ void plan_buffer_line(float x, float y, float z, const float &e, float feed_rate
   // Rest here until there is room in the buffer.
   if (block_buffer_tail == next_buffer_head) {
       do {
-          manage_heater(); 
-          // Vojtech: Don't disable motors inside the planner!
-          manage_inactivity(false);
```
DRracer, line 717: beware - manage_inactivity(**false**); at this spot

DRracer (changes requested):
please double-check the `manage_inactivity( true / false );` at various spots

delay_keep_alive uses `manage_inactivity(true);` in its current form

Firmware/planner.cpp
gudnimg: I suspect this is a bug. It doesnt make sense to allow the motors to be disabled here.

PR #3842, opened 2022-12-18
README: Document the new build system

wavexx:
This gets rid of the outdated documentation and replaces it with instructions for CMake or PF-build.

Some help is needed for windows has I have zero experience with it.

closes #3543

PR #3858, opened 2022-12-29
cmake: Consistent target names

wavexx:
When writing #3842 I realized our target names are a bit inconsistent. This PR addresses this by renaming some of our targets to be more consistent.

- Use ALL_MULTILANG and VARIANT_MULTILANG instead of VARIANT_Multilang
- Use ALL_ENGLISH and VARIANT_ENGLISH instead of VARIANT_En-only
- Rename the intermediate VARIANT_Multilang_* targets to VARIANT_lang_* to avoid confusion


DRracer (approved):
no complaints

PR #3859, opened 2022-12-29
Simplify variant names

wavexx:
- Remove nozzle size and heater type from the variant names (all variants support any nozzle size anyway).
- Also remove board type from MK3/MK3S (both have EINSY as the only type)

This allows to cleanup some rules in cmake and shorten the target names, making them more readable.

- [x] update build server needed before merging.


gudnimg (approved):
Works great on my end (Windows)

wavexx:
Can't approve my own PR, but the REVO changes are ok

3d-gussner (approved):
Rebased and fixed. It is working.

PR #3861, opened 2022-12-29
Revised firmware calibration and upgrade checks (3.12)

wavexx:
Backport of #3855

- Includes updated eeprom offsets
- Stubbed definition of LCDButtonChoice to reduce changes


wavexx:
Also to check: flash and calibrate FW 3.12, then upgrade to FW 3.13: model should still be calibrated.

3d-gussner (approved):
Tested on MK404 and real hardware.
Upgrade and downgrade of firmware works well.

DRracer (approved):

PR #3863, opened 2022-12-30
TM: Also clear the status flags when disabling the model

wavexx:
Disabling the model during a warn/error condition will also stop updating the warning/error flag, keeping the printer in an error state.

Clear all flags as well when changing model settings.


3d-gussner (approved):
Tested on MK404 and works well
To force errors I used `M310 W0.1 E0.2`

DRracer (approved):

PR #3867, opened 2022-12-30
Fix build failure when `TEMP_RESIDENCY_TIME` is not defined

### LINKED ISSUE #3866: [BUG] 3.13 Build fails if `TEMP_RESIDENCY_TIME` is not defined
**Printer type** - Any
**Printer firmware version** -  3.13

**MMU upgrade** - N/A

**SD card or USB/Octoprint**
N/A

**Describe the bug**
The firmware fails to build if preprocessor definition `TEMP_RESIDENCY_TIME` is not available. Particularly we have an undefined variable `tmp_extruder` in `waiting_for_heater()`
  
**Expected behavior**
Build will build :)

gudnimg:
* Fix build when `TEMP_RESIDENCY_TIME` is not defined
* Fix build when `EXTRUDER_WATTS` is defined

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -9792,7 +9792,7 @@ static void wait_for_heater(long codenum, uint8_t extruder) {
 	while ((!cancel_heatup) && ((residencyStart == -1) ||
 		(residencyStart >= 0 && (((unsigned int)(_millis() - residencyStart)) < (TEMP_RESIDENCY_TIME * 1000UL))))) {
 #else
-	while (target_direction ? (isHeatingHotend(tmp_extruder)) : (isCoolingHotend(tmp_extruder) && (CooldownNoWait == false))) {
+	while (target_direction ? isHeatingHotend(active_extruder) : isCoolingHotend(active_extruder)) {
 #endif //TEMP_RESIDENCY_TIME
 		if ((_millis() - codenum) > 1000UL)
 		{ //Print Temp Reading and remaining time every 1 second while heating up/cooling down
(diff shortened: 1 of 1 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -9792,7 +9792,7 @@ static void wait_for_heater(long codenum, uint8_t extruder) {
 	while ((!cancel_heatup) && ((residencyStart == -1) ||
 		(residencyStart >= 0 && (((unsigned int)(_millis() - residencyStart)) < (TEMP_RESIDENCY_TIME * 1000UL))))) {
 #else
-	while (target_direction ? (isHeatingHotend(tmp_extruder)) : (isCoolingHotend(tmp_extruder) && (CooldownNoWait == false))) {
```
wavexx, line 9795: I wouldn't remove CooldownNowait.
See https://github.com/prusa3d/Prusa-Firmware/pull/2060/files#diff-a2e2d6e9dab1e37ea9a2882433e2b8f70af751b3c3445d5e997d11a93d4b28b3R10135 and the comments in the PR.

gudnimg: But its not defined in this scope 🤔the build will fail

gudnimg: Hmm I’m pretty sure I saw the build fail due to the variable being undefined. I can double check tomorrow.

wavexx: Yeah, I'm sure it was broken. I just remembered that we had a PR fixing this by actually using/fixing CooldownNoWait, which is IMHO better.

gudnimg: Fixed. The PR now only resolves the undefined variable `tmp_extruder`.

gudnimg:
Closing. I'll create another PR which will remove the extruder parameter since the extruder ID is always 0.

PR #3873, opened 2023-01-01
Optimisation: implement `lcd_print_pad_P()`

gudnimg:
Used `lcd_print_pad_P` is some obvious places (mainly the menu code). We can probably use this function in more places to further save flash memory

Skimming through the menus I didn't see any issues. This needs to be tested well :) 

Change in memory:
Flash: -496 bytes
SRAM: -4 bytes

### DIFF
--- a/Firmware/lcd.cpp
+++ b/Firmware/lcd.cpp
@@ -538,6 +538,16 @@ char lcd_print_pad(const char* s, uint8_t len)
     return *s;
 }
 
+char lcd_print_pad_P(const char* s, uint8_t len)
+{
+    while (len && pgm_read_byte(s)) {
+        lcd_write(pgm_read_byte(s++));
+        --len;
+    }
+    lcd_space(len);
+    return *s;
+}
+
 void lcd_print(char c, int base)
 {
 	lcd_print((long) c, base);
--- a/Firmware/lcd.cpp
+++ b/Firmware/lcd.cpp
@@ -66,6 +66,8 @@
 #define LCD_RS_FLAG 0x01
 #define LCD_HALF_FLAG 0x02
 
+constexpr uint8_t row_offsets[] PROGMEM = { 0x00, 0x40, 0x14, 0x54 };
+
 FILE _lcdout; // = {0}; Global variable is always zero initialized, no need to explicitly state that.
 
 uint8_t lcd_displayfunction = 0;
@@ -341,8 +343,7 @@ static void FORCE_INLINE lcd_set_current_row(uint8_t row)
 /// @return row offset which the LCD register understands
 static uint8_t __attribute__((noinline)) lcd_get_row_offset(uint8_t row)
 {
-	uint8_t row_offsets[] = { 0x00, 0x40, 0x14, 0x54 };
-	return row_offsets[min(row, LCD_HEIGHT - 1)];
+	return pgm_read_byte(row_offsets[min(row, LCD_HEIGHT - 1)]);
 }
 
 void lcd_set_cursor(uint8_t col, uint8_t row)
--- a/Firmware/lcd.cpp
+++ b/Firmware/lcd.cpp
@@ -329,13 +329,31 @@ void lcd_no_autoscroll(void)
 }
 #endif
 
-void lcd_set_cursor(uint8_t col, uint8_t row)
+/// @brief set the current LCD row
+/// @param row LCD row number, ranges from 0 to LCD_HEIGHT - 1
+static void FORCE_INLINE lcd_set_current_row(uint8_t row)
+{
+	lcd_currline = min(row, LCD_HEIGHT - 1);
+}
+
+/// @brief Calculate the LCD row offset
+/// @param row LCD row number, ranges from 0 to LCD_HEIGHT - 1
+/// @return row offset which the LCD register understands
+static uint8_t __attribute__((noinline)) lcd_get_row_offset(uint8_t row)
 {
 	uint8_t row_offsets[] = { 0x00, 0x40, 0x14, 0x54 };
-	if (row >= LCD_HEIGHT)
-		row = LCD_HEIGHT - 1;    // we count rows starting w/0
-	lcd_currline = row;  
-	lcd_command(LCD_SETDDRAMADDR | (col + row_offsets[row]));
+	return row_offsets[min(row, LCD_HEIGHT - 1)];
+}
+
+void lcd_set_cursor(uint8_t col, uint8_t row)
+{
+	lcd_set_current_row(row);
+	lcd_command(LCD_SETDDRAMADDR | (col + lcd_get_row_offset(lcd_currline)));
+}
+
+void lcd_set_cursor_column(uint8_t col)
+{
+	lcd_command(LCD_SETDDRAMADDR | (col + lcd_get_row_offset(lcd_currline)));
 }
 
 // Allows us to fill the first 8 CGRAM locations
--- a/Firmware/lcd.cpp
+++ b/Firmware/lcd.cpp
@@ -343,7 +343,7 @@ static void FORCE_INLINE lcd_set_current_row(uint8_t row)
 /// @return row offset which the LCD register understands
 static uint8_t __attribute__((noinline)) lcd_get_row_offset(uint8_t row)
 {
-	return pgm_read_byte(row_offsets[min(row, LCD_HEIGHT - 1)]);
+	return pgm_read_byte(row_offsets + min(row, LCD_HEIGHT - 1));
 }
 
 void lcd_set_cursor(uint8_t col, uint8_t row)
--- a/Firmware/lcd.cpp
+++ b/Firmware/lcd.cpp
@@ -538,14 +538,14 @@ char lcd_print_pad(const char* s, uint8_t len)
     return *s;
 }
 
-char lcd_print_pad_P(const char* s, uint8_t len)
+uint8_t lcd_print_pad_P(const char* s, uint8_t len)
 {
     while (len && pgm_read_byte(s)) {
         lcd_write(pgm_read_byte(s++));
         --len;
     }
     lcd_space(len);
-    return *s;
+    return len;
 }
 
 void lcd_print(char c, int base)
--- a/Firmware/lcd.h
+++ b/Firmware/lcd.h
@@ -54,6 +54,7 @@ extern void lcd_printFloat(double number, uint8_t digits);
 
 extern void lcd_print(const char*);
 extern char lcd_print_pad(const char* s, uint8_t len);
+char lcd_print_pad_P(const char* s, uint8_t len);
 extern void lcd_print(char, int = 0);
 extern void lcd_print(unsigned char, int = 0);
 extern void lcd_print(int, int = 10);
--- a/Firmware/lcd.h
+++ b/Firmware/lcd.h
@@ -54,7 +54,12 @@ extern void lcd_printFloat(double number, uint8_t digits);
 
 extern void lcd_print(const char*);
 extern char lcd_print_pad(const char* s, uint8_t len);
-char lcd_print_pad_P(const char* s, uint8_t len);
+
+/// @brief print a string from PROGMEM with left-adjusted padding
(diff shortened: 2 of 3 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -806,7 +806,8 @@ int uart_putchar(char c, FILE *)
 void lcd_splash()
 {
 	lcd_clear(); // clears display and homes screen
-	lcd_printf_P(PSTR("\n Original Prusa i3\n   Prusa Research\n%20.20S"), PSTR(FW_VERSION));
+	lcd_puts_P(PSTR("\n Original Prusa i3\n   Prusa Research\n"));
+	lcd_print_pad_P(PSTR(FW_VERSION), LCD_WIDTH);
```
leptun: Doesn't this move the text to be left aligned? The old one is right aligned

gudnimg: Good catch. The savings here were only minimal. I'll revert this change.

Firmware/lcd.h
```diff
@@ -54,6 +54,7 @@ extern void lcd_printFloat(double number, uint8_t digits);
 
 extern void lcd_print(const char*);
 extern char lcd_print_pad(const char* s, uint8_t len);
+char lcd_print_pad_P(const char* s, uint8_t len);
```
leptun: why not make this one also extern?

Firmware/menu.cpp
```diff
@@ -197,7 +179,7 @@ static void menu_draw_toggle_puts_P(const char* str, const char* toggle, const u
     sprintf_P(lineStr, PSTR("%c%-18.18S"), (settings & 0x01)?'>':' ', str);
     sprintf_P(lineStr + LCD_WIDTH - ((settings & 0x02)?strlen_P(toggle):strlen(toggle)) - 3, (settings & 0x02)?PSTR("[%S]%c"):PSTR("[%s]%c"), toggle, eol);
     if (!(settings & 0x04)) lcd_set_cursor(0, menu_row);
-    fputs(lineStr, lcdout);
+    lcd_print_pad(lineStr, LCD_WIDTH);
```
leptun: Not sure how useful this is here. lineStr is still padded by the sprintf_P a bit higher. It might be worth rewriting this entire function at some point so it becomes more readable and optimal

Firmware/menu.cpp
```diff
@@ -233,7 +215,9 @@ static void menu_draw_item_select_sheet_E(char type_char, const Sheet &sheet)
     lcd_set_cursor(0, menu_row);
     SheetFormatBuffer buffer;
     menu_format_sheet_select_E(sheet, buffer);
-    lcd_printf_P(PSTR("%c%-18.18s%c"), menu_selection_mark(), buffer.c, type_char);
+    lcd_putc(menu_selection_mark());
+    lcd_print_pad_P(buffer.c, LCD_WIDTH - 2);
```
leptun: buffer.c is in ram. Use the non-P version of lcd_print_pad

Firmware/menu.cpp
```diff
@@ -242,25 +226,19 @@ static void menu_draw_item_puts_E(char type_char, const Sheet &sheet)
     lcd_set_cursor(0, menu_row);
     SheetFormatBuffer buffer;
     menu_format_sheet_E(sheet, buffer);
-    lcd_printf_P(PSTR("%c%-18.18s%c"), menu_selection_mark(), buffer.c, type_char);
+    lcd_putc(menu_selection_mark());
+    lcd_print_pad_P(buffer.c, LCD_WIDTH - 2);
```
leptun: same issue

Firmware/menu.cpp
```diff
@@ -242,25 +226,19 @@ static void menu_draw_item_puts_E(char type_char, const Sheet &sheet)
...
+    lcd_putc(type_char);
 }
 
 static void menu_draw_item_puts_P(char type_char, const char* str, char num)
 {
     lcd_set_cursor(0, menu_row);
-    lcd_printf_P(PSTR("%c%-.16S "), menu_selection_mark(), str);
+    lcd_putc(menu_selection_mark());
+    lcd_print_pad_P(str, LCD_WIDTH - 3);
```
leptun: This usage is not correct. This function needs/uses only the length limiting modifier, not also the padding modifier. With lcd_print_pad_P, the number after the string is pushed to the end of the line, which is not the desired effect. See the MMU filament menus where this is used (afaik, might be wrong on this statement)

Firmware/lcd.h
gudnimg: I don't see a reason to use `extern` here. We are including the header file in `.cpp` files this function is used so `extern` is not useful here. We could remove `extern` keyword from most of the functions in the firmware without any issues.

leptun: I agree. I mentioned it just for consistency with the other prototypes around it

Firmware/menu.cpp
gudnimg: I'll re-check this one. I probably only added this because I saw a change in the memory footprint.

Firmware/menu.cpp
```diff
@@ -448,13 +426,9 @@ static void menu_draw_P(char chr, const char* str, int16_t val);
...
-	int text_len = strlen_P(str);
-	if (text_len > 15) text_len = 15;
-	char spaces[LCD_WIDTH + 1] = {0};
-    memset(spaces,' ', LCD_WIDTH);
-	if (val <= -100) spaces[15 - text_len - 1] = 0;
-	else spaces[15 - text_len] = 0;
-	lcd_printf_P(menu_fmt_int3, chr, str, spaces, val);
+    char spaces[LCD_WIDTH + 1] = {0};
+    sprintf_P(spaces, menu_fmt_int3, chr, str, spaces, val);
+    lcd_print_pad_P(spaces, LCD_WIDTH);
```
leptun: this one is not ok. Go for example to "Settings->Temperature" and you will see how it fails / crashes the firmware.

gudnimg: Woops I see `spaces` appears twice there. Need to think about this one 🤔

Firmware/menu.cpp
gudnimg: Fixed [8207561](https://github.com/prusa3d/Prusa-Firmware/pull/3873/commits/82075610040c6fe4ea44ea7834bc7a7027f34e8d)

Firmware/menu.cpp
gudnimg: Fixed [8207561](https://github.com/prusa3d/Prusa-Firmware/pull/3873/commits/82075610040c6fe4ea44ea7834bc7a7027f34e8d)

Firmware/menu.cpp
gudnimg: Took a shot at some refactoring in [9f70de8](https://github.com/prusa3d/Prusa-Firmware/pull/3873/commits/9f70de8ba09b90f81e5745ec702d309808f20ee3) and saved 106 bytes. We can do a lot more refactoring but it will affect a lot more area in the code which I don't think we should attempt in this PR.

Firmware/menu.cpp
gudnimg: I reverted these changes for now.

Firmware/menu.cpp
gudnimg: You are correct 👍 Interestingly it's not an issue with "Extruder X" items.
IMG_2832

IMG_2831

gudnimg: Issue is now fixed. I changed `lcd_print_pad_P` such that it now returns the number of padded pads (the remains of `len`) It can help us determine the next cursor position.

gudnimg:
Difference in memory on Multilang build:

Before:
```
[build] Program:  252682 bytes (96.4% Full)
[build] 
[build] Data:       5648 bytes (68.9% Full)
```

After:
```
[build] Program:  252228 bytes (96.2% Full)
[build] 
[build] Data:       5648 bytes (68.9% Full)
```

Firmware/lcd.cpp
```diff
@@ -329,13 +329,31 @@ void lcd_no_autoscroll(void)
...
+{
+	lcd_currline = min(row, LCD_HEIGHT - 1);
+}
+
+/// @brief Calculate the LCD row offset
+/// @param row LCD row number, ranges from 0 to LCD_HEIGHT - 1
+/// @return row offset which the LCD register understands
+static uint8_t __attribute__((noinline)) lcd_get_row_offset(uint8_t row)
 {
 	uint8_t row_offsets[] = { 0x00, 0x40, 0x14, 0x54 };
```
DRracer: Can we move this array into PROGMEM?

Firmware/lcd.cpp
```diff
@@ -341,7 +343,6 @@ static void FORCE_INLINE lcd_set_current_row(uint8_t row)
 /// @return row offset which the LCD register understands
 static uint8_t __attribute__((noinline)) lcd_get_row_offset(uint8_t row)
 {
-	uint8_t row_offsets[] = { 0x00, 0x40, 0x14, 0x54 };
 	return row_offsets[min(row, LCD_HEIGHT - 1)];
```
leptun: I'm pretty sure you must use pgm_read_byte for the progmem array

gudnimg: You're right, I forgot about that part.

gudnimg: @leptun Fixed. Flash usage increased by 4 bytes. I'll run this again quickly on my machine later this evening.

Firmware/lcd.cpp
gudnimg: Done

Firmware/lcd.cpp
```diff
@@ -329,13 +331,30 @@ void lcd_no_autoscroll(void)
...
-		row = LCD_HEIGHT - 1;    // we count rows starting w/0
-	lcd_currline = row;  
-	lcd_command(LCD_SETDDRAMADDR | (col + row_offsets[row]));
+	lcd_set_current_row(row);
+	lcd_command(LCD_SETDDRAMADDR | (col + lcd_get_row_offset(lcd_currline)));
+}
+
+void lcd_set_cursor_column(uint8_t col)
+{
+	lcd_command(LCD_SETDDRAMADDR | (col + lcd_get_row_offset(lcd_currline)));
```
DRracer, line 357: `lcd_command(LCD_SETDDRAMADDR | (col + lcd_get_row_offset(lcd_currline)));`
seems to be duplicated in these functions - I'm not sure if doing something like this actually saves anything:
```
void lcd_set_cursor(uint8_t col, uint8_t row)
{
	lcd_set_current_row(row);
	lcd_set_cursor_column(col);
}

void lcd_set_cursor_column(uint8_t col)
{
	lcd_command(LCD_SETDDRAMADDR | (col + lcd_get_row_offset(lcd_currline)));
}
```

DRracer: yes, it saves another 22B, but `lcd_set_cursor` must be made `__attribute__((noinline))`

gudnimg: The downside here is sending two commands to the LCD. I dont remember how long it took but I think one command transfer is in the order of 20 to 50us.

gudnimg: IMO I would keep this as is for speed. 22B doesnt seem worth it to make the cursor change 2x slower. 

Maybe we can shorten the command to only modify the column. I need to find the datasheet again 🤔

leptun: I would definitely keep the LCD faster at the cost of 22B. It's already painfully slow. Let's not make it even slower.
The `LCD_SETDDRAMADDR` sets both the row and the column at the same time. Actually, the LCD doesn't really have the concept of rows and columns and that's why we need the row offsets. From the perspective of the LCD, the DDRAM is just a block of memory that we write to and `LCD_SETDDRAMADDR` just updates the r/w pointer to that memory block.

DRracer: Ok, what you wrote makes sense, it would be nice to add some kind of explanation why is the code written as it is. Otherwise, I like it :wink:

DRracer (approved):

PR #3876, opened 2023-01-01
🚧 refactor feedmultiply code for status screen

gudnimg:
Change in memory:
Flash: -64 bytes
SRAM: 0 bytes


leptun (changes requested):
Bug: ENCODER_FEEDRATE_DEADZONE is not handled if going from a `feedmultiply < 100` to a `feedmultiply > 100` without ever rendering the screen with `feedmultiply == 100`. This happens in both ways when passing 100. It makes aiming for 100% a bit harder than before.

gudnimg:
Closing in favor of https://github.com/prusa3d/Prusa-Firmware/pull/4181

PR #3884, opened 2023-01-06
Fix blank LCD while err. Load to Extr. Failed

DRracer:
Because Load to Extr. Failed error is not an MMU error but a printer one, the existing infrastructure has been bent to support such a scenario. During testing it turned out, that some machines fail to draw the error screen due to previous internal states. This PR resets the internal states so that the conditions for drawing the error screen are met.

Cutting was invoked one step later before this PR - so even after cutting the filament tip successfully and the load would have been successful, the operation failed with an error. This is now fixed as well.

Operations CutFilament and EjectFilament now also support suppression of their FullScreenMsgs, which is especially usefull in case when called from a ToolChange.

PFW-1474

### DIFF
--- a/Firmware/mmu2.cpp
+++ b/Firmware/mmu2.cpp
@@ -301,7 +301,7 @@ bool MMU2::ToolChangeCommonOnce(uint8_t slot){
         } else { // Prepare a retry attempt
             unload();
             if( retries == 2 && eeprom_read_byte((uint8_t*)EEPROM_MMU_CUTTER_ENABLED) == EEPROM_MMU_CUTTER_ENABLED_enabled){
-                cut_filament(slot); // try cutting filament tip at the last attempt
+                cut_filament(slot, false); // try cutting filament tip at the last attempt
             }
         }
     }
@@ -451,11 +451,13 @@ void FullScreenMsg(const char *pgmS, uint8_t slot){
     lcd_print(slot + 1);
 }
 
-bool MMU2::cut_filament(uint8_t slot){
+bool MMU2::cut_filament(uint8_t slot, bool enableFullScreenMsg /* = true */){
     if( ! WaitForMMUReady())
         return false;
 
-    FullScreenMsg(_T(MSG_CUT_FILAMENT), slot);
+    if( enableFullScreenMsg ){
+        FullScreenMsg(_T(MSG_CUT_FILAMENT), slot);
+    }
     {
         if( FindaDetectsFilament() ){
             unload();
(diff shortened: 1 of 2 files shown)


Firmware/mmu2.cpp
```diff
@@ -683,6 +689,8 @@ void MMU2::CheckUserInput(){
             // we'll actually wait for it automagically in manage_response and after it finishes correctly,
             // we'll issue another command (like toolchange)
             logic.ClearPrinterError();
+            lastErrorCode = ErrorCode::OK;
+            lastErrorSource = ErrorSourceNone; // this seems to help clearing the error screen
```
gudnimg, line 691: These lines should not be needed. LCD blank issue was fixed in https://github.com/prusa3d/Prusa-Firmware/pull/3907

Firmware/mmu2.cpp
```diff
@@ -312,6 +312,8 @@ void MMU2::ToolChangeCommon(uint8_t slot){
     while( ! ToolChangeCommonOnce(slot) ){ // while not successfully fed into extruder's PTFE tube
         // failed autoretry, report an error by forcing a "printer" error into the MMU infrastructure - it is a hack to leverage existing code
         // @@TODO theoretically logic layer may not need to be spoiled with the printer error - may be just the manage_response needs it...
+        // overcome blocking conditions
+        lastErrorCode = ErrorCode::OK;
```
gudnimg: If this is to fix the LCD blank issue I'm against this change. The LCD blank issue was fixed in https://github.com/prusa3d/Prusa-Firmware/pull/3907

Firmware/mmu2.cpp
```diff
@@ -451,11 +453,13 @@ void FullScreenMsg(const char *pgmS, uint8_t slot){
     lcd_print(slot + 1);
 }
 
-bool MMU2::cut_filament(uint8_t slot){
+bool MMU2::cut_filament(uint8_t slot, bool enableFullScreenMsg /* = true */){
     if( ! WaitForMMUReady())
         return false;
 
-    FullScreenMsg(_T(MSG_CUT_FILAMENT), slot);
+    if( enableFullScreenMsg ){
```
gudnimg, line 458: I like this 👍

Firmware/mmu2_protocol_logic.h
```diff
@@ -145,6 +145,7 @@ class ProtocolLogic {
 
     inline void SetPrinterError(ErrorCode ec){
         explicitPrinterError = ec;
+        errorCode = ErrorCode::OK;
```
gudnimg: Is this required?

gudnimg (changes requested):
There are some conflict with my fix in https://github.com/prusa3d/Prusa-Firmware/pull/3907

Firmware/mmu2.cpp
DRracer: I know, that the issue went away in #3907, but I still think cleaning up after this error is a good idea.
I'm considering moving the PrinterError infrastructure from the logic layer into mmu2.cpp.

Firmware/mmu2.cpp
DRracer: yes, this should be removed :+1:

Firmware/mmu2_protocol_logic.h
DRracer: may be not

Firmware/mmu2.cpp
gudnimg: Cleaning the error code is a good point.

gudnimg (approved):

DRracer:
rearranged commits after review + fixups

PR #3886, opened 2023-01-09
optimisation: re-use `serialecho_temperatures()`

gudnimg:
Extracted optimisation from https://github.com/prusa3d/Prusa-Firmware/pull/3857 since this is a low hanging fruit

Change in memory:
Flash: -82 bytes
SRAM: 0 bytes


leptun (approved):

DRracer (approved):
nice :+1:

Issue #3895, opened 2023-01-12
[BUG] MMU2S/printer can't progress following filament failure intervention 


3dprintandpaint:
*** This is a copy of #3742, which is present in the Beta and describes the problem perfectly.  The issues persists in the final release making the MMU unusable after any attempted intervention ***

Printer type -MK3S+
Printer firmware version - 3.12.0

MMU upgrade - MMU2S
MMU upgrade firmware version - 1.0.6

USB/Octoprint

Describe the bug
When MMU2S is in failure state (e.g., filamant is not loaded, or out-of-filament), then printer displays error message containing like "press button on MMU unit to continue". Anyway pressing any button on MMU2S unit or recovering the MMU2S failure (by loading the filament) is not read out by printer and printer is stuck forever, so only possible solution is to reset it.

To Reproduce
Use printer menu to load filament but do not insert the filament to MMU2S. After a while printer displays error message and it is not possible to recover printer by loading the filament (using MMU2S buttons). Printer is stuck and the only solution is to reset printer.

Expected behavior
When MMU2S failure is resolved (e.g., by loading the filament using MMU2S buttons) printer shall continue.

G-code
N/A

Video
NONE

3d-gussner:
Thanks for reporting this issue. We are looking into it.

3d-gussner:
@3dprintandpaint Can you please try this firmware?
   
[prusa3d_fw_MK3S_MK3S+_t3_12_0-1_5673.hex.zip](https://github.com/prusa3d/Prusa-Firmware/files/10413481/prusa3d_fw_MK3S_MK3S%2B_t3_12_0-1_5673.hex.zip)

If it is working please report back and close this issue?

3d-gussner:
Closing issue as the fix has been merged for FW 3.12.1

PR #3897, opened 2023-01-13
Fix Load filament issues with MMU FW1.0.6

3d-gussner:
Revert `LongTimers` to `uint32_t` to fix issues with MMU2 FW1.0.6.
This PR is only for Firmware 3.12.x NOT 3.13.x !!!  
Should fix https://github.com/prusa3d/Prusa-Firmware/issues/3895

```
Sketch uses 246490 bytes (97%) of program storage space. Maximum is 253952 bytes.
Global variables use 5569 bytes of dynamic memory.
```
before
```
Sketch uses 246296 bytes (96%) of program storage space. Maximum is 253952 bytes.
Global variables use 5572 bytes of dynamic memory.
```
+194 flash
-3 RAM

PFW-1462

### DIFF
--- a/Firmware/mmu.cpp
+++ b/Firmware/mmu.cpp
@@ -76,9 +76,12 @@ int16_t mmu_version = -1;
 
 int16_t mmu_buildnr = -1;
 
-LongTimer mmu_last_request;
-LongTimer mmu_last_response;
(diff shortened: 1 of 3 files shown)


DRracer (approved):
looks correct to me, let's wait for a retest

Issue #3889, opened 2023-01-10
  [ENHANCEMENT] The new Nozzle Change feature should max out the z height 


laskr1999:
Printer:MK3S+
MMU: no

The new nozzle change feature is good, but the problem is, the bed nearly stays under the nozzle. When someone(like me) uses a screwdriver like socket I can't fit it under the nozzle just in a steep angle because of not maxed out z height+bed is under it.
  
**Describe the solution you'd like**
  I think the bed should go to the zero position to the back and the Z to the max 210mm so anything can fit under the nozzle very well/make it configurable so peoples can adjust to their wrenches.

3d-gussner:
@laskr1999 Thanks for the feedback.

> I think the bed should go to the zero position

Agreed on the Z height and Y as far back, but I still would keep the X pos in the middle to have access from all sides.

3d-gussner:
@laskr1999 Thanks for the request. It will be implemented in next release see https://github.com/prusa3d/Prusa-Firmware/pull/3901 and https://github.com/prusa3d/Prusa-Firmware/pull/3902

Please close this issue / feature request.

3d-gussner:
Closing as the enhancement has been merged.
Thanks @laskr1999 again for the feedback :hugs:

PR #3901, opened 2023-01-13
PFW-1478 MK3 Move Z to top during Nozzle change

3d-gussner:
Minor change to allow more work space during the nozzle change, thanks to @laskr1999 requesting the change https://github.com/prusa3d/Prusa-Firmware/issues/3889
In our opinion it makes still sense to move the extruder in the middle of the x-axis.

Before
```
Sketch uses 253040 bytes (99%) of program storage space. Maximum is 253952 bytes.
Global variables use 5658 bytes of dynamic memory.
```
With this PR
```
Sketch uses 253036 bytes (99%) of program storage space. Maximum is 253952 bytes.
Global variables use 5658 bytes of dynamic memory.
```
-4 Flash
+- 0 RAM

- [x] Cherry-pick to MK3_3.12

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -1037,7 +1037,7 @@ void lcd_commands()
                 lcd_update_enabled = true;
                 lcd_draw_update = 2; //force lcd clear and update after the stack unwinds.
                 enquecommand_P(PSTR("G28 W"));
-                enquecommand_P(PSTR("G1 X125 Y10 Z150 F1000"));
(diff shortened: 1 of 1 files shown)


DRracer (approved):
:+1:  
Removing `Y10` makes sense as well to me. The machine first homes with 'G28 W` => Y=0 at that moment => heat bed has been moved far back allowing for easier access to the nozzle and preventing accidental damage to the bed/sheet in case something fell onto it while changing the nozzle.

PR #3904, opened 2023-01-13
Don't enable fan on MINTEMP error

gudnimg:
If the printer is already in a very cold environment, enabling the fan won't help.

### DIFF
--- a/Firmware/temperature.cpp
+++ b/Firmware/temperature.cpp
@@ -536,7 +536,10 @@ void set_temp_error(TempErrorSource source, uint8_t index, TempErrorType type)
 
     // keep disabling heaters and keep fans on as long as the condition is asserted
     disable_heater();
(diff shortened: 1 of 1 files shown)


DRracer (changes requested):
:thinking: can we rely on the fact, that MINTEMP actually means low temperature? I don't think so. Any thermal-related error like this should try to stay as safe as possible, that includes keeping the fans on to cool down the HW, even if there are scenarios when it may look counterintuitive.

gudnimg:
> 🤔 can we rely on the fact, that MINTEMP actually means low temperature? I don't think so. Any thermal-related error like this should try to stay as safe as possible, that includes keeping the fans on to cool down the HW, even if there are scenarios when it may look counterintuitive.

Yeah I kind of agree that we can't rule out the situation where having the fan on is good **if** the thermistor output is very unreliable.

PR #3907, opened 2023-01-14
Fix blank LCD issue

gudnimg:
Also removed the log `"retryAttempts=x"` which can be spammed in logs under some circumstances.

Change in memory:
Flash: -38 bytes
SRAM: 0 bytes

PFW-1474


DRracer (approved):
I don't see any logical error in this PR, however the real cause of the occasional blank LCD's (unreproducible in some cases) remains hidden.

PR #3909, opened 2023-01-14
Shorten the Try-Load-Unload sequence

DRracer:
... and allow easier future tweking in case of need.

Also, save ~30B of code by returning the previous E-motor position.


gudnimg (approved):
Looks good to me 👍

PR #3914, opened 2023-01-15
optimisation: re-use `delay_keep_alive()` in a few places PR2

gudnimg:
Pulled all optimisations from https://github.com/prusa3d/Prusa-Firmware/pull/3857 which don't change any behavior in the firmware.

Change in memory (on multilang build):
Flash: -52 bytes
SRAM: 0 bytes


DRracer (approved):
looks correct :+1:

PR #3915, opened 2023-01-15
MMU: Fix bug with button choice on MMU error screen isn't reset

gudnimg:
PFW-1381

Fix issue where button selection on MMU error screen is not reset correctly

The issue only appears after two error screens with different number of buttons.

Step-by-step description of issue:

1. Trigger an MMU error screen with 3 buttons. Select any button to resolve it.
2. On closing the screen, current_selection would be reset to the default button selection with respect to a error screen with 3 buttons.
3. Trigger an MMU error screen with 2 buttons. Select default button.
4. Notice issue. Firmware thinks you selected the OTHER button. This is due to the fact the screen uses the default choices as if it was a 3 button screen. current_selection still has the values from step 2.

The solution is to reset the button choice when we enter `ReportErrorHookMonitor()` for the first time on a new MMU error screen. I added a new variable reset_button_selection to track this. After the changes in this PR I can no longer reproduce the issue.

Change in memory:
Flash: +16 bytes
SRAM: +1 bytes

### DIFF
--- a/Firmware/mmu2_reporting.cpp
+++ b/Firmware/mmu2_reporting.cpp
@@ -101,6 +101,7 @@ static uint8_t ReportErrorHookMonitor(uint8_t ei) {
     uint8_t ret = 0;
     bool two_choices = false;
     static int8_t enc_dif = lcd_encoder_diff;
+    static uint8_t reset_button_selection = 1;
 
     if (lcd_encoder_diff == 0)
     {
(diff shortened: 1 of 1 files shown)


gudnimg:
I didn't expect SRAM to increase by 2 bytes. Should only increase by 1 byte. 🤔

gudnimg:
Found the issue. PR will need to be re-tested.

The variable `reset_button_selection` is initialized to `0` due to being `static`. When the very first MMU screen is shown after booting the printer, we don't need to clear the choice selection because it is already at the default. It is only after we get a error screen the 2nd time that we need to clear the choice selection and by that point `reset_button_selection = 1`.

gudnimg:
Marking this PR as ready for review. I can confirm the PR fixes the issue.

DRracer (approved):
looks good to me :+1:

PR #3916, opened 2023-01-15
wizard: Do not complain about missing calibration if model is disabled

### LINKED ISSUE #3891: [ENHANCEMENT]3.12.0: Make thermal model disabled also prevent calibration requirement
**Printer type**: MK3S+

With new firmware 3.12 introducing thermal model (TM) a problem comes up with modded printers (like with revo hotend added) which requires TM to be disabled.

That's easy to do via "M310 S0 + M500".

The problem is that doesn't disable thermal calibration requirement after firmware upgrade etc.

Example scenario:
- printer with revo
- 3.12.0 RC2 firmware; M310 S0 + M500 done
- upgrade to final 3.12.0, printer boots and pauses ("Recv: echo:busy: paused for user") requiring manual intervention (on another floor in my case) to be able to control printer via serial again. Printer wanted to do TM calibration (which won't work on revo anyway).

EDIT:
- other scenario, just boot a printer - "Temp model not calibrated yet" pop ups every boot/serial connect

**Solution**

So maybe firmware should not only check EEPROM bit for "is thermal model calibrated" but also should check if user has TM enabled?

That would prevent asking user to do calibration which won't work with revo and as TM is disabled anyway on such printer.

wavexx:
This fixes #3891 without having to set the calibration bit, meaning that if the model is later re-enabled without a real calibration, a prompt is shown as expected.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -1556,7 +1556,7 @@ void setup()
           if (!calibration_status_get(CALIBRATION_STATUS_LIVE_ADJUST))
               lcd_show_fullscreen_message_and_wait_P(_T(MSG_BABYSTEP_Z_NOT_SET));
 #ifdef TEMP_MODEL
-          if (!calibration_status_get(CALIBRATION_STATUS_TEMP_MODEL))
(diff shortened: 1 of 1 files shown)


DRracer (approved):
nice :+1:

PR #3920, opened 2023-01-15
wizard: Do no set "Self test OK" as an alert message

### LINKED ISSUE #3892: [BUG] "Self tests OK" message hides the new thermal model messages
**Printer type** - Prusa MK3S+ with Bear customs and E3D Revo Micro hotend
**Printer firmware version** - 3.12.0 Final
**MMU upgrade** - None

**Describe the bug**
If we run a self test and right after a thermal model calibration then all the messages regarding the calibration are not visible and instead we have the "Self test OK" message. The THERMAL ANOMALY raised in case of problem is also hidden.
  
**To Reproduce**
1. Flash 3.12.0 firmware
2. Start the Thermal Model calibration => fails with THERMAL ANOMALY as I have a Revo
3. LCD menu -> Clear TM error
4. LCD menu -> Calibration -> Selftest => after a successful testing we have the "Self test OK" message displayed
5. LCD menu -> Calibration -> Temp Model cal. => no message from the TM calibration are displayed, even in the case of a thermal anomaly. Note that the hotend is still hot and the calibration is waiting the hotend to be at 50°C as expected
  
**Expected behavior**
Thermal model calibration messages and error should not be hidden by a successful self test.
 

Thank you for the help :)

wavexx:
Alert messages are intended to persist above "info" messages, thus preventing further calibration status updates to be shown (such as thermal model calibration).

Just set the message as a regular status message.

Fixes #3892

@3d-gussner as an extra test we need to ensure there is no sort of "workflow" that requires "Self test OK" to persist above other messages for some odd reason.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -6456,7 +6456,7 @@ bool lcd_selftest()
 	
 	if (_result)
 	{
-		LCD_ALERTMESSAGERPGM(_i("Self test OK"));////MSG_SELFTEST_OK c=20
+		lcd_setstatuspgm(_i("Self test OK"));////MSG_SELFTEST_OK c=20
 		calibration_status_set(CALIBRATION_STATUS_SELFTEST);
 	}
 	else
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -6456,8 +6456,9 @@ bool lcd_selftest()
 	
 	if (_result)
 	{
-		lcd_setstatuspgm(_i("Self test OK"));////MSG_SELFTEST_OK c=20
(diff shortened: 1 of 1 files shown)


3d-gussner:
I tried this before and it "breaks" the `LCD -> Calibration -> Selftest` to jump out of the menu and displaying the "Self test OK" message in the Info screen status line. The `LCD_ALERTMESSAGERPGM` is used as a "hack" to achieve this.
We gonna need a different solution for that.

wavexx:
Is it really the only reason? That's easy to fix.

3d-gussner (approved):
Tested on MK404 and real hardware. Behaves as before

PR #3921, opened 2023-01-15
wizard: Do no set "Self test OK" as an alert message (3.12)

wavexx:
Backport of #3920 for 3.12

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -6661,8 +6661,9 @@ bool lcd_selftest()
 	
 	if (_result)
 	{
-		lcd_setstatuspgm(_i("Self test OK"));////MSG_SELFTEST_OK c=20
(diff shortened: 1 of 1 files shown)


3d-gussner:
See https://github.com/prusa3d/Prusa-Firmware/pull/3920#issuecomment-1383590585

3d-gussner (approved):
Tested on MK404 and real hardware

DRracer (commented):
Looks correct to be, but :thinking: I'm afraid what else does this change break

wavexx:
AFAIK we don't have many possible workflows here. Only two paths looking at calls to lcd_selftest()

- manual run via menu:
  * this now matches existing behavior (essentially just exit to status)
- run during the wizard:
  * if all is ok, "self test ok" is not even shown
  * if the next step of the wizard fails, I expect the _new_ failure to be shown, not the "self test ok" message

Issue #3890, opened 2023-01-11
[BUG] Clear TM error acknowledgement doesn't allow to calibrate the thermal model


gregsaun:
**Printer type** - Prusa MK3S+ with Bear customs and E3D Revo Micro hotend
**Printer firmware version** - 3.12.0 Final
**MMU upgrade** - None

**Describe the bug**
I am unable to calibrate the thermal model for a safer hotend like the E3D Revo. Clear TM error acknowledgement seems to only unlock the LCD standard menu but doesn't allow to calibrate the thermal model. The documentation says:

> Added Clear TM error
> 
> To “unlock” the printer in case a “Thermal anomaly” is detected, there is a new option in the menu LCD menu -> Clear TM error to temporarily clear/acknowledge the issue. As this is only a “temporary” solution proper calibration is recommended. In case the temp model calibration doesn't solve this, please contact Prusa support or open a [GitHub issue](https://github.com/prusa3d/Prusa-Firmware/issues/new/choose).
  
**To Reproduce**
1. Flash the 3.12.0 firmware
2. Follow the Thermal Model calibration wizard => raise the THERMAL ANOMALY error
3. LCD menu -> Clear TM error
4. LCD menu -> calibration -> Temp Model cal. => raise the THERMAL ANOMALY error
  
**Expected behavior**
From the LCD menu we should be able to calibrate the thermal model for modifications on the hotend/extruder (new extruder body, new hotend, adding a sock) as well as the use of a safer hotend like E3D Revo. The release notes are not clear enough on how to proceed in such case.

arekm:
TM is incompatible with Revo at this moment (see https://github.com/prusa3d/Prusa-Firmware/issues/3636). If you want to use this firmware with Revo then you need to disable TM.

You can do that using serial console and issuing M310 S0 followed by M500.

There is no menu for this (due to: "We removed that option from the menu (we want to prevent ordinary users from disabling the TM easily/accidentally)").

Unfortunately that M310 S0/M500 doesn't seem to survive firmware upgrades (RC2->final upgrade and TM wizard came up again on first boot), so needs to be repeated after each upgrade. (Added https://github.com/prusa3d/Prusa-Firmware/issues/3891 for this)

gregsaun:
Thank you for your answer.

[...]

PR #3922, opened 2023-01-15
TM: Set the cal status bit when running `M310 A` directly

wavexx:
Running `M310 A` should set the TM calibration bit status if the autotune procedure was successful.

Partially addresses #3891 and #3890, since running `M310 A F0` should *still* count as valid calibration data and is not something that can be replicated by using the LCD (which enforces self-test).


DRracer (approved):
makese sense to me, approved :+1:

PR #3927, opened 2023-01-17
When IR sensor is used then show its status (LOADED or EMPTY) in

emard:
LCD menu Calibration -> Show end stops

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -3358,7 +3358,11 @@ void lcd_temp_cal_show_result(bool result) {
 }
 
 static void lcd_show_end_stops() {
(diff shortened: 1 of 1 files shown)


wavexx:
Don't we have this info in Support -> Sensor status?

emard:
I was buildins firmware for MK2.5 the menu system is
then slightly different I will have to again check if endstops
XYZ are on one menu and filament sensor on other

Maybe better idea is instead of having 2 separate menus
one for XYZ endstops other for sensors, both can be on one
place and fit on one screen....

emard:
I checked, really there is menu
Support -> Sensor info
and it shows sensor status, great it's already there
It was not on the place where I would intuitively search for
good practice to recompile firmware

Never mind, let's discard this pull request

PR #3928, opened 2023-01-17
Optimisation: Purge conv2str

gudnimg:
Change in memory:
Flash: -288 bytes
SRAM: -8 bytes


Firmware/cardreader.cpp
```diff
@@ -578,9 +577,9 @@ void CardReader::getStatus(bool arg_P)
         SERIAL_PROTOCOL('/');
         SERIAL_PROTOCOLLN(filesize);
         uint16_t time = ( _millis() - starttime ) / 60000U;
-        SERIAL_PROTOCOL(itostr2(time/60));
+        SERIAL_PROTOCOL(time / 60);
         SERIAL_PROTOCOL(':');
-        SERIAL_PROTOCOLLN(itostr2(time%60));
```
gudnimg, line 583: Not sure why we were limiting the output here to 2 characters. Doesn't make sense to me so I just removed the limitation.

gudnimg:
Ran some tests on the PID calibration status line. The UI looks identical to me:

| Before  |  After |
|---|---|
| IMG_2852  | IMG_2850  |
|  IMG_2851 |  IMG_2849 |

DRracer (approved):
nice :+1:

PR #3929, opened 2023-01-17
Disable M47 if TMC2130 drivers are used

leptun:
M47 only makes sense for printers that use actual endstops. The "virtual endstop" of the TMC driver only gives pulses during the stall event, so it doesn't make sense to show the state of the DIAG pins on screen.
It also made little sense that this menu actually showed the X/Y/Z endstop pin states on the einsy, where X and Y are not connected to anything, and Z is connected to the PINDA. Since we have the sensors submenu in support, this menu/mcode can be disabled safely on the MK3/S.

flash: -126B
RAM: 0B


gudnimg (approved):
This makes sense to me 👍

DRracer (approved):

PR #3930, opened 2023-01-17
Cleanup: remove `code_value_float()`

gudnimg:
The function adds a bit of overhead compared to `code_value()`

I suspect this function is a relic of the past. If we need to search for the `'E'` character. Why not just do it externally? It makes no sense to be apply this overhead on every use of `code_value_float()`. For floats we use `code_value()`

Change in memory:
Flash: -82 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -2050,7 +2050,7 @@ static float probe_pt(float x, float y, float z_before) {
     *  K<factor>                  Set advance K factor
     */
 inline void gcode_M900() {
-    float newK = code_seen('K') ? code_value_float() : -2;
+    float newK = code_seen('K') ? code_value() : -2;
 #ifdef LA_NOCOMPAT
     if (newK >= 0 && newK < LA_K_MAX)
         extruder_advance_K = newK;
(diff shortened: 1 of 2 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -7925,7 +7925,7 @@ SERIAL_PROTOCOLPGM("\n\n");
     case 862: // M862: print checking
           float nDummy;
           uint8_t nCommand;
-          nCommand=(uint8_t)(modff(code_value_float(),&nDummy)*10.0+0.5);
+          nCommand=(uint8_t)(modff(code_value(),&nDummy)*10.0+0.5);
```
DRracer, line 7928: we could save even more by not using `modff` ... in this case it is only here to detect `862.something`

DRracer (approved):
I don't see anything obviously wrong with this PR, approved

Firmware/Marlin_main.cpp
gudnimg: I have a separate PR for that :D https://github.com/prusa3d/Prusa-Firmware/pull/3795

PR #3936, opened 2023-01-20
TM: Do not restrict autotune temperature

wavexx:
Fix a wrong type that prevents running TM autotune at temps >127


gudnimg (approved):
Looks good to me. We're assigning `A` with `code_value_short()` which is `int16_t`

PR #3938, opened 2023-01-21
Optimisation: Don't print floats with `lcd_print()`

gudnimg:
Two optimisations:

1. When printing a number with `MarlinSerial::printNumber()` and `lcd_printNumber()`, there can be at most 8 * 4 = 32 digits (characters). We can use a `uint8_t` instead of `uint32_t` index to keep track of the number of digits.
2. Don't print `float` with `lcd_print()`. Currently there is only one place in the firmware which does this. See the menu `Support->XYZ cal. details`. This is very inefficient and we're better off using `lcd_printf_P()`.

Change in memory (on multilang MK3S build):
Flash: -594 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -2597,12 +2597,11 @@ static void lcd_menu_xyz_skew()
 	 _i("Slight skew"), _deg(bed_skew_angle_mild),  ////MSG_SLIGHT_SKEW c=14
 	 _i("Severe skew"), _deg(bed_skew_angle_extreme)  ////MSG_SEVERE_SKEW c=14
 	);
+	lcd_set_cursor(15, 0);
 	if (angleDiff < 100){
-		lcd_set_cursor(15,0);
 		lcd_printf_P(_N("%3.2f\x01"), _deg(angleDiff));
-	}
-	else{
-		lcd_puts_at_P(15,0, _T(MSG_NA));
+	} else {
+		lcd_puts_P(_T(MSG_NA));
 	}
     if (lcd_clicked())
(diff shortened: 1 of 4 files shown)


Firmware/ultralcd.cpp
```diff
@@ -2623,19 +2622,10 @@ static void lcd_menu_xyz_offset()
...
-    world2machine_read_valid(vec_x, vec_y, cntr);
-
-    for (uint_least8_t i = 0; i < 2; i++)
-    {
-        lcd_set_cursor((cntr[i] < 0) ? 13 : 14, i+2);
-        lcd_print(cntr[i]);
-        lcd_puts_at_P(18, i + 2, PSTR("mm"));
+    for (uint8_t i = 0; i < 2; i++) {
+        lcd_set_cursor(0, i + 2);
+        lcd_printf_P(PSTR("%c           %6.2fmm"), 'X' + i, eeprom_read_float((float*)(EEPROM_BED_CALIBRATION_CENTER+4*i)));
```
DRracer: :thinking: it looks like the formatting string is correct, there is:
- 1 axis designation character
- 11 spaces
- max 6 digits floating point number (padded with spaces if shorter)
- 2 characters 'mm'

may be we could abuse the float formatting placeholder to avoid the extra spaces as well, something like this:
`"%c%17.2fmm"`

DRracer (approved):
nice :+1: 

just please have a look at the formatting string if we can get rid of those extra spaces

Firmware/ultralcd.cpp
gudnimg: Done 👍 Works fine on my end.

Issue #3933, opened 2023-01-20
[BUG]Hotend Fan spins full power


Gamma4D:
**MK3S+** - [e.g. MK3S, MK3, MK2.5S, MK2.5, MK2S, MK2]
**3.12.1-RC1-5682** -  [e.g. 3.8.1, 3.8.1-RC1, ...]

**no** - [e.g. MMU2S, MMU2, MMU1]
**no** - [e.g. 1.0.6, 1.0.6-RC2, ...]

**SD card**
  Please let us know if you print via SD card or USB/Octoprint

**Describe the bug**
After flashing the firmware, the hotend fan spins full power.
In normal case, except during booting, it becomes 4000RPM over 50C and 0RPM below 50C.
But now it spins about 8280RPM.
It produces a loud noise in printing.

**To Reproduce**
Even after rebooting and performing calibration again, it does not go away.
I don't know how to reproduce this error.

**Expected behavior**
  
**G-code**
  Please attach a G-code. This will make it easier for us to replicate the error.

**Video**
  Please attach a video. It usually helps to solve the problem.

3d-gussner:
@Gamma4D Thanks for reporting this issue.
Do you have a  stock printer or did you modify something?
If stock, can you see if your hotend fan is from Noctua or a Delta fan (more powerful and louder)?
Please also check in LCD `Settings -> HW setup` long press knob in the `Experimental` menu if `ALTFAN det.` is On or Off

H0Z0:
I also experienced this, with 3.12 fw the altfan det. might not work. It is turned ON in the experimental menu, but if I try to turn it OFF, after restarting the printer, it will be ON again. But it will not control the speed. It spins at 8280RPM.
However, it works fine on the same printer with 3.11 fw. The ON/OFF state remains there after the printer is switched on/off and also controls the speed.

gudnimg:
> It is turned ON in the experimental menu, but if I try to turn it OFF, after restarting the printer, it will be ON again.

I can confirm this on latest MK3 branch (3.13). MK3S+, the setting was turned ON, I toggled it OFF and then rebooted the printer via the CPU reset button on the LCD panel. And the setting was back to ON again.

Edit: I suspect the issue was introduced in this commit. https://github.com/prusa3d/Prusa-Firmware/commit/da6e56a17e3db80ab63f758a6fc65ffb81d5b90e We're no longer reading the EEPROM value during boot-up.

Gamma4D:
[...]

PR #3939, opened 2023-01-21
Fix an issue where Altfan Override setting in EEPROM is not respected at boot-up

gudnimg:
See comments in this Github issue: https://github.com/prusa3d/Prusa-Firmware/issues/3933 I could reproduce this issue on my end with 3.13.

Change in memory:
Flash: -2 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/fancheck.cpp
+++ b/Firmware/fancheck.cpp
@@ -164,6 +164,9 @@ bool extruder_altfan_detect()
 {
     // override isAltFan setting for detection
     altfanStatus.isAltfan = 0;
+
+    // During initialisation, use the EEPROM value
+    altfanStatus.altfanOverride = eeprom_read_byte((uint8_t *)EEPROM_ALTFAN_OVERRIDE);
     setExtruderAutoFanState(3);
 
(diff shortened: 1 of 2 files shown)


leptun:
@gudnimg Please see the last commit. I was able to reduce the flash usage a bit by using the new eeprom_init_default methods

gudnimg:
Nice 👍 thanks

Firmware/fancheck.cpp
```diff
@@ -164,6 +164,9 @@ bool extruder_altfan_detect()
 {
     // override isAltFan setting for detection
     altfanStatus.isAltfan = 0;
+
+    // During initialisation, use the EEPROM value
+    altfanStatus.altfanOverride = eeprom_init_default_byte((uint8_t*)EEPROM_ALTFAN_OVERRIDE, 0);
```
DRracer, line 169: good, using the new eeprom infrastructure :+1:

DRracer (approved):

wavexx (approved):

PR #3942, opened 2023-01-23
PFW-1482: wizard: Disable TM warnings during calibration (3.12)

wavexx:
Backport of #3941 for 3.12

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -1050,6 +1050,7 @@ void lcd_commands()
             [[fallthrough]];
 
         case 3:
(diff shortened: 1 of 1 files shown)


DRracer (approved):

3d-gussner:
This doesn't work with manual gcode `M310 A F1`. Is that something we want or not?

wavexx:
IMHO it's ok if you run it by hand (you have full control there so I think you're wise enough to understand it's just a warning).

wavexx:
Also because running the model with warning set to a zero value is the easiest way to dump the internal value as the model is running (something I don't want to remove).

Issue #1434, opened 2018-12-31
Feature Request: Action commands


danielegger1:
I would like to get "Action" commands when some print issue occur, for example crash detection.
In Octoprint it is possible to react on printer "actions" but for that the printer must sent threw the terminal a message like "action:crashdetected". This can be added for several different things like filament runout, power panic, etc. This way I can send a message from octoprint to my phone (via telegram for example) with a picture, or I can pause the print make a alert or whatever I want.

![image](https://user-images.githubusercontent.com/16063664/50558816-77ea3480-0cf1-11e9-9356-d9fd2e959512.png)

Send: N21272 G1 X140.372 Y123.808 E0.00075*103
Recv: echo:enqueing "G1 E-1.000 F2700.000"
Recv: echo:enqueing "G1 Z  10.199 F 800.000"
Recv: echo:enqueing "CRASH_DETECTEDX"

NEW = action:crashdetected

Recv: tmc2130_home_enter(axes_mask=0x01)
Recv: echo:busy: processing
Recv: echo:busy: processing
Recv:   0 step=30 mscnt= 482
Recv: tmc2130_goto_step 0 29 2 1000
Recv: tmc2130_home_exit tmc2130_sg_homing_axes_mask=0x01
Recv: tmc2130_home_enter(axes_mask=0x02)
Recv: echo:busy: processing
Recv:   0 step=18 mscnt= 291
Recv: tmc2130_goto_step 1 17 2 1000
Recv: tmc2130_home_exit tmc2130_sg_homing_axes_mask=0x02
Recv: echo:enqueing "CRASH_RECOVER"
Recv: T:215.2 E:0 W:1
Recv: T:215.3 E:0 W:0
Recv: echo:busy: processing
Recv: echo:busy: processing
Recv: echo:busy: processing
Recv: Resend: 21242
Recv: ok
Send: N21242 G1 X143.783 Y126.239 E0.03352*101
Recv: ok

danielegger1:
I would also like to get a GCode command (M118 for example) which sends just a action:userAction command.
This way I could easily add this Gcode line on q certain layer height to let octoprint send me a message

DanaViolet:
I would also like to see this feature implemented. I would like for the printer to pause when the filament runout sensor is tripped, but this does not happen when printing with octoprint.

napter:
[...]

Issue #1381, opened 2018-12-05
[Feature request] Add M118 command


NewbieYnet414:
Hello I'm creating this ticket because ticket #921 is still opened without any response from developers.
**Wanted feature**
M118 command implemented as in vanilla Marlin.

**Explanation**
Please can you implement M118 command used to instruct host/G-code sender about events handled by printer firmware like unpausing print job from printer. This feature is already implemented in vanilla Marlin and is pretty handy.

**Background story**
I'm running 3D printing service on dormitory as activity of civil association. We are printing non-stop so I set up Octoprint for managing filament usage, remote controlling but in first place for queuing print jobs because in first month of usage we used 4kg of filament. And we are loosing lots of time on loading jobs to Octoprint. So using print queue I can speed up loading jobs via M0/M1 after each job pause FW let someone clean heatbed and confirm that with rotary button and start next job. Problem is that job is paused twice first on printer FW and after that on Octoprint which could be resumed by sending M118 A1 action:resume to printer and echoing back //action:resume.

johnbiggs:
That does sound helpful. I wonder if adding a button to the Pi and writing an octopi plugin to detect that button would be faster than waiting for the Devs?

NewbieYnet414:
That is not solution but workaround in first place. And probably not easy to do because Octoprint is running on VPS because slicing on raspberry will be way too slow. Raspberry is used only for camera and socat tty bridge between raspi and Octoprint on VPS.

Also why should someone write (reinvent wheel with square shape that can ride only on catenary shaped road) plugin that does only one thing of many that can be done with M118 which is supported by Marlin, Repetier, Octoprint, Pronterface.

johnbiggs:
I’m just sharing the reality of the situation. It’s probably faster for you to implement a workaround than wait for this. Additionally the command seems like it requires a lot of memory which the Devs have none left.

NewbieYnet414:
Hello @inquisitiveKangaroo any update on implementing M118?

NewbieYnet414:
Hello,

hopefully, you will add this in the next release after more than 6 months from first ticket opened.

```
case 118: //M118
  {

    bool hasE = false,
    hasA = false;
    uint8_t strchr_pos = 5;

    for (uint8_t i = 2; i--;) {
      if ((strchr_pointer[strchr_pos] == 'A' ||
          strchr_pointer[strchr_pos] == 'E') &&
[...]

Issue #921, opened 2018-07-11
Please add:  M118 - Serial print 


ieatacid:
The OctoPrint [ActionCommands plugin](https://github.com/benlye/OctoPrint-ActionCommandsPlugin) relies on serial output from M118 to execute the commands that are specified to it. 

Please add it back.

Is there a specific reason it was removed?

Edit: This is regarding MK3 firmware. I'm unsure of M118 availability on other variants.

petrzelka:
When printing from SD and using OctoPrint to monitor the print, one could use M118 in Slic3r in BEFORE_LAYER_CHANGE custom gcode to let OctoPrint know about Z change. 
This way (with a trivial OctoPrint plugin that I can create) it could do timelapse on z change, which currently isn't possible when priting from SD card.

PR #3946, opened 2023-01-24
Add M118 Gcode support

RoboMagus:
A long standing Feature Request, ported `M118` support from [Marlin](https://github.com/MarlinFirmware/Marlin/blob/bugfix-2.1.x/Marlin/src/gcode/host/M118.cpp).

Related issues:
- #921
- #1381
- #1434

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -6525,7 +6525,7 @@ Sigma_Exit:
     - `String` - Message string. If omitted, a blank line will be sent.
     */
     case 118: {
-        bool hasE, hasA = false;
+        bool hasE = false, hasA = false;
         char *p = strchr_pointer;
         
         for (uint8_t i = 2; i--;) {
@@ -6540,7 +6540,7 @@ Sigma_Exit:
         }
 
         if (hasE) SERIAL_ECHO_START;
-        if (hasA) SERIAL_ECHO("//");
+        if (hasA) SERIAL_ECHOPGM("//");
 
         SERIAL_ECHOLN(p);
     }
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3936,6 +3936,7 @@ extern uint8_t st_backlash_y;
 //!@n M114 - Output current position to serial port
 //!@n M115 - Capabilities string
 //!@n M117 - display message
+//!@n M118 - Serial print
 //!@n M119 - Output Endstop status to serial port
 //!@n M123 - Tachometer value
 //!@n M126 - Solenoid Air Valve Open (BariCUDA support by jmil)
@@ -6509,6 +6510,31 @@ Sigma_Exit:
         const char *src = strchr_pointer + 4; // "M117"
         lcd_setstatus(*src == ' '? src + 1: src);
         custom_message_type = CustomMsg::M117;
+    }
+    break;
+
+    /*!
+    ### M118 - Serial print <a href="https://reprap.org/wiki/G-code#M118:_Echo_message_on_host">M118: Serial print</a>
+    */
+    case 118: {
+        bool hasE, hasA = false;
+        char *p = strchr_pointer;
+        
+        for (uint8_t i = 2; i--;) {
+          // A1, E1, and Pn are always parsed out
+          if (!((p[0] == 'A' || p[0] == 'E') && p[1] == '1')) break;
+          switch (p[0]) {
+            case 'A': hasA = true; break;
+            case 'E': hasE = true; break;
+          }
+          p += 2;
+          while (*p == ' ') ++p;
+        }
+
+        if (hasE) SERIAL_ECHO_START;
+        if (hasA) SERIAL_ECHO("//");
+
+        SERIAL_ECHOLN(p);
     }
     break;
 
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -6526,7 +6526,7 @@ Sigma_Exit:
     */
     case 118: {
         bool hasE = false, hasA = false;
-        char *p = strchr_pointer;
+        char *p = strchr_pointer + 5;
         
         for (uint8_t i = 2; i--;) {
           // A1, E1, and Pn are always parsed out
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -6515,6 +6515,14 @@ Sigma_Exit:
 
     /*!
     ### M118 - Serial print <a href="https://reprap.org/wiki/G-code#M118:_Echo_message_on_host">M118: Serial print</a>
+    #### Usage
+
+        M118 [ A1 | E1 ] [ String ]
+
+    #### Parameters
+    - `A1` - Prepend // to denote a comment or action command. Hosts like OctoPrint can interpret such commands to perform special actions. See your host’s documentation.
+    - `E1` - Prepend echo: to the message. Some hosts will display echo messages differently when preceded by echo:.
+    - `String` - Message string. If omitted, a blank line will be sent.
     */
     case 118: {
         bool hasE, hasA = false;


gudnimg:
Adding M118 only uses around 58 bytes of flash memory. 👍 Shorter than I expected. It looks like the E and A parameters come from Marlin and not Reprap wiki.

Firmware/Marlin_main.cpp
```diff
@@ -6512,6 +6513,31 @@ SERIAL_PROTOCOLPGM("\n\n");
     }
     break;
 
+    /*!
+    ### M118 - Serial print <a href="https://reprap.org/wiki/G-code#M118:_Echo_message_on_host">M118: Serial print</a>
+    */
```
gudnimg, line 6526: Since we are handling E and A parameters. I think we ought to document what they do (taken from here https://marlinfw.org/docs/gcode/M118.html). You can see a similar format used for other gcodes in `Marlin_main.cpp`.

### M118 - 
    #### Usage
    
        M118 [ A1 | E1 ] [ String ]
    
    #### Parameters
    - `A1` - Prepend // to denote a comment or action command. Hosts like OctoPrint can interpret such commands to perform special actions. See your host’s documentation.
    - `E1` - Prepend echo: to the message. Some hosts will display echo messages differently when preceded by echo:.
    - `String` - Message string. If omitted, a blank line will be sent.
    */

RoboMagus: Thanks for the quick feedback. 
Indeed adding usage information to the comment was something I overlooked. Added now ;)

Firmware/Marlin_main.cpp
```diff
@@ -6512,6 +6513,39 @@ SERIAL_PROTOCOLPGM("\n\n");
...
+          switch (p[0]) {
+            case 'A': hasA = true; break;
+            case 'E': hasE = true; break;
+          }
+          p += 2;
+          while (*p == ' ') ++p;
+        }
+
+        if (hasE) SERIAL_ECHO_START;
+        if (hasA) SERIAL_ECHO("//");
```
DRracer: `SERIAL_ECHOPGM("//")` ... may be we even have this string already

Firmware/Marlin_main.cpp
```diff
@@ -6512,6 +6513,39 @@ SERIAL_PROTOCOLPGM("\n\n");
...
+
+        SERIAL_ECHOLN(p);
```
DRracer, line 6545: This call relies on the fact, that `p` ends with a `\0`. Can we always assume that?

Firmware/Marlin_main.cpp
```diff
@@ -6512,6 +6513,39 @@ SERIAL_PROTOCOLPGM("\n\n");
...
+    - `E1` - Prepend echo: to the message. Some hosts will display echo messages differently when preceded by echo:.
+    - `String` - Message string. If omitted, a blank line will be sent.
+    */
+    case 118: {
+        bool hasE, hasA = false;
+        char *p = strchr_pointer;
+        
+        for (uint8_t i = 2; i--;) {
+          // A1, E1, and Pn are always parsed out
+          if (!((p[0] == 'A' || p[0] == 'E') && p[1] == '1')) break;
```
DRracer, line 6533: This line assumes there are always 2 characters in the input buffer (well, the memory of the buffer is valid all the time, but I'm not sure if it gets cleared after each G-code). Can we presume safely, that the code survives malformed syntax like `M118 A`?

DRracer (changes requested):
Please address my comments in the code, thank you.

Firmware/Marlin_main.cpp
RoboMagus: `get_command()` terminates all strings on either newline characters, start of comments, or when its buffer is full. That should make it safe to use `strchr_pointer` and its derivative `p` as null terminated strings.

Firmware/Marlin_main.cpp
RoboMagus: Could not find occurence of `"//"` elsewhere, but changed to use macro `SERIAL_ECHOPGM`

Firmware/Marlin_main.cpp
RoboMagus: To be honest I don't really like how `A1` and `E1` are used instead of just plain old `A` and `E`, but just to adhere to the protocol that's already out there I would propose sticking to it.
Similarly we should stick to the same method of parsing it, such that the results replicate between stock marlin and Prusa FW. If the reference just prints the parameter flags when the command is malformed so should we.

As it stands, whenever the command flags are malformed, the entire string including the flags is just echoed over serial (depending on what parts of the command are malformed). Based on some preliminary testing the execution of this code should be safe.

I've got a code snipped in [Godbolt.org](https://godbolt.org/z/E859fTTsc) to validate that.

3d-gussner:
@RoboMagus Have you tested your PR?
I merged your PR locally and used the MK404 SIM with OctoPrint to review it.
If you are on Linux please use `./PF-build.sh -m1` to compile the firmware and start the MK404 SIM with this firmware.

Sending `M118 A1 action:cancel` (or any other action command)  wasn't processed by OctoPrint. If I am not mistaken [OctoPrint expects](https://docs.octoprint.org/en/master/features/action_commands.html?highlight=action) `// action:<command>` to trigger anything.

As MINI firmware is based on Marlin 2.x I would keep the syntax `E1` and `A1` as different favors of gcode are always a pain for Slicer developers. If we want to get any chance to get into Prusa Slicer we should not have differences here between MINI and MK2.5/3/S/+ Prusa printers.
Alternative would be to change `E` and `A` parameters in Marlin first and then here. I could imagine that they used the number as a delimiter in the string.    

Please try to test PRs and add a short test log to the PRs, this helps a lot to review these.

Here my test logs:
```
2023-01-25 09:02:07,804 - Recv: echo: 3.13.0-ALPHA1-6894

2023-01-25 09:02:07,813 - Recv: echo: Last Updated: Jan 25 2023 09:01:27 | Author: (none, default config)
2023-01-25 09:02:07,816 - Recv: echo: Free Memory: 2516  PlannerBufferBytes: 1776


2023-01-25 09:04:18,341 - Send: M118 E1 Hello World
2023-01-25 09:04:18,351 - Recv: echo:M118 E1 Hello World


2023-01-25 09:05:20,069 - Send: M118 A1 action:cancel
2023-01-25 09:05:20,073 - Recv: echo:M118 A1 action:cancel

2023-01-25 09:07:18,781 - Send: M118 A1 action:pause
2023-01-25 09:07:18,784 - Recv: echo:M118 A1 action:pause
```

Here the log when pressing Pause in OctoPrint with some G-Cose scripts

```
2023-01-25 09:07:47,197 - Send: M601
2023-01-25 09:07:47,198 - Recv: ok
2023-01-25 09:07:47,199 - Changing monitoring state from "Pausing" to "Paused"
2023-01-25 09:07:47,201 - Recv: echo:enqueing "G1 E-1.000 F2700"
2023-01-25 09:07:47,202 - Recv: // action:paused
2023-01-25 09:07:47,202 - Printer signalled that it paused, switching state...
```

RoboMagus:
Apologies, I had only tested the code snippets separately as I do not have a 'spare' printer available at the moment to test on.
In testing I had not taken into account that the `M118` is still part of the string pointed to by `strchr_pointer`.

I cannot get the MK404 sim to work, but with the offset added should work now.

3d-gussner:
> I cannot get the MK404 sim to work, but with the offset added should work now.

What are your problems with MK404? Are you on Linux? The `./MK404-build.sh` script should guide you through the needed packages and then compile it correctly.

RoboMagus:
I'm using WSL and after working through all missing dependencies and disabling display I cannot get the Serial connection to work.

The script immediately exits:
```
No bootloader specified, or specified file not found. Skipping!
ALSA lib confmisc.c:767:(parse_card) cannot find card '0'
ALSA lib conf.c:4732:(_snd_config_evaluate) function snd_func_card_driver returned error: No such file or directory
ALSA lib confmisc.c:392:(snd_func_concat) error evaluating strings
ALSA lib conf.c:4732:(_snd_config_evaluate) function snd_func_concat returned error: No such file or directory
ALSA lib confmisc.c:1246:(snd_func_refer) error evaluating name
ALSA lib conf.c:4732:(_snd_config_evaluate) function snd_func_refer returned error: No such file or directory
ALSA lib conf.c:5220:(snd_config_expand) Evaluate error: No such file or directory
ALSA lib pcm.c:2642:(snd_pcm_open_noupdate) Unknown PCM default
Failed to open audio: No available audio device
ScriptHost: NOTE: Duplicate context name (Thermistor) with different pointer. Incrementing ID...
ScriptHost: NOTE: Duplicate context name (Thermistor) with different pointer. Incrementing ID...
ScriptHost: NOTE: Duplicate context name (Thermistor) with different pointer. Incrementing ID...
ScriptHost: NOTE: Duplicate context name (Fan) with different pointer. Incrementing ID...
Creating pinspec for atmega2560
Prusa_MK3S_flash.bin: Read 262144 bytes.
Loading 262144 bytes of xflash
Loading 4096 bytes of EEPROM
Read 4096 bytes
Loaded 234976 bytes from HEX file: /mnt/c/Projects/Prusa-Firmware/../PF-build-hex/FW3130-ALPHA1-Build6054/BOARD_EINSY_1_0a/FW3130-ALPHA1-Build6054-1_75mm_MK3S-EINSy10a-E3Dv6full-EN_FARM.hex
Initialized VCD file Prusa_MK3S_VCD.vcd
uart_pty_init bridge on port *** /dev/pts/1 ***
/tmp/simavr-uart2 now points to: /dev/pts/1
note: export SIMAVR_UART_XTERM=1 and install picocom to get a terminal
uart_pty_init bridge on port *** /dev/pts/2 ***
uart_pty_init bridge on port *** /dev/pts/3 ***
SD file Prusa_MK3S_SDcard.bin does not exist. Will not create it.
SD card image (Prusa_MK3S_SDcard.bin) could not be mounted (error -1 ).
Init on ADC 0 start temp: 25
Init on ADC 2 start temp: 25
Init on ADC 3 start temp: 25
Init on ADC 6 start temp: 25
ScriptHost: NOTE: Duplicate context name (VSrc) with different pointer. Incrementing ID...
MK3S - adding IR sensor.
/tmp/simavr-uart0 now points to: /dev/pts/2
note: export SIMAVR_UART_XTERM=1 and install picocom to get a terminal
/tmp/simavr-uart1 now points to: /dev/pts/3
note: export SIMAVR_UART_XTERM=1 and install picocom to get a terminal
Validating script...
Script validation finished.
Waiting for board to finish...
Starting atmega2560 execution...
10M idle cycles is 346184 ns (0.346184 ns per tick)
atmega2560finished (2).
Wrote 262144 bytes of flash
Wrote 4096 bytes of EEPROM to Prusa_MK3S_eeprom.bin
Wrote 262145 bytes of xflash to Prusa_MK3S_xflash.bin
Stopping Einsy_atmega2560
Done
~uart_pty
~uart_pty
~uart_pty
Done
```

3d-gussner:
WSL or WSL2? I heard that the new version works better. I used to use WSL and at the end I decided for dual boot and now barely use Windows.
Did you build it in WSl using `./MK404-build.sh` ?

RoboMagus:
WSL2 indeed, which is better than nothing, but still causes too many LFS / Device related issues for my liking.
Meanwhile I do have the simulator running, but no UART interaction seems to be possible. (The issues seem to be related with my system and MK404 than with the changes in this PR)

That is after building with the MK404 build script.

I'm fairly confident though that with the latest commit the code in this PR the command should work as expected. Do you have the chance to check again if behavior is now as expected?

If not possible I'll hope to get to take some time during this weekend to flash the new FW on printer, as otherwise I could not test this myself.

RoboMagus:
Had one minor fix left ('M' also still part of `strchr_pointer` which I had not taken into account), but these changes have successfully tested on my Prusa MK3s together with Octoprint.

Here's a snippet from the Terminal:
```
[...]
Send: M118 Here is a test string that is echoed to Host from Prusa MK3s
Recv: Here is a test string that is echoed to Host from Prusa MK3s
Recv: ok
[...]
Send: M118 E1 Here is a test string that is ECHOED to Host from Prusa MK3s
Recv: echo:Here is a test string that is ECHOED to Host from Prusa MK3s
Recv: ok
[...]
Send: M118 A1 action:pause
Recv: //action:pause
Pausing on request of the printer...
Recv: ok
[...]
Send: M118 A1 E1 Here's a string with A1, E1
Recv: echo://Here's a string with A1, E1
Recv: ok
[...]
Send: M118 E1 E1 Here's a string with duplicate E1
Recv: echo:Here's a string with duplicate E1
Recv: ok
[...]
Send: M118 E1E2 Here's malformed command string 1
Recv: echo:E2 Here's malformed command string 1
Recv: ok
[...]
Send: M118 A2 Here's malformed command string 2
Recv: A2 Here's malformed command string 2
Recv: ok
[...]
Send: M118  A1 Here's malformed command string with an additional space before the parameter
Recv:  A1 Here's malformed command string with an additional space before the parameter
Recv: ok
[...]
Send: M118  E1 Here's malformed command string with an additional space before the parameter
Recv:  E1 Here's malformed command string with an additional space before the parameter
Recv: ok
[...]
Send: M118 E1 Here's malformed command string with an additional space before the parameter
```

RoboMagus:
@DRracer, After some additional testing with printing files containing `M118` commands from SD card, I noticed that these lines are split (this is not the case when performing `M118` over Serial).

[This line](https://github.com/prusa3d/Prusa-Firmware/blob/51374965e413c18668d88df9056ca60bee6ed2ec/Firmware/cmdqueue.cpp#L548) appears to be the culprit, but I do not understand why a colon (`:`) would also be used as the start of comment in GCode. Cause any file I've seen thus far uses only semicolon (`;`) to indicate start of comment. This appears to be a bug.
~Seeing as you authored this bit of code~, could you clarify what's happening here please?

*Edit:*
Seems like this colon has been there since the code was ported from Marlin 1.x and appears to be the definition of _legacy_ code. There was a bug filed for this back in 2015: [Ticket by Foosel](https://github.com/MarlinFirmware/Marlin/issues/1568), and subsequently fixed in PR [1570](https://github.com/MarlinFirmware/Marlin/pull/1570).

I would propose to port these fixes also, as the behavior as it currently stands is not intended.

DRracer:
@RoboMagus thank you for your effort getting your PR tuned and merged.

You are right, the `:` check is some kind of legacy code. Theoretically we could implement escaped sequences just like [PR#1570](https://github.com/MarlinFirmware/Marlin/pull/1570) .

However, I'd like to avoid touching the G-code skipping state machine, it took us a lot of effort to get it up to the speed where it is now. That implies that escaping a semicolon `;` will not be supported.

RoboMagus:
The colon issue should be fixed by PR #3978. This indeed does not add any escape sequence handling to avoid further complication of the GCode reading. I have not yet found any GCode that includes `;` in any other way than just comments.

With that little bug squashed the M118 implementation is working as expected on my MK3s at least.

RoboMagus:
@DRracer,  are these proposed changes OK to be merged?
Should include #3978 as well then.

gudnimg:
@RoboMagus Sorry for the late review/response. But I'll try look at your PR over the weekend (and test on my printer). 👍 It looks good to me just by looking at the code but I want to test it on the hardware to understand the changes better.

RoboMagus:
Thanks for the update @gudnimg. 
I figured it could be delayed as we're waiting for the next FW version before merging, but good to see its on the radar.

I've been running modified firmware on my MK3s  for over a month with these changes incorporated (and have been extensively using with a W.I.P. Octoprint plugin) and have not ran into issues yet. Hope your testing yields the same results 😃.

gudnimg:
Current memory changes:

Flash: +190 bytes
SRAM: 0 bytes

I wonder why it needs so much memory 🤔

gudnimg:
I tried my best to make M118 fail but these results look good me. I think we should try to reduce the memory usage if we can. Note the timestamps below are just from Arduino IDE, not from firmware.

I noticed M118 will skip spaces in the beginning of the message:

`M118 A1 E1                 hello world`
Yields:
```
11:05:41.865 -> echo://hello world
11:05:41.865 -> ok
```
Not sure if this is expected behavior.

Tests:
Input `M118`
Output: 
```
10:46:04.530 -> 
10:46:04.530 -> ok
```
----
Input `M118 A`
Output:
```
10:47:32.066 -> A
10:47:32.066 -> ok
```
----
Input `M118 E`
Output:
```
10:48:05.096 -> E
10:48:05.096 -> ok
```
----
Input `M118 A256`
Output:
```
10:48:45.314 -> A256
10:48:45.314 -> ok
```
----
Input `M118 A128`
Output:
```
10:51:25.372 -> //28
10:51:25.372 -> ok
```
----
Input `M118 A1 E192391932`
Output:
```
10:52:07.666 -> echo://92391932
10:52:07.666 -> ok
```
----
Input: `M118 A1 E1 Averyveryveryveryveryverylongstring`
Output:
```
10:54:35.958 -> echo://Averyveryveryveryveryverylongstring
10:54:35.958 -> ok
```
----
Input `M118 *`
Output:
```
11:01:39.973 -> Error:No Line Number with checksum, Last Line: 0
11:01:39.973 -> Resend: 1
11:01:39.973 -> ok
```

gudnimg:
Our new CMake builds also give us the .asm output. Here it is for this PR just to share it:

```asm
case 118: {
        bool hasE = false;
        bool hasA = false;
        char *p = strchr_pointer + 5;
   1a54e:	00 91 4c 03 	lds	r16, 0x034C	; 0x80034c <strchr_pointer>
   1a552:	10 91 4d 03 	lds	r17, 0x034D	; 0x80034d <strchr_pointer+0x1>
   1a556:	0b 5f       	subi	r16, 0xFB	; 251
   1a558:	1f 4f       	sbci	r17, 0xFF	; 255
   1a55a:	83 e0       	ldi	r24, 0x03	; 3
    - `E1` - Prepend echo: to the message. Some hosts will display echo messages differently when preceded by echo:.
    - `String` - Message string. If omitted, a blank line will be sent.
    */
    case 118: {
        bool hasE = false;
        bool hasA = false;
   1a55c:	f1 2c       	mov	r15, r1
    - `A1` - Prepend // to denote a comment or action command. Hosts like OctoPrint can interpret such commands to perform special actions. See your host’s documentation.
    - `E1` - Prepend echo: to the message. Some hosts will display echo messages differently when preceded by echo:.
    - `String` - Message string. If omitted, a blank line will be sent.
    */
    case 118: {
        bool hasE = false;
   1a55e:	40 e0       	ldi	r20, 0x00	; 0
   1a560:	81 50       	subi	r24, 0x01	; 1
        bool hasA = false;
        char *p = strchr_pointer + 5;
        
        for (uint8_t i = 2; i--;) {
   1a562:	e9 f0       	breq	.+58     	; 0x1a59e <process_commands() [clone .part.41]+0x3d7e>
          // A1, E1, and Pn are always parsed out
          if (!((p[0] == 'A' || p[0] == 'E') && p[1] == '1')) break;
   1a564:	d8 01       	movw	r26, r16
   1a566:	9c 91       	ld	r25, X
   1a568:	29 2f       	mov	r18, r25
   1a56a:	2b 7f       	andi	r18, 0xFB	; 251
   1a56c:	21 34       	cpi	r18, 0x41	; 65
   1a56e:	b9 f4       	brne	.+46     	; 0x1a59e <process_commands() [clone .part.41]+0x3d7e>
   1a570:	11 96       	adiw	r26, 0x01	; 1
   1a572:	2c 91       	ld	r18, X
   1a574:	21 33       	cpi	r18, 0x31	; 49
   1a576:	99 f4       	brne	.+38     	; 0x1a59e <process_commands() [clone .part.41]+0x3d7e>
          switch (p[0]) {
   1a578:	91 34       	cpi	r25, 0x41	; 65
   1a57a:	71 f0       	breq	.+28     	; 0x1a598 <process_commands() [clone .part.41]+0x3d78>
   1a57c:	95 34       	cpi	r25, 0x45	; 69
   1a57e:	09 f4       	brne	.+2      	; 0x1a582 <process_commands() [clone .part.41]+0x3d62>
            case 'A': hasA = true; break;
            case 'E': hasE = true; break;
   1a580:	41 e0       	ldi	r20, 0x01	; 1
          }
          p += 2;
   1a582:	98 01       	movw	r18, r16
   1a584:	2e 5f       	subi	r18, 0xFE	; 254
   1a586:	3f 4f       	sbci	r19, 0xFF	; 255
   1a588:	89 01       	movw	r16, r18
   1a58a:	2f 5f       	subi	r18, 0xFF	; 255
   1a58c:	3f 4f       	sbci	r19, 0xFF	; 255
          while (*p == ' ') ++p;
   1a58e:	f8 01       	movw	r30, r16
   1a590:	90 81       	ld	r25, Z
   1a592:	90 32       	cpi	r25, 0x20	; 32
   1a594:	c9 f3       	breq	.-14     	; 0x1a588 <process_commands() [clone .part.41]+0x3d68>
   1a596:	e4 cf       	rjmp	.-56     	; 0x1a560 <process_commands() [clone .part.41]+0x3d40>
        
        for (uint8_t i = 2; i--;) {
          // A1, E1, and Pn are always parsed out
          if (!((p[0] == 'A' || p[0] == 'E') && p[1] == '1')) break;
          switch (p[0]) {
            case 'A': hasA = true; break;
   1a598:	ff 24       	eor	r15, r15
   1a59a:	f3 94       	inc	r15
   1a59c:	f2 cf       	rjmp	.-28     	; 0x1a582 <process_commands() [clone .part.41]+0x3d62>
          }
          p += 2;
          while (*p == ' ') ++p;
        }

        if (hasE) SERIAL_ECHO_START;
   1a59e:	44 23       	and	r20, r20
   1a5a0:	21 f0       	breq	.+8      	; 0x1a5aa <process_commands() [clone .part.41]+0x3d8a>
   1a5a2:	8c ef       	ldi	r24, 0xFC	; 252
   1a5a4:	9b ea       	ldi	r25, 0xAB	; 171
   1a5a6:	0e 94 f6 89 	call	0x113ec	; 0x113ec <serialprintPGM(char const*)>
        if (hasA) SERIAL_ECHOPGM("//");
   1a5aa:	ff 20       	and	r15, r15
   1a5ac:	21 f0       	breq	.+8      	; 0x1a5b6 <process_commands() [clone .part.41]+0x3d96>
   1a5ae:	86 e2       	ldi	r24, 0x26	; 38
   1a5b0:	96 e8       	ldi	r25, 0x86	; 134
   1a5b2:	0e 94 f6 89 	call	0x113ec	; 0x113ec <serialprintPGM(char const*)>

        SERIAL_ECHOLN(p);
   1a5b6:	c8 01       	movw	r24, r16
   1a5b8:	0e 94 5a 94 	call	0x128b4	; 0x128b4 <MarlinSerial::println(char const*)>
   1a5bc:	0c 94 6e c4 	jmp	0x188dc	; 0x188dc <process_commands() [clone .part.41]+0x20bc>
```

RoboMagus:
> Current memory changes:
> 
> Flash: +190 bytes SRAM: 0 bytes
> 
> I wonder why it needs so much memory 🤔

How do you find the increase in mem usage? 
It's indeed strange that that's such a significant increase for such a relatively small addition. Especially since about 2 months ago this increase was __MUCH__ smaller (your first comment):
> Adding M118 only uses around 58 bytes of flash memory

All I can think of that's been changed in commits since then has been to use `SERIAL_ECHOPGM` i.s.o. `SERIAL_ECHO` 🤔.

RoboMagus:
> I noticed M118 will skip spaces in the beginning of the message:
> ...
> Not sure if this is expected behavior.

This behaviour is copied from the [Marlin implementation](https://github.com/MarlinFirmware/Marlin/blob/241c06218822d63e47710acfa6885351fe495bc5/Marlin/src/gcode/host/M118.cpp#L52) to ensure behaviour does not deviate between implementations.

I cannot think of a usecase where spaces would be usefull to retain myself, but if there would be one it'd be best in my opinion to keep implementation consistent rather than deviate based edge usecases.

gudnimg:
> How do you find the increase in mem usage?

When I build the firmware via Vscode + CMake I get this result:

Keep in mind the number for the memory in Flash (label Program in the image) doesn't include the bootloader which eats up 8192 bytes of the free memory. As you can see we don't have that much memory free which is why I'm constantly looking for optimisations so we can squeeze more into the firmware 😅 
image

I usually build the latest current MK3 branch, take a screenshot of the result. Then switch to the branch I'm working on and use my screenshot to compare with the result on my branch.

Note I am applying your changes to the latest MK3 branch, maybe something changed since you opened the PR which is increasing the code size but I don't see how.

I have found this alone takes 70B? hmm
`char *p = strchr_pointer + 5;`

I've also found one potential optimisation which may help in this PR. And that's to create a common function to skip whitespace such that:

`while (*src == ' ') ++src;` becomes `src = cmd_skip_whitespace(src);` I'm looking into it now.

gudnimg:
I'm seeing this only is adding 70B 🤔 Strange the compiler doesn't optimize this away.
image

gudnimg:
It looks like the memory difference is due to the compiler or linker optimizing differently. So the issue is not your PR :) Probably some function is being inlined somewhere by LTO (not sure how to find it though, the asm file is huge).

gudnimg (approved):

RoboMagus:
@DRracer / @3d-gussner  
Is there still work required for this PR to be merged?

DRracer (approved):
Now with some more space we can finally merge this PR.
The proposed changes are scope-limited.

Future improvements may include:
- extracting M118 implementation away from Marlin_main.cpp (which may help the compiler to optimize the code better)
- adding unit tests to see how malformed M118 records manage to break the whole thing

DRracer:
@RoboMagus Thank you for keeping your PR up to date and responding to our reviews.

PR #3947, opened 2023-01-24
Add active sheet reporting to M503

RoboMagus:
Adds active sheet index and name reporting to M503.

Currently there is no way for connected host (Octoprint) to retrieve active sheet information.
For preventive checks / pauses / user prompts when changing materials and thus sheets, retrieving active sheet information is required.

Adding this information to M503 instead off comming up with a separate GCode made the most sense.

### DIFF
--- a/Firmware/ConfigurationStore.cpp
+++ b/Firmware/ConfigurationStore.cpp
@@ -106,6 +106,15 @@ void Config_PrintSettings(uint8_t level)
 #ifdef THERMAL_MODEL
     thermal_model_report_settings();
 #endif
+    // Report Active sheet
+    {
+      const int8_t sheetNR = eeprom_read_byte(&(EEPROM_Sheets_base->active_sheet));
+      char sheetName[8] = {0};
+      eeprom_read_block(sheetName, EEPROM_Sheets_base->s[sheetNR].name, 7);
+      printf_P(PSTR(
+          "%SActive sheet: %d %s\n"),
+          echomagic, sheetNR, sheetName);
+    }
 }
 #endif
 


RoboMagus:
Just after this PR I discovered M862, so opened a different PR as that GCode might be more suitable for this addition. 

Keeping both open for now until feedback on which path to take.

wavexx:
Commenting here for either:

> For preventive checks / pauses / user prompts when changing materials and thus sheets, retrieving active sheet information is required.

How do you store the current/expected sheet normally (in the slicer/octoprint)? I'm curious in how this helps.
I sometimes swap the sheet but forget to change it on the printer. I have written labels on the sheets to remind me of this, but it doesn't really help ;)

RoboMagus:
Currently there is not yet any way to track this in PrusaSlicer or Octoprint as far as I'm aware. But that would actually not make sense until the ability to retreive the information from the printer and verify this exists.

That's what I hope to achieve with this (or the other) PR.

I was actually working on a plugin for Octoprint to remind me of switching sheets based on the material used for a new print, but without a way of communicating this with the printer quickly stalled my progress.

In PrusaSlicer there is a bit of GCode injected to set the Nozzle Size used when slicing, which could then show a prompt on the printer when it does not match with what's currently used on the printer. I could imagine similar settings implemented in PrusaSlicer for the steel sheets, but only after functionality is available in FW first. 
Though that would actually make PR #3948 the better choice instead of this one 😄.

3d-gussner:
@RoboMagus can you please rebase this PR to see how many resources it uses? Or close it if you think that https://github.com/prusa3d/Prusa-Firmware/pull/3948 is better as you write above.

RoboMagus:
@3d-gussner, For my personal usecase (doing basically everything through Octoprint) this addition to `M503` is just fine. Of the 2 options this is probably also the smallest cost.
However for workflows that rely more on manual printer interaction the case can be made for the advantage of provided checks in `M862`.

vintagepc:
FYI the ability to query and  set an active sheet is available already as part of https://github.com/prusa3d/Prusa-Firmware/pull/4328

RoboMagus:
Closing in favor of #4567

RoboMagus:
Closing in favor of #4567

Issue #3482, opened 2022-04-24
[FEATURE REQUEST] Optional Steel Sheet confirm before print starts 


livingkurt:
**Problem:**
I go from a textured steel sheet to a smooth steel sheet a lot in my day to day. Every so often I forget to switch the steel sheet setting on the printer from textured sheet to smooth sheet while a smooth sheet remains on the printer. Since the textured sheet is thinner than the smooth, and based on the first layer settings I like for each sheet, this causes the nozzle to dig into my smooth sheet sometimes making the sheet unusable. It doesn't happen too often but when it does it is a real bummer. I try my best to remember to double check before starting the print but I can't 100% prevent it from happening again.
  
**Solution:**
I am suggesting an optional feature that you can set through the "Settings" on the printer itself, that would be an On/Off Boolean setting, called something like "Steel Sheet Check". 
This check is to prevent the above issue from ever happening for those who wish to be notified.

**How it will work:**
When this setting is on, the effect will take place right after the bed and nozzle heat up and before the homing and bed leveling.
You will be welcomed with a message: 

Is this steel sheet correct? : <current_steel_sheet>"
Yes
Change setting
Change steel sheet

**Yes**
If you select "Yes" the print will continue as if nothing happened.

**Change setting**
If you select "Change setting" the steel sheet menu will show up on the printer allowing you to choose the steel sheet for your print. 

Smooth1
Texture1

Once selected the same "Is this steel sheet correct? : <current_steel_sheet>" menu will show up in case you need to change the steel sheet as well. If everything looks correct then you can click Yes.

**Change steel sheet**
If you select "Change steel sheet" the printer head will move out of the way and the Y-axis will put the sheet forward to allow for the steel sheet replacement. A confirm will now show that you will select when you have finished replacing your steel sheet. 

Change steel sheet
Confirm

Once confirmed the "Is this steel sheet correct? : <current_steel_sheet>" menu will show up in case you need to change the settings as well. If everything looks correct then you can click Yes.

**Final Thoughts**
Overall I think this would be the only way to 100% prevent the nozzle from digging into my precious smooth steel sheet.

PR #3948, opened 2023-01-24
Add active steel sheet check to M862

RoboMagus:
Adds check / query for active steel sheet to M862 as `M862.6`.

Alternative to my other PR #3947, which adds the same information to M503. 
Just after creating that PR I discovered M862 which could be the beter option of the 2.

I'll leave both PRs open for now until feedback provides insight on which of the routes to take.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -7992,7 +7992,7 @@ Sigma_Exit:
       - M862.3 { P"<model_name>" | Q }
       - M862.4 { P<fw_version> | Q }
       - M862.5 { P<gcode_level> | Q }
-      - M862.6 { P<active_sheet> | Q }
+      - M862.7 { P<active_sheet> | Q }
     
     When run with P<> argument, the check is performed against the input value.
     When run with Q argument, the current value is shown.
@@ -8075,7 +8075,7 @@ Sigma_Exit:
                     else if(code_seen('Q'))
                          SERIAL_PROTOCOLLN(GCODE_LEVEL);
                     break;
-               case ClPrintChecking::_ActiveSheet:      // ~ .6
+               case ClPrintChecking::_ActiveSheet:      // ~ .7
                     if(code_seen('P')) {
                       uint8_t sheetIdx = code_value_uint8();
                       active_sheet_check(sheetIdx);
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -7992,6 +7992,7 @@ Sigma_Exit:
       - M862.3 { P"<model_name>" | Q }
       - M862.4 { P<fw_version> | Q }
       - M862.5 { P<gcode_level> | Q }
+      - M862.6 { P<active_sheet> | Q }
     
     When run with P<> argument, the check is performed against the input value.
     When run with Q argument, the current value is shown.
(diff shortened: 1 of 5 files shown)


DRracer:
@RoboMagus this is an interesting idea :+1: . We must think what it may break, but generally I like it.

RoboMagus:
@DRracer, I also just came accros Issue #3482, which got me thinking to provide a multi choice prompt on the display when the sheet in EEPROM does not match the expected value in GCode:
- Cancel print
- Change active sheet to expected value (after user switches the sheet)
- Continue with current settings.

The main problem I see there is the lack of screen realestate and the inability to insert sheet names in the messages while keeping multi language options.

But after seeing @livingkurt usecase I would see a huge benefit in having such an option at the start of a print (Even though that was not my initial goal with this PR)

Firmware/Marlin_main.cpp
```diff
@@ -7893,6 +7893,7 @@ SERIAL_PROTOCOLPGM("\n\n");
       - M862.3 { P"<model_name>" | Q }
       - M862.4 { P<fw_version> | Q }
       - M862.5 { P<gcode_level> | Q }
+      - M862.6 { P<active_sheet> | Q }
```
3d-gussner: Needs to be changed as https://reprap.org/wiki/G-code#M862.6:_Firmware_features is used by buddy firmware

Firmware/util.h
```diff
@@ -31,7 +31,8 @@ enum class ClPrintChecking:uint_least8_t
     _Model=2,
     _Smodel=3,
     _Version=4,
-    _Gcode=5
+    _Gcode=5,
+    _ActiveSheet=6
```
3d-gussner: see above

3d-gussner (changes requested):
Thanks, please rebase this PR and fix the comments. We like the idea and thinking of merging it.

Firmware/Marlin_main.cpp
RoboMagus: Updated to sub-code 7 instead

Firmware/util.cpp
```diff
@@ -389,6 +389,29 @@ void gcode_level_check(uint16_t nGcodeLevel) {
...
+    // SERIAL_ECHOPGM("actual  : ");
+    // SERIAL_ECHOLN(activeSheetIdx);
+    // SERIAL_ECHOPGM("expected: ");
+    // SERIAL_ECHOLN(sheetIdx);
+        
+    render_M862_warnings(
+        _T(MSG_GCODE_DIFF_SHEET_CONTINUE)
+        ,_T(MSG_GCODE_DIFF_SHEET_CANCELLED)
+        ,(uint8_t)oCheckMode
+    );
```
RoboMagus, line 412: I think these 2 choices may not be enough and / or may be confusing for the end user.

I've been toying with the idea of multiple choises here. A rough implementation as such: 
```

    if (oCheckMode == 1) { // Warning, stop print if user selects 'No'
        bool requestedSheetInitialized = eeprom_is_sheet_initialized(sheetIdx);
        uint8_t choice = lcd_show_multiscreen_message_with_choices_and_wait_P(_i("Current active sheet differs from the G-Code."), true, 0,
            _i("Cancel print."));
            _i("Ignore and continue with current sheet."),
            (requestedSheetInitialized ? _i("Requested sheet installed. Update EEPROM and continue.") : nullptr),
        if(choice == 2) {
            eeprom_update_byte(&(EEPROM_Sheets_base->active_sheet), sheetIdx); // and continue print
        }
        else if(choice == 0 || choice == 0xFF) { // User cancel or Timeout
            lcd_print_stop();
        }
    } else if (oCheckMode == 2) { // Strict, always stop print
        lcd_show_fullscreen_message_and_wait_P(_i("Current active sheet differs from the G-code. Print cancelled."));
        lcd_print_stop();
    }
```

This at least gives the user the added option for changing to the requested sheet on the spot instead of having to cancel and re-start the print.

The remaining problem is the lack of verbosity about the current sheet and the requested sheet. Being able to print the names would be great, but that would cost too much. Let allone the issue in doing so for other languages....

Oh, and screen realestate too!

Any thoughts on this?

3d-gussner:
@RoboMagus Please check if https://github.com/prusa3d/Prusa-Firmware/pull/4567 is sufficient to get similar function you are looking for.

RoboMagus:
> @RoboMagus Please check if #4567 is sufficient to get similar function you are looking for.

Thanks for the heads up. I'm currently unable to fully look into or test this, but at first glance this looks like it would suit my use-case. I'll close this as it looks like #4567 should do just fine.

PR #3950, opened 2023-01-25
Refactor mmu2.cpp and related interfaces

DRracer:
The general idea is to keep platform specific implementation away from the MMU state machines as much as we can. That would enable unit testing the top level MMU state machine and integration into other projects (Marlin2) as well if needed.

### DIFF
--- a/Firmware/mmu2.cpp
+++ b/Firmware/mmu2.cpp
@@ -3,45 +3,43 @@
 #include "mmu2_error_converter.h"
 #include "mmu2_fsensor.h"
 #include "mmu2_log.h"
+#include "mmu2_marlin.h"
+#include "mmu2_marlin_macros.h"
 #include "mmu2_power.h"
 #include "mmu2_progress_converter.h"
 #include "mmu2_reporting.h"
 
-#include "cardreader.h" // for IS_SD_PRINTING
-#include "Marlin.h"
-#include "language.h"
-#include "messages.h"
-#include "sound.h"
-#include "stepper.h"
 #include "strlen_cx.h"
-#include "temperature.h"
-#include "ultralcd.h"
 #include "SpoolJoin.h"
 
+#ifdef __AVR__
 // As of FW 3.12 we only support building the FW with only one extruder, all the multi-extruder infrastructure will be removed.
 // Saves at least 800B of code size
 static_assert(EXTRUDERS==1);
+
+constexpr float MMM_TO_MMS(float MM_M){ return MM_M / 60.0f; }
+#endif
 
 namespace MMU2 {
 
 template<typename F>
 void waitForHotendTargetTemp(uint16_t delay, F f){
-    while (((degTargetHotend(active_extruder) - degHotend(active_extruder)) > 5)) {
+    while (((thermal_degTargetHotend() - thermal_degHotend()) > 5)) {
         f();
-        delay_keep_alive(delay);
+        safe_delay_keep_alive(delay);
     }
 }
 
 void WaitForHotendTargetTempBeep(){
     waitForHotendTargetTemp(3000, []{ });
-    Sound_MakeSound(e_SOUND_TYPE_StandardPrompt);
+    MakeSound(Prompt);
 }
 
 MMU2 mmu2;
 
 MMU2::MMU2()
     : is_mmu_error_monitor_active(false)
-    , logic(&mmu2Serial, MMU2_TOOL_CHANGE_LOAD_LENGTH)
+    , logic(&mmu2Serial, MMU2_TOOL_CHANGE_LOAD_LENGTH, MMU2_LOAD_TO_NOZZLE_FEED_RATE)
     , extruder(MMU2_NO_TOOL)
     , tool_change_extruder(MMU2_NO_TOOL)
     , resume_position()
@@ -52,24 +50,15 @@ MMU2::MMU2()
     , loadFilamentStarted(false)
     , unloadFilamentStarted(false)
     , loadingToNozzle(false)
-    , inAutoRetry(false)
-    , retryAttempts(MAX_RETRIES)
     , toolchange_counter(0)
     , tmcFailures(0)
 {
 }
 
 void MMU2::Start() {
-#ifdef MMU_HWRESET
-    WRITE(MMU_RST_PIN, 1);
-    SET_OUTPUT(MMU_RST_PIN); // setup reset pin
-#endif //MMU_HWRESET
-
     mmu2Serial.begin(MMU_BAUD);
 
     PowerOn(); // I repurposed this to serve as our EEPROM disable toggle.
-    Reset(ResetForm::ResetPin);
-
     mmu2Serial.flush(); // make sure the UART buffer is clear before starting communication
 
     extruder = MMU2_NO_TOOL;
@@ -78,7 +67,7 @@ void MMU2::Start() {
     // start the communication
     logic.Start();
 
-    ResetRetryAttempts();
+    logic.ResetRetryAttempts();
 }
 
 void MMU2::Stop() {
@@ -94,10 +83,17 @@ void MMU2::StopKeepPowered(){
 
 void MMU2::Reset(ResetForm level){
     switch (level) {
-    case Software: ResetX0(); break;
-    case ResetPin: TriggerResetPin(); break;
-    case CutThePower: PowerCycle(); break;
-    default: break;
+    case Software:
+        ResetX0();
+        break;
+    case ResetPin:
+        TriggerResetPin();
+        break;
+    case CutThePower:
+        PowerCycle();
+        break;
+    default:
+        break;
     }
 }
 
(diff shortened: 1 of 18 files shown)


Firmware/mmu2.h
```diff
@@ -172,52 +163,46 @@ class MMU2 {
...
-            return { 0, 0, 0}; 
+            return { 0, 0, 0 };
         }
     }
 
     // Helper variable to monitor knob in MMU error screen in blocking functions e.g. manage_response
     bool is_mmu_error_monitor_active;
 
     /// Method to read-only mmu_print_saved
-    inline bool MMU_PRINT_SAVED() const { return mmu_print_saved != SavedState::None; }
```
gudnimg: Any specific reason this `inline` was removed? I don't see any change in memory so the compiler is probably inlining this already.

I propose to add `inline` back here to reduce the git diff.

Firmware/mmu2_log.h
```diff
@@ -20,18 +20,36 @@ void LogEchoEvent_P(const char *msg);
...
+            SERIAL_ECHO_START;   \
+            SERIAL_MMU2();       \
+            SERIAL_ECHOLN(S);    \
+        } while (0)
+    #define MMU2_ERROR_MSGLN(S) MMU2_ECHO_MSGLN(S) //!@todo Decide MMU2 errors  on serial line
+    #define MMU2_ECHO_MSGRPGM(S) \
+        do {                     \
+            SERIAL_ECHO_START;   \
+            SERIAL_MMU2();       \
+            SERIAL_ECHO(S);      \
```
gudnimg: ⚠️ This must be `SERIAL_ECHORPGM(S)` for PGM strings, otherwise the serial message is corrupted. This negatively impacts the printer and it appears to run more slowly due the corruption seems to "spam" the serial stream.

gudnimg (changes requested):

Firmware/mmu2.cpp
```diff
@@ -861,49 +819,56 @@ StepStatus MMU2::LogicStep(bool reportErrors) {
...
-        plan_buffer_line_curposXYZE(pgm_read_float(&(step->feedRate)));
-        st_synchronize();
+        const float es = pgm_read_float(&(step->extrude));
+        const feedRate_t fr_mm_m = pgm_read_float(&(step->feedRate));
+        planner_set_current_position_E(planner_get_current_position_E() + es);
+        planner_line_to_current_position(MMM_TO_MMS(fr_mm_m));
+
         step++;
     }
+    planner_synchronize(); // it looks like it's better to sync the moves at the end - smoother move (if the sequence is not too long).
```
gudnimg, line 845: `st_synchronize()` is included with `planner_line_to_current_position()` so I think `planner_synchronize()` doesn't help, only increases code size.

gudnimg (changes requested):

Firmware/mmu2.cpp
```diff
@@ -861,49 +819,56 @@ StepStatus MMU2::LogicStep(bool reportErrors) {
...
+
     const E_Step *step = sequence;
     for (uint8_t i = 0; i < steps; i++) {
-        current_position[E_AXIS] += pgm_read_float(&(step->extrude));
-        plan_buffer_line_curposXYZE(pgm_read_float(&(step->feedRate)));
-        st_synchronize();
+        const float es = pgm_read_float(&(step->extrude));
+        const feedRate_t fr_mm_m = pgm_read_float(&(step->feedRate));
+        planner_set_current_position_E(planner_get_current_position_E() + es);
+        planner_line_to_current_position(MMM_TO_MMS(fr_mm_m));
```
gudnimg: ⚠️ All the extruder sequences are already in units mm/s. `MMM_TO_MMS()` here will make the feedrate 60 times slower which explains why I'm seeing "Load to Nozzle" operations take over 4 minutes to complete :) 

Either remove the `/ 60.F` from the extruder sequences, or remove `MMM_TO_MMS()`.

ℹ️ Removing `MMM_TO_MMS()` saves 12 bytes of flash.

gudnimg (changes requested):

Firmware/mmu2_log.h
gudnimg: Fixed in [f79dfac](https://github.com/prusa3d/Prusa-Firmware/pull/3950/commits/f79dfac133e99d4335e6aaa3eef6c4e937db7a45)

Firmware/mmu2_protocol_logic.cpp
```diff
@@ -495,7 +509,7 @@ StepStatus ProtocolLogic::IdleStep() {
     return Finished;
 }
 
-ProtocolLogic::ProtocolLogic(MMU2Serial *uart, uint8_t extraLoadDistance)
+ProtocolLogic::ProtocolLogic(MMU2Serial *uart, uint8_t extraLoadDistance, uint8_t pulleySlowFeedrate)
```
gudnimg, line 512: Just a thought. If we add more registers here, adding a new parameter each time isn't very scalable. We should consider  creating a struct of some kind to reduce the parameters 🤔

Firmware/mmu2_protocol_logic.cpp
```diff
@@ -510,12 +524,15 @@ ProtocolLogic::ProtocolLogic(MMU2Serial *uart, uint8_t extraLoadDistance)
...
-    , regs8 { 0, 0, 0 }
-    , regs16 { 0, 0 }
-    , initRegs8 { extraLoadDistance }
     , regIndex(0)
-    , mmuFwVersion { 0, 0, 0 }
-{}
+    , retryAttempts(MAX_RETRIES)
+    , inAutoRetry(false) {
+    // @@TODO currently, I don't see a way of writing the initialization better :(
+    // I'd like to write something like: initRegs8 { extraLoadDistance, pulleySlowFeedrate }
```
gudnimg, line 531: Could we pass a struct by reference?

Firmware/mmu2_protocol_logic.h
```diff
@@ -330,31 +362,27 @@ class ProtocolLogic {
     static constexpr uint8_t regs8Count = 3;
     static_assert(regs8Count > 0); // code is not ready for empty lists of registers
     static const uint8_t regs8Addrs[regs8Count] PROGMEM;
-    uint8_t regs8[regs8Count];
+    uint8_t regs8[regs8Count] = { 0, 0, 0 };
```
gudnimg, line 365: is `regs8` not zero initialised by the compiler? Adding {0, 0, 0} increases code size by 12 bytes.

Same applies to `regs16` and `mmuFwVersion`

Firmware/mmu2_marlin1.cpp
```diff
@@ -0,0 +1,123 @@
...
+        current_position[Z_AXIS] = z;
+        plan_buffer_line_curposXYZE(feedRate_mm_s);
+        st_synchronize();
+}
+
+void nozzle_park() {
+    current_position[X_AXIS] = MMU_ERR_X_PAUSE_POS;
+    current_position[Y_AXIS] = MMU_ERR_Y_PAUSE_POS;
+    planner_line_to_current_position(NOZZLE_PARK_XY_FEEDRATE);
+    st_synchronize();
```
gudnimg: `st_synchronize()` is already included with `planner_line_to_current_position()`.

gudnimg: Should maybe `plan_buffer_line_curposXYZE` be used here?

Firmware/mmu2.cpp
```diff
@@ -1048,4 +1012,24 @@ void MMU2::OnMMUProgressMsgSame(ProgressCode pc){
...
+//void MMU2::LogEchoEvent(const char *msg) {
+//    MMU2_ECHO_MSG(msg);
+//    SERIAL_ECHOLN();
+//}
+
+void MMU2::LogEchoEvent_P(const char *msg_P) {
+    MMU2_ECHO_MSGRPGM(msg_P);
+    SERIAL_ECHOLN();
+}
+
```
gudnimg: Why are we re-defining these functions? I see they're present in `mmu2_log.cpp`, shouldn't we use those?

gudnimg:
I've gone through all the code changes line-by-line. The only two blockers I see is some inconsistency in using st_synchronize(). And the problem with `planner_line_to_current_position(MMM_TO_MMS(fr_mm_m));`

If those two are resolved I am ready to approve this PR.

Firmware/mmu2_protocol_logic.cpp
DRracer: I agree, for now it is good enough, in the future we can introduce some data structure

Firmware/mmu2_protocol_logic.cpp
DRracer: yes, but it wouldn't help with the initialization of the `initRegs8` array

Firmware/mmu2_protocol_logic.h
DRracer: :thinking: it has been like this before ... just the initialization was written in the constructor.
And I'd sacrifice those 12B for the sake of safety just in case this class gets used somewhere else in multiple instances.

gudnimg (approved):

PR #3962, opened 2023-01-27
Convert `fanSpeed` and `saved_fan_speed` to `uint8_t`

gudnimg:
The fan speed variables only range from 0 to 255 and should be `uint8_t` (1 byte). Even in the planner/block this applies and sync the code a little bit with Marlin 2.

Change in memory (Multilang MK3S build):
Flash: -324 bytes
SRAM: -20 bytes

### DIFF
--- a/Firmware/menu.cpp
+++ b/Firmware/menu.cpp
@@ -172,12 +172,10 @@ static void menu_draw_toggle_puts_P(const char* str, const char* toggle, const u
     //xxxxxcba
     //a = selection mark. If it's set(1), then '>' will be used as the first character on the line. Else leave blank
     //b = toggle string is from progmem
-    //c = do not set cursor at all. Must be handled externally.
     uint8_t is_progmem = settings & 0x02;
     const char eol = (toggle == NULL) ? LCD_STR_ARROW_RIGHT[0] : ' ';
     if (toggle == NULL) toggle = _T(MSG_NA);
     uint8_t len = 4 + (is_progmem ? strlen_P(toggle) : strlen(toggle));
-    if (!(settings & 0x04)) lcd_set_cursor(0, menu_row);
     lcd_putc((settings & 0x01) ? '>' : ' ');
     lcd_print_pad_P(str, LCD_WIDTH - len);
     lcd_putc('[');
@@ -429,11 +427,10 @@ const char menu_fmt_float31[] PROGMEM = "%-12.12S%+8.1f";
 
 const char menu_fmt_float13[] PROGMEM = "%c%-13.13S%+5.3f";
 
-template<typename T>
-static void menu_draw_P(char chr, const char* str, int16_t val);
 
-template<>
-void menu_draw_P<int16_t*>(char chr, const char* str, int16_t val)
+
+template <typename T>
+void menu_draw_P(char chr, const char* str, T val)
 {
 	// The LCD row position is controlled externally. We may only modify the column here
 	lcd_putc(chr);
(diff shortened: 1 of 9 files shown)


gudnimg:
Ran a MMU print with these changes. Print went very smooth and finished on time. Also tested editing the fan speed while printing, works fine as expected.

Note I also tweaked the PrusaSlicer profile a little bit to shrink the Wipe Tower.

image

Firmware/ultralcd.cpp
```diff
@@ -5581,7 +5581,7 @@ static void lcd_tune_menu()
 	MENU_ITEM_EDIT_int3_P(_T(MSG_NOZZLE), &target_temperature[0], 0, HEATER_0_MAXTEMP - 10);//3
 	MENU_ITEM_EDIT_int3_P(_T(MSG_BED), &target_temperature_bed, 0, BED_MAXTEMP - 10);
 
-	MENU_ITEM_EDIT_int3_P(_T(MSG_FAN_SPEED), &fanSpeed, 0, 255);//5
+	MENU_ITEM_EDIT_int3_P(_T(MSG_FAN_SPEED), (int16_t*)&fanSpeed, 0, 255);//5
```
leptun: This needs to be checked carefully. I am suspicious of that int16_t* cast. We need to check that the edit menu doesn't overwrite something around `fanSpeed`.

Firmware/ultralcd.cpp
```diff
@@ -5581,7 +5581,7 @@ static void lcd_tune_menu()
...
 	MENU_ITEM_EDIT_int3_P(_i("Flow"), &extrudemultiply, 10, 999);//6////MSG_FLOW c=15
```
vintagepc: Yes, this is definitely dangerous. You'll zero the adjacent byte in memory.

Firmware/ultralcd.cpp
```diff
@@ -5732,7 +5732,7 @@ static void lcd_control_temperature_menu()
 #if TEMP_SENSOR_BED != 0
   MENU_ITEM_EDIT_int3_P(_T(MSG_BED), &target_temperature_bed, 0, BED_MAXTEMP - 3);
 #endif
-  MENU_ITEM_EDIT_int3_P(_T(MSG_FAN_SPEED), &fanSpeed, 0, 255);
+  MENU_ITEM_EDIT_int3_P(_T(MSG_FAN_SPEED), (int16_t*)&fanSpeed, 0, 255);
```
vintagepc: ditto

vintagepc (changes requested):

gudnimg:
If we can get that menu change [34c1e55](https://github.com/prusa3d/Prusa-Firmware/pull/3962/commits/34c1e55863650c3859567a2d73283109eefeb9d6) through that would eliminate one annoying edge case in the menu code. I didn't see before that it only affects one menu item.

Gives me more hope of finishing https://github.com/prusa3d/Prusa-Firmware/pull/3985 😆

gudnimg:
Rebased to sync with MK3 and fix conflict

Firmware/menu.cpp
```diff
@@ -452,20 +449,8 @@ void menu_draw_P<int16_t*>(char chr, const char* str, int16_t val)
...
-    {
-        menu_draw_toggle_puts_P(str, _T(MSG_OFF), 0x04 | 0x02 | (chr=='>'));
-    }
-    else
-    {
-        lcd_printf_P(menu_fmt_float13, chr, str, factor);
-    }
-}
+template void menu_draw_P<int16_t>(char chr, const char* str, int16_t val);
+template void menu_draw_P<uint8_t>(char chr, const char* str, uint8_t val);
```
DRracer, line 453: removal of the special code for linearity correction has been approved :tada:

PR #3964, opened 2023-01-28
build: Report target sizes only once in AVR format

wavexx:
cmake currently reports the binary sizes twice, once without the real mcu set and the second in AVR format.

The double reporting is confusing, and we really seem to prefer the AVR format. Change the `report_size` function to use the current mcu and remove the double (confusing) reporting.

This results in:

```
MK3S-EINSy10a_ENGLISH AVR Memory Usage
----------------
Device: atmega2560

Program:  235188 bytes (89.7% Full)
(.text + .data + .bootloader)

Data:       5662 bytes (69.1% Full)
(.data + .bss + .noinit)
```

Note the current target name in the header

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -235,11 +235,12 @@ if(CMAKE_CROSSCOMPILING)
   set(CMAKE_C_FLAGS_RELEASE ${CMAKE_CXX_FLAGS_RELEASE})
 
   # mcu and target-related settings
+  set(TARGET_MCU "atmega2560")
   add_compile_options(
-    -mmcu=atmega2560 -DF_CPU=16000000L -DARDUINO=10819 -DARDUINO_AVR_PRUSA_EINSY_RAMBO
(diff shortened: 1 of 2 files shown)


wavexx:
@gudnimg does it work as expected on vscode?

gudnimg:
Before:

image


After:
image


We lose information regarding the different regions like `.text`, `.bss` etc. It would be useful in my opinion to keep those stats.

wavexx:
Looks like `echo -n` is also broken. Ok, closing then.

gudnimg:
@wavexx using the built-in echo_append command in CMake seems to work on my end 👍 Probably more portable that way too.
image

gudnimg:
https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-E-arg-echo

gudnimg:
The empty `echo` string `""` also doesn't work on Vscode on Windows (I never noticed this). But using the built-in `echo` for that, seems to work too. The `"ECHO is on"` string goes away.

image

wavexx:
Looks a good alternative to me. Can you make a PR to fix that echo?

PR #3968, opened 2023-01-28
Fix an issue with using echo on Windows

gudnimg:
* Use CMake's built-in `echo` command to ensure behavior is the same between different platforms
* Append the target name to the memory usage results post build. This is done using CMake's built-in command `echo_append`

**Before:**
Notice the string `"ECHO is on"` string appears. It should not be there.
image

**After:**
image

### DIFF
--- a/cmake/Utilities.cmake
+++ b/cmake/Utilities.cmake
@@ -58,7 +58,7 @@ endfunction()
 function(report_size target)
   add_custom_command(
     TARGET ${target} POST_BUILD
-    COMMAND echo "" # visually separate the output
+    COMMAND ${CMAKE_COMMAND} -E echo "" # visually separate the output
     COMMAND "${CMAKE_SIZE_UTIL}" -B "$<TARGET_FILE:${target}>"
     USES_TERMINAL
     )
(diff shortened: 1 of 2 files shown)


wavexx (approved):

wavexx:
Can you maybe also do what we did in #3964 by putting the $target name before "AVR memory usage"?

When having multiple build it makes it more obvious IMHO.

gudnimg:
Ok I’ll add it later this evening.

gudnimg:
@wavexx it looks like this now:

image

cmake/Utilities.cmake
```diff
@@ -58,7 +58,7 @@ endfunction()
 function(report_size target)
   add_custom_command(
     TARGET ${target} POST_BUILD
-    COMMAND echo "" # visually separate the output
+    COMMAND ${CMAKE_COMMAND} -E echo "" # visually separate the output
```
wavexx: I almost think this extra echo here is not useful. The output looks already separated to me.

gudnimg: Looks like this without it. I don't mind deleting this line. image

PR #3984, opened 2023-02-04
Set the fanSpeed=0 in order to stop the fan during the manual actions on a filament change…

jfestrada:
Sometimes the M600 is inside a gcode and you are not in front of the print so can be a little time with the fan cooling the hotend.
I was reviewing the code of the gcode_M600 and this already take care about the restore to the previous fan speed but never stop it, so I added the stop of the fan when the M600 is executed without MMU.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3525,7 +3525,11 @@ static void gcode_M600(bool automatic, float x_position, float y_position, float
     st_synchronize();
 
     //Beep, manage nozzle heater and wait for user to start unload filament
-    if(!mmu_enabled) M600_wait_for_user(HotendTempBckp);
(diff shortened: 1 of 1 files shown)


gudnimg:
@jfestrada We added a similar fix to 3.13 (but not 3.12) here https://github.com/prusa3d/Prusa-Firmware/pull/3746 I understand this PR fixes the same issue?

We had discussed about adding this to 3.12 but decided to keep it for 3.13

jfestrada:
@gudnimg  yes I think that's the same, so this PR could be closed if you are not planning to release a new minor version. Let me know and I will close it.

leptun:
There aren't any plans for another minor release of 3.12. So I think the PR can be closed.

PR #3985, opened 2023-02-04
🚧 PFW-880 Re-work menus to allow editing `uint8_t` and `float` values

gudnimg:
There are number of improvements in this PR:

* `lcd_encoder` was changed from `int32_t` to `int8_t`. This is the biggest optimisation by far, it comes down to the idea that we should use `lcd_encoder `for anything other than to track changes in the LCD's knob rotation.
    * Add new veriable `currentValue` to the menu edit data to replace `lcd_encoder`. Previously we were using `lcd_encoder` to hold the temperorary unsaved value on the screen.
* The above will break the Live K adjustment menu (which is by default not included in the firmware). I fixed it by making the menu code support editing float values. This adds a little bit overhead but I think it is worth it as we can apply this menu code to other functions in the firmware.
* `MENU_ITEM_EDIT_int3_P()` renamed to `MENU_ITEM_EDIT_P()`.

TODO:
Add callback functionality. There are several different variants that needs to be configured and they must be configurable such that only one of the below callbacks are used.
- [ ] add callback to save value on menu exit
    * Extrusion multiplier menu
- [ ] add callback to save value live
    * K adjust menu
    * Move X, Y, Z
    * Z-live adjust
    * E move

Change in memory:
Flash: -820 bytes
SRAM: -4 bytes

### DIFF
--- a/Firmware/menu.cpp
+++ b/Firmware/menu.cpp
@@ -455,7 +455,7 @@ void menu_draw_item_P(char chr, const char* str, int16_t val)
 	lcd_putc(chr);
 	uint8_t len = lcd_print_pad_P(str, LCD_WIDTH - 1);
 	lcd_set_cursor_column((LCD_WIDTH - 1) - len + 1);
-	lcd_putc(':');
+	lcd_putc(' ');
 
 	// The value is right adjusted, set the cursor then render the value
 	if (val < 10) { // 1 digit
@@ -470,10 +470,7 @@ void menu_draw_item_P(char chr, const char* str, int16_t val)
 
 void __attribute__((noinline)) menu_draw_edit_P(char chr, const char* str, int16_t val, uint8_t isFloat)
 {
-    menu_data_edit_t* _md = (menu_data_edit_t*)&(menu_data[0]);
-    if (val <= _md->minEditValue) {
-        menu_draw_toggle_puts_P(str, _T(MSG_OFF), 0x04 | 0x02 | (chr=='>'));
-    } else if (isFloat) {
+    if (isFloat) {
         float temp = menu_edit_convert_to_float(val, 3);
         lcd_printf_P(menu_fmt_float13, chr, str, temp);
     } else {
--- a/Firmware/menu.cpp
+++ b/Firmware/menu.cpp
@@ -442,11 +442,10 @@ void menu_draw_item_P(char chr, const char* str, int16_t val)
 	lcd_print(val);
 }
 
-template<typename T>
-void menu_draw_edit_P(char chr, const char* str, T val)
+void menu_draw_edit_P(char chr, const char* str, int16_t val)
 {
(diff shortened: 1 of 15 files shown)


gudnimg:
Since `minEditValue` and `maxEditValue` are always constant. Maybe we can move these to `PROGMEM`? 🤔 Maybe that will help simplify our code structure so we have a "list" of limits for specific menus.

wavexx:
Would this help shrinking #3494 ?

gudnimg:
@wavexx I think we can do this with little overhead.

An idea I'm playing with is define the menu data in PROGMEM like this:

```
static const constexpr int16_t TEMPERATURE_NOZZLE_EDIT_MENU[] PROGMEM = {
    0, // Minimum value
    HEATER_0_MAXTEMP, // Maximum value
    LCD_JUMP_HOTEND_TEMP // Jump value
};

static const constexpr int16_t TEMPERATURE_BED_EDIT_MENU[] PROGMEM = {
    0, // Minimum value
    BED_MAXTEMP - 3, // Maximum value
    LCD_JUMP_BED_TEMP // Jump value
};

static const constexpr int16_t * menuEditTemplates[] PROGMEM = {
    TEMPERATURE_NOZZLE_EDIT_MENU,
    TEMPERATURE_BED_EDIT_MENU,
};
```

gudnimg:
The only menu I see which depends on limits which are configurable at run time is the blacklight levels.

gudnimg:
@wavexx I've got a working solution which:

* Defines the edit limit in PROGMEM
* Allows defining the limits at run time (for backlight menu) -- I supsect this is eating up most flash right now since I need to handle PROGMEM and non-PROGMEM differently.
* Implements a jump value as is done in #3494 

This adds 250 bytes of flash. But we're still saving 576 bytes.

I think I can optimise this a bit more.

gudnimg:
Current memory stats:

```
[build] Program:  250132 bytes (95.4% Full)
[build] 
[build] Data:       5634 bytes (68.8% Full)
```

Feel free to suggest any optimisations. :)

Firmware/menu.cpp
```diff
@@ -423,23 +425,37 @@ uint8_t menu_item_gcode_P(const char* str, const char* str_gcode)
...
 
-template<>
-void menu_draw_P<int16_t*>(char chr, const char* str, int16_t val)
+void menu_draw_item_P(char chr, const char* str, int16_t val)
 {
 	// The LCD row position is controlled externally. We may only modify the column here
 	lcd_putc(chr);
 	uint8_t len = lcd_print_pad_P(str, LCD_WIDTH - 1);
 	lcd_set_cursor_column((LCD_WIDTH - 1) - len + 1);
-	lcd_putc(':');
```
gudnimg, line 442: Note I removed the `':'` here so UI for integers and floats are the same.

gudnimg:
Too many things are broken, so I will close this for now.

PR #3995, opened 2023-02-07
optimisation: re-use `adjust_bed_reset()`

gudnimg:
Moved `adjust_bed_reset()` into `eeprom.cpp` since we're only working with EEPROM

Change in memory:
Flash: -42 bytes
SRAM: 0 bytes


DRracer (approved):
looks good to me :+1:

PR #3957, opened 2023-01-26
TM: PTC model support

wavexx:
This is a first attempt to support runtime PTC heating elements such as the Revo. There is no full calibration support yet.

We expose two existing model simulation parameters: the dampening filter and model lag as M310 "D" and "L" respectively, as they need to be tuned for a faster response. Besides parameter handling, there is no actual runtime cost to these.

As usual, to keep things fast we have to cheat. To actual PTC heating element has a quadratic response according to the datasheet, but we actually add just a new linear term to the power estimation:

2023-01-26T235657

The data above was actually carved out of the Revo datasheet image using "g3data" btw. 

It *should* be enough, and the values I got from the graph seem to match in behavior the logs we have on #3636. It defaults to an identity, so existing calibration data will work unchanged.

The new U/V/D/L parameters seem to be pretty consistent across samples, which is good.

Since the calibration doesn't handle the new terms yet it needs to be "primed" for the new values so that C/R can be done by the existing calibration function.

The starting values for the Revo Six seem to be `P40 U-0.0014 V1.05 D0.15 L270`.
With this in mind, it should be possible to test this by doing:

```
M310 S0 ; disable the model if hasn't been yet
M310 P40 U-0.0014 V1.05 D0.15 L270 ; set Revo parameters
M310 A F0 ; calibrate C/R
M310 S1 ; enable the model
```

**I wasn't able to test this on any hardware yet**. I'm only basing this on the sample logs, so I'm not sure if this works as it is or requires further tweaks. It's also unknown whether the existing warning/error threshold still make sense (from the logs they should still fall within range), since this needs a practical test.

PR #4006, opened 2023-02-10
PFW-1499 TM: PTC model support (3.12)

wavexx:
Backport of #3957 for 3.12


3d-gussner (approved):
Latest changes work even better than before on stock MK3S with E3D v6

PR #4007, opened 2023-02-10
Improve sound handling for MMU error screen

gudnimg:
An idea I have to re-use M600 sound code for error screen.

The requirements are the same as for the M600 sound prompts.

* Loud mode => should continuously beep (loud prompt sound)
* Silent mode => should be silent
* Once mode => should beep only once
* Assist/Blind => Same as Loud mode but beep with slightly lower tone (same sound as when clicking or rotation the knob)

**TODO:**

- [x] Tune beeping period to be same as before in M600
- [x] Test M600 thoroughly (with all sounds settings)
- [x] Test MMU error screen (with all sound settings)

Testing:
* MMU Error screen:
    * Loud mode ✔️ 
    * Once mode ✔️  
    * Silent mode ✔️ 
    * Assist mode ✔️ 

Change in memory:
Flash: -148 bytes
SRAM: +4 bytes

PFW-831

### DIFF
--- a/Firmware/mmu2_reporting.cpp
+++ b/Firmware/mmu2_reporting.cpp
@@ -253,7 +253,14 @@ void ReportErrorHook(CommandInProgress /*cip*/, uint16_t ec, uint8_t /*es*/) {
                 ReportErrorHookState = ReportErrorHookStates::RENDER_ERROR_SCREEN;
                 break;
             case 2:
-                ReportErrorHookState = ReportErrorHookStates::DISMISS_ERROR_SCREEN;
+                // Exit error screen and enable lcd updates
+                lcd_set_custom_characters();
+                lcd_update_enable(true);
+                lcd_return_to_status();
+                sound_wait_for_user_reset();
+                // Reset the state in case a new error is reported
+                mmu2.is_mmu_error_monitor_active = false;
+                ReportErrorHookState = ReportErrorHookStates::RENDER_ERROR_SCREEN;
                 break;
             default:
                 break;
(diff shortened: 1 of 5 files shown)


Firmware/mmu2_reporting.cpp
```diff
@@ -252,13 +253,7 @@ void ReportErrorHook(CommandInProgress /*cip*/, uint16_t ec, uint8_t /*es*/) {
...
                 break;
             case 2:
-                // Exit error screen and enable lcd updates
-                lcd_set_custom_characters();
-                lcd_update_enable(true);
-                lcd_return_to_status();
-                // Reset the state in case a new error is reported
-                mmu2.is_mmu_error_monitor_active = false;
-                ReportErrorHookState = ReportErrorHookStates::RENDER_ERROR_SCREEN;
+                ReportErrorHookState = ReportErrorHookStates::DISMISS_ERROR_SCREEN;
```
DRracer: How is this change related to sounds?

gudnimg: I would have needed to add a call to `sound_wait_for_user_reset();` here. When I wrote the error screen code this change above didn't save memory, but for some reason now it saves 10 bytes.

Firmware/mmu2_reporting.cpp
```diff
@@ -256,6 +257,7 @@ void ReportErrorHook(CommandInProgress /*cip*/, uint16_t ec, uint8_t /*es*/) {
                 lcd_set_custom_characters();
```
DRracer, line 257: Are you going to extract this block into a separate function?
It is the same like `case (uint8_t)ReportErrorHookStates::DISMISS_ERROR_SCREEN:`

DRracer (approved):

PR #4008, opened 2023-02-11
Optimisation: Remove duplicate code in `Sound_MakeCustom()`

gudnimg:
Minor code duplication I noticed could be improved.

Change in memory:
Flash: -48 bytes
SRAM: 0 bytes


DRracer (approved):
it took me a while to verify the logic, but it seems to be ok

PR #4010, opened 2023-02-11
backlight: always allow backlight to dim/wake

### LINKED ISSUE #2777: [BUG] FW 3.9.0 - Backlight doesn't turn to bright mode when interacting after filament runout and cooldown
**Printer type** - MK3S
**Printer firmware version**-  3.9.0

**MMU Upgrade** - None

**SD card or USB/Octoprint** SD card

**Describe the bug**
When the filament sensor trips mid-print and the hot-end cools down after a timeout, the screen will stay in dim mode until the hot-end reaches full temperature even though the user had to press the knob to start the unload procedure.

I have my backlight dim mode set to 0 (off) with a 120 second timeout.

With my dim brightness set to 0, this made it very difficult to tell whether the printer was heating up or if it was stuck/off.
  
**To Reproduce**
1. Set backlight dim setting to something low (or 0).  Set your backlight timeout set to something low for testing.
2. Start a print via SD card
3. Cut the filament strand early to trip the filament runout sensor
4. Wait until the screen dims and the hot-end cools down (the extruder fan will stop)
5. Press the knob on the printer to begin the unload process

The printer should start to warm up the hot-end but the screen will stay dimmed/off until it reaches full temperature.
  
**Expected behavior**
That the screen will light up immediately when a user interacts with the printer after a filament runout (and hot-end cooldown timeout) to show the temperature of the hot-end as it heats back up.

gudnimg:
This commit adds the ability for the firmware to dim and wake the backlight when LCD updates are disabled. Such as in the MMU error screen or when rendering full screen messages which typically disable the LCD updates to prevent the status screen from rendering.

Fixes #2777

Fixes an issue where the backlight may stay dim on MMU error screen even though the knob is pressed or rotated. Only happens when the Sound settings are set to Silent.

Change in memory:
Flash: -6 bytes
SRAM: +1 byte

### DIFF
--- a/Firmware/lcd.cpp
+++ b/Firmware/lcd.cpp
@@ -15,6 +15,7 @@
 #include "fastio.h"
 //-//
 #include "sound.h"
+#include "backlight.h"
 
 #define LCD_DEFAULT_DELAY 100
 
@@ -638,6 +639,7 @@ int8_t lcd_encoder_diff = 0;
 uint8_t lcd_buttons = 0;
 uint8_t lcd_button_pressed = 0;
 uint8_t lcd_update_enabled = 1;
+static bool lcd_backlight_wake_trigger; // Flag set by interrupt when the knob is pressed or rotated
 
 uint32_t lcd_next_update_millis = 0;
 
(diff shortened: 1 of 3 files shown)


wavexx (approved):

Firmware/lcd.cpp
```diff
@@ -776,6 +787,7 @@ void lcd_buttons_update(void)
 	if (lcd_buttons & EN_B) enc |= B10;
 	if (enc != lcd_encoder_bits)
 	{
+		lcd_backlight_wake_trigger = true; // flag event, knob rotated
```
leptun: I'm not sure this is entirely correct. I feel like this will make the backlight wake even if the encoder was moved only 1/4 of a tick of rotation. That might be enough to create random wakes of the backlight if the knob is between two encrot steps and oscillates between them due to noise/vibrations.

leptun:
Besides the comment I wrote above, the PR looks fine to me.

Firmware/lcd.cpp
gudnimg: That’s a good point. 👍

gudnimg: Resolved with [8a926fa](https://github.com/prusa3d/Prusa-Firmware/pull/4010/commits/8a926fad103b1409701a092b98f5137ceff6ad5e)

leptun (approved):
Tested on MK3S 👍

PR #4012, opened 2023-02-11
 Don't show Pause/Stop menus while processing T-codes

gudnimg:
PFW-1449

Change in memory:
Flash: +18 bytes
SRAM: +1 bytes


DRracer (approved):
that's exactly what I had in mind :+1:

PR #4013, opened 2023-02-12
mmu: Fix crash when print is paused on toolchange

### LINKED ISSUE #3989: [BUG] 3.12.1 MK3S Firmware Crash on MMU2 Recovery
**Printer type** - MK3S
**Printer firmware version** -  3.12.1-5686

**MMU upgrade** - MMU2S
**MMU upgrade firmware version** - 1.0.6-372

**SD card or USB/Octoprint**
  SD Print with Octoprint monitoring via USB

**Describe the bug**
Since upgrading to 3.12.1 from 3.10, (I did not use 3.11), each time I have had an MMU error the printer has had a firmware crash immediately following MMU recovery and the print aborted.  Both times this was several hours into an 11 hour or longer print, but I do not have timing details.    This has happened 3 times and I am reverting to 3.10 for a final attempt at the print.  

The other change is that the gcode was sliced with prusaslicer 2.6.0 alpha2.

I have captured a D21 dump however.
  
**To Reproduce**
Unknown, but I have not had a successful MMU recovery with the new firmware.  Unfortunately I am unable to help reporduce immediately as I need to try to get a successful print this weekend. Once done, I will reflash 3.12.1 and try to repro with minimal steps if the crash dump does not show something obvious. 
  
**Expected behavior**
 Once MMU recovery is complete, the print should continue
  
**G-code**
[helicopter_4color_split_edit-2+sprus-treesupport.2_0.6n_0.3mm_PLA_MK3SMMU2S_11h37m.gcode.zip](https://github.com/prusa3d/Prusa-Firmware/files/10611353/helicopter_4color_split_edit-2%2Bsprus-treesupport.2_0.6n_0.3mm_PLA_MK3SMMU2S_11h37m.gcode.zip)

I was unable to issue D21 with pronterface, but have attached a log of printer booting and D21 from Octoprint.

[3.12.1_crashdump.txt](https://github.com/prusa3d/Prusa-Firmware/files/10611313/3.12.1_crashdump.txt)

### LINKED ISSUE #3990: FW crash if pausing when filament change (MMU)
**Printer type** - MK3S
**Printer firmware version** -  3.12.1

**MMU upgrade** - MMU2S
**MMU upgrade firmware version**  - 1.0.6

**SD card or USB/Octoprint**
SD card

**Describe the bug**
  Prusa FW crashed if you pause when filament is changing. After crashing, the printer informs that I can resume the printing, but their is no option to resume.
  
**To Reproduce**
 1. start a multi material print
 2. when the first filament finished printing and is unloaded from the nozzle, just before the selector of MMU is about to move to the next filament, pause the print. The FW wil crash.
  
**Expected behavior**
- no crash
- option to resume printing
  
**G-code**
[MMU-Tip_5colors-GCode.zip](https://github.com/prusa3d/Prusa-Firmware/files/10660531/MMU-Tip_5colors-GCode.zip)

**Video**
It happen every time, no matter the gcode. I can post a video, on request.

gudnimg:
The solution is to unblock the planner after the print has been saved into RAM and the command queue has been cleared. This will ensure the watchdog does not time out (which causes the FW crash).

Additionally this fixes an issue where the printer does not park the extruder when the print is paused.

Fixes #3989, #3990

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -11385,6 +11385,10 @@ void stop_and_save_print_to_ram(float z_move, float e_move)
 	card.sdprinting = false;
 //	card.closefile();
 	saved_printing = true;
+
+  // Enable stepper driver interrupt to move Z axis. This should be fine as the planner and
+  // command queues are empty, SD card printing is disabled, usb is inhibited.
+  planner_aborted = false;
   // We may have missed a stepper timer interrupt. Be safe than sorry, reset the stepper timer before re-enabling interrupts.
   st_reset_timer();
 	sei();


wavexx:
This is the wrong thing to do. planner_aborted should only be reset at the beginning of loop(), except in power panic where we are sure we will never go back to loop().

Maybe the real problem is that after pause, you need to exit all busy loops until you're called again?

gudnimg:
> Maybe the real problem is that after pause, you need to exit all busy loops until you're called again?

Indeed it looks like this is a problem in 3.13 too, the MMU code doesnt return the loop() until the current MMU operation is done. I’m not sure why there is a watchdog reset in 3.12.

wavexx (changes requested):
stop_and_save_print_to_ram() shouldn't clear planner_aborted as it can lead to recursive behavior when it's used.

PR #4020, opened 2023-02-16
Disable MMU2S by default

DRracer:
On printers without the MMU no error screen should occur and no attempts of communication with the MMU should be performed -> EEPROM_MMU_ENABLED should default to 0.

PFW-1418


wavexx (approved):

3d-gussner (approved):
Works well after Factory reset

PR #4025, opened 2023-02-17
optimisation: extract common M861 code

gudnimg:
Change in memory:
Flash: -162 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3803,15 +3803,17 @@ static void gcode_G92()
 /// @brief Helper function to reduce code size in M861 
 /// by extracting common code into one function
 static void gcode_M861_print_pinda_cal_eeprom() {
-    int16_t usteps;
+    int16_t usteps = 0;
+    float mm = 0;
     static const char comma_sep[] PROGMEM = ", ";
     for (uint8_t i = 0; i < 6; i++) {
-        usteps = 0;
         if(i > 0) {
             usteps = eeprom_read_word((uint16_t*) EEPROM_PROBE_TEMP_SHIFT + (i - 1));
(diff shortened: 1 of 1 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -3800,6 +3800,26 @@ static void gcode_G92()
...
+static void gcode_M861_print_pinda_cal_eeprom() {
+    int16_t usteps;
+    for (uint8_t i = 0; i < 6; i++) {
+        usteps = 0;
+        if(i > 0) {
+            usteps = eeprom_read_word((uint16_t*) EEPROM_PROBE_TEMP_SHIFT + (i - 1));
+        }
+        float mm = ((float)usteps) / cs.axis_steps_per_unit[Z_AXIS];
+        i == 0 ? SERIAL_PROTOCOLPGM("n/a") : SERIAL_PROTOCOL(i - 1);
+        SERIAL_PROTOCOLPGM(", ");
```
DRracer: Please check, that gcc is able to merge the duplicate `", "` strings.

gudnimg: @DRracer do you mean by adding `", "` to `messages.cpp`?

DRracer: Even by extracting to a `static const char something[] PROGMEM;` locally ...

DRracer: ... and yes, having a collection of these tiny strings at one spot and reusing them across the whole firmware might actually bring some savings.

gudnimg: Done ✔️

DRracer (approved):
looks ok to me :+1:

PR #4027, opened 2023-02-18
Many minor optimisations for 3.13/3.14

gudnimg:
WIP. I'm adding a few minor optimisations to this branch which we probably don't want to create a separate PR for each since they save little < 10B each.

Current savings:
Flash: -420bytes
SRAM:  -2 bytes


gudnimg:
Rebased to sync with MK3 branch and fix merge conflict.

Also dropped one questionable optimisation.

gudnimg:
Rebased  to sync with MK3 branch and resolve conflicts.

Firmware/ultralcd.cpp
```diff
@@ -3224,17 +3224,16 @@ void lcd_temp_cal_show_result(bool result) {
...
+
+	if (result) {
 		SERIAL_ECHOLNPGM("PINDA calibration done. Continue with pressing the knob.");
 		lcd_show_fullscreen_message_and_wait_P(_T(MSG_PINDA_CALIBRATION_DONE));
-		eeprom_update_byte((unsigned char *)EEPROM_TEMP_CAL_ACTIVE, 1);
-	}
-	else {
-		eeprom_update_byte((uint8_t*)EEPROM_CALIBRATION_STATUS_PINDA, 0);
+	} else {
 		SERIAL_ECHOLNPGM("PINDA calibration failed. Continue with pressing the knob.");
```
leptun, line 3228: Can this serial message be extracted to a single PROGMEM item? If I recall, these PSTR strings are not deduplicated by the linker script

Firmware/util.cpp
```diff
@@ -247,36 +247,15 @@ ClCheckVersion oCheckVersion=ClCheckVersion::_None;
...
-    if (oCheckVersion == ClCheckVersion::_Undef) {
-        oCheckVersion = ClCheckVersion::_Warn;
-        eeprom_update_byte((uint8_t *)EEPROM_CHECK_VERSION, (uint8_t)oCheckVersion);
-    }
-    oCheckGcode = (ClCheckGcode)eeprom_read_byte((uint8_t *)EEPROM_CHECK_GCODE);
-    if (oCheckGcode == ClCheckGcode::_Undef) {
-        oCheckGcode = ClCheckGcode::_Warn;
-        eeprom_update_byte((uint8_t *)EEPROM_CHECK_GCODE, (uint8_t)oCheckGcode);
-    }
+    oCheckMode = (ClCheckMode)eeprom_init_default_byte((uint8_t *)EEPROM_CHECK_MODE, farm_mode ? (uint8_t) ClCheckMode::_Strict : (uint8_t)ClCheckMode::_Warn);
```
leptun: I don't think this change is completely correct. We can't rely on farm_mode during eeprom initialization since farm mode will most likely not be enabled during the first eeprom initialization, so the default value would always be loaded instead.
The old code also seems to enforce a few of these settings even if the eeprom is already initialized. The new code only sets defaults when uninitialized. I think we should try to preserve the old behaviour.

The EEPROM_NOZZLE_DIAMETER_uM_DEFAULT initialization looks as if it behaves correctly, but I'm a bit confused by the old code. It's overly complicated and I still don't understand why we need two eeprom "nozzle diameter" settings.

The EEPROM_CHECK_MODEL, EEPROM_CHECK_VERSION and EEPROM_CHECK_GCODE code looks correct 👍.

leptun:
Please see the comments I wrote above. Besides those, everything looks good to me 👍.

Firmware/ultralcd.cpp
leptun: Scrap this. Not sure how I didn't see that one string says "done" and one says "failed". So you may ignore the comment above.

Firmware/util.cpp
gudnimg: Regarding `EEPROM_CHECK_MODE`, does the change below look OK? Surprisingly there is no change in memory usage.

The code below will always at least initialise the EEPROM value. But if farm mode is enabled, it's overwritten if it hasn't been done already.
image

leptun: I think that change is correct, but I didn't check it too carefully.
There is still EEPROM_NOZZLE_DIAMETER below it that doesn't look correct. We can't have a ternary operator on farm_mode in one of those eeprom _init_default functions.

gudnimg: I havent forgotten the nozzle setting. Just needed a fresh 🧠 and sleep. I need to re-review how these nozzle eeprom settings are used. It’s very confusing at only a quick glance. Needs a more thorough “glance” ;) 

I may delay this for the weekend to priorities the menu testing. But will see how the week goes. :)

gudnimg:
Rebased to resolve conflict

Firmware/util.cpp
gudnimg: Hopefully resolved with [bc4089f](https://github.com/prusa3d/Prusa-Firmware/pull/4027/commits/bc4089fe6517803a8fb11be7ff33b5a531533ce8)

gudnimg:
Rebased to pull in latest fixes on the MK3 branch. Such as the editing menu fixes.

leptun (approved):

Firmware/Marlin_main.cpp
```diff
@@ -183,13 +183,12 @@ bool mesh_bed_leveling_flag = false;
...
 uint32_t total_filament_used;
 HeatingStatus heating_status;
-uint8_t heating_status_counter;
 bool loading_flag = false;
 
 #define XY_NO_RESTORE_FLAG (mesh_bed_leveling_flag || homing_flag)
 
 
-bool fan_state[2];
+
```
wavexx: Would be nice to remove those extra useless lines.

Firmware/Marlin_main.cpp
```diff
@@ -6472,10 +6464,9 @@ SERIAL_PROTOCOLPGM("\n\n");
           SERIAL_ECHOPGM(" based on Marlin FIRMWARE_URL:https://github.com/prusa3d/Prusa-Firmware PROTOCOL_VERSION:");
           SERIAL_ECHOPGM(PROTOCOL_VERSION);
           SERIAL_ECHOPGM(" MACHINE_TYPE:");
-          SERIAL_ECHOPGM(CUSTOM_MENDEL_NAME); 
-          SERIAL_ECHOPGM(" EXTRUDER_COUNT:"); 
-          SERIAL_ECHOPGM(STRINGIFY(EXTRUDERS)); 
-          SERIAL_ECHOPGM(" UUID:"); 
+          SERIAL_ECHOPGM(CUSTOM_MENDEL_NAME);
+          SERIAL_ECHOPGM(" EXTRUDER_COUNT:1");
```
wavexx: Since we still have "EXTRUDERS" we can probably do:

``SERIAL_ECHOPGM(" EXTRUDER_COUNT:" STRINGIFY(EXTRUDERS)));``

which should still avoid the extra call.

wavexx (approved):

PR #4028, opened 2023-02-18
Merge duplicate strings in cmdqueue.cpp

DRracer:
Original idea from @GilesBathgate, PR #3322 . This is just a rework on top of current MK3 branch. And, we don't want to have these strings translated...

### DIFF
--- a/Firmware/cmdqueue.cpp
+++ b/Firmware/cmdqueue.cpp
@@ -248,6 +248,9 @@ void cmdqueue_dump_to_serial()
 }
 #endif /* CMDBUFFER_DEBUG */
 
+static const char bufferFull[] PROGMEM = "\" failed: Buffer full!";
+static const char enqueingFront[] PROGMEM = "Enqueing to the front: \"";
+
 //adds an command to the main command buffer
 //thats really done in a non-safe way.
 //needs overworking someday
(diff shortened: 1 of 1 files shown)


GilesBathgate:
@DRracer I don't know much about your compile toolchain, but does the compiler support LTO? (`-flto` in clang). It should optimise for smaller size code in conjunction with the optimise for size compiler flag `-Os`

leptun:
@GilesBathgate We already have LTO enabled. Without LTO the build would exceed the available flash.

DRracer:
@GilesBathgate yes, we are using LTO ... last time I checked, without LTO the firmware exceeds the CPU Flash size by ~50KB :wink: 

And currently we are using avr-gcc-7.3.x which is pretty recent (C++14 and a much more reliable LTO phase than gcc 4.9 which has been used before)

GilesBathgate:
@DRracer There might be some more of these duplicate strings, I tried:

`grep "\".*\"" * -Roh | sort | uniq -d | awk '{ print length, $0 }' | sort -n | cut -d" " -f2-`
Which currently yields: https://gist.github.com/GilesBathgate/58fb9c6182b2856fd728e2d4a29fc051

Although I suspect there are mostly false positives.

DRracer:
@GilesBathgate good point, I looked up a few of them and at least this one: `lcd_puts_at_P(0,3,_n("1/4"));` occurs twice in the source code and also twice in the binary.

@gudnimg a good idea for optimization :wink:

GilesBathgate:
@DRracer Yes cross-referencing them with the ones in the binary using the `strings` command could be a good linting approach ;)

GilesBathgate:
@DRracer Yes cross-referencing them with the ones in the binary using the `strings` command could be a good linting approach ;) Its odd that something like -Os or -flto doesn't remove them :shrug:

DRracer:
-Os and -flto works fine as long as you don't play with multiple sections (which is what we need for the translations). Then the automatic merging of strings at link time stops working.

leptun (approved):
looks good to me

PR #4030, opened 2023-02-18
optimsation: selftest: reduce code size in LCD rendering

gudnimg:
* Move "OK" to PROGMEM
* Treat `_indicator` as a single byte character, not a pointer to a character.

Tested on MK3S+ hardware, I did not see any difference in the LCD rendering.

Change in memory (MK3S+ Multilang):
Flash: -62 bytes
SRAM: -6 bytes


DRracer (approved):
nice :+1:

PR #4032, opened 2023-02-19
🚧 Improve MMU "Unload Filament" menu behavior

gudnimg:
Disable LCD updates while unloading to prevent the user from triggering another unload while the previous unload is not done yet.

Additionally, the following scenarios also show the fullscreen:
* When print is stopped, this restores 3.11 behavior
* When M702 is executed

Change in memory:
Flash: +36 bytes
SRAM: 0 bytes


gudnimg:
There is an issue where the fullscreen Preheat screen which shows the temperature status isn't being rendered. PR https://github.com/prusa3d/Prusa-Firmware/pull/4034 will fix that.

PR #4033, opened 2023-02-19
preheat: improve target temperature threshold

gudnimg:
Use 5°C threshold to be consistent with other parts of the firmware.

Relying on 95% of the target temperature creates
a dependency on the temperature. I don't see any other part of the firmware do this except the Preheat menu.

| Material       | Set Target           | Current Threshold  | Proposed Threshold |
| ------------- |-------------| -----|----|
| PLA     | 215°C  | 204.25 °C | 210°C |
| PETG   | 230°C  | 218.5 °C | 225°C |
| ABS     | 255°C  | 242.25 °C | 250°C |
| ASA     | 260°C  | 247 °C | 255°C |
| PC     | 275°C  | 261.25 °C | 270°C |

My proposal is we instead use a constant
`TEMP_HYSTERESIS = 5`, which is consistent with
M109, and behavior when restoring print from RAM
and some of the MMU code (like unload function)

Change in memory:
Flash: -14 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -11424,7 +11424,7 @@ void M600_wait_for_user(float HotendTempBckp) {
 				}
 				break;
 			case 2: //waiting for nozzle to reach target temperature
-				if (fabs(degTargetHotend(active_extruder) - degHotend(active_extruder)) < 1) {
+				if (fabs(degTargetHotend(active_extruder) - degHotend(active_extruder)) < TEMP_HYSTERESIS) {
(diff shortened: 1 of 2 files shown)


gudnimg:
Reason for me wanting this change is related to https://github.com/prusa3d/Prusa-Firmware/pull/4032, there is few second delay before the Full Screen message is rendered because the Preheat menu doesn't wait for the hotend to reach as high a temperature closer to the target.

DRracer (approved):

PR #4034, opened 2023-02-19
menu: always consume the knob click

gudnimg:
It's not a good idea to leave a button event unconsumed. This has been an issue for a long time but has never been addressed.

This issue is a blocker for the following PR https://github.com/prusa3d/Prusa-Firmware/pull/4032, Preheat screen is immediately canceled due to the unconsumed button click.

Change in memory:
Flash: -64 bytes
SRAM: 0 bytes


gudnimg:
Closing in favor of https://github.com/prusa3d/Prusa-Firmware/pull/4046

PR #4036, opened 2023-02-19
optimisation: MMU silent mode toggling & init

gudnimg:
Creating a separate PR for this optimisation due to saving some valuable SRAM.

Change in memory:
Flash: -16 bytes
SRAM: -2 bytes


gudnimg:
I wonder if we can get rid of `SilentModeMenu` as well since we have the value in EEPROM.

leptun (approved):

3d-gussner (approved):
Tested with `D3 Ax0da9 C1` on MK404 and the MMU Mode toggle menu works now.

PR #4037, opened 2023-02-19
Optimisation: Disable M351 when using TMC2130 and code related to it

gudnimg:
It looks to me this code is only used when not using TMC2130 steppers. If I hopefully didn't miss something, then this is pretty nice memory savings.

Change in memory (MK3S+ multilang build)
Flash: -250 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -8378,6 +8378,7 @@ Sigma_Exit:
     }
     break;
 
+#ifndef TMC2130
     /*!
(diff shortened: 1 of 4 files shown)


leptun (approved):
Checked on MK2.5. Everything still seems to be working correctly

3d-gussner (approved):
Tested with MK404
MK3S `M351` unknown :+1: 
MK2.5S
```
Send: M351
Recv: MS1,MS2 Pins
Recv: X: 11
Recv: Y: 11
Recv: Z: 11
Recv: E0: 11
Recv: ok
```

PR #4041, opened 2023-02-20
MK3_3.12: Set `Is filament loaded? Yes|No` default to `No` without a timeout.

3d-gussner:
- [x] Create MK3 branch PR
Normally I would prepare the MK3 PR first and cherry-pick it, but as the `lcd_show_fullscreen_message_yes_no_and_wait_P` are different it doesn't make sense.


gudnimg (approved):
Looks good to me 👍 false = No option in 3.12

PR #4043, opened 2023-02-20
MK3: Set `Is filament loaded? Yes|No` default to `No` without a timeout.

3d-gussner:
Same as https://github.com/prusa3d/Prusa-Firmware/pull/4041 for MK3_3.12 branch but newer


gudnimg (approved):
Looks good to me 👍

PR #4044, opened 2023-02-20
PFW-1460 `PF-build.sh` with `cmake` as default

3d-gussner:
- [X] Switch to cmake
  - [X] Keep Arduino IDE 1.8.5 and 1.8.19 as an option to build the firmware.
- [X]  `MK404-build.sh` after build still works
- [x] Update for https://github.com/prusa3d/Prusa-Firmware/pull/3859
- [ ] Rebase after cmake fixes
- [ ] update build server


CMakeLists.txt
```diff
@@ -2,52 +2,75 @@ cmake_minimum_required(VERSION 3.19)
...
-include(cmake/ProjectVersion.cmake)
-resolve_version_variables()
+set(PROJECT_DEV_VERSION
+    "<auto>"
+    CACHE
+      STRING
+      "Development version to check sliced gcode match firmware."
+    )
+if(PROJECT_DEV_VERSION STREQUAL "<auto>")
+  set(PROJECT_DEV_VERSION "${FW_DEV_VERSION}")
```
vintagepc, line 35: I've never been a fan of this sort of arbitrary string comparison to key. CMake defaults to false for blank strings so my preference would be either use that or adding additional OPTION bools to be explicit about it rather than magic strings.

PF-build.sh
```diff
@@ -530,32 +531,37 @@ fi
...
+    fi
+    if [ "$ARDUINO_ENV" == "1.8.19" ]; then
+        BOARD_URL="https://raw.githubusercontent.com/prusa3d/Arduino_Boards/master/IDE_Board_Manager/package_prusa3d_index.json"
+        #BOARD_URL="https://raw.githubusercontent.com/3d-gussner/Arduino_Boards/master/IDE_Board_Manager/package_prusa3d_index.json"
+    else
+        BOARD_URL="https://raw.githubusercontent.com/prusa3d/Arduino_Boards/master/IDE_Board_Manager/package_prusa3d_index.json"
+    fi
+    BOARD_FILENAME="prusa3dboards"
+    if [ "$ARDUINO_ENV" == "1.8.19" ]; then
+        BOARD_FILE_URL="https://raw.githubusercontent.com/prusa3d/Arduino_Boards/master/IDE_Board_Manager/prusa3dboards-$BOARD_VERSION.tar.bz2"
```
vintagepc, line 552: is this even needed anymore? don't we fetch it in bootstrap?

vintagepc:
... I don't know, all together this feels... unnecessarily complicated. But I realise there is a legacy boat anchor we're trying to drag around and continue to support.  The sooner we can clean it up the better.

wavexx:
I would rebase this against the current MK3 branch, but without PR #3665. I think the two are almost independent from each other, and if you look at the changes in the last 3 commits it's much more readable.

I still see the code inside has the code handling "!= cmake", which I would deliberately drop. As @vintagepc says, I think the point is that we want to make PF-build ultra friendly so you can just confirm the steps, but still use all the existing code as much as possible and get exactly the same build. This includes using bootstrap to get and setup the dependencies.

With this in mind, I expect PF-build to become very simple.

PF-build.sh
3d-gussner: I try to keep Arduino IDE active for one firmware release and don't want to change anything to that.

wavexx: Fair enough, but then this should be rebased on the current MK3 branch, without PR 3665. It's hard to see here what is changed for PF-build and all the other version data.

PR #4046, opened 2023-02-22
Refactor LCD knob rotation and click event handling

gudnimg:
This should do everything I wanted to do with https://github.com/prusa3d/Prusa-Firmware/pull/3473 but ALSO does it for LCD screens which have the LCD update disabled :) It's a win-win.

This commit adds the ability for firmware to make sounds when the knob is clicked or rotated when LCD updates are disabled. The improvement here is the sound is being made with one line of code regardless whether or not LCD updates are enabled or disabled. Should make things more consistent.

Some long standing issues fixed:

* Fixed an issue where menu items don't consume knob click events consistently.
* Fixed an issue when clicking on a menu item, the buzzer can produce two or more beeps. We can now better control whether a submenu should create a feedback sound. We don't want to do this when clicking the knob, otherwise there are two beeps.
* Using `lcd_buttons` to check for a knob click was very unreliable since it is cleared on the next Temperature ISR service. Now fixed.

Change in memory:
Flash: -196 bytes
SRAM: -9 bytes

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -9478,11 +9478,11 @@ void manage_inactivity(bool ignore_stepper_queue/*=false*/) //default argument s
   // handle longpress
   if(lcd_longpress_trigger)
   {
+      lcd_consume_click(); // Reset trigger to prevent recursion
       // long press is not possible in modal mode, wait until ready
       if (lcd_longpress_func && lcd_update_enabled)
       {
           lcd_longpress_func();
-          lcd_longpress_trigger = 0;
       }
   }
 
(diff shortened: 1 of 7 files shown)


gudnimg:
Threw myself into the menu / LCD knob rabbit hole again. But so far I have fixed all the issues I have found.

`lcd_encoder` is now updated regardless whether or not LCD updates are disabled or enabled. This helps save 9B of RAM!

To solve the issue I had to add these PRs into this one:

* https://github.com/prusa3d/Prusa-Firmware/pull/4034
* https://github.com/prusa3d/Prusa-Firmware/pull/4048

Some long standing issues fixed:

* Fixed an issue where menu items don't consume knob click events consistently.
* Fixed an issue when clicking on a menu item, the buzzer can produce two or more beeps. We can now better control whether a submenu should create a feedback sound. We don't want to do this when clicking the knob, otherwise there are two beeps.
* Using `lcd_buttons` to check for a knob click was very unreliable since it is cleared on the next Temperature ISR service. Now fixed.

DRracer (approved):
Even though there are so many changes at various places, the PR is well split into individual commits which make the review much easier.
After going through the proposed changes I don't see anything obviously wrong with this PR -> approving.

PR #4047, opened 2023-02-22
Optimisation: change `lcd_encoder` from `int32_t` to  `int16_t`

gudnimg:
The 2nd commit is from https://github.com/prusa3d/Prusa-Firmware/pull/3985

Allowing `lcd_encoder` to be `int32_t` is very inefficient in terms of program size. It should never exceed more than 127 (int8_t). This PR also removes `lcd_encoder` from the Menu Editing menu items.

Change in memory:
Flash: -912 bytes
SRAM: -2 bytes

### DIFF
--- a/Firmware/menu.cpp
+++ b/Firmware/menu.cpp
@@ -470,14 +470,14 @@ static void _menu_edit_P(void)
 	menu_data_edit_t* _md = (menu_data_edit_t*)&(menu_data[0]);
 	if (lcd_draw_update)
 	{
-		if (lcd_encoder < _md->minEditValue) lcd_encoder = _md->minEditValue;
-		else if (lcd_encoder > _md->maxEditValue) lcd_encoder = _md->maxEditValue;
+		// Constrain the value in case it's outside the allowed limits
+		_md->currentValue = constrain(_md->currentValue, _md->minEditValue, _md->maxEditValue);
 		lcd_set_cursor(0, 1);
-		menu_draw_P(' ', _md->editLabel, lcd_encoder);
+		menu_draw_P(' ', _md->editLabel, _md->currentValue);
 	}
 	if (lcd_clicked())
 	{
-		*((T)(_md->editValue)) = lcd_encoder;
+		*((T)(_md->editValue)) = _md->currentValue;
 		menu_back_no_reset();
 	}
 }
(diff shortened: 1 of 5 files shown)


DRracer (approved):
nice :+1: 
I don't see anything obviously wrong with this PR -> approved

Firmware/menu.cpp
```diff
@@ -65,7 +65,7 @@ void menu_goto(menu_func_t menu, const uint32_t encoder, const bool feedback, bo
 
 void menu_start(void)
 {
-    if (lcd_encoder > 0x8000) lcd_encoder = 0;
+    if (lcd_encoder > INT8_MAX) lcd_encoder = 0;
```
leptun: This change is a bit concerning. The SD card menu can easily exceed 127 items.

leptun (changes requested):

Firmware/menu.cpp
DRracer: Is this really the total number of menu items?
Anyway, we can change it to `INT16_MAX`

leptun: I may be wrong, but I think that line can be removed entirely. I don't really see the point of it. The intention seems to be to check that the lcd_encoder doesn't get too large, in which case it's reset to 0. That's all nice and good, except now that lcd_encoder is a int16_t, the range of lcd_encoder is reduced. So the maximum positive value it can have is 0x7fff. If that check is removed entirely, then lcd_encoder is allowed to overflow and become negative, something which is checked a bit lower in the code and it's handled almost the same way as the code I'm proposing to remove. This should work around the issue for menus with many items, while still working more or less the same as before.

DRracer: ok, makes sense ... I'll remove the line

Firmware/ultralcd.cpp
```diff
@@ -2848,7 +2848,7 @@ bool lcd_calibrate_z_end_stop_manual(bool only_z)
                 lcd_encoder_diff = 0;
                 if (! planner_queue_full()) {
                     // Only move up, whatever direction the user rotates the encoder.
-                    current_position[Z_AXIS] += fabs(lcd_encoder);
+                    current_position[Z_AXIS] += lcd_encoder;
```
leptun: Is this correct? Does the new code always move up? Because I have a feeling it won't preserve the old behaviour. Looks like it still needs an abs().

DRracer: `abs()` call added

DRracer:
rebased onto MK3

leptun (approved):

PR #4048, opened 2023-02-23
buttons: Save 4 clock cycles in Temperature ISR

gudnimg:
* Use `newbutton` for encoder bits. No need to parse `EN_A` and `EN_B` bits twice.
* Don't set `lcd_encoder_bits` unless `newbutton` has changed.
* Add `ENCODER_SPIN()` to make the code a bit more readable. I spotted this in Marlin https://github.com/MarlinFirmware/Marlin/blob/d3527f5de42f087bfe313aa475b0bf5d07559fed/Marlin/src/lcd/marlinui.cpp#L1394 and think it is a good idea 👍 

Change in memory:
Flash: -8 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/lcd.cpp
+++ b/Firmware/lcd.cpp
@@ -818,8 +818,9 @@ void lcd_buttons_update(void)
 		if (abs(lcd_encoder_diff) >= ENCODER_PULSES_PER_STEP) {
 			lcd_backlight_wake_trigger = true; // flag event, knob rotated
 		}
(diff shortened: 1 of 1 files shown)


gudnimg:
Closing this in favor of https://github.com/prusa3d/Prusa-Firmware/pull/4046 

We can still merge this one separately if we want. In that case feel free to re-open.

PR #4050, opened 2023-02-24
Introduce MMU EEPROM reset feature

DRracer:
This new feature requires MMU FW v 2.1.7 (which brings some more small changes). It is not being used yet in the MK3S FW, but we may decide to add it as an option into some menu.

MMU-193


gudnimg (approved):
Looks good to me 👍

PR #4054, opened 2023-02-26
M200: Don't process T parameter, active extruder should always be selected

gudnimg:
The active extruder is always selected regardless if the user has an MMU or not. This simplifies the M200 code a little bit and we save some precious memory too.

There is no need for us to update the RepRap wiki since there is no specific handling documented for Prusa firmware. :) 

Regarding the T parameter, we can rely on a similar definition as Marlin:

> T[index] Select the target extruder. If omitted, the active extruder.

Where Prusa Firmware simply assumes the T parameter is always omitted, and so active extruder is always used.

Change in memory:
Flash: -60 bytes
SRAM: 0 bytes


DRracer (changes requested):
I'm not sure if we can safely do this change.
Having M200 T0 , T1, T2... at the start for the MMU code is something to expect (not saying it is being extensively used though)

PR #4057, opened 2023-02-28
Implement variant of enquecommand which accepts format string

gudnimg:
This eliminates many local buffers.

This is just an idea at the moment. Adding `vsnprintf_P()` to the firmware adds about 200 bytes. But we can use `enquecommandf()` in so many places we end up saving a lot of memory.

Change in memory:
Flash: -420 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3594,7 +3594,7 @@ static void gcode_M600(bool automatic, float x_position, float y_position, float
     
         // Recover feed rate
         feedmultiply = feedmultiplyBckp;
-        enquecommandf(MSG_M220, feedmultiplyBckp);
+        enquecommandf_P(MSG_M220, feedmultiplyBckp);
     }
     
     lcd_setstatuspgm(MSG_WELCOME);
@@ -10776,16 +10776,16 @@ void recover_print(uint8_t automatic) {
   // and second also so one may remove the excess priming material.
   if(eeprom_read_byte((uint8_t*)EEPROM_UVLO) == 1)
   {
-      enquecommandf(PSTR("G1 Z%.3f F800"), current_position[Z_AXIS] + 25);
+      enquecommandf_P(PSTR("G1 Z%.3f F800"), current_position[Z_AXIS] + 25);
   }
 
   // Home X and Y axes. Homing just X and Y shall not touch the babystep and the world2machine
   // transformation status. G28 will not touch Z when MBL is off.
 	enquecommand_P(PSTR("G28 X Y"));
   // Set the target bed and nozzle temperatures and wait.
-  enquecommandf(PSTR("M104 S%d"), target_temperature[active_extruder]);
-  enquecommandf(PSTR("M140 S%d"), target_temperature_bed);
-  enquecommandf(PSTR("M109 S%d"), target_temperature[active_extruder]);
+  enquecommandf_P(PSTR("M104 S%d"), target_temperature[active_extruder]);
+  enquecommandf_P(PSTR("M140 S%d"), target_temperature_bed);
+  enquecommandf_P(PSTR("M109 S%d"), target_temperature[active_extruder]);
 	enquecommand_P(MSG_M83); //E axis relative mode
 
     // If not automatically recoreverd (long power loss)
@@ -10793,7 +10793,7 @@ void recover_print(uint8_t automatic) {
         //Extrude some filament to stabilize the pressure
         enquecommand_P(PSTR("G1 E5 F120"));
         // Retract to be consistent with a short pause
-        enquecommandf(G1_E_F2700, default_retraction);
+        enquecommandf_P(G1_E_F2700, default_retraction);
     }
 
 	printf_P(_N("After waiting for temp:\nCurrent pos X_AXIS:%.3f\nCurrent pos Y_AXIS:%.3f\n"), current_position[X_AXIS], current_position[Y_AXIS]);
@@ -10916,7 +10916,7 @@ void restore_print_from_eeprom(bool mbl_was_active) {
 
 	MYSERIAL.print(filename);
   strcat_P(filename, PSTR(".gco"));
-  enquecommandf(MSG_M23, filename);
+  enquecommandf_P(MSG_M23, filename);
 	uint32_t position = eeprom_read_dword((uint32_t*)(EEPROM_FILE_POSITION));
 	SERIAL_ECHOPGM("Position read from eeprom:");
 	MYSERIAL.println(position);
@@ -10927,7 +10927,7 @@ void restore_print_from_eeprom(bool mbl_was_active) {
     float pos_y = eeprom_read_float((float*)(EEPROM_UVLO_CURRENT_POSITION + 4));
     if (pos_x != X_COORD_INVALID)
     {
-        enquecommandf(PSTR("G1 X%f Y%f F3000"), pos_x, pos_y);
+        enquecommandf_P(PSTR("G1 X%f Y%f F3000"), pos_x, pos_y);
     }
 
     // Enable MBL and switch to logical positioning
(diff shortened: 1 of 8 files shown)


Firmware/cmdqueue.cpp
```diff
@@ -251,6 +252,19 @@ void cmdqueue_dump_to_serial()
 static const char bufferFull[] PROGMEM = "\" failed: Buffer full!";
 static const char enqueingFront[] PROGMEM = "Enqueing to the front: \"";
 
+
+void enquecommandf(const char *fmt, ...)
```
leptun: proposal for another name: enqueuecommandf_P() since the fmt is from PROGMEM. This is how I've tried naming anything PGM related in the past.

gudnimg: Done ✔️

gudnimg: Note sure about changing "enque" to "enqueue" without also applying that to the other "enque" functions.

Firmware/cmdqueue.cpp
```diff
@@ -251,6 +252,19 @@ void cmdqueue_dump_to_serial()
...
+{
+    // MAX_CMD_SIZE is 96, but for formatting
+    // string we usually don't need more than 30 bytes
+    char cmd_buffer[30];
```
leptun, line 260: Idea: Could we maybe work around the need for allocating an array on the stack? I think it would be more efficient if we made a function in the cmdqueue code to get a pointer for writing (so it allocates a place in the cmdqueue that can fit MAX_CMD_SIZE, and then we format with vsnprintf_P on that pointer and then we "finalize" the command by telling the cmdqueue what is the length of the string that was added, so that less than MAX_CMD_SIZE is reserved after doing this enqueuing operation. This way you don't allocate the buffer on the stack (less stack footprint, so less chaces of overwriting static variables) and it also avoids having to copy data from one temporary buffer to the cmdqueue, so it should be faster (marginally, but still faster nonetheless).

Firmware/cmdqueue.cpp
leptun: Oh, I didn't see that typo. Good catch. I think we should name it like the other functions and just deal with renaming everything at a later time (if ever).

gudnimg:
Rebased to sync with MK3 branch. ~50 commits

Firmware/cmdqueue.cpp
DRracer: This is actually a good idea, because the formatted string ends up in the cmd queue anyway, so why not writing directly into the target memory area ... might not be simple to achieve in the current API though...

DRracer: The tricky part here would be knowing the formatted string in advance to prepare the buffer's internal indices correctly.
The funny point is, that `enquecommand` actually may silently skip a g-code line entirely when `cmdqueue_could_enqueue_back` fails to find enough space in the buffer.

Firmware/Marlin_main.cpp
```diff
@@ -11163,15 +11141,13 @@ void stop_and_save_print_to_ram(float z_move, float e_move)
         // A snprintf would have been a safer call, but since it is not used
         // in the whole program, its implementation would bring more bytes to the total size
         // The behavior of dtostrf 8,3 should be roughly the same as %-0.3
-        sprintf_P(buf, G1_E_F2700, e_move);
-        enquecommand(buf, false);
+        enquecommandf(G1_E_F2700, e_move);
```
DRracer: here, the original buffer size was expected to be 48 bytes, but:
`G1_E_F2700` looks like this: `"G1 E%-0.3f F2700"`
that's 10 characters + up to ~8+1+3 characters for the E coordinate ... should fit into those 30 bytes of buffer in `enquecommandf`

Firmware/Marlin_main.cpp
```diff
@@ -11163,15 +11141,13 @@ void stop_and_save_print_to_ram(float z_move, float e_move)
...
     }
 
     if(z_move)
     {
         // Then lift Z axis
-        sprintf_P(buf, PSTR("G1 Z%-0.3f F%-0.3f"), saved_pos[Z_AXIS] + z_move, homing_feedrate[Z_AXIS]);
-        enquecommand(buf, false);
+        enquecommandf(PSTR("G1 Z%-0.3f F%-0.3f"), saved_pos[Z_AXIS] + z_move, homing_feedrate[Z_AXIS]);
```
DRracer: `"G1 Z%-.3f F%-.3f"` should produce no more than:
6 + 3+1+3 + 8+1+3 bytes -> that's up to 25 bytes (considering the fact, that the Z coordinate is no more than 999.999)

Firmware/Marlin_main.cpp
```diff
@@ -10939,46 +10927,37 @@ void restore_print_from_eeprom(bool mbl_was_active) {
     float pos_y = eeprom_read_float((float*)(EEPROM_UVLO_CURRENT_POSITION + 4));
     if (pos_x != X_COORD_INVALID)
     {
-        sprintf_P(cmd, PSTR("G1 X%f Y%f F3000"), pos_x, pos_y);
-        enquecommand(cmd);
+        enquecommandf(PSTR("G1 X%f Y%f F3000"), pos_x, pos_y);
```
DRracer: This format string should be clamped to some fixed size ... a single %f could theoretically produce up to 19 characters ... and having 2 of them in the format string will overflow the 30 bytes buffer in `enquecommandf`.

However, we can safely limit the X and Y range to machine limits, so having `X%-.3f` should be safe and precise enough.

Firmware/Marlin_main.cpp
```diff
@@ -10939,46 +10927,37 @@ void restore_print_from_eeprom(bool mbl_was_active) {
...
     }
 
     // Enable MBL and switch to logical positioning
     if (mbl_was_active)
         enquecommand_P(PSTR("PRUSA MBL V1"));
 
     // Move the Z axis down to the print, in logical coordinates.
-    sprintf_P(cmd, PSTR("G1 Z%f"), eeprom_read_float((float*)(EEPROM_UVLO_CURRENT_POSITION_Z)));
-	enquecommand(cmd);
+    enquecommandf(PSTR("G1 Z%f"), eeprom_read_float((float*)(EEPROM_UVLO_CURRENT_POSITION_Z)));
```
DRracer: `%-.3f`

Firmware/Marlin_main.cpp
```diff
@@ -10939,46 +10927,37 @@ void restore_print_from_eeprom(bool mbl_was_active) {
...
 
     // Restore acceleration settings
     float acceleration = eeprom_read_float((float*)(EEPROM_UVLO_ACCELL));
     float retract_acceleration = eeprom_read_float((float*)(EEPROM_UVLO_RETRACT_ACCELL));
     float travel_acceleration = eeprom_read_float((float*)(EEPROM_UVLO_TRAVEL_ACCELL));
-    sprintf_P(cmd, PSTR("M204 P%f R%f T%f"), acceleration, retract_acceleration, travel_acceleration);
-    enquecommand(cmd);
+    enquecommandf(PSTR("M204 P%f R%f T%f"), acceleration, retract_acceleration, travel_acceleration);
```
DRracer: Accelerations are usually ordinary numbers, no need to keep extensive amount of decimal places
`enquecommandf_P(PSTR("M204 P%-.1f R%-.1f T%-.1f")...`

Firmware/Marlin_main.cpp
```diff
@@ -10939,46 +10927,37 @@ void restore_print_from_eeprom(bool mbl_was_active) {
...
 
   // Unretract.
-    sprintf_P(cmd, G1_E_F2700, default_retraction);
-    enquecommand(cmd);
+    enquecommandf(G1_E_F2700, default_retraction);
   // Recover final E axis position and mode
     float pos_e = eeprom_read_float((float*)(EEPROM_UVLO_CURRENT_POSITION_E));
-    sprintf_P(cmd, PSTR("G92 E%6.3f"), pos_e);
-    enquecommand(cmd);
+    enquecommandf(PSTR("G92 E%6.3f"), pos_e);
```
DRracer: `E%-.3f`

DRracer (approved):
Approving the original idea from @gudnimg

3d-gussner (approved):
Tested on Mk404
First layer cal and PID tuning

PR #4063, opened 2023-03-04
PFW-1503 Improve live Z babystep menu handling

gudnimg:
Short history:
* I broke the behavior in here https://github.com/gudnimg/Prusa-Firmware/commit/2276217cbb47501a0558d988446c8d1a863a3384 with optimisations. This created a bug where the menu would close and not save the last Z value into EEPROM.
* The bug above was fixed recently here https://github.com/prusa3d/Prusa-Firmware/pull/4005 by @3d-gussner but later we discovered this created another bug where `menu_leave` is set to `1`, but it is then never set to `0` again. This bug appears only when first layer calibration is run twice consequatively and the LCD will freeze or stop behaving correctly.

In this PR I have a proposed solution which fixes the reported problem:

Improvements
====

* Add `BABYSTEP_ALLOWED` function to unify checks whether or not babystepping is allowed, if not the menu is dismissed and hidden from the user.
* The Z baby step menu closes automatically when first layer calibration is done. This is expected behavior but additionally now we don't need to wait for a timeout or close the menu manually by setting a variable.
* Do not allow the Z baby step menu to timeout when LcdCommands != Idle. This covers first layer calibration. In all other situations it times out as normal (after 90 seconds).

Change in memory:
Flash: -182 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -354,7 +354,11 @@ extern LongTimer safetyTimer;
 
 #define PRINT_PERCENT_DONE_INIT 0xff
 
-extern bool printer_active();
+// Returns true if there is a print running. It does not matter if
+// the print is paused, that still counts as a "running" print.
+bool printJobOngoing();
+
+bool printer_active();
 
 //! Beware - mcode_in_progress is set as soon as the command gets really processed,
 //! which is not the same as posting the M600 command into the command queue
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -370,7 +370,7 @@ bool check_fsensor();
 //! 2) Not allowed during Homing (printer busy)
 //! 3) Not allowed during Mesh Bed Leveling (printer busy)
 //! 4) Allowed if there are queued blocks OR there is a print job running
-bool BABYSTEP_ALLOWED();
+bool babystep_allowed();
 
 extern void calculate_extruder_multipliers();
 
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -373,7 +373,9 @@ bool check_fsensor();
 //! 1) Z-axis position is less than 2.0mm (only allowed during the first couple of layers)
 //! 2) Not allowed during Homing (printer busy)
 //! 3) Not allowed during Mesh Bed Leveling (printer busy)
-//! 4) Allowed if there are queued blocks OR there is a print job running
+//! 4) Allowed if:
+//!         - First Layer Calibration is running
+//!         - OR there are queued blocks, printJob is running and it's not paused, and Z-axis position is less than 2.0mm (only allowed during the first couple of layers)
 bool babystep_allowed();
 
 extern void calculate_extruder_multipliers();
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -366,7 +366,7 @@ extern bool printer_active();
 bool check_fsensor();
 
 //! Condition where Babystepping is allowed:
-//! 1) Axis are trusted: Z-axis position must be known
+//! 1) Z-axis position is less than 2.0mm (only allowed during the first couple of layers)
 //! 2) Not allowed during Homing (printer busy)
 //! 3) Not allowed during Mesh Bed Leveling (printer busy)
 //! 4) Allowed if there are queued blocks OR there is a print job running
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -365,6 +365,13 @@ extern bool printer_active();
 //! I'd normally change this macro, but who knows what would happen in the MMU :)
 bool check_fsensor();
 
+//! Condition where Babystepping is allowed:
+//! 1) Axis are trusted: Z-axis position must be known
+//! 2) Not allowed during Homing (printer busy)
+//! 3) Not allowed during Mesh Bed Leveling (printer busy)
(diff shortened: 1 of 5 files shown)


gudnimg:
I suspect my `blocks_queued()` condition is not sufficient alone. Sometimes the screen is being dismissed. Need to think about how to add more to this check without increasing the code size too much.

Firmware/Marlin.h
```diff
@@ -365,6 +365,13 @@ extern bool printer_active();
 //! I'd normally change this macro, but who knows what would happen in the MMU :)
 bool check_fsensor();
 
+//! Condition where Babystepping is allowed:
+//! 1) Axis are trusted: Z-axis position must be known
+//! 2) Not allowed during Homing (printer busy)
+//! 3) Not allowed during Mesh Bed Leveling (printer busy)
+//! 4) Allowed if there are queued blocks OR there is a print job running
+bool BABYSTEP_ALLOWED();
```
DRracer: /nitpick/: I'd avoid naming the function in uppercase letters

Firmware/ultralcd.cpp
```diff
@@ -5342,7 +5332,7 @@ static void lcd_main_menu()
     MENU_ITEM_FUNCTION_P(PSTR("power panic"), uvlo_);
 #endif //TMC2130_DEBUG
 
-    if ( ( IS_SD_PRINTING || usb_timer.running() || (lcd_commands_type == LcdCommands::Layer1Cal)) && (current_position[Z_AXIS] < Z_HEIGHT_HIDE_LIVE_ADJUST_MENU) && !homing_flag && !mesh_bed_leveling_flag) {
+    if ( BABYSTEP_ALLOWED() ) {
```
DRracer: beware: the `current_position[Z_AXIS] < Z_HEIGHT_HIDE_LIVE_ADJUST_MENU` is important. The check was not about the known positions of Z, but about allowing to enter the live-z menu during printing of the first few layers (and not allow any later)

Firmware/ultralcd.cpp
```diff
@@ -7457,13 +7447,11 @@ void menu_lcd_longpress_func(void)
 
     // explicitely listed menus which are allowed to rise the move-z or live-adj-z functions
     // The lists are not the same for both functions, so first decide which function is to be performed
-    if ( (moves_planned() || IS_SD_PRINTING || usb_timer.running() )){ // long press as live-adj-z
-        if (( current_position[Z_AXIS] < Z_HEIGHT_HIDE_LIVE_ADJUST_MENU ) // only allow live-adj-z up to 2mm of print height
-        && ( menu_menu == lcd_status_screen // and in listed menus...
+    if (BABYSTEP_ALLOWED()){ // long press as live-adj-z
```
DRracer: oh, this is the primary spot for the `current_position[Z_AXIS] < Z_HEIGHT_HIDE_LIVE_ADJUST_MENU` check

Firmware/Marlin.h
gudnimg: Fixed in [65e9e86](https://github.com/prusa3d/Prusa-Firmware/pull/4063/commits/65e9e864c0eb8b1fd9bdac9ac549ed4b981e1976)

Firmware/ultralcd.cpp
gudnimg: Github marks this as outdated due to me rebasing the branch to sync with MK3. This comment is still relevant 👍

Firmware/ultralcd.cpp
gudnimg: Github marks this as outdated due to me rebasing the branch to sync with MK3. This comment is still relevant 👍

Firmware/ultralcd.cpp
gudnimg: [e3b9e9e](https://github.com/prusa3d/Prusa-Firmware/pull/4063/commits/e3b9e9efe0aeb9361faa7a7e93c877f250c540fe) This is now applied everywhere so it's consistant in behavior.

Firmware/ultralcd.cpp
gudnimg: [e3b9e9e](https://github.com/prusa3d/Prusa-Firmware/pull/4063/commits/e3b9e9efe0aeb9361faa7a7e93c877f250c540fe) This is now applied everywhere so it's consistant in behavior.

DRracer (approved):
Even though the conditions are a bit different compared to the previous implementation, it looks acceptable to me :+1: 

We should gather more feedback during testing

3d-gussner (changes requested):
On MK404 during LCD -> Calibration -> First layer cal. it jumps out quite fast to status screen, while it should stay in the Live Z-adjust menu until it is finished.

gudnimg:
> On MK404 during LCD -> Calibration -> First layer cal. it jumps out quite fast to status screen, while it should stay in the Live Z-adjust menu until it is finished.

Ah I think it is this condition kicking in: `current_position[Z_AXIS] < Z_HEIGHT_HIDE_LIVE_ADJUST_MENU)` I saw this before but forgot about it when I added the condition back. 

The extruder lifts up above 2mm while moving to the start position. 

Unfortuately I can’t fix it until this evening (after work). Feel free to add a patch if possible.

gudnimg:
@3d-gussner I've fixed the issue, please re-test when you have time :)

3d-gussner (approved):
Tested on MK404

PR #4065, opened 2023-03-05
Firmware_fr: Updating PO file

ErwanAliasr1:
This commit is about fixing the french PO file by
- Adding proper accents on words to ease the reading
- Fix some typos
- Enforce the same abbreviations on the same word
- Add one missing translation
- Fix non-french sentences


3d-gussner:
@ErwanAliasr1 Thanks for the translation improvements. But there is an official update coming soon, so please wait until that has been merged and then it would be great if you review it.

ErwanAliasr1:
@3d-gussner Ok I'll drop the PR. What version will bring improvement ? I could make a pass if needed.

3d-gussner:
It should be bit shorter as we have some new messages to be translated and ran out of space.
French is quite big compared to the other languages.
Like German too formal and polite 😉

PR #4072, opened 2023-03-09
Fix AutoLoad filament menu not shown when FSensor is [Off]

3d-gussner:


Firmware/ultralcd.cpp
```diff
@@ -5428,7 +5428,7 @@ static void lcd_main_menu()
 #endif //MMU_HAS_CUTTER
         } else {
 #ifdef FILAMENT_SENSOR
-            if (fsensor.getAutoLoadEnabled()) {
+            if (fsensor.getAutoLoadEnabled() && fsensor.isEnabled()) {
```
leptun: Just a nitpick, I would personally put the two things that are && together the other way around, so that if the fsensor is not enabled, then getAutoLoadEnabled() doesn't get called

3d-gussner: @leptun Feel free to modify this PR

DRracer (approved):

PR #4076, opened 2023-03-11
PFW-1504 Add visualization on LCD

gudnimg:
This PR adds visualization for the try-load-unload sequence.

Change in memory:
Flash: +194 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/mmu2_marlin1.cpp
+++ b/Firmware/mmu2_marlin1.cpp
@@ -27,6 +27,10 @@ bool planner_any_moves() {
 }
 
 float planner_get_machine_position_E_mm(){
+    return current_position[E_AXIS];
+}
+
+float stepper_get_machine_position_E_mm(){
     return st_get_position_mm(E_AXIS);
 }
 
--- a/Firmware/mmu2_marlin1.cpp
+++ b/Firmware/mmu2_marlin1.cpp
@@ -27,7 +27,7 @@ bool planner_any_moves() {
 }
 
 float planner_get_machine_position_E_mm(){
-    return current_position[E_AXIS];
+    return st_get_position_mm(E_AXIS);
 }
 
 float planner_get_current_position_E(){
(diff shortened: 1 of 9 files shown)


Firmware/mmu2_marlin1.cpp
```diff
@@ -27,7 +27,7 @@ bool planner_any_moves() {
 }
 
 float planner_get_machine_position_E_mm(){
-    return current_position[E_AXIS];
+    return st_get_position_mm(E_AXIS);
```
DRracer, line 34: we should introduce another function `stepper_get_machine_position_E_mm`

Firmware/mmu2_reporting.cpp
```diff
@@ -289,6 +289,20 @@ void ReportProgressHook(CommandInProgress cip, uint16_t ec) {
     }
 }
 
+void TryLoadUnloadProgressbarInit() {
+    // Clear the status line
+    lcd_set_cursor(0, 3);
+    lcd_space(LCD_WIDTH);
+
+    // Reset cursor position
+    lcd_set_cursor(0, 3);
```
DRracer: Are you sure no other drawing function accesses the LCD while doing the progress bar?
Other functions may screw the cursor and we may start printing to some other position.

gudnimg: That's a good point. I'll add back `lcd_putc_at`, it will increase code size probably but it should be safer for now.

Firmware/mmu2_marlin1.cpp
gudnimg: Fixed in [9f02bec](https://github.com/prusa3d/Prusa-Firmware/pull/4076/commits/9f02bec7f293092cd5acd811cbe01e622e1f2534)

Firmware/mmu2_reporting.cpp
gudnimg: Fixed with [78ea5ee](https://github.com/prusa3d/Prusa-Firmware/pull/4076/commits/78ea5ee5da5b330877e0c17beff5c52b0df77a68)

DRracer (approved):
looks good to me :+1:

gudnimg:
There is one issue with my current implementation. Sometimes it will not render the last pixel due to delays in the while loop. I'm currently looking into how to correct this behavior.

DRracer (approved):

PR #4077, opened 2023-03-11
optimisation: reduce code size in status screen

gudnimg:
Had this stashed locally, thought I should commit in case we need the memory.

Good ideas from Marlin 2. IMO this new code is a little bit more readable.

Change in memory:
Flash: -36 bytes
SRAM: 0 bytes


gudnimg:
Closing in favor of https://github.com/prusa3d/Prusa-Firmware/pull/4156

PR #4080, opened 2023-03-13
PFW-1504 fix conflicting UI issues

gudnimg:
Proposal to fix some of the issues with the initial implementation it is safer to use the status line code to print the message so there aren't any conflicts in the LCD cursor position.

Allow inserting a byte into any position in the LCD status message

Also, add a variable to control from which index in the array should the message start printing. This is very useful for progress bars and messages which continually update. I think we can save some memory by applying this to Mesh Bed Leveling later.

Change in memory:
Flash: +208 bytes
SRAM: +1 byte

### DIFF
--- a/Firmware/lcd.cpp
+++ b/Firmware/lcd.cpp
@@ -546,14 +546,14 @@ void lcd_print(const char* s)
 	while (*s) lcd_write(*(s++));
 }
 
-char lcd_print_pad(const char* s, uint8_t len)
+uint8_t lcd_print_pad(const char* s, uint8_t len)
 {
     while (len && *s) {
         lcd_write(*(s++));
         --len;
     }
     lcd_space(len);
-    return *s;
+    return len;
 }
 
 uint8_t lcd_print_pad_P(const char* s, uint8_t len)
--- a/Firmware/lcd.h
+++ b/Firmware/lcd.h
@@ -56,7 +56,7 @@ extern void lcd_space(uint8_t n);
 extern void lcd_printNumber(unsigned long n, uint8_t base);
 
 extern void lcd_print(const char*);
-extern char lcd_print_pad(const char* s, uint8_t len);
+extern uint8_t lcd_print_pad(const char* s, uint8_t len);
 
 /// @brief print a string from PROGMEM with left-adjusted padding
 /// @param s string from PROGMEM.
--- a/Firmware/mmu2.cpp
+++ b/Firmware/mmu2.cpp
@@ -305,6 +305,7 @@ bool MMU2::VerifyFilamentEnteredPTFE() {
     }
 
     Disable_E0();
+    TryLoadUnloadProgressbarDeinit();
 
     if (fsensorState) {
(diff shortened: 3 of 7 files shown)


gudnimg:
Got everything working finally when printing and running toolchanges. Synced with MK3 branch (will hopefully re-test tomorrow)

gudnimg:
Oof, after a rebase to sync with MK3 branch. The flash consumption increased from 94B to 114B. 🤔

gudnimg:
Converting to draft. The solution for the 4 second delay in https://github.com/prusa3d/Prusa-Firmware/pull/4098 is not very nice and not even in my initial PFW-1504 PR (though it 'works'). It tempting to re-write the status line code but I probably don't have time to do it :) + I would 100% be creating more bugs than solving existing ones.

I also don't like how much memory this PR uses.

Need to think more about this.

gudnimg:
Adding to this PR discussion comments from @wavexx in https://github.com/prusa3d/Prusa-Firmware/pull/4098

> Reading through commit https://github.com/prusa3d/Prusa-Firmware/commit/bbfc4b9f7853525d36a74e09ffc14366bcd9edbc it seems that we touch `lcd_status_message_timeout` with the intention to avoid the display to be updated so that we can keep the previous changes for the filament status on-screen.

> 
> `lcd_status_message_timeout` is used to prevent the message from being changed depending on priority, but it still needs to be displayed in this function. This is what is causing the screen to become unresponsive then the thermal timeout is happening. We can't use the timer like this, and we can't combine it for this purpose.

> 
> `INFO` needs to have higher priority, and can't be expired by something like the filament sensor.

> 
> We could introduce a new message priority (something like `LCD_STATUS_NOCHANGE`) inbetween `INFO` and `NONE`, and handle that with a different timeout. The priority needs to be handled in `lcd_message_check` (which currently only handles `INFO`, since other priorities do not expire) and use it to see if you can actually set/reset the message priority and then expire it properly.
> 
> I am referring to the progress drawn by `TryLoadUnloadProgressbar()`, just to clarify.

gudnimg:
Feel free to review this PR for RC1/RC2. It works so far, only thing left to do is find a better solution for the time delay. I won't be able to look into it until next week.

gudnimg:
Should we move this to RC2?

leptun:
If it's not ready till tomorrow, then yes. I'd like to avoid blocking the RC1 testing task on Monday unless something is broken and needs fixing asap.

gudnimg:
The issue addressed in this PR is not a blocker IMO. It's only UI related. The LCD visualization code needs to be changed anyway due to changes on the MK4 side so I think it's best to wait until then.

gudnimg:
Dropped one commit. Also synced branch to latest MK3.

gudnimg:
Rebased to sync the branch. Resolved one conflict.

gudnimg:
Rebased to sync with MK3 branch (48 commits)

3d-gussner (approved):
Looks good to me

PR #4073, opened 2023-03-09
WIP waiting for community feedback: MK3 PFW-1445 shorten translations

3d-gussner:
Shorten messages

PR #4082, opened 2023-03-14
PFW-1445: Shorten and updated translations 

3d-gussner:
Based on https://github.com/prusa3d/Prusa-Firmware/pull/4073


jmd73:
@3d-gussner There is a error in the swedish translation (it is visible in the first "Fix Swedish".)

In row 738, there is an extra "d" in the word "ladddades". The word should be "laddades".
bild

leptun (approved):

PR #4091, opened 2023-03-18
🚧 optimisation: Simplify if statement check

gudnimg:
Let's call:
`m_started` := X
`msPeriod` := Y

Then `m_started <=  m_started + msPeriod` becomes
```math
X ≤  X + Y\\
```
Subtract X from both sides gives
```math
0 ≤ Y
```
or equivalently:
```math
Y ≥ 0
```
For this reason,
`m_started <=  m_started + msPeriod` is equivalent to `msPeriod >= 0`

Using ≥ instead of ≤ seems to save 2 bytes

Change in memory:
Flash: -40 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Timer.cpp
+++ b/Firmware/Timer.cpp
@@ -46,7 +46,7 @@ bool Timer<T>::expired(T msPeriod)
     if (!m_isRunning) return false;
     bool expired = false;
     const T now = _millis();
-    if (m_started <=  m_started + msPeriod)
+    if (msPeriod >= 0)
     {
         if ((now >= m_started + msPeriod) || (now < m_started))
         {


leptun:
I'm not sure this is correct. What `m_started <=  m_started + msPeriod` is checking is whether adding `msPeriod` to `m_started` causes an unsigned integer overflow (carry set). Your new simplified check doesn't check for overflow at all and will be always true anyways (an unsigned integer is indeed always >= 0), so the else branch is discarded.

leptun:
@gudnimg Might be worth looking into this: https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html
Maybe using these builtin functions results in some small savings. The current code checks the result of the addition against `m_started`, but only checking for overflow on addition should be needed. The result of the addition also doens't need to be stored anywhere, so some optimizations can be done also there.

gudnimg:
> @gudnimg Might be worth looking into this: https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html Maybe using these builtin functions results in some small savings. The current code checks the result of the addition against `m_started`, but only checking for overflow on addition should be needed. The result of the addition also doens't need to be stored anywhere, so some optimizations can be done also there.

I'll take a closer look. This function seems overly complicated for what it is supposed to be doing. We can probably handle the rollover event more efficiently.

gudnimg:
@leptun 

This saves 56 bytes:

I haven't used that builtin before, hopefully I'm not using in incorrectly. :) 

I'm not satisfied with the overflow handling in case `m_started > now` 🤔 Wonder if we can improve this somehow so we aren't creating false "expired" events.

```cpp
template<typename T>
bool Timer<T>::expired(T msPeriod)
{
    if (!m_isRunning) return false;
    bool expired = false;
    const T now = _millis();
    T elapsed; // elapsed duration
    bool overflowEvent = __builtin_sub_overflow(now, m_started, &elapsed);
    if (overflowEvent || int32_t(msPeriod - elapsed) < 0)
    {
        // Overflow can be triggered if m_started > now, in that case expire the timer
        expired = true;
    }
    if (expired) m_isRunning = false;
    return expired;
}
```

leptun:
I think this can be simplified further still. I need to check something. Will come back with some code

gudnimg:
Here is another attempt from me, only saves 4 bytes

```cpp
template<typename T>
bool Timer<T>::expired(T msPeriod)
{
    if (!m_isRunning) return false;
    bool expired = false;
    const T now = _millis();
    T elapsed; // elapsed duration
    if (m_started > now) { 
        // Unsigned integer overflow
        // Example: now = +10
        //          m_started = +247
        //          +10 + (0xFF - 247) = +10 + 8 = 18
        static constexpr T max_integer = sizeof(T) == 2 ? UINT16_MAX : UINT32_MAX;
        elapsed = now + (max_integer - m_started);
    } else {
        elapsed = now - m_started;
    }
    
    if (elapsed >= msPeriod)
    {
        expired = true;
    }
    if (expired) m_isRunning = false;
    return expired;
}
```

gudnimg:
This check in my previous attempt was probably not working correctly `int32_t(msPeriod - elapsed) < 0`

PR #4093, opened 2023-03-18
Restore previous Z-lift in M701/M702 + other PROGMEM optimisations

gudnimg:
Previously when the firmware called M701/M702 manually there was no Z lift. But after we added support for the Z parameter we set the default to 50mm.

Change strings `"M701"` to `"M701 Z0"` and `"M702"` to `"M702 Z0"` to restore the previous behavior from before 3.13.

Also pulled the gcodes into PROGMEM in message.cpp along with M83 and M84 to save memory. We can maybe save more memory with other gcodes.

Change in memory:
Flash: -124 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -10789,7 +10789,7 @@ void recover_print(uint8_t automatic) {
         //Extrude some filament to stabilize the pressure
         enquecommand_P(PSTR("G1 E5 F120"));
         // Retract to be consistent with a short pause
-        sprintf_P(cmd, PSTR("G1 E%-0.3f F2700"), default_retraction);
+        sprintf_P(cmd, G1_E_F2700, default_retraction);
         enquecommand(cmd);
     }
 
@@ -10946,7 +10946,7 @@ void restore_print_from_eeprom(bool mbl_was_active) {
     enquecommand(cmd);
 
   // Unretract.
-    sprintf_P(cmd, PSTR("G1 E%0.3f F2700"), default_retraction);
+    sprintf_P(cmd, G1_E_F2700, default_retraction);
     enquecommand(cmd);
   // Recover final E axis position and mode
     float pos_e = eeprom_read_float((float*)(EEPROM_UVLO_CURRENT_POSITION_E));
@@ -11157,7 +11157,7 @@ void stop_and_save_print_to_ram(float z_move, float e_move)
         // A snprintf would have been a safer call, but since it is not used
         // in the whole program, its implementation would bring more bytes to the total size
         // The behavior of dtostrf 8,3 should be roughly the same as %-0.3
-        sprintf_P(buf, PSTR("G1 E%-0.3f F2700"), e_move);
+        sprintf_P(buf, G1_E_F2700, e_move);
         enquecommand(buf, false);
     }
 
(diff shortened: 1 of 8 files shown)


gudnimg:
Ran a first layer cal as a sanity check just to confirm this is actually not breaking anything. Works well.

Firmware/Marlin_main.cpp
```diff
@@ -10946,7 +10946,7 @@ void restore_print_from_eeprom(bool mbl_was_active) {
     enquecommand(cmd);
 
   // Unretract.
-    sprintf_P(cmd, PSTR("G1 E%0.3f F2700"), default_retraction);
+    sprintf_P(cmd, G1_E_F2700, default_retraction);
```
gudnimg, line 10949: Note there is one change here. There was a missing `"-"` modifier. It just makes sure the E value is left adjusted (same as was already done in the two other places this same gcode is used).

3d-gussner (approved):

Firmware/Marlin_main.cpp
```diff
@@ -11150,15 +11149,15 @@ void stop_and_save_print_to_ram(float z_move, float e_move)
     {
         // First unretract (relative extrusion)
         if(!saved_extruder_relative_mode){
-            enquecommand(PSTR("M83"), true);
```
leptun, line 11153: hmmmmm. Wasn't this actually broken in the past? It looks to me as if the wrong enqueuecommand function type was used in the past, treating a progmem reference as a ram reference.

Firmware/messages.cpp
```diff
@@ -228,3 +228,17 @@ const char MSG_ADVANCE_K[] PROGMEM_N1 = "Advance K:"; ////c=13
 const char MSG_POWERPANIC_DETECTED[] PROGMEM_N1 = "POWER PANIC DETECTED"; ////c=20
 const char MSG_LCD_STATUS_CHANGED[] PROGMEM_N1 = "LCD status changed";
 const char MSG_UNKNOWN_CODE[] PROGMEM_N1 = "Unknown %c code: %s\n";
+
+// Common G-gcodes
+const char G1_E_F2700[] PROGMEM_N1 = "G1 E%-0.3f F2700";
+const char G28W0[] PROGMEM_N1 = "G28 W0";
```
leptun: Just wondering, what would happen if we ommit these "0" argument values? Can we save just a few more bytes from this? So instead of "G28 W0", we use "G28 W". Same for the M701 and M702 a few lines lower

Firmware/ultralcd.cpp
```diff
@@ -3804,7 +3804,7 @@ static void lcd_wizard_load() {
         loading_flag = true;
     }
     gcode_M701(FILAMENTCHANGE_FIRSTFEED, 0);
-    //enquecommand_P(PSTR("M701"));
+    //enquecommand_P(MSG_M701_NO_LIFT); // is enqueuecommand_P safe here?
```
leptun, line 3807: Why would it not be safe? In any case, it doesn't look like the cmdqueue is needed anymore since gcode_M701 is called directly just above.

leptun (approved):

Firmware/messages.cpp
3d-gussner: Every char less saves bytes :wink:

3d-gussner: I always use `G28 W` is the `0` read at all?

3d-gussner: It save additional 2 bytes

3d-gussner (approved):

PR #4094, opened 2023-03-19
Reduce code size in First Layer Calibration

gudnimg:
* Create a common function to enqueue an array of `PROGMEM` commands. The array is required to be in program memory.
* Added `preheat_cmd` into `PROGMEM`, this required adding all it's commands into `PROGMEM` as well, `PSTR()` was not accepted by the compiler.

Change in memory:
Flash: -60 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/first_lay_cal.cpp
+++ b/Firmware/first_lay_cal.cpp
@@ -47,12 +47,16 @@ static const char zero_extrusion[] PROGMEM = "G92 E0";
 //! @brief Wait for preheat
 void lay1cal_wait_preheat()
 {
-    const char * const preheat_cmd[] =
+    static const char preheat_cmd_2[] PROGMEM = "M190";
+    static const char preheat_cmd_3[] PROGMEM = "M109";
+    static const char preheat_cmd_4[] PROGMEM = "G28";
+
+    static const char * const preheat_cmd[] PROGMEM =
     {
         MSG_M107,
-        PSTR("M190"),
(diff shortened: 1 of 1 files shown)


Firmware/first_lay_cal.cpp
```diff
@@ -33,25 +33,34 @@ static constexpr float spacing(float layer_height, float extrusion_width, float
...
+}
+
 static const char extrude_fmt[] PROGMEM = "G1 X%d Y%d E%-.5f";
 static const char zero_extrusion[] PROGMEM = "G92 E0";
 
 //! @brief Wait for preheat
 void lay1cal_wait_preheat()
 {
-    const char * const preheat_cmd[] =
+    static const char preheat_cmd_2[] PROGMEM = "M190";
```
leptun, line 50: Does extracting these result in more optimal code?

leptun (approved):
The changes look good but I need to also test them on the printer (all busy) to confirm if everything still works.

Firmware/first_lay_cal.cpp
gudnimg: I didn't look at the assembly but extracting it was required. The compiler didn't like `PSTR("M190")` and complained about dynamic initialisation when I made `preheat_cmd[]` `static` and in `PROGMEM`.

I recall this did save some 30 or 40 bytes.

DRracer (approved):

PR #4098, opened 2023-03-20
PFW-1515 Fix an issue with Thermal Anomaly message

gudnimg:
Issue introduced with PFW-1504

Problem with `expired_cont()` and `expired()` is they will stop the timer once expired is true. For the purpose of just delaying the next progress message when rendering the MMU visualisation, this was not the intended behavior. We just want to wait for 4 seconds to pass. For now don't apply this delay/waiting when the custom message type is `Status`.

Change in memory:
Flash: +40 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -562,7 +562,9 @@ void lcdui_print_status_line(void) {
             scrollstuff = 0;
         }
     } else { // Otherwise check for other special events
-        if (!lcd_status_message_timeout.expired_cont(LCD_STATUS_DELAYED_TIMEOUT))
+        if ( custom_message_type != CustomMsg::Status
+        && lcd_status_message_timeout.running()
+        && lcd_status_message_timeout.elapsed() < LCD_STATUS_DELAYED_TIMEOUT)
         {
             return; // Nothing to do, waiting for delay to expire
         }


3d-gussner (approved):

Firmware/ultralcd.cpp
```diff
@@ -562,7 +562,9 @@ void lcdui_print_status_line(void) {
             scrollstuff = 0;
         }
     } else { // Otherwise check for other special events
-        if (!lcd_status_message_timeout.expired_cont(LCD_STATUS_DELAYED_TIMEOUT))
+        if ( custom_message_type != CustomMsg::Status
+        && lcd_status_message_timeout.running()
```
leptun, line 566: I don't understand why this change was necessary. Why not just add the check like this?
```
if (custom_message_type != CustomMsg::Status && !lcd_status_message_timeout.expired_cont(LCD_STATUS_DELAYED_TIMEOUT) {
    return;
}
```
This way, the timer expired_cont method gets called only if custom_message_type != CustomMsg::Status.
But maybe that was not the desired effect. Since I didn't quite understand how this code works, I can't say for sure.

wavexx:
Reading through commit bbfc4b9f7853525d36a74e09ffc14366bcd9edbc it seems that we touch lcd_status_message_timeout with the intention to avoid the display to be updated so that we can keep the previous changes for the filament status on-screen.

lcd_status_message_timeout is used to prevent the message from being _changed_ depending on priority, but it still needs to be displayed in this function. This is what is causing the screen to become unresponsive then the thermal timeout is happening. We can't use the timer like this, and we can't combine it for this purpose.

INFO needs to have higher priority, and can't be expired by something like the filament sensor.

We could introduce a new message priority (something like LCD_STATUS_NOCHANGE) inbetween INFO and NONE, and handle that with a different timeout. The priority needs to be handled in lcd_message_check (which currently only handles INFO, since other priorities do not expire) and use it to see if you can actually set/reset the message priority and then expire it properly.

wavexx:
I am referring to the progress drawn by TryLoadUnloadProgressbar(), just to clarify.

Firmware/ultralcd.cpp
gudnimg: I was worried this may interfere with other alert messages. If all those use Status, then I agree 👍

PR #4101, opened 2023-03-21
Reduce calls to `__divsf3` when calculating feedrate

gudnimg:
For non-time critical code it is more efficient to call a function rather than inlining each division operation.

Most of the time there is no call to `__divsf3` but this PR optimizes the few exceptions.

Change in memory:
Flash: -122 bytes
SRAM: 0 bytes


DRracer (approved):
nice :+1:

PR #4102, opened 2023-03-22
Fix HBed on Load for MMU Loading test

3d-gussner:
During testing I have found the LCD Settings -> Load Test ignores the Settings -> Hbed on load state. This fixes the issue


leptun:
NOK. The printer just sat on the "Testing filament x" screen while the nozzle was heating up. No temperatures were shown. But at least now it didn't heat the bed, so that part works correctly. The test did end up finishing correctly, so just the heatup screen seems to not be working.

leptun (approved):
Nevermind. It's ok. Issue I encountered is fixed by #4090

3d-gussner:
@leptun Thanks for the review I have been testing with the latest PRs for RC1

PR #4108, opened 2023-03-24
Ensure E-Cool off during MMU2 load/unload.

### LINKED ISSUE #3497: [BUG] Experimental feature "E-cool mode" has odd effects on extruder on MMU2S load sequence.
**Printer type** - MK3S+
**Printer firmware version** -  3.11.0

**MMU upgrade** - MMU2S

**SD card or USB/Octoprint**
  Octoprint

**Describe the bug**
  With E-cool mode enabled in the experimental features menu, the loading sequence causes the extruder to make a clicking sound.
  
**To Reproduce**
  With an MK3S+ MMU2S printer, enable the E-cool mode in the experimental features menu, then do a first-layer calibration on any filament, listen to the extruder during the load sequence.
  
**Expected behavior**
  No observable difference from E-cool mode being off.
  
**G-code**
  On-board first-layer calibration sufficient to reproduce issue.

**Video**
  [Video of issue on YouTube (apologies for the vertical video).](https://www.youtube.com/shorts/Sud9mxOqvwY)

ggppjj:
Temporarily toggles E-Cool off during MMU2 load/unload sequence to avoid issue with the many small steps used during those sequences. Meant to fix #3497.

### DIFF
--- a/Firmware/mmu2.cpp
+++ b/Firmware/mmu2.cpp
@@ -12,6 +12,12 @@
 #include "strlen_cx.h"
 #include "SpoolJoin.h"
 
+#ifdef TMC2130
+#include "tmc2130.h"
+#include "Prusa_Farm.h"
+#include "stepper.h"
+#endif //TMC2130
+
 #ifdef __AVR__
 // As of FW 3.12 we only support building the FW with only one extruder, all the multi-extruder infrastructure will be removed.
 // Saves at least 800B of code size
@@ -451,6 +457,7 @@ bool MMU2::unload() {
         return false;
 
     WaitForHotendTargetTempBeep();
+    bool previous_ecool_state = load_unload_ecool_toggle(false);
 
     {
         FSensorBlockRunout blockRunout;
@@ -472,6 +479,7 @@ bool MMU2::unload() {
         extruder = MMU2_NO_TOOL;
         tool_change_extruder = MMU2_NO_TOOL;
     }
+    (void)load_unload_ecool_toggle(previous_ecool_state);
     return true;
 }
 
@@ -900,10 +908,49 @@ void MMU2::execute_extruder_sequence(const E_Step *sequence, uint8_t steps) {
 
 void MMU2::execute_load_to_nozzle_sequence() {
     planner_synchronize();
+    bool previous_ecool_state = load_unload_ecool_toggle(false);
     // Compensate for configurable Extra Loading Distance
     planner_set_current_position_E(planner_get_current_position_E() - (logic.ExtraLoadDistance() - MMU2_FILAMENT_SENSOR_POSITION));
     execute_extruder_sequence(load_to_nozzle_sequence, sizeof(load_to_nozzle_sequence) / sizeof(load_to_nozzle_sequence[0]));
+    (void)load_unload_ecool_toggle(previous_ecool_state);
 }
+
+bool MMU2::load_unload_ecool_toggle(bool enable){
+    // Check to see if E-cool mode is enabled and if so disable it during load sequence due to
+    // incompatibility with the MMU2 load sequence's many small moves.
+    // Fix for #3497.
+    if (enable == false) {
+        #ifdef TMC2130
+            if (farm_mode 
+            || (( eeprom_read_byte((uint8_t *)EEPROM_ECOOL_ENABLE) == EEPROM_ECOOL_MAGIC_NUMBER ) 
+                && ( eeprom_read_byte((uint8_t *)EEPROM_EXPERIMENTAL_VISIBILITY) == 1 ))
+            ){
+                st_synchronize();
+                if (tmc2130_wait_standstill_xy(1000)) {}
+                cli();
+                tmc2130_init(TMCInitParams(false)); 
+                st_reset_timer();
+                sei();
+                // Return that E-cool had previously been enabled.
+                return true;
+            }
+            else { // Return that E-cool had previously been disabled.
+                return false;
+            }
+        }
+        else { // Re-enable e-cool after it had been disabled.
+            st_synchronize();
+            if (tmc2130_wait_standstill_xy(1000)) {}
+            cli();
+            tmc2130_init(TMCInitParams(true)); 
+            st_reset_timer();
+            sei();
+        #endif //TMC2130
+    }
+    // Return that E-cool had previously been disabled.
+    // Unused feedback currently.
+    return false;
+}    
 
 void MMU2::ReportError(ErrorCode ec, ErrorSource res) {
     // Due to a potential lossy error reporting layers linked to this hook
--- a/Firmware/mmu2.cpp
+++ b/Firmware/mmu2.cpp
@@ -14,7 +14,7 @@
 
 #ifdef TMC2130
 #include "tmc2130.h"
-#include "Prusa_Farm.h"
+#include "Prusa_farm.h"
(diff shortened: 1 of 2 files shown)


Firmware/mmu2.cpp
```diff
@@ -900,11 +908,50 @@ void MMU2::execute_extruder_sequence(const E_Step *sequence, uint8_t steps) {
...
+        }
+        else { // Re-enable e-cool after it had been disabled.
+            st_synchronize();
+            if (tmc2130_wait_standstill_xy(1000)) {}
+            cli();
+            tmc2130_init(TMCInitParams(true)); 
+            st_reset_timer();
+            sei();
+        #endif //TMC2130
+    }
```
gudnimg, line 949: Can this be reduced to this?: 
```cpp
#ifdef TMC2130
if (farm_mode
    || (( eeprom_read_byte((uint8_t *)EEPROM_ECOOL_ENABLE) == EEPROM_ECOOL_MAGIC_NUMBER ) 
    && ( eeprom_read_byte((uint8_t *)EEPROM_EXPERIMENTAL_VISIBILITY) == 1 )))
{
    st_synchronize();
    if (tmc2130_wait_standstill_xy(1000)) {}
    cli();
    tmc2130_init(TMCInitParams(enable)); 
    st_reset_timer();
    sei();
    // Return that E-cool had previously been enabled.
    return true;
} else { // Return that E-cool had previously been disabled.
    return false;
}
#endif //TMC2130
```
We need to reduce the flash memory usage as much as possible. The compiler may not be smart enough to notice duplicated code.

ggppjj: I'll modify that slightly to always return false if TMC2130 is not defined, but otherwise that would be great!

ggppjj: Haha, nevermind about that last comment, it was already set up for that.

ggppjj: OK, so I've implemented the following: 
```C++
bool MMU2::load_unload_ecool_toggle(bool enable){
    // Check to see if E-cool mode is enabled and if so disable it during load sequence due to
    // incompatibility with the MMU2 load sequence's many small moves.
    // Fix for #3497.
    #ifdef TMC2130
        if (farm_mode
            || (( eeprom_read_byte((uint8_t *)EEPROM_ECOOL_ENABLE) == EEPROM_ECOOL_MAGIC_NUMBER )
            && ( eeprom_read_byte((uint8_t *)EEPROM_EXPERIMENTAL_VISIBILITY) == 1 )))
        {
            st_synchronize();
            if (tmc2130_wait_standstill_xy(1000)) {}
            cli();
            tmc2130_init(TMCInitParams(enable));
            st_reset_timer();
            sei();
            // Return that E-cool had previously been enabled.
            return true;
        } else { // Return that E-cool had previously been disabled, or always false if not TMC2130.
    #endif //TMC2130
            return false;
    #ifdef TMC2130
        }
    #endif //TMC2130
}
```

I'm not certain how to deal with the unused parameter for non-TMC2130 platforms, though.

gudnimg: Also changing ```
if (farm_mode
    || (( eeprom_read_byte((uint8_t *)EEPROM_ECOOL_ENABLE) == EEPROM_ECOOL_MAGIC_NUMBER ) 
    && ( eeprom_read_byte((uint8_t *)EEPROM_EXPERIMENTAL_VISIBILITY) == 1 )))
``` to `if(FarmOrUserECool())` should save some memory too.

But we are trying to not include `ultralcd.h` into `mmu2.cpp` or `mmu2.h`.

I don't see any reason why `FarmOrUserECool()` should be in `ultralcd.cpp` 🤔 Maybe we can move it to `Marlin_main.cpp`

gudnimg: >I'm not certain how to deal with the unused parameter for non-TMC2130 platforms, though.

IMO the code should only be run on TMC2130 printers. the `#ifdef` /`#endif` should otherwise prevent the code from being included in the firmware image.

ggppjj: That was my initial thought, but it being in `ultralcd.cpp` did stop me. I'd ideally want to also move the E-cool enable/disable there also, as I think it's implemented in `ultralcd.cpp` also, then things might get smaller also without the duplicate toggling logic.

gudnimg: I agree, this shouldn't be in the MMU2 class since this is technically on the printer side and not the MMU. Even so it seems strange to me to have this in `ultralcd.cpp` since this function is not UI related. Maybe part of this can be in `tmc2130.cpp` but its not clear to me at the moment if that's a clean solution.

ggppjj: Do you know if there's a way to see if the currently running TMC2130 parameters are different from the setting in the EEPROM? I think I see a way forward that should be hopefully good if so.

leptun:
@gudnimg Wouldn't it be better if instead of disabling E-Cool while loading to the gears. we just make the move to load to gears continuous (really long) and interrupt it with planner_abort_hard() when the fsensor switches as we discussed some time ago? This way you don't fiddle with the E-Cool mode and you make the move less jittery and consistent even in the normal E-mode.

ggppjj:
Ooh, that does sound way better to me. Is planner_abort_hard() a new thing from after this revision of the MMU? For giggles, I'll start a new branch myself and poke around with another draft PR if all goes well initially.

gudnimg:
> @gudnimg Wouldn't it be better if instead of disabling E-Cool while loading to the gears. we just make the move to load to gears continuous (really long) and interrupt it with planner_abort_hard() when the fsensor switches as we discussed some time ago? This way you don't fiddle with the E-Cool mode and you make the move less jittery and consistent even in the normal E-mode.

I havent tried it yet. I’ll add it to my to-do list for tomorrow to try it out 👍

ggppjj:
Ah, it seems like the planner logic for the load sequence is handled in the MMU2 firmware, which makes sense. Well, this seems to be where this journey ends. I'll keep this a draft until I hear back whether you were successful in testing E-Cool with a larger E value during the load. 

Thanks for your time to both of you!

gudnimg:
This seems to work. The catch is we must reset `planner_aborted` to `false` without returning to the main `loop()`. Not sure yet if this is safe though. @wavexx has explained that it should be reset by returning to the top level main `loop()` but we can't do that when running MMU toolchange.
[abort_moves.patch](https://github.com/prusa3d/Prusa-Firmware/files/11068344/abort_moves.patch)

EDIT:

Better optimized patch, uses less memory:
[abort_moves_v2.patch](https://github.com/prusa3d/Prusa-Firmware/files/11068448/abort_moves_v2.patch) it uses 18 bytes of Flash memory.

gudnimg:
I can confirm that aborting the moves fixes the issue (day and night difference). I'll try running a multi-color print and see if everything works OK after a number of toolchanges.

gudnimg:
With my patch above

Before:

https://youtu.be/4BohuSCnrIU

After:

https://youtu.be/K3Io_GBRNfo

ggppjj:
That looks excellent! Hopefully the print test goes well but it seems like it should be exactly the solution needed here. I'll close this PR in favor of your solution, thanks for everyone's efforts on this MMU invigoration. I've been following the PRs and milestones closely, and I'm very much excited by what's coming down the pipeline!

PR #4070, opened 2023-03-08
planner: optimise acceleration limit checking

gudnimg:
Rename `acceleration_st` to `acceleration_steps_per_s2` to be same as Marlin 2

Store the acceleration into a local variable `accel` while performing the limit checks.

When limit checks are done we can assign the block it's acceleration. Especially
`block->acceleration_steps_per_s2` is now only written to once, instead of reading/writing to it often in the limit checks.

Change in memory:
Flash: -142 bytes
SRAM: 0 bytes

PR #4111, opened 2023-03-25
Fix clicking noise during Toolchange when E-cool mode is enabled

### LINKED ISSUE #3497: [BUG] Experimental feature "E-cool mode" has odd effects on extruder on MMU2S load sequence.

gudnimg:
Idea to fix https://github.com/prusa3d/Prusa-Firmware/issues/3497 by using the re-written MMU code on the MK3 side (3.13).

Kudos goes to @leptun for this idea, this is my attempt at implementing it.

Currently we are queuing many small 2 mm E-moves until the filament sensor triggers. We do this because we never abort any queued moves, so making small 2 mm moves ensures we do not move the E-motor more than 2 mm past the filament sensor.

Unfortunately, when E-cool mode is enabled, the E-motor will create audible clicking sounds (similar one hears during jam or a loose grub screw). A workaround for this is to queue one or more very long moves. Where _very long_ is something of the order of hundreds of millimeters. I have it set to 350 mm but it's just a random constant really. Keep in mind the firmware will block too large E-moves, if I recall correctly it was anything above > 450 mm (see `PREVENT_LENGTHY_EXTRUDE`)

In order to use very long moves, we must somehow stop the E-motor from moving once the filament sensor triggers. In other words, throw away what's left of the current E-motor move. For this simple purpose  we can use `planner_abort_hard()` but we must set `planner_aborted` to `false` afterwards because the code architecture does not allow the main `loop()` to run until the Toolchange command is done processing.

Video demonstration
=====

**BEFORE:** https://youtu.be/4BohuSCnrIU (Notice the clicking sound when the MMU changes state from Loading to Gears, to Loading to Fsensor)

**AFTER:** https://youtu.be/K3Io_GBRNfo (Clicking noise is gone, sound is smooth like normal) 

------

TO-DO list:
=====

- [x] Run a multi-color print with at least 100 toolchanges
- [x] Add more comments to the code, e.g. document function and our intention with the abort. It's only for E-moves.

-----

Change in memory:
Flash: +18 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/mmu2.cpp
+++ b/Firmware/mmu2.cpp
@@ -922,10 +922,12 @@ void MMU2::ReportError(ErrorCode ec, ErrorSource res) {
     switch (logic.Progress()) {
     case ProgressCode::UnloadingToFinda:
         unloadFilamentStarted = false;
+        planner_abort_queued_moves();  // Abort excess E-moves to be safe
         break;
     case ProgressCode::FeedingToFSensor:
         // FSENSOR error during load. Make sure E-motor stops moving.
         loadFilamentStarted = false;
+        planner_abort_queued_moves(); // Abort excess E-moves to be safe
         break;
     default:
         break;
@@ -1040,6 +1042,10 @@ void MMU2::OnMMUProgressMsgSame(ProgressCode pc) {
             case FilamentState::AT_FSENSOR:
                 // fsensor triggered, finish FeedingToExtruder state
                 loadFilamentStarted = false;
+
+                // Abort any excess E-move from the planner queue
+                planner_abort_queued_moves();
+
                 // After the MMU knows the FSENSOR is triggered it will:
                 // 1. Push the filament by additional 30mm (see fsensorToNozzle)
                 // 2. Disengage the idler and push another 2mm.
@@ -1048,7 +1054,7 @@ void MMU2::OnMMUProgressMsgSame(ProgressCode pc) {
             case FilamentState::NOT_PRESENT:
                 // fsensor not triggered, continue moving extruder
                 if (!planner_any_moves()) { // Only plan a move if there is no move ongoing
-                    MoveE(2.0f, MMU2_LOAD_TO_NOZZLE_FEED_RATE);
+                    MoveE(350.0f, MMU2_LOAD_TO_NOZZLE_FEED_RATE);
                 }
                 break;
             default:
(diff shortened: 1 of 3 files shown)


gudnimg:
Printed 5 Catan Seafarers ships. Its a handy test print which I use often since it doesnt take more than 5 hours. 

The results are good. 

On top of this PR I added two other PRs to test:

* https://github.com/prusa3d/Prusa-Firmware/pull/4070
* https://github.com/prusa3d/Prusa-Firmware/pull/3962

This print might not be the best to test the acceleration PR but I wanted to print something with it anyway. 

The boats look good 😃 E-cool mode was active the entire time. 

image

image

All 5 MMU slots were loaded with filament. But in this print only 3 colors are used: Blue, White and Yellow.

leptun:
Test print successful also here. 2 colors, 5 loaded. 133 toolchanges. No issues in the purge tower.

image

Firmware/mmu2.cpp
```diff
@@ -1048,7 +1054,7 @@ void MMU2::OnMMUProgressMsgSame(ProgressCode pc) {
             case FilamentState::NOT_PRESENT:
                 // fsensor not triggered, continue moving extruder
                 if (!planner_any_moves()) { // Only plan a move if there is no move ongoing
-                    MoveE(2.0f, MMU2_LOAD_TO_NOZZLE_FEED_RATE);
+                    MoveE(350.0f, MMU2_LOAD_TO_NOZZLE_FEED_RATE);
```
leptun, line 1062: Please add the note about the long extrusion prevention here.

leptun (approved):

Firmware/mmu2.cpp
gudnimg: Thanks I forgot about it

gudnimg: Done

PR #4112, opened 2023-03-25
Fix a few ambiguous overloaded calls

gudnimg:
Noticed this when exploring another optimization

By specifying exactly which overloaded function to use we save some memory. I suspect `uint8_t` was being promoted to `int16_t` or `uint16_t` (will need to look at the assembly to see what is happening).
Seems to have something to do with doing arithmetic in the function argument

Change in memory:
Flash: -156 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -583,7 +583,7 @@ void lcdui_print_status_line(void) {
                 lcd_space(LCD_WIDTH);
                 lcd_puts_at_P(0, 3, _T(MSG_CALIBRATE_Z_AUTO));
                 lcd_puts_P(PSTR(" : "));
-                lcd_print(custom_message_state - 10);
+                lcd_print(uint8_t(custom_message_state - 10));
             } else {
                 if (custom_message_state == 3) {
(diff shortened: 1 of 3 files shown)


Firmware/ultralcd.cpp
```diff
@@ -4823,7 +4823,7 @@ uint8_t choose_menu_P(const char *header, const char *item, const char *last_ite
         for (uint_least8_t i = 0; i < ordinary_items; i++)
         {
             lcd_set_cursor(2 + item_len, i+1);
-            lcd_print(first + i + 1);
```
gudnimg, line 4826: Note over 100B are saved here, I havent looked at the assembly but maybe I am overlooking something here.

gudnimg: Looking at the assembly, it looks like with this change the compiler or linker has decided to inline `lcd_print(unsigned long, int)` here.

DRracer: We should tag `lcd_print` as `__attribute__((noinline))` to prevent future surprises

leptun (approved):

DRracer (approved):

PR #4114, opened 2023-03-27
Don't duplicate silent mode menu

gudnimg:
Create a common function for the menu

This saves a lot of memory and 50 lines of code!

Tested on MK3S+

Change in memory:
Flash: -110 bytes
SRAM: 0 bytes


DRracer (approved):
nice :+1:

PR #4117, opened 2023-03-29
Fix fsensor runout conditions

leptun:
And also reorder autoload and runout conditions.
This should make it so that you can't get a filament runout event during the Wizzard or First Layer Cal.

### DIFF
--- a/Firmware/Filament_sensor.cpp
+++ b/Firmware/Filament_sensor.cpp
@@ -109,9 +109,9 @@ bool Filament_sensor::checkFilamentEvents() {
 void Filament_sensor::triggerFilamentInserted() {
     if (autoLoadEnabled
         && (eFilamentAction == FilamentAction::None)
-        && (! MMU2::mmu2.Enabled() ) // quick and dirty hack to prevent spurious runouts while the MMU is in charge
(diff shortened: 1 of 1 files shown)


Firmware/Filament_sensor.cpp
```diff
@@ -109,9 +109,9 @@ bool Filament_sensor::checkFilamentEvents() {
 void Filament_sensor::triggerFilamentInserted() {
     if (autoLoadEnabled
         && (eFilamentAction == FilamentAction::None)
-        && (! MMU2::mmu2.Enabled() ) // quick and dirty hack to prevent spurious runouts while the MMU is in charge
         && !(
-            moves_planned() != 0
+            MMU2::mmu2.Enabled() // quick and dirty hack to prevent spurious runouts while the MMU is in charge
+            || moves_planned() != 0
```
gudnimg, line 114: I like this. Should be more efficent to run one `!()` instead of two  `!() && !()` 👍

gudnimg (approved):
Looks good to me, but I haven't tested this on a printer.

DRracer (approved):

PR #4119, opened 2023-03-30
Postponed kill() message with softReset

leptun:
This is a prototype implementation. Relevant comments copied from slack:

---

@leptun
I have a proposal that might fix a lot of our issues regarding kill():
Current situation. When kill is called, we turn off the interrupts, disable all heaters, show an LCD message and we halt. This has the problem that not all MCU peripherals are reset, pins stay in their previous state (see the BEEPER issue with the thermal model), the message can't be multi-page since the timebase is disabled at that point (no interrupts).
Proposal: We display the kill message just like with the crash dump message. So when kill gets called, we store the kill message pointer somewhere (might be uninitialized ram or eeprom), we trigger a softReset() so that a watchdog reset gets executed and all MCU peripherals/registers get reset. On startup we then check if the firmware reset due to kill() and we show the kill message early during startup, with the timebase and interrupts enabled. This solves both the beeper issue and the multi-page error message issue. I also think this approach is safer in general since it is guaranteed that watchdog brings the printer into a safe state, unlike whatever we do now in kill()

Now the question remains if this new approach would ask the user to reset the printer. I'd say no. In this case, the user should only need to press the knob to confirm the message and then the firmware should finish booting up.

And so that the octoprint users don't complain about the kill messages preventing bootup, the kill message pending flag (whatever that will be) should be cleared before showing the message so that a subsequent printer reset doesn't show the message again on startup.

---

The kill message pointer is stored in eeprom in this prototype.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -1041,11 +1041,11 @@ static void fw_crash_init()
 
 static void fw_kill_init() {
     if (eeprom_read_byte((uint8_t*)EEPROM_KILL_PENDING_FLAG) == KILL_PENDING_FLAG) {
-        PGM_P kill_msg = (PGM_P)eeprom_read_word((uint16_t*)EEPROM_KILL_MESSAGE);
         // clear pending message event
         eeprom_write_byte((uint8_t*)EEPROM_KILL_PENDING_FLAG, EEPROM_EMPTY_VALUE);
 
         // display the kill message
+        PGM_P kill_msg = (PGM_P)eeprom_read_word((uint16_t*)EEPROM_KILL_MESSAGE);
         lcd_show_fullscreen_message_and_wait_P(kill_msg);
     }
 }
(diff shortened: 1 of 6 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -1591,6 +1602,9 @@ void setup()
     // report crash failures
     fw_crash_init();
 
+    // report kill() events
```
wavexx: I would move this much earlier during startup. Definitely before the wizard, so that a bug anywhere doesn't preclude the kill() message.

leptun: I moved it earlier. This is as early as I'm confortable. It must be after st_init since that enables interrupts (needed by the full screen message).

wavexx (approved):

leptun:
Note: some of the kill messages need to be changed since a reset is no longer necessary. So instead of something like "reset to continue", it needs to be "click to continue". Or just remove that part entirely since it's a regular full screen message now and it's implied that you click to continue.

3d-gussner (approved):
Tested on MK404 and I really like it.

PR #4122, opened 2023-04-02
Fix MMU2_CHECK_FILAMENT_PRESENCE_EXTRUSION_LENGTH for MMU2

leptun:
Fix flipped sign so that the filament is pushed far enough into the hotend when checking if the load was successful.
Tested with PETG on MK2.5 with MMU2. No filament was extruded by this check out of the nozzle. The change should have no effect on MMU2S since the filament sensor position is 0 in that config.


gudnimg (approved):
Looks good to me

PR #4123, opened 2023-04-02
Minor UI fixes for MMU status line during toolchange

gudnimg:
Minor MMU specific fixes pulled from https://github.com/prusa3d/Prusa-Firmware/pull/4080

- Fixes issue where toolchange visualization is cut-off in the middle due to the status message being set incorrectly by `ReportProgressHook()`. The function `EndReport()` takes care of changing `CustomMsg::MMUProgress` to `CustomMsg::Status`
- Fixes a minor issue where `MSG_WELCOME` is always written to the status line after MMU command is done. This should only be done when NOT printing.

Change in memory:
Flash: +2 bytes
SRAM:  0 bytes

### DIFF
--- a/Firmware/mmu2_reporting.cpp
+++ b/Firmware/mmu2_reporting.cpp
@@ -20,7 +20,9 @@ void BeginReport(CommandInProgress /*cip*/, uint16_t ec) {
 
 void EndReport(CommandInProgress /*cip*/, uint16_t /*ec*/) {
     // clear the status msg line - let the printed filename get visible again
-    lcd_setstatuspgm(MSG_WELCOME); // should be seen only when the printer is not printing a file
+    if (!printJobOngoing()) {
+        lcd_setstatuspgm(MSG_WELCOME);
+    }
     custom_message_type = CustomMsg::Status;
 }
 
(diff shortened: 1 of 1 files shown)


gudnimg:
Note for reviewers. It may be hard to see `MSG_WELCOME` on the screen because it goes away fast. But you should see “LCD status changed” in the logs. With this PR that log line should be gone when printing only.

3d-gussner (approved):
Tested on MK404 with MMU and it improves the look during the print.
Only minor comment would be that the last tool change operation message stays on LCD status line while the printer is printing.
Maybe it would be better to show the filename bit earlier.

leptun (approved):
I've tested it on a printer and it does work correctly, but sometimes a badly timed lcd_refresh will cause random characters from surrounding display positions to appear in the status line. There isn't really much we can do here unfortunately as that's just the unfortunate effect of a LCD reinitialization if you don't redraw the entire screen contents.

As for the Idler disengaged message, it now stays on the status screen longer than before, but the text eventually switches to the filename, so it's still fine. I guess #4080 should address this problem.

PR #4128, opened 2023-04-06
Fix factory reset menu overwritten by status screen

leptun:
Bug introduced in https://github.com/prusa3d/Prusa-Firmware/commit/0cac7612db26d8d2c027b86907a0a03d1a3db651 with `delay_keep_alive(0);`

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -803,6 +803,7 @@ void lcd_splash()
 
 void factory_reset() 
 {
+	LcdUpdateDisabler lcdUpdateDisabler;
 	KEEPALIVE_STATE(PAUSED_FOR_USER);
 	if (!READ(BTN_ENC))
 	{


gudnimg (approved):
LGTM.

I suspect this increases flash usage quite a bit? Either way it can be improved later.

leptun:
26B of flash extra. Sounds reasonable to me

leptun:
choose_menu_P and lcd_alright also seem to be affected by the same bug. I'm tempted to drop this change and just revert the delay_keep_alive(0); changes since those seem to add bugs everywhere since delay_keep_alive calls lcd_update internally, while the old code didn't.

gudnimg:
I wonder if that will break buzzer sound feedback. The code now expects `lcd_update()` to produce encoder sounds.

Maybe we can add a flag to signal an event that a fullscreen message or a blocking menu is running and so overwrite `lcd_update_enabled`. Or re-purpose `lcd_update_enabled` for this event.

leptun:
Hmmm. I didn't consider that. I'll test in the morning if clicks work. If they don't, then we can change things further, but otherwise I'll leave it like this.

gudnimg:
It’s nice to see you found this issue as I forgot to consider lcd updates (i.e. Status screen) in those functions (the issue you observed) but I think this principle we have now on the MK3 branch is a step in a better direction. The principle being In other words making LCD buttons ISR routine set a flag to indicate when to make feedback sound. 

I am actually worried button clicks wont work on some screens now when they don’t call lcd_update(). All the menus are covered in this case, its only blocking menus like M600 and factory reset etc.

I wonde if we create a “full screen” flag we can also solve the issue where the MMU error screen might be overwritten by a fullscreen. We could use to flag to dismiss the error screen temporarily…. I’m just thinking out loud at the moment :)

PR #4129, opened 2023-04-06
Revert delay_keep_alive(0) optimizations

leptun:
delay_keep_alive calls lcd_update internally, while the old code didn't. This results in blocking screens/menus being overwritten by the lcd update. This PR reverts recent optimizations that resulted in overwritten menus.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -2214,7 +2214,8 @@ uint8_t lcd_alright() {
     lcd_consume_click();
     while (1)
     {
-        delay_keep_alive(0);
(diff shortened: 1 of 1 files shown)


gudnimg:
I think this is a weekend task. We need some way to keep `lcd_update(0)` there to update `lcd_encoder` even when LCD updates are disabled. 

Maybe its time we implemented an LCD update function which is totally independent from `lcd_update_enabled` as that variable is really only used for the status screen. Knob itself (encoder) doesnt really care what the status screen does or if its disabled.

leptun:
Closing in favor of #4143

Issue #4126, opened 2023-04-06
[BUG]MK3 PAT9125 sensor not recognized


emard:
on MK3 filament sensor PAT9125 is not longer
recognized as filament presence source however
its hardware readings are ok, seen in
support->sensor info

happened somewhere between v7089 and v7122

leptun:
is it autoload, runout or jam detection that isn't working?

emard:
printer "thinks" as no filament present, it constantly
asks to replace filament.

in Support->Sensor info there is only line
PINDA: 0/1
this line was at v7089, I tried v7122 and v7174
now this line is missing:
Fil.Sensor: 0/1

Next lines seem ok
B: 80   Xd: 0
S: 17   Yd: 20102

when I remove or insert filament they change correctly

leptun:
Oke. I'll check that issue out soon. Fixing now other worse bugs that were introduced in the latest merging spree (which you might have also noticed).

emard:
Yes :) I noticed knob press clicks but doesn't enter SD directory,
some obvious bug, should be fixed soon or already it is

leptun:
Didn't know about that one. Adding it to the list now :)

- [x] LCD knob rotation
- [x] Factory reset menu overwritten
- [x] Settings menu corruption
- [x] PAT9125 thinks no filament is present
- [x] SD menu click issue

gudnimg:
> Yes :) I noticed knob press clicks but doesn't enter SD directory,
some obvious bug, should be fixed soon or already it is

It may be something is consuming the button event too soon.

Does the SD directory work if you rotate the knob? If so that indicates a problem with `lcd_draw_update`. Should always be set if either clicked or rotated.

emard:
When I click 2nd time on the SD directory it enters it,
but then when I try to print file.gcode, no matter how
much I click knob, it won't start printing.

PR #4130, opened 2023-04-07
SD menu do not consume click in _scrolling state

leptun:
The click is handled by the _standard state. If the click is consumed in _scrolling, then on click while scrolling, the only thing that will happen will be that the menu will switch to the _standard rendering, but have no click event on the file and it would just start scrolling again after a small delay.
This PR makes it so that a click in the _scrolling state isn't consumed and is handled on the next lcd update event in the _standard state.

Related comment: https://github.com/prusa3d/Prusa-Firmware/issues/4126#issuecomment-1499672012


gudnimg (approved):
Looks good to me 👍

PR #4131, opened 2023-04-07
Fix toggle menu items rendering

leptun:
The cursor was never set for the current item, so the toggle menu items would be rendered on the wrong lines.


gudnimg (approved):
Looks correct to me.

PR #4132, opened 2023-04-07
Fix filament autoloading

### LINKED ISSUE #4126: [BUG]MK3 PAT9125 sensor not recognized

leptun:
Closes #4126. The autoload action is a submenu. Correctly switch to that submenu.


gudnimg (approved):
Looks good to me.

PR #4133, opened 2023-04-09
Reduce XYZ cal code size

gudnimg:
Notes
* `_X_` is equal to `count_position[X_AXIS]`
* `X_AXIS_MASK` is equal to `_BV(X_AXIS)` i.e. `(1 << X_AXIS)`

Change in memory:
Flash: -46 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/xyzcal.cpp
+++ b/Firmware/xyzcal.cpp
@@ -415,22 +415,12 @@ void print_hysteresis(int16_t min_z, int16_t max_z, int16_t step){
 	}
 }
 
-void update_position_1_step(uint8_t axis, uint8_t dir){
-	if (axis & X_AXIS_MASK)
-		_X_ += dir & X_AXIS_MASK ? -1 : 1;
-	if (axis & Y_AXIS_MASK)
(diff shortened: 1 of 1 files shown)


leptun (commented):
I feel like this will not work. The old code handles multiple axis bits being set at the same time. The new code only works with one axis at a time. I think the code can still be optimized, but it required using a for loop and checking all axis bits for changes, just like before.
But in any case, I'll run a test on MK3S and check if it works or not.

gudnimg:
> The old code handles multiple axis bits being set at the same time.

That's a good point.

leptun:
NOK. Nozzle crashed hard during the zig-zag move.

gudnimg:
@leptun Thanks for confirming. I'll re-open if I see any better way.

leptun:
Please see the code I just pushed. This seems to work well on the printer and I think it's also a bit more optimal than what was initially there. Less optimal than your code, but at least this seems to work

gudnimg:
Nice we're still saving 46 bytes then 👍

leptun (approved):

PR #4134, opened 2023-04-09
tmc2130: simplify parsing axis mask

gudnimg:
There is no need to check for every axis bit with each for-loop iteration since we know only one of the bits is set. Instead check for one axis bit at a time.

The mask computation was also optimized in the iterations over all axes.

Change in memory:
Flash: -34 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/tmc2130.cpp
+++ b/Firmware/tmc2130.cpp
@@ -272,8 +272,7 @@ void tmc2130_home_enter(uint8_t axes_mask)
 		tmc2130_wait_standstill_xy(1000);
 	for (uint8_t axis = X_AXIS; axis <= Z_AXIS; axis++) //X Y and Z axes
 	{
-		uint8_t mask = (X_AXIS_MASK << axis);
-		if (axes_mask & mask)
+		if (const uint8_t mask = axes_mask & _BV(axis); mask)
 		{
 			tmc2130_sg_homing_axes_mask |= mask;
 			//Configuration to spreadCycle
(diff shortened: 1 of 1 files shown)


Firmware/tmc2130.cpp
```diff
@@ -272,17 +272,15 @@ void tmc2130_home_enter(uint8_t axes_mask)
...
-		if (axes_mask & mask)
+		if (const uint8_t mask = axes_mask & _BV(axis); mask)
 		{
 			tmc2130_sg_homing_axes_mask |= mask;
 			//Configuration to spreadCycle
 			tmc2130_wr(axis, TMC2130_REG_GCONF, TMC2130_GCONF_NORMAL);
 			tmc2130_wr(axis, TMC2130_REG_COOLCONF, (((uint32_t)tmc2130_sg_thr_home[axis]) << 16));
 			tmc2130_wr(axis, TMC2130_REG_TCOOLTHRS, __tcoolthrs(axis));
 			tmc2130_setup_chopper(axis, tmc2130_mres[axis], tmc2130_current_h[axis], tmc2130_current_r_home[axis]);
-			if (mask & (X_AXIS_MASK | Y_AXIS_MASK | Z_AXIS_MASK))
```
gudnimg, line 284: **Note:** I think this if statement is always true since the mask will always include one of the axis bits.

Firmware/tmc2130.cpp
```diff
@@ -272,17 +272,15 @@ void tmc2130_home_enter(uint8_t axes_mask)
 		tmc2130_wait_standstill_xy(1000);
 	for (uint8_t axis = X_AXIS; axis <= Z_AXIS; axis++) //X Y and Z axes
 	{
-		uint8_t mask = (X_AXIS_MASK << axis);
```
gudnimg, line 275: **Note:** `(X_AXIS_MASK << axis)` is equal to `(1 << axis)` which is equal to `_BV(axis)`

leptun (approved):
Tested on MK3S. All OK:
- Homing
- Belt test
- Z calibration
- Crash detection
- Z crash detection during Z homing

PR #4136, opened 2023-04-09
Enable Analog IR sensor again

leptun:
This is still WIP. I'm making a PR so that I don't forget about it / lose the changes.
The IR analog changes were lost at some point during rebasing. This PR brings the support back.

### DIFF
--- a/Firmware/temperature.cpp
+++ b/Firmware/temperature.cpp
@@ -2093,9 +2093,9 @@ void adc_callback()
 #ifdef VOLT_BED_PIN
     current_voltage_raw_bed = adc_values[ADC_PIN_IDX(VOLT_BED_PIN)]; // 6->9
 #endif
-#ifdef IR_SENSOR_ANALOG
-    current_voltage_raw_IR = adc_values[ADC_PIN_IDX(VOLT_IR_PIN)];
-#endif //IR_SENSOR_ANALOG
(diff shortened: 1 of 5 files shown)


Firmware/temperature.cpp
```diff
@@ -2093,9 +2093,9 @@ void adc_callback()
 #ifdef VOLT_BED_PIN
     current_voltage_raw_bed = adc_values[ADC_PIN_IDX(VOLT_BED_PIN)]; // 6->9
 #endif
-#ifdef IR_SENSOR_ANALOG
-    current_voltage_raw_IR = adc_values[ADC_PIN_IDX(VOLT_IR_PIN)];
-#endif //IR_SENSOR_ANALOG
+#if defined(FILAMENT_SENSOR) && (FILAMENT_SENSOR_TYPE == FSENSOR_IR_ANALOG)
+    fsensor.voltUpdate(adc_values[ADC_PIN_IDX(VOLT_IR_PIN)]);
```
gudnimg, line 2097: Just spotted `current_voltage_raw_IR` on MK3 branch. Completely unused variable :) 👍

gudnimg (approved):
Changes seem to work OK on my MK3S+ with MMU

IR sensor voltage is at 4.4V when no filament is present, and goes down to 0.6V-0.8V when I trigger the filament sensor. I can slightly see values between 4.4V and 0.6V when I trigger the sensor slowly but the transition is very quick.

I didn't test with the non-MMU filament sensor setup.

PR #4137, opened 2023-04-10
Optimization menu click return

leptun:
Do not return early in menus if items are clicked. Further optimizations are possible.
This PR also fixes some bugs that were introduced that are similar to @gudnimg's commit in this PR. The old code would not handle those optimizations properly and result in multiple items getting clicked under certain conditions.

flash: -1172B
ram: 0B

### DIFF
--- a/Firmware/menu.cpp
+++ b/Firmware/menu.cpp
@@ -140,6 +140,7 @@ void menu_submenu_no_reset(menu_func_t submenu, const bool feedback)
 uint8_t menu_item_ret(void)
 {
 	lcd_draw_update = 2;
+	menu_item++;
 	return 1;
 }
 
--- a/Firmware/menu.cpp
+++ b/Firmware/menu.cpp
@@ -77,6 +77,11 @@ void menu_start(void)
 
 void menu_end(void)
 {
+	if (menu_row >= LCD_HEIGHT)
+	{
+		// Early abort if the menu was clicked. The current menu might have changed because of the click event
+		return;
+	}
 	if (lcd_encoder >= menu_item)
 	{
 		lcd_encoder = menu_item - 1;
@@ -141,6 +146,11 @@ uint8_t menu_item_ret(void)
 {
 	lcd_draw_update = 2;
 	menu_item++;
+	
+	//prevent the rest of the menu items from rendering or getting clicked
+	menu_row = LCD_HEIGHT; // early exit from the MENU_BEGIN() for loop at the end of the current cycle
+	menu_line = 0; // prevent subsequent menu items from rendering at all in the current MENU_BEGIN() for loop cycle
+	menu_clicked = 0; // prevent subsequent items from being able to be clicked in case the current menu or position was changed by the clicked menu item
 	return 1;
 }
 
--- a/Firmware/menu.cpp
+++ b/Firmware/menu.cpp
@@ -142,7 +142,7 @@ void menu_submenu_no_reset(menu_func_t submenu, const bool feedback)
 	}
 }
 
-uint8_t menu_item_ret(void)
+void menu_item_ret(void)
 {
 	lcd_draw_update = 2;
 	menu_item++;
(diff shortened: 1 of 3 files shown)


Firmware/menu.cpp
```diff
@@ -140,6 +145,12 @@ void menu_submenu_no_reset(menu_func_t submenu, const bool feedback)
 uint8_t menu_item_ret(void)
 {
 	lcd_draw_update = 2;
+	menu_item++;
+	
+	//prevent the rest of the menu items from rendering or getting clicked
+	menu_row = LCD_HEIGHT; // early exit from the MENU_BEGIN() for loop at the end of the current cycle
+	menu_line = 0; // prevent subsequent menu items from rendering at all in the current MENU_BEGIN() for loop cycle
+	menu_clicked = 0; // prevent subsequent items from being able to be clicked in case the current menu or position was changed by the clicked menu item
```
gudnimg, line 153: I don’t think we need to reset `menu_clicked`. It should be reset automatically on next `menu_begin` … `menu_end` iteration.

Firmware/menu.cpp
```diff
@@ -77,6 +77,11 @@ void menu_start(void)
 
 void menu_end(void)
 {
+	if (menu_row >= LCD_HEIGHT)
```
gudnimg, line 80: Shouldn’t this be `if(menu_clicked)` ?

Firmware/menu.cpp
leptun: I had to do this because of the following scenario:
- You click a back item in a menu
- That action changes the lcd_encoder to the parent menu value from the menu stack let's say that lcd_encoder gets set to 3 by this
- The current menu MENU_BEGIN for loop iteration still continues though, so you end up at some point to the menu_item==3==lcd_encoder, which will be clicked as well.

So to avoid a second click action in the wrong menu item if you click on an item that executes menu_goto and changes lcd_encoder, we clear the click event in menu_item_ret().

Firmware/menu.cpp
leptun: It can't be menu_clicked because it is cleared in menu_item_ret(). I had to add this workaround since the menu_end() uses lcd_encoder, but lcd_encoder is invalid in the for loop cycle after a click event, so it would result in broken behaviour if I didn't do that extra check. I hope this made sense. The entire menu code is really confusing and I'd love to improve this MENU_BEGIN/END at some point since I see a lot of unnecessary complexity that ends up in needing these workarounds.

Firmware/menu.cpp
gudnimg: Makes sense. Will think a bit more about this.

I would really love to see the menu code be decoupled from `lcd_encoder` in the future. `lcd_encoder` should, in my mind, only be used to track changes in the encoder position. In that way the menu code would have its own separate encoder variable in the menu stack. But this is quite a bit outside the scope of this PR.

gudnimg:
Gave this PR a short test run on my machine yesterday. so far so good. Only issues I found thus far are not related to this PR 👍

I will try to continue to test this PR over the week. Meanwhile I’m also testing the menu changes from the last merge spree 😅

leptun:
This PR fixes some bugs with the menus as well. Some of the optimizations that were merged last week don't work properly without this PR.

gudnimg:
> This PR fixes some bugs with the menus as well. Some of the optimizations that were merged last week don't work properly without this PR.

Ok good to know. I’ll do my testing on this branch.

gudnimg (approved):
Approving this PR 👍 

I noticed a very reproducible case where the menus were double clicking. By toggling the TMC2130 Normal/Silent mode in menus, it would always open the HWsetup menu. But with this PR it is fixed.

Started a short print and kept the Support menu always on, also opened all the menus I could see and it caused no issues.

I think this PR is good as it is, we can fix issues in separate PRs

leptun:
That double feedback was caused by your optimization of that code where menu items were extracted to functions since the early return didn't work anymore in that case. That is no longer the case with this PR, hence why the bug was fixed. What was happening is that multiple menu items would get clicked. That's why you heard multiple click sounds and why you ended up in the wrong menu.

Thanks for the review. I'll merge this now so that we have functional Settings and Tune menus in the MK3 branch.

PR #4138, opened 2023-04-11
Fix broken editing menu

gudnimg:
* `uint8_t` values must be rendered with base 10 (`DEC`) and not 0.
* Fix an issue where editing menus were not updating and seemed completely broken. Issue was introduced by https://github.com/prusa3d/Prusa-Firmware/pull/4047/commits/764fcec19b53499fe23970c9664ff832f162a06b


Firmware/ultralcd.cpp
```diff
@@ -4040,13 +4040,11 @@ void lcd_settings_linearity_correction_menu(void)
     );
 	MENU_ITEM_BACK_P(_T(MSG_SETTINGS));
 #ifdef TMC2130_LINEARITY_CORRECTION_XYZ
-	//tmc2130_wave_fac[X_AXIS]
-
-	MENU_ITEM_EDIT_int3_P(_i("X-correct:"),  &tmc2130_wave_fac[X_AXIS],  TMC2130_WAVE_FAC1000_MIN-TMC2130_WAVE_FAC1000_STP, TMC2130_WAVE_FAC1000_MAX);////MSG_X_CORRECTION c=13
-	MENU_ITEM_EDIT_int3_P(_i("Y-correct:"),  &tmc2130_wave_fac[Y_AXIS],  TMC2130_WAVE_FAC1000_MIN-TMC2130_WAVE_FAC1000_STP, TMC2130_WAVE_FAC1000_MAX);////MSG_Y_CORRECTION c=13
-	MENU_ITEM_EDIT_int3_P(_i("Z-correct:"),  &tmc2130_wave_fac[Z_AXIS],  TMC2130_WAVE_FAC1000_MIN-TMC2130_WAVE_FAC1000_STP, TMC2130_WAVE_FAC1000_MAX);////MSG_Z_CORRECTION c=13
+	MENU_ITEM_EDIT_int3_P(_i("X-correct"), &tmc2130_wave_fac[X_AXIS], TMC2130_WAVE_FAC1000_MIN-TMC2130_WAVE_FAC1000_STP, TMC2130_WAVE_FAC1000_MAX);////MSG_X_CORRECTION c=13
```
leptun, line 4043: Please check if the translation files need updating after modifying the string here.

gudnimg: You're right, I'll add another commit with that change

gudnimg:
@3d-gussner I suspect this fix https://github.com/prusa3d/Prusa-Firmware/pull/4138/commits/334128a48102d83a7f1c8014813b0162dd421018 explains the reason why https://github.com/prusa3d/Prusa-Firmware/pull/4112 was not working. Because for `int8_t (char)` and `uint8_t (unsigned char)` the code uses `base = 0` but for other types it uses `base = 10` or `DEC` for decimal. 

Its my understanding the base 0 should be used for ASCII. But in that case I think we should instead use `lcd_putc` for letters and only use `lcd_print` for numbers. Then there is no need to use this trick `int + ‘0’`

leptun:
> I think we should instead use lcd_putc for letters and only use lcd_print for numbers

That sounds like a good idea. It might even be a bit more efficient.

gudnimg:

leptun (approved):
Tested last evening on MK3S. Most of the edit menus work, but the linearity correction menu has some odd behaviour since the value of 29 is considered as "off". We need to look further into those menu items in a followup PR.
Besides that, I think this PR is ready for merging.

3d-gussner (approved):
Tested on MK404

only minor change would be to update the MSG_?_CORRECTION c=13 to c=15

PR #4140, opened 2023-04-12
:construction: Introduce Unload from FSensor state

DRracer:
and sync progress_codes.h with the MMU repo / PR 271

Ideally,  the ramming sequence should be shortened and an extra 40mm unload move at 50mm/s should be planned on top of it while the MMU is already doing something. That should prevent extra thick strings from being held by the Bondtech gears and smoother and faster operation as well.


3d-gussner:
@DRracer Any update on this one?

ggppjj:
Apologies if this is unrelated, but would this change the behavior of auto-reload attempts if the filament load has an error that would trigger it? 

I believe at the moment that the code currently triggers ramming on auto-reload, and that currently this leaves blobs on prints that aren't particularly recoverable that grow larger for each attempt. It sounds like this change could affect that behavior positively. 

For reference, there is an open bug here for what I've described. 

https://github.com/prusa3d/Prusa-Firmware/issues/4655

3d-gussner:
@DRracer I am closing this for the moment as https://github.com/prusa3d/Prusa-Firmware-MMU/pull/271 has been closed and no update.

DRracer:
ok, I'll revive it later if necessary

PR #4143, opened 2023-04-14
Fix lcd_update() FW crashes

gudnimg:
My proposal to fix issue in https://github.com/prusa3d/Prusa-Firmware/pull/4129.

The main change is `lcd_encoder` is no longer updated by calling `lcd_update()`. This way checking for knob rotation is now decoupled from `lcd_update()` and can always be called in `manage_inactivity()` as the encoder flags are updated through the temperature ISR. We may even remove a few `lcd_update()` calls now with this change but they need to be evaluated one by one.

By using `manage_inactivity()` the encoder rotation event can be checked in both the main loop and during blocking operations such as factory reset menu or "choose filament" menu (MMU only).

**Other changes:**
* Fixes an issue where encoder did not work in manual fan check menu. It was missing a call to `manage_inactivity()`. Fix was tested on MK3S+.

----

Tests on MK3S+ with MMU
* ✅ First layer calibration
* ✅ MMU error screen (2 button choices). 
    * ✔️ The M600-like sound prompts also work correctly.
    * ✔️ MORE button works and and cycling through the details error message works and upon cycling through the message the error screen is shown correctly again.
* ✅ LCD menu item: Load to Nozzle
* ✅ M600 when not printing
* ✅ M600 when printing
* ✅ Pause, Resume and Stop print
* ✅ Z Calibration
* ✅ Selftest
* ✅ XYZ Calibration
* ✅ Belt test
----

Change in memory:
Flash: +16 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -9437,6 +9437,9 @@ void manage_inactivity(bool ignore_stepper_queue/*=false*/) //default argument s
   check_axes_activity();
   MMU2::mmu2.mmu_loop();
 
+  lcd_knob_update();
+  backlight_update();
+
   // handle longpress
   if(lcd_longpress_trigger)
   {
--- a/Firmware/lcd.cpp
+++ b/Firmware/lcd.cpp
@@ -690,11 +690,7 @@ void lcd_quick_feedback(void)
   lcd_beeper_quick_feedback();
 }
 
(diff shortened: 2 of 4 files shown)


gudnimg:
Marking this as ready for review. I can no longer find any crashes.

3d-gussner (approved):

PR #4150, opened 2023-04-15
Fix a typo in `PSU_Delta` define

gudnimg:
Symbol names are case sensitive: "PSU_delta" --> "PSU_Delta"

I'm not sure what the consequences of this is. It's been like this for a long time.


leptun:
I'd wait with this. Maybe we just delete PSU_Delta entirely after 3.13.0 releases if my new Z standstill stealthchop code doesn't regress. I don't want to change the PSU_Delta code if we aren't going to test it in case we revert in the future and encounter bugs because of this.

3d-gussner (approved):

PR #4153, opened 2023-04-16
MMU: Remove unused LoadingToNozzleRAII

gudnimg:
The variable is not used anywhere. I don't see it either in the buddy firmware repo except in `mmu2mk404.cpp` but that code seems outdated.

Change in memory:
Flash: -14 bytes
SRAM: -1 byte


DRracer (approved):
:thinking: it is not used in Buddy FW anymore as well ... Originally it was intended to distinguish between the different loading distances in ToolChange and LoadToNozzle.

Approved :+1:

PR #4154, opened 2023-04-16
MMU: Always reset `custom_message_type` when command is done processing

gudnimg:
* Fix issue where `Tx` gcode wasn't using `BeginReport()`/`EndReport()`. This resulted in the SD filename not being shown on the status line when printing -- only affects single material MMU gcodes. Issue introduced with https://github.com/prusa3d/Prusa-Firmware/pull/4123
* Don't set `custom_message_type` in `ReportProgressHook()`. We need to use `BeginReport()`/`EndReport()` instead to ensure `custom_message_type` is reset to `CustomMsg::Status` once the MMU command is done processing. This will for example allow showing SD file name again when printing.

Changes tested on MK3S+

Change in memory:
Flash: -8 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/mmu2_reporting.cpp
+++ b/Firmware/mmu2_reporting.cpp
@@ -270,7 +270,6 @@ void ReportErrorHook(CommandInProgress /*cip*/, uint16_t ec, uint8_t /*es*/) {
 
 void ReportProgressHook(CommandInProgress cip, uint16_t ec) {
     if (cip != CommandInProgress::NoCommand) {
-        custom_message_type = CustomMsg::MMUProgress;
         lcd_setstatuspgm( _T(ProgressCodeToText(ec)) );
     }
 }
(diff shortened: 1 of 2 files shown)


3d-gussner (approved):
Tested on MK404

Firmware/mmu2_reporting.cpp
```diff
@@ -270,7 +270,6 @@ void ReportErrorHook(CommandInProgress /*cip*/, uint16_t ec, uint8_t /*es*/) {
 
 void ReportProgressHook(CommandInProgress cip, uint16_t ec) {
     if (cip != CommandInProgress::NoCommand) {
-        custom_message_type = CustomMsg::MMUProgress;
```
DRracer, line 273: Are you sure that removal of this line doesn't break the try-load error reporting? I remember we had a hard time getting that piece of code running reliably.

gudnimg: We use `ReportingRAII rep(CommandInProgress::ToolChange);` everywhere now, this (and only this) sets the custom message type. It shouldnt be reset unless the Toolchange command is done and we return to the main marlin loop.

But I havent tested the try-load error specifically. Do I understand correctly you mean the scenario when the Printer reports an error and not the MMU?

gudnimg:
Closing in favor of https://github.com/prusa3d/Prusa-Firmware/pull/4080/commits/25c7ca4a0efc5b54df4169bce83017c9b5e20dee

PR #4156, opened 2023-04-21
Fix `M0` click not consumed

leptun:
Fix the need to click the knob twice to dismiss `M0`. Also some code optimization.

Flash: -90B
SRAM: -2B


gudnimg:
Note if we merge, we can probably close https://github.com/prusa3d/Prusa-Firmware/pull/4077 :)

gudnimg (approved):
Using the changes in this PR fixes the issue (which is easily reproduced). Now only one knob click is needed :) And long-presses are ignored which is good.

PR #4157, opened 2023-04-21
Menu optimizations

leptun:
Various menu optimizations

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -4054,27 +4054,17 @@ static void lcd_fsensor_settings_menu() {
 
 #endif //FILAMENT_SENSOR
 
-static void settingsSpoolJoin()
+static void menuitems_MMU_settings_common()
 {
     MENU_ITEM_TOGGLE_P(MSG_SPOOL_JOIN, SpoolJoin::spooljoin.isSpoolJoinEnabled() ? _T(MSG_ON) : _T(MSG_OFF), SpoolJoin::spooljoin.toggleSpoolJoin);
-}
-
-#define SETTINGS_SPOOLJOIN \
-do\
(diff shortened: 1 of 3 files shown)


Firmware/ultralcd.cpp
```diff
@@ -3456,8 +3441,7 @@ static void crash_mode_switch()
     {
         lcd_crash_detect_enable();
     }
-	if (printJobOngoing() || (lcd_commands_type == LcdCommands::Layer1Cal)) menu_goto(lcd_tune_menu, 9, true, true);
-	else menu_goto(lcd_settings_menu, 9, true, true);
+    menu_back();
```
gudnimg: The call to `menu_back()` should be removed. There is no need for a menu change here :)

gudnimg (changes requested):

Firmware/ultralcd.cpp
```diff
@@ -3423,10 +3411,7 @@ static void lcd_silent_mode_set() {
   // Wait until the planner queue is drained and the stepper routine achieves
   // an idle state.
   st_synchronize();
-  if (tmc2130_wait_standstill_xy(1000)) {}
```
gudnimg, line 3396: Proposal for change:

Guard the `lcd_display_message_fullscreen_P` call by `blocks_queued()`. If there are no planned moves, there is no need to render this message, it creates a weird LCD draw noticable by the user.

This does add about 10B of flash but I think its worth it for slightly cleaner UI. I'm not sure if `tmc2130_wait_standstill_xy` should be guarded as well, you're more familiar with it than me.

image

gudnimg (changes requested):

gudnimg (approved):

PR #4158, opened 2023-04-21
Cleanup: remove unused extruder macros

gudnimg:
The following macros are removed.
* `enable_e1()`
* `enable_e2()`
* `disable_e1()`
* `disable_e2()`

Also removed some code guarded by `EXTRUDER > 1`.

The firmware doesn't support multiple extruders so I propose these be removed as cleanup.

No change in memory

### DIFF
--- a/Firmware/Configuration_adv.h
+++ b/Firmware/Configuration_adv.h
@@ -453,14 +453,6 @@ const unsigned int dropsegments=5; //everything with less than this number of st
   #define THERMISTORHEATER_0 TEMP_SENSOR_0
   #define HEATER_0_USES_THERMISTOR
 #endif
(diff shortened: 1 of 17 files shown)


leptun:
@gudnimg I suggest that you also remove TEMP_SENSOR_1 and TEMP_SENSOR_2 code

gudnimg:
I think that's most of it now.

gudnimg:
Added one minor commit. I don't expect more commits to this PR since it is quite large at this point 😅

leptun:
👍 I'll review the PR tomorrow and merge it if everything is ok.

leptun (approved):

PR #4160, opened 2023-04-22
Rewrite the Tcodes description

gudnimg:
The old description had typos and was a bit confusing for new users

The new text is a suggestion from David Buzek at Prusa Research

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -8628,13 +8628,13 @@ Sigma_Exit:
   // end if(code_seen('M')) (end of M codes)
   /*!
   -----------------------------------------------------------------------------------------
-  # T Codes
-  T<extruder nr.> - select extruder in case of multi extruder printer. select filament in case of MMU_V2.
-  #### For MMU_V2:
-  T<n> Gcode to extrude at least 38.10 mm at feedrate 19.02 mm/s must follow immediately to load to extruder wheels.
-  @n T? Gcode to extrude shouldn't have to follow, load to extruder wheels is done automatically
-  @n Tx Same as T?, except nozzle doesn't have to be preheated. Tc must be placed after extruder nozzle is preheated to finish filament load.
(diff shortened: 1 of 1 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -8628,13 +8628,13 @@ SERIAL_PROTOCOLPGM("\n\n");
...
-  @n T? Gcode to extrude shouldn't have to follow, load to extruder wheels is done automatically
-  @n Tx Same as T?, except nozzle doesn't have to be preheated. Tc must be placed after extruder nozzle is preheated to finish filament load.
-  @n Tc Load to nozzle after filament was prepared by Tc and extruder nozzle is already heated.
+  T<extruder nr.> - select extruder in case of multi extruder printer. Selects filament position 1-5 (T0-T4) in case of MMU.
+
+  For MMU2/S / MMU3:
+    T<extruder nr.> - Selects the filament position. A Gcode to load a filament to the nozzle must follow.
+    Tx - Printer asks user to select a filament position. Then loads the filament from the MMU unit into the extruder wheels only. G-code to heat up the nozzle follows.
+    Tc - Loads the filament tip from the extruder wheels into the nozzle.
+    T? - acts the same as Tx
```
leptun: Doesn't `T?` act the same way as `Tx` followed by `Tc`?

gudnimg: I think you're right. The Tcodes code isn't super obvious to me. I'll double check.

gudnimg: It looks like that's only the case when the MMU is disabled and used for single material.

gudnimg: Hold up.... the MMU2 code is calling load to nozzle. 🤔 image

I suspect we can drop this if-else then:
image

gudnimg: @leptun fixed. I'll address the potential optimisation later in a separate PR. This PR is only meant as "cleanup"

leptun (approved):

PR #4161, opened 2023-04-22
MMU: Simplify `T?` processing

gudnimg:
We can use `bool MMU2::tool_change(char code, uint8_t slot)` to process `T?`. This way we don't need to call `load_filament_to_nozzle()` in `TCodes.cpp`. This also reduces the code complexity quite a bit.

Tested on MK3S+:
* ✅ T0
* ✅ Tx and Tc by starting a print
* ✅ T?
* ✅ Load to Nozzle from LCD

Change in memory:
Flash: -64 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Tcodes.cpp
+++ b/Firmware/Tcodes.cpp
@@ -20,18 +20,7 @@ inline void TCodeInvalid() {
     SERIAL_ECHOLNPGM("Invalid T code."); 
 }
 
-struct SChooseFromMenu {
-    uint8_t slot:7;
-    uint8_t loadToNozzle:1;
-    inline constexpr SChooseFromMenu(uint8_t slot, bool loadToNozzle):slot(slot), loadToNozzle(loadToNozzle){}
-    inline constexpr SChooseFromMenu():slot(0), loadToNozzle(false) { }
-};
-
-SChooseFromMenu TCodeChooseFromMenu() {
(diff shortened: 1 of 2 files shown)


gudnimg:
Assigning this to 3.13 RC2. I think it's too risky for RC1.

DRracer (changes requested):
I'm afraid replacing `load_filament_to_nozzle` with a plain `tool_change` is not correct. 

`load_filament_to_nozzle` really pushes filament into the tip of the nozzle while a `tool_change` doesn't.

Or am I missing something?

gudnimg:
> I'm afraid replacing `load_filament_to_nozzle` with a plain `tool_change` is not correct.
> 
> `load_filament_to_nozzle` really pushes filament into the tip of the nozzle while a `tool_change` doesn't.
> 
> Or am I missing something?

It looks like to me we are currently duplicating the code by handling `T?` in `Tcodes.cpp` and in `tool_change()`

The proposed changes now calls a variant of `tool_change` which handles the `T?` character and calls `load_filament_to_nozzle`. Remember `tool_change` is overloaded.

`MMU2::mmu2.tool_change(strchr_pointer[index], choose_menu_P(_T(MSG_SELECT_FILAMENT), _T(MSG_FILAMENT)));`

gudnimg:
Marking this with 🚧 as the PR will need to be re-tested to verify it doesnt break edge cases.

3d-gussner (approved):
Tested on MK404

DRracer (approved):

PR #4162, opened 2023-04-23
Fix `lcd_encoder_diff` getting out of sync with the knob hard steps

leptun:
Two issues were fixed:
- lcd_encoder_diff should not be cleared to 0 whenever a step is made. Consider the following scenario: lcd_encoder_diff is equal to 6. That means there is one lcd_encoder step to be made, but there are also the two remaining lcd_encoder_diff steps. With the old code, lcd_encoder would be incremented and lcd_encoder_diff would be set to 0, so those remaining 2 substeps would be reset to 0. In this case, the knob would get out of sync with the hard steps of the physical knob.
- lcd_encoder_diff was not being modified atomically. It is updated from an ISR, but also from the main context.

Flash: +22B
SRAM: 0B

### DIFF
--- a/Firmware/lcd.cpp
+++ b/Firmware/lcd.cpp
@@ -684,9 +684,11 @@ void lcd_knob_update() {
 	if (lcd_backlight_wake_trigger) {
 		lcd_backlight_wake_trigger = false;
 		backlight_wake();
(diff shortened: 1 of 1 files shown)


gudnimg (approved):
Looks good to me. 👍

PR #4163, opened 2023-04-23
Fix issue where Tx gcode doesnt reset custom_message_type

gudnimg:
We need to call BeginReport and EndReport
otherwise the SD filename is not shown when printing

This only affects single material MMU gcodes

Similar to https://github.com/prusa3d/Prusa-Firmware/pull/4154 but only fix this one bug. Don't touch any other area of the code.

Things to test:
* Printing with single material MMU profile (Tx/Tc)

Change in memory:
Flash: -4 bytes
SRAM: 0 bytes


leptun (approved):
lgtm. Tested on MK2.5_MMU2SR with and without this patch. The issue is indeed gone now

PR #4164, opened 2023-04-23
Cleanup language menu + 1 minor optimisation in 1st Layer Cal menu

gudnimg:
⚠️ I have not tested these changes yet.

* lcd_draw_update = 2 is now redundant when calling `menu_goto()`. This also means `lcd_clear()` is redundant.
* Change for loop index type to 1 byte.

Change in memory:
Flash: -50 bytes
SRAM: 0 bytes


3d-gussner (approved):
Tested on MK404

- [X]  Languages load correctly
- [X] First layer cal as well

DRracer (approved):

PR #4165, opened 2023-04-24
Remove last <binary.h> dependencies

leptun:
Finally get rid of all instances of the <binary.h> macros.


3d-gussner (approved):
Tested on MK404 and all chars look good.

PR #4166, opened 2023-04-24
M701/M702: Set default Z value to 0

gudnimg:
Set default Z value to 0 to maintain backwards compatibility with older gcodes.

What needs to be tested:
* M600 with and without MMU
* See how print finish behavior is with MMU. Should now be like 3.12
* Any other functionality which uses M701 and M702.

Change in memory:
Flash: -24 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -1805,10 +1805,10 @@ switch(eFilamentAction)
                // FALLTHRU
           case FilamentAction::Load:
                loading_flag=true;
-               enquecommand_P(MSG_M701_NO_LIFT);      // load filament
+               enquecommand_P(MSG_M701);      // load filament
                break;
           case FilamentAction::UnLoad:
-               enquecommand_P(MSG_M702_NO_LIFT);      // unload filament
+               enquecommand_P(MSG_M702);      // unload filament
                break;
           case FilamentAction::MmuLoad:
           case FilamentAction::MmuLoadingTest:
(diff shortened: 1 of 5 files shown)


Firmware/ultralcd.cpp
```diff
@@ -3717,8 +3717,7 @@ static void lcd_wizard_load() {
         lcd_puts_at_P(0, 2, _T(MSG_LOADING_FILAMENT));
         loading_flag = true;
     }
-    gcode_M701(FILAMENTCHANGE_FIRSTFEED, 0);
```
gudnimg, line 3720: Note I hope this change works OK. Otherwise we need another call to `raise_z_above(MIN_Z_FOR_LOAD)` here.

wavexx: This seems to differ. "M701" will call gcode_M701(FILAMENTCHANGE_FIRSTFEED, 0xff) which seems to bypass ``MMU2::mmu2.load_filament_to_nozzle(mmuSlotIndex);``. Here we force slot 0 instead.
I can't test this at the moment to verify...

gudnimg: You're correct!

I also noticed if slot `0xFF` is given with the MMU, then the M701 will act like the single color setup which is not good. I should probably split that MMU if statement into two nested if's even though it may take 4-8 bytes.
image

gudnimg: Should hopefully be fixed now. Needs to be verified.

wavexx: Looks good now.

It's almost a shame we have to put the if for all those raise_z_above().

During M701/M702 the Z parameter is always relative to the current position?

gudnimg: > During M701/M702 the Z parameter is always relative to the current position?

Yes I believe that's correct. Because we use `raise_z` for the lift, which applies a delta to whatever the current position is. The delta is then also used to revert the Z move similar to how we unpark the XY axis in M600.

wavexx (approved):
Looks good now, but I couldn't test it on a real mmu

PR #4167, opened 2023-04-24
Lcd encoder improvements

leptun:
Lcd encoder code optimization. Instead of nested switch statements in an ISR, just use a small lookup table. This lookup table also correctly handles the scenario where the knob turns two substeps from one ISR to the next. In the old code, the lcd_encoder_diff would become out of phase by 2 ticks if that happened. The new code keeps lcd_encoder_diff in sync with the hard encoder ticks at all times. Even so, it's not perfect since there is no way to determine in which direction the encoder rotated when that happens, but at least now it doesn't get out of sync.


3d-gussner (approved):
Testing on real hardware and the knob responds as expected.

PR #4169, opened 2023-04-25
Move nozzle to center during Thermal model cal.

3d-gussner:


Firmware/ultralcd.cpp
```diff
@@ -902,9 +902,24 @@ void lcd_commands()
...
+            lcd_commands_step = 5;
+            break;
+
+        case 5:
+            enquecommand_P(PSTR("G1 X125 Y105 Z1 F8000"));
+            lcd_commands_step = 4;
+            break;
+
+        case 4:
+            st_synchronize();
```
leptun, line 915: This looks wrong. st_synchronise will block before the previos gcode commands get a chance to be parsed and added to the planner properly. I would replace this with an M400 call.

Firmware/ultralcd.cpp
```diff
@@ -902,9 +902,24 @@ void lcd_commands()
         switch (lcd_commands_step)
         {
         case 0:
-            lcd_commands_step = 3;
+            lcd_commands_step = 6;
             [[fallthrough]];
 
+        case 6:
+            enquecommand_P(G28W);
```
leptun: I don't think this many steps are needed. Steps 4, 5 and 6 can be condensed into 3 enquecommand_P() calls for:
```
G28...
G1...
M400

```
This is possible since the cmdqueue should be big enough for all these gcodes to be enqueued at once.

leptun (changes requested):

Firmware/ultralcd.cpp
leptun: Actually, nevermind. The st_sync will only execute once the cmdqueue is empty, so it should be safe. However, I still think the other optimization I mentioned should be applied at some point in the future, if not now.

leptun (approved):
I reconsidered my review. While it is not optimal, it should work for now.

Firmware/ultralcd.cpp
3d-gussner: Right the M400 cost 10 bytes.

Firmware/ultralcd.cpp
3d-gussner: Combining G28W and G1... saves some bytes but the `st_synchronize()` in the same case doesn't work as the TM cal will stop as it isn't IDLE.

Firmware/ultralcd.cpp
wavexx: I would replace this with

```
if(!blocks_queued()) { lcd_commands_step = 3; }
break;
```

so this avoids the nested wait loop. We already check for the cmdbuffer at the uppermost if(), so it we should be covered.

But it's a minor thing. This should also be fine.

PR #4172, opened 2023-04-28
Allow `Statistics` menu while running 1st Layer cal

gudnimg:
There is no problem with opening the menu during first layer calibration. Removing this condition simplifies the code a little bit.

Changes tested on MK3S+ with MMU.

Change in memory:
Flash: -8 bytes
SRAM: 0 bytes


3d-gussner (changes requested):
Needs a rebase and the we can approve and merge it.

DRracer:
rebased onto MK3

3d-gussner (approved):

PR #4173, opened 2023-04-28
optimisation: use `disable_heater()` in more places

gudnimg:
This is just an idea. Needs to be verified.

Change in memory:
Flash: -84 bytes
SRAM: 0 bytes


3d-gussner (changes requested):
Please rebase

gudnimg:
Rebased to sync with MK3 branch (297 commits)

gudnimg:
The change in power panic needs to be tested carefully.

3d-gussner (approved):
Checked and makes sense

PR #4174, opened 2023-04-28
optimisation: use `CHECK_ALL_HEATERS` in more places

gudnimg:
Note that `!a && !b` is equal to `!(a || b)`

* a = `bool(target_temperature[0])`
* b = `bool(target_temperature_bed)`

CHECK_ALL_HEATERS is essentially `(a || b)`

1. `(degTargetBed() == 0) && (degTargetHotend(0) == 0)` is equal to `(!a && !b)` or `!(a || b)` => `!(CHECK_ALL_HEATERS)`
2. `!degTargetBed() && !degTargetHotend(0)` is equal to same as 1) => `!(CHECK_ALL_HEATERS)`
3. `degTargetBed() || degTargetHotend(0)` is equal to `(a || b)` => `(CHECK_ALL_HEATERS)`

Change in memory:
Flash: -162 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -7805,9 +7805,7 @@ Sigma_Exit:
 		cancel_heatup = false;
 
 		bool is_pinda_cooling = false;
-		if ((degTargetBed() == 0) && (degTargetHotend(0) == 0)) {
-		    is_pinda_cooling = true;
-		}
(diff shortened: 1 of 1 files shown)


3d-gussner (approved):
Tested on MK404 :heavy_check_mark: 
To speed up the test set the `DEFAULT_SAFETYTIMER_TIME_MINS 2` instead of 30 min
Previous and with this PR behave the same. :heavy_check_mark:

PR #4175, opened 2023-04-29
cleanup: Remove unused `LcdCommands` state `LoadFilament`

gudnimg:
This may help the compiler generate smaller code

Change in memory:
Flash: -2 bytes
SRAM: 0 bytes


leptun (approved):
Lgtm

PR #4176, opened 2023-04-29
cleanup: more compact gcode/mcode

gudnimg:
Remove 70 lines of code by making
the gcode/mcode parsing more compact in a few selected places

We can probably do a lot more but this is a small step the right direction.

No change in memory

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -5414,10 +5414,7 @@ void process_commands()
     */
     case 32:
     {
-      if(card.sdprinting) {
-        st_synchronize();
-
-      }
+      if(card.sdprinting) st_synchronize();
 
       const char* namestartpos = (strchr(strchr_pointer + 4,'!'));   //find ! to indicate filename string start.
       if(namestartpos==NULL)
@@ -5427,7 +5424,7 @@ void process_commands()
       else
         namestartpos++; //to skip the '!'
 
-      bool call_procedure=(code_seen('P'));
+      bool call_procedure = code_seen('P');
 
       if(strchr_pointer>namestartpos)
         call_procedure=false;  //false alert, 'P' found within filename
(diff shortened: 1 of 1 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -5427,17 +5424,18 @@ void process_commands()
...
       if(strchr_pointer>namestartpos)
         call_procedure=false;  //false alert, 'P' found within filename
 
       if( card.cardOK )
       {
         card.openFileReadFilteredGcode(namestartpos,!call_procedure);
-        if(code_seen('S'))
-          if(strchr_pointer<namestartpos) //only if "S" is occuring _before_ the filename
-            card.setIndex(code_value_long());
+        if(code_seen('S') && strchr_pointer < namestartpos) {
```
3d-gussner, line 5435: Why `{}` here while other places removed?

3d-gussner: The comment can be above

Firmware/Marlin_main.cpp
```diff
@@ -7306,9 +7254,10 @@ SERIAL_PROTOCOLPGM("\n\n");
       float temp = 150.0;
       int e = 0;
       int c = 5;
-      if (code_seen('E')) e = code_value_short();
-        if (e < 0)
-          temp = 70;
+      if (code_seen('E')) {
```
3d-gussner, line 7257: Why not?
if (code_seen('E')) e = code_value_short();
if (e < 0) temp = 70;

3d-gussner (commented):
See comments

DRracer (changes requested):
While doing such a refactoring, please reformat the code segments according to our code style rules. Especially the "parsing" code is horrible and often mis-indented.

gudnimg:
Closing this PR, we're probably better of just reformatting the entire file

PR #4177, opened 2023-04-29
🚧 Simplify `mFilamentItem`

gudnimg:
Ideas to simplify `mFilamentItem()`

Motivation for these changes
* Reduce calls to M701 and M702.
* Reduce duplicate code

Change in memory:
Flash: -120 bytes
SRAM: -1 byte

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -57,7 +57,6 @@ static void lcd_backlight_menu();
 
 FilamentAction eFilamentAction=FilamentAction::None; // must be initialized as 'non-autoLoad'
 static bool bFilamentPreheatState; // True if target temperature is above min_temp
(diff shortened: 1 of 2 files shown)


Firmware/ultralcd.cpp
```diff
@@ -1761,75 +1759,39 @@ void lcd_print_target_temps_first_line(){
     }
 }
 
-static void mFilamentPrompt()
+static void waitFilamentLoop()
 {
-uint8_t nLevel;
+    for(;;) {
+        manage_heater();
+        manage_inactivity(true);
```
gudnimg, line 1767: Note: lcd updates are enabled (since this is run inside the menus), so we probably don't want to call `lcd_update(0)`.

I'm not sure if we should disable lcd updates all together here.

I decided to go with a blocking loop since there shouldn't be any printing going on while waiting for the user.

PR #4180, opened 2023-04-30
Fix regression in `menu_draw_P`

gudnimg:
The new code did not take into account the minus sign.

The template for `menu_draw_P` is removed since there is no benefit in `uint8_t` version of this function since the value
is converted to two bytes anyway by `lcd_printf_P`. Most of the memory savings come from this change.

Change in memory:
Flash: -92 bytes
SRAM: 0 bytes


3d-gussner (approved):
Nice fix and even with saving of resources.
Mon May  8 13:42:17 2023
Mon May  8 13:42:11 2023

Issue #2602, opened 2020-04-13
[BUG] G28 or G80 moves E when temp is >=180


ochm:
I'm not sure if it is a bug or feature, but IMHO mes bed leveling should not to do extrusion at the end.

**Printer type** MK3S
**Printer firmware version**-  3.9.0-RC1-3272
**Describe the bug**
When the nozzle is preheated under 180, the G28, or G80 does mesh bed leveling, 
and the extruder is parked at 0,0. Correct.
But when the nozzle is preheated to or over 180, after G28 or G80, the extruder is parked at 0,0 and additionaly does short move of E. 
  
**To Reproduce**
No E move at the end of mesh bed leveling:
M109 S160
G28 ;or G80

Extra E move at the end of mesh bed leveling:
M109 S180
G28 ;or G80

**Expected behavior**
As i wrote above, i don't know if it is bug or feature, 
but IMHO there is no reason for move E after mesh bed leveling at any temperature of nozzle.

leptun:
Isn't the Extruder retracted at the beginning of MBL? Maybe what you are seeing is simply the unretract movement.

Panayiotis-git:
IIRC, if the hotend temperature is greater than the `EXTRUDE_MINTEMP`, the firmware retracts the filament at the beginning of the MBL.

ochm:
Not, the E moves after MBL and the movement is in forward (extrusion) direction, not retraction.
You can see it at video in a moment...

leptun:
@ochm Well, yes. That is what an unretract movement is.
Retraction pulls filament (backwards), unretraction pushes filament (forwards).
I believe this behavior was done so that during MBL the oozing of filament would be minimal.

ochm:
http://www.och.cz/tmp/E-moves-after-MBL.mov you can see extraction of some filament after MBL. There is absolutely no reason to do it.

And there is no retraction before MBL.
M862.1 P0.40 ; nozzle diameter check
G90 ; use absolute coordinates
M83 ; extruder relative mode
M104 S180 ; set extruder temp
M140 S85 ; set bed temp
M190 S85 ; wait for bed temp
M109 S180 ; wait for extruder temp
G28 W ; home all without mesh bed level
G80 ; mesh bed leveling
G1 Y-3.0 F1000.0 ; go outside print area
M109 S230; wait for extruder temp
G92 E0.0
G1 X60.0 E9.0 F1000.0 ; intro line
[...]

Issue #2579, opened 2020-04-05
[QUESTION] Code weirdness in G80


leptun:
```
		if (degHotend(active_extruder) > EXTRUDE_MINTEMP && temp_cal_active == true && calibration_status_pinda() == true && target_temperature_bed >= 50) {
			current_position[E_AXIS] += default_retraction;
			plan_buffer_line_curposXYZE(400, active_extruder);
		}
```
This snippet of code at the end of G80 makes no sense. Why only unretract when Temperature calibration is used?

Panayiotis-git:
At the beginning of the G80 command there is the following code:
```
#ifndef PINDA_THERMISTOR
		if (run == false && temp_cal_active == true && calibration_status_pinda() == true && target_temperature_bed >= 50)
		{
			temp_compensation_start();
			run = true;
			repeatcommand_front(); // repeat G80 with all its parameters
			enquecommand_front_P((PSTR("G28 W0")));
			break;
		}
        run = false;
#endif //PINDA_THERMISTOR
```
The `temp_compensation_start()` function, does a retraction:
```
	if (degHotend(active_extruder) > EXTRUDE_MINTEMP) {
		current_position[E_AXIS] -= default_retraction;
	}
	plan_buffer_line_curposXYZE(400, active_extruder);
```

I think that the unretract should be enclosed in a:
`#ifndef PINDA_THERMISTOR`

leptun:
Thanks, that makes a lot more sense.
Now that I think about it, why don’t we always retract during MBL? It just seems logical doing so.

vintagepc:
On the topic of hidden gems, I found this lurking in lcd.cpp:

```
#ifdef VT100
 uint8_t lcd_escape[8];
#endif
```
What is a VT100,you ask? 🤣 
https://en.wikipedia.org/wiki/VT100

leptun:
@vintagepc it’s not hidden code, it is my code :).
The vt100 codes are escape sequence of characters that trigger a certain terminal command. In our case, we had home and set cursor that were useful for the lcd.
These were quite wasteful, so I changed the code to not use them anymore. I commented out the code so that I could save close to 1KB of flash

Edit: fix misinformation

vintagepc:
Interesting - either way it gave me a good laugh!

PR #4183, opened 2023-05-01
MBL print area

### LINKED ISSUE #3224: [ENHANCEMENT] Add repetition limit on enforcing Z calibration during mesh bed leveling
During mesh bed leveling, if there is an error with the z-height [("Enforcing the Z calibration")](https://help.prusa3d.com/en/article/mesh-bed-leveling_112163/#enforcing-the-z-calibration), this can go on indefinitely. For example, if the magnetic bed is misaligned and not flat (eg, it is accidentally on a screw in the back of the print bed), mesh bed leveling will continuously re-try and enforce the z calibration. Today, I can into work and found that the printer had been running the mesh bed/z-enforcement for 20 hours.

The obvious solution to this is a timeout or limit on the repetitions. For example, it will enforce the Z calibration no more than five times. If it hits that limit, it turns off the heaters and prompts the user on the screen.

(This occurred on a MK3S with firmware version 3.10.0-4481)

leptun:
Similar implementation of the print area UBL from the Buddy firmware, but instead of using `M555` for defining the print area, `G80` uses additional parameters. The `G80` print area is defined with `X`, `Y`, `W` and `H`, just like with `M555`. If the print area is not specified, the whole bed is probed. Gcode future and backward compatibility is maintained since old firmware will ignore these extra parameters.

With this PR, if a print area is provided, the 7x7 MBL will only probe those points. This results in a quicker, precise MBL where precision is needed. Besides the print area probed points, the rest of the 7x7 mesh uses interpolation of the data sorted in eeprom during Z calibration so that the toolhead can safely exit the print area and not cause a nozzle crash with a warped bed.

**Tasks after merge**
- [ ] Update rep-rap Gcode documentation
- [x] Update PrusaSlicer profiles
- [x] Update translation (`kill(PSTR("Mesh bed leveling failed. Please run Z calibration."));`)
- [x] Add `O` in `G80 [ N | R | L | R | F | B | X | Y | W | H ]` list

https://user-images.githubusercontent.com/17808203/235438936-5a1d4750-c54e-4ca1-b04d-6058d9f225a9.MOV

image (25)

https://user-images.githubusercontent.com/17808203/235439037-07113fb1-cba4-4f4a-847b-9709a6b06f1b.MOV

In case the 3x3 mesh in eeprom from Z calibration is missing/invalid, the 3x3 points are also probed besides the print area:

https://user-images.githubusercontent.com/17808203/235439240-6f7d648f-204e-4c27-b208-c8271e6d7415.MOV

Another interesting feature enabled by this PR is the ability to load and activate the mesh stored in eeprom without probing any point ~(besides point 0)~. This can be achieved with `G80 W` since it results in the print area being completely outside the bed.

Closes https://github.com/prusa3d/Prusa-Firmware/issues/3224

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -2800,14 +2800,6 @@ static void gcode_G80()
     static bool run = false; // thermistor-less PINDA temperature compensation is running
 #endif // ndef PINDA_THERMISTOR
 
-#ifdef SUPPORT_VERBOSITY
-    int8_t verbosity_level = 0;
-    if (code_seen('V')) {
-        // Just 'V' without a number counts as V1.
-        char c = strchr_pointer[1];
-        verbosity_level = (c == ' ' || c == '\t' || c == 0) ? 1 : code_value_short();
-    }
-#endif //SUPPORT_VERBOSITY
     // Firstly check if we know where we are
     if (!(axis_known_position[X_AXIS] && axis_known_position[Y_AXIS] && axis_known_position[Z_AXIS])) {
         // We don't know where we are! HOME!
@@ -2877,15 +2869,7 @@ static void gcode_G80()
     current_position[X_AXIS] = BED_X0;
     current_position[Y_AXIS] = BED_Y0;
 
-#ifdef SUPPORT_VERBOSITY
-    if (verbosity_level >= 1)
-    {
-        bool clamped = world2machine_clamp(current_position[X_AXIS], current_position[Y_AXIS]);
-        clamped ? SERIAL_PROTOCOLPGM("First calibration point clamped.\n") : SERIAL_PROTOCOLPGM("No clamping for first calibration point.\n");
-    }
-#else //SUPPORT_VERBOSITY
     world2machine_clamp(current_position[X_AXIS], current_position[Y_AXIS]);
-#endif //SUPPORT_VERBOSITY
 
     int XY_AXIS_FEEDRATE = homing_feedrate[X_AXIS] / 20;
     plan_buffer_line_curposXYZE(XY_AXIS_FEEDRATE);
@@ -2901,11 +2885,6 @@ static void gcode_G80()
     uint8_t mesh_point = 0; //index number of calibration point
     int Z_LIFT_FEEDRATE = homing_feedrate[Z_AXIS] / 40;
     bool has_z = is_bed_z_jitter_data_valid(); //checks if we have data from Z calibration (offsets of the Z heiths of the 8 calibration points from the first point)
-#ifdef SUPPORT_VERBOSITY
-    if (verbosity_level >= 1) {
-        has_z ? SERIAL_PROTOCOLPGM("Z jitter data from Z cal. valid.\n") : SERIAL_PROTOCOLPGM("Z jitter data from Z cal. not valid.\n");
-    }
-#endif // SUPPORT_VERBOSITY
     int l_feedmultiply = setup_for_endstop_move(false); //save feedrate and feedmultiply, sets feedmultiply to 100
     while (mesh_point != MESH_NUM_X_POINTS * MESH_NUM_Y_POINTS) {
         // Get coords of a measuring point.
@@ -2922,11 +2901,6 @@ static void gcode_G80()
             uint16_t z_offset_u = eeprom_read_word((uint16_t*)(EEPROM_BED_CALIBRATION_Z_JITTER + 2 * ((ix/3) + iy - 1)));
             z0 = mbl.z_values[0][0] + *reinterpret_cast<int16_t*>(&z_offset_u) * 0.01;
             mbl.set_z(ix, iy, z0);
-#ifdef SUPPORT_VERBOSITY
-            if (verbosity_level >= 1) {
-                printf_P(PSTR("Bed leveling, point: %d, calibration Z stored in eeprom: %d, calibration z: %f \n"), mesh_point, z_offset_u, z0);
-            }
-#endif // SUPPORT_VERBOSITY
         }
 
         if ((nMeasPoints == 3) && !isOn3x3Mesh) {
@@ -2951,20 +2925,8 @@ static void gcode_G80()
         current_position[X_AXIS] = x_pos;
         current_position[Y_AXIS] = y_pos;
 
-        //printf_P(PSTR("[%f;%f]\n"), current_position[X_AXIS], current_position[Y_AXIS]);
-
-
-#ifdef SUPPORT_VERBOSITY
-        if (verbosity_level >= 1) {
-            bool clamped = world2machine_clamp(current_position[X_AXIS], current_position[Y_AXIS]);
-            SERIAL_PROTOCOL(mesh_point);
-            clamped ? SERIAL_PROTOCOLPGM(": xy clamped.\n") : SERIAL_PROTOCOLPGM(": no xy clamping\n");
-        }
-#else //SUPPORT_VERBOSITY
         world2machine_clamp(current_position[X_AXIS], current_position[Y_AXIS]);
-#endif // SUPPORT_VERBOSITY
 
-        //printf_P(PSTR("after clamping: [%f;%f]\n"), current_position[X_AXIS], current_position[Y_AXIS]);
         plan_buffer_line_curposXYZE(XY_AXIS_FEEDRATE);
         st_synchronize();
         if (planner_aborted)
@@ -2981,7 +2943,6 @@ static void gcode_G80()
             break;
         }
         if (init_z_bckp - current_position[Z_AXIS] < 0.1f) { //broken cable or initial Z coordinate too low. Go to MESH_HOME_Z_SEARCH and repeat last step (z-probe) again to distinguish between these two cases.
-            //printf_P(PSTR("Another attempt! Current Z position: %f\n"), current_position[Z_AXIS]);
             current_position[Z_AXIS] = MESH_HOME_Z_SEARCH;
             plan_buffer_line_curposXYZE(Z_LIFT_FEEDRATE);
             st_synchronize();
@@ -2999,45 +2960,19 @@ static void gcode_G80()
             puts_P(PSTR("Bed leveling failed. Sensor triggered too high."));
             break;
         }
-#ifdef SUPPORT_VERBOSITY
-        if (verbosity_level >= 10) {
-            SERIAL_ECHOPGM("X: ");
-            MYSERIAL.print(current_position[X_AXIS], 5);
-            SERIAL_ECHOLNPGM("");
-            SERIAL_ECHOPGM("Y: ");
-            MYSERIAL.print(current_position[Y_AXIS], 5);
-            SERIAL_PROTOCOLPGM("\n");
-        }
-#endif // SUPPORT_VERBOSITY
-        float offset_z = 0;
 
 #ifdef PINDA_THERMISTOR
-        offset_z = temp_compensation_pinda_thermistor_offset(current_temperature_pinda);
-#endif //PINDA_THERMISTOR
-        //			#ifdef SUPPORT_VERBOSITY
-        /*			if (verbosity_level >= 1)
-                    {
-                    SERIAL_ECHOPGM("mesh bed leveling: ");
-                    MYSERIAL.print(current_position[Z_AXIS], 5);
-                    SERIAL_ECHOPGM(" offset: ");
-                    MYSERIAL.print(offset_z, 5);
-                    SERIAL_ECHOLNPGM("");
-                    }*/
-        //			#endif // SUPPORT_VERBOSITY
+        float offset_z = temp_compensation_pinda_thermistor_offset(current_temperature_pinda);
         mbl.set_z(ix, iy, current_position[Z_AXIS] - offset_z); //store measured z values z_values[iy][ix] = z - offset_z;
+#else
+        mbl.set_z(ix, iy, current_position[Z_AXIS]); //store measured z values z_values[iy][ix] = z;
+#endif //PINDA_THERMISTOR
 
         custom_message_state--;
         mesh_point++;
         lcd_update(1);
     }
     current_position[Z_AXIS] = MESH_HOME_Z_SEARCH;
-#ifdef SUPPORT_VERBOSITY
-    if (verbosity_level >= 20) {
-        SERIAL_ECHOLNPGM("Mesh bed leveling while loop finished.");
-        SERIAL_ECHOLNPGM("MESH_HOME_Z_SEARCH: ");
-        MYSERIAL.print(current_position[Z_AXIS], 5);
-    }
-#endif // SUPPORT_VERBOSITY
     plan_buffer_line_curposXYZE(Z_LIFT_FEEDRATE);
     st_synchronize();
     if (mesh_point != MESH_NUM_X_POINTS * MESH_NUM_Y_POINTS) {
@@ -3067,7 +3002,6 @@ static void gcode_G80()
 #endif // TMC2130
             enable_z_endstop(bState);
         } while (st_get_position_mm(Z_AXIS) > MESH_HOME_Z_SEARCH); // i.e. Z-leveling not o.k.
-        //               plan_set_z_position(MESH_HOME_Z_SEARCH); // is not necessary ('do-while' loop always ends at the expected Z-position)
 
         custom_message_type = custom_message_type_old;
         custom_message_state = custom_message_state_old;
@@ -3077,19 +3011,12 @@ static void gcode_G80()
         return;
     }
     clean_up_after_endstop_move(l_feedmultiply);
-    //		SERIAL_ECHOLNPGM("clean up finished ");
 
 #ifndef PINDA_THERMISTOR
     if(eeprom_read_byte((uint8_t *)EEPROM_TEMP_CAL_ACTIVE) && calibration_status_pinda() == true) temp_compensation_apply(); //apply PINDA temperature compensation
 #endif
     babystep_apply(); // Apply Z height correction aka baby stepping before mesh bed leveing gets activated.
-    //		SERIAL_ECHOLNPGM("babystep applied");
     bool eeprom_bed_correction_valid = eeprom_read_byte((unsigned char*)EEPROM_BED_CORRECTION_VALID) == 1;
-#ifdef SUPPORT_VERBOSITY
-    if (verbosity_level >= 1) {
-        eeprom_bed_correction_valid ? SERIAL_PROTOCOLPGM("Bed correction data valid\n") : SERIAL_PROTOCOLPGM("Bed correction data not valid\n");
-    }
-#endif // SUPPORT_VERBOSITY
     const constexpr uint8_t sides = 4;
     int8_t correction[sides] = {0};
     for (uint8_t i = 0; i < sides; ++i) {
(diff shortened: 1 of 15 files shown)


leptun:
In order to use the feature, change the G80 call in the slicer startup script to this:
```
G80 X{first_layer_print_min[0]} Y{first_layer_print_min[1]} W{(first_layer_print_max[0]) - (first_layer_print_min[0])} H{(first_layer_print_max[1]) - (first_layer_print_min[1])}; mesh bed leveling
```

MBL 7x7 must also be enabled on the printer. Otherwise, the entire bed will be probed with 3x3.

Firmware/Marlin_main.cpp
```diff
@@ -2905,7 +2906,7 @@ static void gcode_G80()
 
         // Move Z up to the probe height of the current Z point.
         const float z0 = mbl.z_values[iy][ix];
-        const float init_z_bckp = !has_z ? MESH_HOME_Z_SEARCH : z0 + 0.35;
+        const float init_z_bckp = !has_z ? MESH_HOME_Z_SEARCH : z0 + MESH_HOME_Z_SEARCH_FAST;
         if (init_z_bckp > current_position[Z_AXIS])
             current_position[Z_AXIS] = init_z_bckp;
         plan_buffer_line_curposXYZE(Z_LIFT_FEEDRATE);
```
gudnimg: Can we add `plan_buffer_line_curposXYZE` call and `st_synchronize` under the if statement? It looks like to me this planning is not needed unless the if statement is true.

leptun: You are correct. Pushing the change now 👍

leptun:
I pushed a lot of changes to the PR. I fixed the point counter behavior. It now computes in advance how many points will be probed, so it no longer skips the counter on the status screen.
I also improved a lot the eeprom mesh code and the code that determines the approximate range where the current probed point should be located. I am now using the upsampled eeprom mesh (in the past it was just hardcoded to current_position=0.f for all points not on the 3x3 mesh for some weird reason). So now I finally fixed the issue where certain warped beds have lots and lots and lots of rechecks on certain points. This was especially an issue on MK2.5(S) since the bed there was generally more warped.
I also improved how the Z is lifted to the next point. Previously it was computed relative to the height of the last measured point. Now I compute it based on the upsampled mesh from eeprom. This results in a much more consistent MBL on warped beds, while still being super fast.
I also no longer have to probe the first point. Only the print area needs to be probed now. So `G80 W` doesn't probe any point now, but this time for real :slightly_smiling_face:. I tested it on my machine with a calibrated eeprom mesh and the first layer was still spot on.

Commands I use for testing:
```
G80 R1 N3                       ; regular 3x3 MBL
G80 R1 N7                       ; regular 7x7 MBL
G80 R3 N7 W                     ; load eeprom MBL. No probing
G80 R3 N7 R1 X20 Y10 W10 H10    ; Small 2x2 mesh near origin
G80 R3 N7 X115 Y91 W32 H24      ; Small 3x3 mesh at the center of the bed
G80 R3 N7 X100 Y80 W70 H70      ; Small 4x4 mesh near the center of the bed
G80 R3 N7 X70 Y50 W109 H109     ; Small 5x5 mesh at the center of the bed
```

Useful tool for visualizing the bed data. Dump the mesh with G81. Beware that the Y axis is flipped:
http://lokspace.eu/3d-printer-auto-bed-leveling-mesh-visualizer/

leptun:
Marked it as ready for review. Don't really have anything else I'd like to change besides maybe some optimizations here and there. Functionality wise, everything seems to be working fine. I still need to test how MBL performs with no Z calibration data. All other tests were performed on two printers: MK3S+_MMU3 and MK2.5_MMU2SR.

leptun:
Up to date video of a centered small 3x3 mesh. Counter works fine, first point is no longer measured. Still a bit disappointed that the superpinda needs to slow down on certain points due to the higher hysteresis. Pinda V2 would not have those slowdowns as is the case on the MK2.5 printer.

https://user-images.githubusercontent.com/17808203/236611576-491db0dd-bd9f-49e0-aa0d-96ef5c3dd046.MOV

ojrik:
Hello. Very nice work. It would be good to remove the return to Home, at the end of leveling.

leptun:
@ojrik Thank you. I need to keep the return to origin for backwards compatibility with old gcodes. But I understand it could be a useful setting to disable somehow if the purge line is printed next to the object. I'm thinking I could add an "O<0,1>" parameter to specify if return to origin is needed after G80, with a default of 1. Still need to think about the implications of this.

ojrik:
Nice thought.
I have this PR applied in Firmware 3.12.2, it works well on MK2.5S + MMU2S, MK3S + MMU2S.

leptun:
@ojrik I added the `O` parameter like I described in my previous comment. You may now add `O0` to the G80 command to disable the return to origin.

> I have this PR applied in Firmware 3.12.2, it works well on MK2.5S + MMU2S, MK3S + MMU2S.

I'm surprised the PR can be rebased so easily onto 3.12. In any case this PR is for 3.14 at the earliest since the 3.13 release is already packed full of changes. Thank you also for testing the PR on your printers. Feedback is greatly appreciated. Do you mind if you also post the G81 output from both printers of the commands listed in this comment? https://github.com/prusa3d/Prusa-Firmware/pull/4183#issuecomment-1536586069. I want to see how closely the mesh saved in eeprom matches the real hardware. Also, did you perform Z/XYZ calibration recently on any of those printers?

ojrik:
Do you mean something like this?
The XYZ calibration was about a month ago when I mounted the Super Pinda on the MK3S MMU2S.
[Test MBL.txt](https://github.com/prusa3d/Prusa-Firmware/files/11419264/Test.MBL.txt)

leptun:
Yes, something like that. I'm also interested in `G80 O0 N7 W`. This should load the Z calibration mesh from eeprom. This mesh data was stored there during Z calibration. The Z calibration is different than MBL and is performed at the end of XYZ calibration or manually via the menu ("Calibration->Calibrate Z"). That calibration probes 9 MBL points and stores them in eeprom. That data is later used by G80 while probing to validate measurements. With this PR, we rely much more on this eeprom data than in the past, so it's critical that the data in eeprom is correct.
Since you ran the XYZ calibration a month ago, the data in eeprom should be quite fresh still. Unless you did some work on the printer hardware in the meantime, it should still be fine.

ojrik:
Zadal som príkaz G80 N7 O0 W, potom G81 my vyšlo toto.

Recv:   0.12000  0.11556  0.11222  0.11000  0.10889  0.10889  0.11000
Recv:   0.10556  0.10296  0.10037  0.09778  0.09519  0.09259  0.09000
Recv:   0.09889  0.09556  0.09185  0.08824  0.08333  0.07852  0.07333
Recv:   0.10000  0.09333  0.08667  0.08000  0.07380  0.06667  0.06000
Recv:   0.10889  0.09630  0.08481  0.07528  0.06519  0.05704  0.05000
Recv:   0.12556  0.10444  0.08630  0.07111  0.05889  0.04963  0.04333
Recv:   0.15000  0.11778  0.09111  0.07000  0.05444  0.04444  0.04000

leptun:
Wow, your printer is considerably straighter than mine :).
Were the measurements in `Test.MBL.txt` done while the printer is hot or cold?

ojrik:
cold

ojrik:
I use Prusa Leveling Guide in Octoprint to set BED.

ojrik:
MK2.5S Test
[Test MBL1.txt](https://github.com/prusa3d/Prusa-Firmware/files/11419679/Test.MBL1.txt)

leptun:
Oy. That's one bent bed 😅. Does this PR help with rechecks during MBL? Both 7x7 and 3x3. In my case on the MK2.5, the points on the last row always had to be rechecked. With this PR, those issues were gone for me. Is this also the case for you?

ojrik:
I rarely use MK2.5S, I just tested your PR with FIRMWARE 3.12.2, FW 3.13.0, I don't have FW for MMU2S

leptun:
If all goes well, the new MMU firmware should be released at the end of this month. You can still use FW 3.13.0 on your printer, but you have to disable the MMU in settings

ojrik:
I know that, I tried that too, I have it adjusted according to this PR.

ojrik:
I use Visual Studio Code. With FW3.13.0 I have some problems compiling. MK2.5S date 1970-01-01, languages ​​not applicable. MK3S also date, and some small things in FW. But that's a small point.

leptun:
The compile date is intentional. It's done like this so that the FW build is consistent, so if you build now and again later, the same compiled hex will be output. The build server should use the correct time and date, so those builds are not affected.
MK2.5S language not applicable: which cmake target are you using for building?
What other fw issues are you encountering with the mk3s?

ojrik:
Here is my Test MK2.5S FW3.12.2
Fantastic, because it's greasy and the bed is so bent.

ojrik:
You need to modify the O Parameter.
so that O0 is the Home return, so G80 remains unchanged. I found a bug with First Layer Calib. So that the kernel FW does not have to be modified.

leptun:
Are you sure you used the latest commits? I initially pushed a wrong commit where the logic was flipped, but then I force pushed a fix.
At the moment it's like this:
- No `O`: Return to origin, old behaviour
- `O0`: No return to origin
- `O1` (or any other O value besides 0): Return to origin.

I don't see what could be broken because of this. Can you please make a video or point me to where this is failing?

gudnimg:
A few tests on my end using a stock MK3S+ with Superpinda

Current XYZ calibration statistics: (I have not re-run it with this PR)
----

* Y distance from min: Left: 11.80mm, Right: 11.54mm
* measured skew 0.00°
* Point offset: X = -1.44mm, Y = 1.32mm
----

Running various Gcode using this PR (Satin sheet):
* Z live adjust value is -1.117
* Nozzle and Bed are cold
* Note that the results below are all taken sequentially without rebooting the printer.

G80 followed by a G81 (Each point is measured 3 times):
image

_Repeat_ G80 followed by a G81:
image

_Repeat_ one more time, G80 followed by a G81:
image

----

`G80 R1 N3` followed by G81:
image


`G80 R1 N7` followed by G81:
image


`G80 R3 N7 W` followed by G81:
Z lifts up and goes back down (is that required?)
image


`G80 R3 N7 R1 X20 Y10 W10 H10` followed by G81
image


`G80 R3 N7 X115 Y91 W32 H24` followed by G81
image


`G80 R3 N7 X100 Y80 W70 H70` followed by G81
image


`G80 R3 N7 X70 Y50 W109 H109` followed by G81
image

gudnimg:
One minor thing that can be improved. If `meshPointsToProbe = 0` as in the case of running `G80 R3 N7 W`, I feel the Z-axis should not be moved. We can probably leave the G80 code earlier.

Also in this case calling `go_home_with_z_lift()` due to the O parameter being missing is redundant since the G80 makes sure to call `G28 W` before running mesh bed leveling (so the printer is already at home). Probably a nitpick but just an observation.

ojrik:
I apologise. It's okay now. My fault.

leptun:
@ojrik No worries.

@gudnimg
> One minor thing that can be improved. If meshPointsToProbe = 0 as in the case of running G80 R3 N7 W, I feel the Z-axis should not be moved. We can probably leave the G80 code earlier.

Not sure if I want to add extra code to handle that scenario. I'm saying this since the `G80 N7 W` sequence for loading the eeprom mesh was just an interesting side effect of the code that I noticed. What that command actually means is run a 7x7 MBL with X and Y min positions at -Infinity, with a width of 0 and max_y pos of Infinity. So it basically defines a print area that is completely outside the bed, so no point ends up being probed. This however is not true in case the eeprom mesh is invalid, so I wouldn't rely on this for anything else besides testing. Do you think we should make this a feature, so being able to load the eeprom mesh at will? There is still the thing that some motion will have to happen when babystepping will be applied, so there will be a bit of Z motion.

> Also in this case calling go_home_with_z_lift() due to the O parameter being missing is redundant since the G80 makes sure to call G28 W before running mesh bed leveling (so the printer is already at home). Probably a nitpick but just an observation.

Not sure I understand this. The O parameter controls whether `go_home_with_z_lift()` is called AFTER probing, not before.

leptun:
@ojrik Can you please post the modified startup script that you use with G80 O0? I want to play with that a bit myself

ojrik:
You mean Prusaslicer

leptun:
Yes. Sorry, forgot to say it's the gcode startup script in the slicer I was referring to.

ojrik:
M115 U3.12.0 ; tell printer latest fw version

M83 ; extruder relative mode
G21 ; set units to millimeters
M107 ; turn off fan
M84 E ; turn off E motor
{if filament_type[initial_tool]=="PC" or filament_type[initial_tool]=="NYLON"}
M104 S{first_layer_temperature[initial_tool]-25} ; set extruder temp for bed leveling
M109 R{first_layer_temperature[initial_tool]-25} ; wait for temp
{elsif filament_type[initial_tool]=="FLEX"}
M104 S210 ; set extruder temp for bed leveling
M109 R210 ; wait for temp
{else}
M104 S170 ; set extruder temp for bed leveling
M109 R170 ; wait for temp
{endif}
G28 W; home all without mesh bed level
;
; MBL
;
M84 E ; turn off E motor
G80 N7 O0 X{first_layer_print_min[0]} Y{first_layer_print_min[1]} W{(first_layer_print_max[0]) - (first_layer_print_min[0])} H{(first_layer_print_max[1]) - (first_layer_print_min[1])}; mesh bed leveling

; go outside print area
G0 X{(min(print_bed_max[0], first_layer_print_min[0]))} Y{(max(0, first_layer_print_min[1])) - 4.5} Z{30} F5000

G1 Z.2 F1000
G92 E0
;
; Extrude purge line
;
{if filament_type[initial_tool]=="FLEX"}
G1 E4 F2400 ; deretraction
{else}
G1 E2 F2400 ; deretraction
{endif}
; intro line
G1 X{(min(print_bed_max[0], first_layer_print_min[0])) + 40} E{40 * 0.13} F1000
G1 Y{(max(0, first_layer_print_min[1])) - 4.5 - 1.5} E{1.5 * 0.15} F1000
G1 X{(min(print_bed_max[0], first_layer_print_min[0]))} E{40 * 0.3} F800
G92 E0
M221 S100 ; set flow to 100%

ojrik:
MK2.5S

leptun:
Added some changes so that https://github.com/prusa3d/Prusa-Firmware/issues/3224 is fixed. Also made the thresholds for throwing "Some problem encountered, Z-leveling enforced ..." tighter since now we can. I skewed the Z axis on my MK2.5 by 3 electrical fullsteps on one of the Z leadscrews and it properly detected the issue.

Placeholder message after repeated failures:
image

leptun:
Fixed the G80 unretract issues explained in these issues:
- https://github.com/prusa3d/Prusa-Firmware/issues/2579
- https://github.com/prusa3d/Prusa-Firmware/issues/2602

leptun:
@ojrik after the latest changes, are you encountering any issues on MK2.5? I made the thresholds for errors tigher, so it might trigger easier on warped beds such as yours.

ojrik:
At the moment, I have not yet applied the final modifications.

ojrik:
No problems for me yet.
The general problem is often with printing that the heating head is not heated when leveling the bed. Unheated filament will then be left on the nozzle, and therefore the Z calibration will start.

ojrik:
On the other hand, I have a heated MBL heating head for the MK2.5S and MK3S and MMU2S in the Prusaslicer, as does the Prusa MINI and MK4 and XL.

leptun:
Rebased onto the latest MK3 branch

3d-gussner (approved):
Tested on MK404 and real printer

Firmware/Marlin_main.cpp
```diff
@@ -4957,29 +4858,26 @@ void process_commands()
     Default 3x3 grid can be changed on MK2.5/s and MK3/s to 7x7 grid.
     #### Usage
 	  
-          G80 [ N | R | V | L | R | F | B ]
+          G80 [ N | R | L | R | F | B | X | Y | W | H ]
```
leptun, line 4861: Just noticed I forgot to add `O` here

sarusani:
@leptun 
I noticed that when I use "G80 C0 N3" the bed gets probed once at each point as expected. (1 probe, 0 repeats) But G81 wont return any data.

Log with  "G80 C0 N3":
```
Send: G80 C0 N3
echo:Enqueing to the front: "G28 W"
Recv: tmc2130_home_enter(axes_mask=0x01)
Recv:   0 step=30 mscnt= 480
Recv: tmc2130_goto_step 0 61 2 1000
Recv: tmc2130_home_exit tmc2130_sg_homing_axes_mask=0x01
Recv: tmc2130_home_enter(axes_mask=0x02)
Recv: echo:busy: processing
Recv:   0 step=18 mscnt= 288
Recv: tmc2130_goto_step 1 33 2 1000
Recv: tmc2130_home_exit tmc2130_sg_homing_axes_mask=0x02

Send: G81
Recv: Num X,Y: 7,7
Recv: Z search height: 5.0f
Recv: Measured points:
Recv:   0.00000  0.00000  0.00000  0.00000  0.00000  0.00000  0.00000
Recv:   0.00000  0.00000  0.00000  0.00000  0.00000  0.00000  0.00000
Recv:   0.00000  0.00000  0.00000  0.00000  0.00000  0.00000  0.00000
Recv:   0.00000  0.00000  0.00000  0.00000  0.00000  0.00000  0.00000
Recv:   0.00000  0.00000  0.00000  0.00000  0.00000  0.00000  0.00000
Recv:   0.00000  0.00000  0.00000  0.00000  0.00000  0.00000  0.00000
Recv:   0.00000  0.00000  0.00000  0.00000  0.00000  0.00000  0.00000
```

Log with  "G80 C1 N3":
```
Send: G80 C1 N3
Recv: echo:Enqueing to the front: "G28 W"
Recv: tmc2130_home_enter(axes_mask=0x01)
Recv:   0 step=29 mscnt= 479
Recv: tmc2130_goto_step 0 61 2 1000
Recv: tmc2130_home_exit tmc2130_sg_homing_axes_mask=0x01
Recv: tmc2130_home_enter(axes_mask=0x02)
Recv:   0 step=18 mscnt= 290
Recv: tmc2130_goto_step 1 33 2 1000
Recv: tmc2130_home_exit tmc2130_sg_homing_axes_mask=0x02

Send: G81
Recv: Num X,Y: 7,7
Recv: Z search height: 5.0f
Recv: Measured points:
Recv:   0.14750  0.20333  0.24167  0.26250  0.26583  0.25167  0.22000
Recv:   0.14167  0.19525  0.22775  0.23917  0.22951  0.19877  0.14694
Recv:   0.13917  0.18654  0.21127  0.21333  0.19275  0.14951  0.08361
Recv:   0.14000  0.17722  0.19222  0.18500  0.15556  0.10389  0.03000
Recv:   0.14417  0.16728  0.17062  0.15417  0.11793  0.06191  -0.01389
Recv:   0.15167  0.15673  0.14645  0.12083  0.07988  0.02358  -0.04806
Recv:   0.16250  0.14556  0.11972  0.08500  0.04139  -0.01111  -0.07250
```

leptun:
C0 is not a valid parameter. Only C >= 1 is allowed. The first probing try is always discarded. That's why C1 probes twice. That initial probe is used for detecting deviations only afaik. In any case, this PR didn't touch that part of the code.

sarusani:
👍 Thanks for the info!

leptun:
Fyi, PrusaSlicer 2.7.0-alpha1 now has the updated startup gcode for the MK3/S/+ and MK2.5/S variants to make use of the MBL print area feature.

PR #4184, opened 2023-05-01
Fix regression where time remaining can overflow on status screen

gudnimg:
If the remaining print time is more than 10.9 hours at 100% then increasing the feedrate multiplier
to 101% or more will yield an unrealistic time due to overflow occuring.

Video demonstration of the bug in action: https://www.youtube.com/shorts/PLXyA_Jf9ck

When my print reached 10h51m@100% the overflow went away, I believe this supports my suspicion. This bug was introduced here https://github.com/prusa3d/Prusa-Firmware/pull/3971/commits/040087923aabbd1abaeaa2b93733017343e2eb6d#diff-5fbfdca43c192573e31d8e78d255c510fb6e213b118cf38173277ec2a2163880R504

Change in memory:
Flash +2 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -446,9 +446,9 @@ void lcdui_print_time(void)
     //if remaining print time estimation is available print it else print elapsed time
     int chars = 0;
     if (printer_active()) {
-        uint16_t print_t = PRINT_TIME_REMAINING_INIT;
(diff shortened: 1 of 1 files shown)


gudnimg:
Tested the changes on MK3S+. And this patch fixes the issue on my end.

leptun (approved):
Lgtm. I'm surprised how optimal the code still is

wavexx (approved):

PR #4188, opened 2023-05-04
Always allow Live Adjust Z in Settings

leptun:
While the printer is idle, it should be possible to change the Live Adjust Z value from Settings.
The Settings menu is accessible only while the printer is idle, so it is safe to show the menu item at all times.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -4470,7 +4470,7 @@ static void lcd_settings_menu()
     MENU_ITEM_TOGGLE_P(_T(MSG_RPI_PORT), (selectedSerialPort == 0) ? _T(MSG_OFF) : _T(MSG_ON), lcd_second_serial_set);
 #endif //HAS_SECOND_SERIAL
 
-    MENU_ITEM_SUBMENU_P(_T(MSG_BABYSTEP_Z), lcd_babystep_z);
+    if (!isPrintPaused) MENU_ITEM_SUBMENU_P(_T(MSG_BABYSTEP_Z), lcd_babystep_z);
 
 #if (LANG_MODE != 0)
 	MENU_ITEM_SUBMENU_P(_T(MSG_SELECT_LANGUAGE), lcd_language_menu);
@@ -5148,7 +5148,8 @@ static void lcd_main_menu()
     MENU_ITEM_FUNCTION_P(PSTR("power panic"), uvlo_);
 #endif //TMC2130_DEBUG
 
-    if ( babystep_allowed() )
+    // Menu is never shown when idle
+    if (babystep_allowed_strict() && (printJobOngoing() || lcd_commands_type == LcdCommands::Layer1Cal))
         MENU_ITEM_SUBMENU_P(_T(MSG_BABYSTEP_Z), lcd_babystep_z);//8
 
     if (farm_mode)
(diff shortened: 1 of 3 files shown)


Firmware/ultralcd.cpp
```diff
@@ -4525,8 +4525,7 @@ static void lcd_settings_menu()
     MENU_ITEM_TOGGLE_P(_T(MSG_RPI_PORT), (selectedSerialPort == 0) ? _T(MSG_OFF) : _T(MSG_ON), lcd_second_serial_set);
 #endif //HAS_SECOND_SERIAL
 
-	if ( babystep_allowed() )
-		MENU_ITEM_SUBMENU_P(_T(MSG_BABYSTEP_Z), lcd_babystep_z);
+    MENU_ITEM_SUBMENU_P(_T(MSG_BABYSTEP_Z), lcd_babystep_z);
```
gudnimg: In 3.12 this was not allowed during homing or mesh bed leveling. Shouldnt that still apply?

leptun: Isn't this menu inaccessible during homing and mbl anyway?

gudnimg: From looking at the code it seems to me the settings menu is accessible during homing unless `printJobOngoing()` returns `false`.

But I haven't checked this on my printer.

gudnimg: The Z adjust menu is dismissed immediately when opening it. So the current solution does not fix the menu.

I think editing `babystep_allowed()` is unavoidable
image

gudnimg: On the bright side this does cover mesh bed leveling and homing if that ever happens :)

gudnimg: Here is my attempt at patching this. Seems to work. I assume `lcd_commands_type == LcdCommands::Idle` is always true while printing.
[PR4188.patch](https://github.com/prusa3d/Prusa-Firmware/files/11401100/PR4188.patch)

3d-gussner: I like the patch but the result is that the Live adjust Z is also shown as 1st menu point
Live_adjust_Z_in_main_menu

leptun: @gudnimg @3d-gussner feel free to push directly to this branch if/when you find a solution

gudnimg:
Synced the branch with MK3. Remember to call `git pull --rebase` next time ;)

Firmware/ultralcd.cpp
gudnimg: I have a potential fix. Will push it once I have verified it tomorrow.

gudnimg:
I believe I've fixed the behavior so it's the same as on 3.12 and before. Below are some notes I wrote to test all the scenarios:

----
Live Z adjust menu tests

When idle (after bootup, print completion etc.):
* Main menu -> Hidden
* Settings menu -> Shown
* Long-press -> Should activate Z-axis move and not live Z adjust

Mesh-bed leveling:
* live Z adjust is always hidden and not accessible
* Long-press is ignored
* Settings menu is hidden

Homing:
* live Z adjust is always hidden and not accessible
* Long-press is ignored
* Settings menu is hidden

First layer calibration:
* When heating nozzle:
    * Shown in Main menu when heating
    * Long press trigger Z-lift and not live Z adjust
* Main menu -> Shown
* Settings menu -> Shown
* Z-adjust menu is shown after loading filament
* Upon 1st layer calibration completion, the Z-adjust menu is automatically dismissed when heater target temperature is set to zero.
* User can open the z-adjust menu after completing 

When printing:
* Main menu -> Shown (Only when Z-axis is below 2mm)
* Settings menu -> Shown
* Long press triggers Z-adjust menu only when Z-axis is belo 2mm, else long press is ignored.

When a print is paused:
* live Z adjust is always hidden and not accessible

gudnimg:
@leptun @3d-gussner please have another look in case I missed something.

3d-gussner (approved):
Tested and approved

PR #4191, opened 2023-05-05
PFW-1392 optimize TMC2130 configuration

gudnimg:
An idea from the MMU FW where we use a union to gather the configuration bits into one dword. Makes it a bit easier to copy all the information.

Change in memory:
Flash: -664 bytes
SRAM: +4 bytes

### DIFF
--- a/Firmware/power_panic.cpp
+++ b/Firmware/power_panic.cpp
@@ -61,10 +61,12 @@ void uvlo_() {
 
     // Minimise Z and E motor currents (Hold and Run)
 #ifdef TMC2130
-    tmc2130_set_current_h(Z_AXIS, 20);
-    tmc2130_set_current_r(Z_AXIS, 20);
-    tmc2130_set_current_h(E_AXIS, 20);
-    tmc2130_set_current_r(E_AXIS, 20);
+    currents[Z_AXIS].iHold = 20;
+    currents[Z_AXIS].iRun = 20;
+    tmc2130_setup_chopper(Z_AXIS, tmc2130_mres[Z_AXIS]);
+    currents[E_AXIS].iHold = 20;
+    currents[E_AXIS].iRun = 20;
+    tmc2130_setup_chopper(E_AXIS, tmc2130_mres[E_AXIS]);
 #endif //TMC2130
 
     if (!sd_print_saved_in_ram && !isPartialBackupAvailable)
@@ -225,8 +227,9 @@ static void uvlo_tiny() {
     disable_e0();
 
 #ifdef TMC2130
-    tmc2130_set_current_h(Z_AXIS, 20);
-    tmc2130_set_current_r(Z_AXIS, 20);
+    currents[Z_AXIS].iHold = 20;
+    currents[Z_AXIS].iRun = 20;
+    tmc2130_setup_chopper(Z_AXIS, tmc2130_mres[Z_AXIS]);
 #endif //TMC2130
 
     // Stop all heaters
--- a/Firmware/tmc2130.cpp
+++ b/Firmware/tmc2130.cpp
@@ -140,6 +140,39 @@ static ShortTimer tmc2130_overtemp_timer;
 #define TMC2130_REG_ENCM_CTRL  0x72 // 2 bits
 #define TMC2130_REG_LOST_STEPS 0x73 // 20 bits
 
+#define _GET_PWR_X      (READ(X_ENABLE_PIN) == X_ENABLE_ON)
+#define _GET_PWR_Y      (READ(Y_ENABLE_PIN) == Y_ENABLE_ON)
+#define _GET_PWR_Z      (READ(Z_ENABLE_PIN) == Z_ENABLE_ON)
+#define _GET_PWR_E      (READ(E0_ENABLE_PIN) == E_ENABLE_ON)
+
+#define _SET_PWR_X(ena) WRITE(X_ENABLE_PIN, ena?X_ENABLE_ON:!X_ENABLE_ON)
+#define _SET_PWR_Y(ena) WRITE(Y_ENABLE_PIN, ena?Y_ENABLE_ON:!Y_ENABLE_ON)
+#define _SET_PWR_Z(ena) WRITE(Z_ENABLE_PIN, ena?Z_ENABLE_ON:!Z_ENABLE_ON)
+#define _SET_PWR_E(ena) WRITE(E0_ENABLE_PIN, ena?E_ENABLE_ON:!E_ENABLE_ON)
+
+#define _GET_DIR_X      (READ(X_DIR_PIN) == INVERT_X_DIR)
+#define _GET_DIR_Y      (READ(Y_DIR_PIN) == INVERT_Y_DIR)
+#define _GET_DIR_Z      (READ(Z_DIR_PIN) == INVERT_Z_DIR)
+#define _GET_DIR_E      (READ(E0_DIR_PIN) == INVERT_E0_DIR)
+
+#define _SET_DIR_X(dir) WRITE(X_DIR_PIN, dir?INVERT_X_DIR:!INVERT_X_DIR)
+#define _SET_DIR_Y(dir) WRITE(Y_DIR_PIN, dir?INVERT_Y_DIR:!INVERT_Y_DIR)
+#define _SET_DIR_Z(dir) WRITE(Z_DIR_PIN, dir?INVERT_Z_DIR:!INVERT_Z_DIR)
+#define _SET_DIR_E(dir) WRITE(E0_DIR_PIN, dir?INVERT_E0_DIR:!INVERT_E0_DIR)
+
+#ifdef TMC2130_DEDGE_STEPPING
+static constexpr uint8_t default_dedge_bit = 1;
+#define _DO_STEP_X      TOGGLE(X_STEP_PIN)
+#define _DO_STEP_Y      TOGGLE(Y_STEP_PIN)
+#define _DO_STEP_Z      TOGGLE(Z_STEP_PIN)
+#define _DO_STEP_E      TOGGLE(E0_STEP_PIN)
+#else // !TMC2130_DEDGE_STEPPING
+static constexpr uint8_t default_dedge_bit = 0;
+#define _DO_STEP_X      { WRITE(X_STEP_PIN, !INVERT_X_STEP_PIN); TMC2130_MINIMUM_DELAY; WRITE(X_STEP_PIN, INVERT_X_STEP_PIN); }
+#define _DO_STEP_Y      { WRITE(Y_STEP_PIN, !INVERT_Y_STEP_PIN); TMC2130_MINIMUM_DELAY; WRITE(Y_STEP_PIN, INVERT_Y_STEP_PIN); }
+#define _DO_STEP_Z      { WRITE(Z_STEP_PIN, !INVERT_Z_STEP_PIN); TMC2130_MINIMUM_DELAY; WRITE(Z_STEP_PIN, INVERT_Z_STEP_PIN); }
+#define _DO_STEP_E      { WRITE(E0_STEP_PIN, !INVERT_E_STEP_PIN); TMC2130_MINIMUM_DELAY; WRITE(E0_STEP_PIN, INVERT_E_STEP_PIN); }
+#endif // TMC2130_DEDGE_STEPPING
 
 uint16_t tmc2130_rd_TSTEP(uint8_t axis);
 uint16_t tmc2130_rd_MSCNT(uint8_t axis);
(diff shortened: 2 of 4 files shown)


gudnimg:
Added a similar optimisation for PWMCONF register. Hopefully the PR can still be reviewed by looking at the two commits separately.

Firmware/tmc2130.cpp
```diff
@@ -183,33 +212,32 @@ void tmc2130_init(TMCInitParams params)
...
-            tmc2130_wr_TPWMTHRS(axis, TMC2130_TPWMTHRS_E);
-            SERIAL_ECHOLNRPGM(eMotorCurrentScalingEnabled);
-        }
+    if( ! params.enableECool ){
+        tmc2130_wr(E_AXIS, TMC2130_REG_GCONF, TMC2130_GCONF_SGSENS);
+    } else {
+        tmc2130_wr(E_AXIS, TMC2130_REG_COOLCONF, (((uint32_t)tmc2130_sg_thr[E_AXIS]) << 16));
+        tmc2130_wr(E_AXIS, TMC2130_REG_TCOOLTHRS, 0);
+        tmc2130_wr(E_AXIS, TMC2130_REG_GCONF, TMC2130_GCONF_SILENT);
+        tmc2130_wr_PWMCONF(E_AXIS, pwmconf_Ecool.dw);
```
gudnimg: Note. I removed the for-loop here so I don't need to keep an array of `pwmconf_Ecool` for all axis. It had no impact on memory usage so I assume the for-loop was optimised out by the compiler anyway.

gudnimg:
Added:

```
static_assert(sizeof(ChopConfU::S) == 4);
static_assert(sizeof(ChopConfU) == 4);
```

Just so the code is the same in both MK3 and MMU FWs.

gudnimg:
Regarding the last commit, it seems like `TMC2130_REG_COOLCONF` was also not reset correctly.

In `tmc2130_home_enter()` it's set to `(((uint32_t)tmc2130_sg_thr_home[axis]) << 16)` but then in `tmc2130_home_exit()` it's not reset to `(((uint32_t)tmc2130_sg_thr[axis]) << 16) | ((uint32_t)1 << 24));` when using Silent mode. Should be fixed now.

@leptun please let me know if this was actually intentional :) I'm just comparing the code to `tmc2130_init()`

Looking at the datasheet I don't think my change will cause any new issues (bit 24 seems to activate a filter)
image

gudnimg:
Another issue fixed: `tmc2130_setup_chopper()` is not called in `tmc2130_home_exit()` when in Silent Mode, only difference I see is the running current will be lower than expected after homing.

```
#define TMC2130_CURRENTS_R {16, 20, 35, 30}  // default running currents for all axes
#define TMC2130_CURRENTS_R_HOME {8, 10, 20, 18}  // homing running currents for all axes
```

I'll try to modify M913 to read from the registers to confirm this.

Edit:

The registers are Write only, so I'm now printing onto UART the values being written to the registers. I don't see the issue I  described 🤔 Will continue to investigate.

Firmware/tmc2130.cpp
```diff
@@ -379,56 +393,84 @@ void tmc2130_check_overtemp()
...
+			uint32_t vhighfs : 1;
+			uint32_t vhighchm : 1;
+			uint32_t sync : 4;
+			uint32_t mres : 4;
+			uint32_t intpol : 1;
+			uint32_t dedge : 1;
+			uint32_t diss2g : 1;
+			uint32_t reserved : 1;
+		} s;
+		uint32_t dw;
```
DRracer: I'd recommend adding a constexpr constructor to set all the members. It might look cumbersome, but the compiler should optimize most of the initialization code below better.

Firmware/tmc2130.cpp
```diff
@@ -379,56 +393,84 @@ void tmc2130_check_overtemp()
...
+	chopconf.s.vsense = 0;
+	chopconf.s.vhighfs = 0;
+	chopconf.s.vhighchm = 0;
+	chopconf.s.sync = 0;
+	chopconf.s.mres = mres;
 	if (axis == E_AXIS)
 	{
 #if defined(TMC2130_INTPOL_E) && (TMC2130_INTPOL_E == 0)
-        intpol = 0;
+		chopconf.s.intpol = 0;
```
DRracer: I'd recommend extracting this `intpol` code-fu into a separate constexpr function and solve the #define hell there. It looks to me, that:
`chopconf.s.intpol = 0`
most of the time, unless some defines exist

Firmware/tmc2130.cpp
```diff
@@ -379,56 +393,84 @@ void tmc2130_check_overtemp()
...
 	if (current_r <= 31)
 	{
-		tmc2130_wr_CHOPCONF(axis, toff, hstrt, hend, fd3, 0, rndtf, chm, tbl, 1, 0, 0, 0, mres, intpol, dedge, 0);
-		tmc2130_wr(axis, TMC2130_REG_IHOLD_IRUN, 0x000f0000 | ((current_r & 0x1f) << 8) | (current_h & 0x1f));
-	}
-	else
-	{
-		tmc2130_wr_CHOPCONF(axis, toff, hstrt, hend, fd3, 0, rndtf, chm, tbl, 0, 0, 0, 0, mres, intpol, dedge, 0);
-		tmc2130_wr(axis, TMC2130_REG_IHOLD_IRUN, 0x000f0000 | (((current_r >> 1) & 0x1f) << 8) | ((current_h >> 1) & 0x1f));
+		chopconf.s.vsense = 1;
```
DRracer: this logic can be also hidden in the constructor of `chopconf` ... I think we have it like that in the MMU code

DRracer (changes requested):
Generally, I'm surprised how much is the compiler unable to understand the shift operations, which have been here before. Rewriting them into unions make the code considerably shorter.

There are a few remarks about further improving the code style while doing this optimization.

Otherwise, this PR can be approved.

Firmware/tmc2130.cpp
gudnimg: I'm thinking about adding this `struct` to the MK3 firmware:

```cpp
struct MotorCurrents {
    bool vSense; ///< VSense current scaling
    uint8_t iRun; ///< Running current
    uint8_t iHold; ///< Holding current

    constexpr MotorCurrents(uint8_t ir, uint8_t ih)
        : vSense((ir < 32) ? 1 : 0)
        , iRun((ir < 32) ? ir : (ir >> 1))
        , iHold((ir < 32) ? ih : (ih >> 1)) {}
};
```
It will calculate the vsense flag for us. But since both Run and Hold currents can be updated separately at run time we need to make sure the Vsense bit is always re-calculated.

We can also drop the Run and Hold current arguments for the function `tmc2130_setup_chopper()` since the settings are already stored in SRAM globally. In other words, `void tmc2130_setup_chopper(uint8_t axis, uint8_t mres, uint8_t current_h, uint8_t current_r)` could be changed to `void tmc2130_setup_chopper(uint8_t axis, uint8_t mres)` should reduce some CPU stack allocation.

At the cost of 4 bytes of SRAM, we can save ~50 bytes of Flash by creating a global array of these structs for each axis. It also seems to make the code cleaner in general. I'm not ready to push it just yet, but it's maybe one way to resolve your comment :)

Firmware/tmc2130.cpp
gudnimg: I do agree, I'll try reduce this `ifdef` mess. It may take me some time, because I need to make sure I'm not changing any settings.

Firmware/tmc2130.cpp
gudnimg: @leptun quick question about `iHold((ir < 32) ? ih : (ih >> 1)) {}`

I wonder about the scenario where both `iHold, iRun = 31`, but what if the user suddenly changed `iRun = 33`, does `iHold` need to be updated with the >> bit shift?

leptun: Yes. vsense gets applied to both IHOLD and IRUN, so IHOLD also needs to be shifted if IRUN got shifted. It is enough to check ir<32 for both fields since always IRUN >= IHOLD.

gudnimg: `IRUN >= IHOLD` this was also something I remember you mentioned about the MMU FW. I don't think we are currently enforcing this.

leptun: Yeah, it might be a good idea to constrain that value, but I would skip that if it doesn't get optimized well.

Firmware/power_panic.cpp
```diff
@@ -61,10 +61,12 @@ void uvlo_() {
...
-    tmc2130_set_current_h(Z_AXIS, 20);
-    tmc2130_set_current_r(Z_AXIS, 20);
-    tmc2130_set_current_h(E_AXIS, 20);
-    tmc2130_set_current_r(E_AXIS, 20);
+    currents[Z_AXIS].iHold = 20;
+    currents[Z_AXIS].iRun = 20;
+    tmc2130_setup_chopper(Z_AXIS, tmc2130_mres[Z_AXIS]);
+    currents[E_AXIS].iHold = 20;
+    currents[E_AXIS].iRun = 20;
+    tmc2130_setup_chopper(E_AXIS, tmc2130_mres[E_AXIS]);
```
gudnimg, line 69: **Note:** this reduces the calls to `tmc2130_setup_chopper()` from 4 calls to only 2. Should reduce some of the SPI activity. 👍 We can change the Hold and Run currents in one go, no need to update each current setting separately.

Firmware/tmc2130.cpp
gudnimg: Done

Firmware/tmc2130.cpp
gudnimg: Done

Firmware/tmc2130.cpp
gudnimg: @DRracer I've refactored this code now, should be much better now 👍

DRracer (approved):
:+1: approved

ojrik:
Greetings. I tried this merged PR today. I have 2 MK3S - one has a 12V power supply, the other 24V, the motors have no power on both machines, and when I want to adjust the M911 and M912 the values ​​are low, if I enter example 32 for the Z axis and the E axis, the values ​​are half as small. Well thank you.

DRracer:
@ojrik thanks for your report, we'll fix that

gudnimg:
> the motors have no power on both machines

@ojrik Could you elaborate on this? Is this happening immediately after powering on the printer? It's normal for some of the motors (like E-axis) to be disabled most of the time but _all_ motors sounds abnormal.

On my MK3S all axis are off except the Z-axis after powering on the printer (mine is 24V). I can print without any issues.

ojrik:
Hello. What I found out on the MK3S+ 24V when printing, after the first layer, the e-axis starts to click and the z-axis starts to hum slightly when lifting, the motors have no power, I gently touch the x-axis and it jumps right away.

ojrik:
MK3S+ 12V motory osi z pri navrate domov hned preskakuje a bzučí. Doteraz som nemal žiadný problém zo strojom až teraz.

ojrik:
The MK3S+ 12V z-axis motors immediately jump and buzz when returning home. So far I have had no problem with the machine until now.

ojrik:
mk3s 24v defined values
#define TMC2130_CURRENTS_H {20, 24, 35, 35}
#define TMC2130_CURRENTS_R {20, 24, 35, 35}
#define TMC2130_CURRENTS_R_HOME {12, 12, 20, 25} ,
 values ​​of M913
Send: M913
Recv: tmc2130_print_currents()
Recv: H R
Recv: X 20 20
Recv: Y 24 24
Recv: Z 17 17
Recv: E 17 17
Recv: ok

ojrik:
if I enter a change to H, R will change to this:
Send: M911 Z35 E35
Recv: ok
Send: M913
Recv: tmc2130_print_currents()
Recv: H R
Recv: X 20 12
Recv: Y 24 12
Recv: Z 35 20
Recv: E 35 17
Recv: ok

ojrik:
if I enter a change to R, then H changes to this:
Send: M912 X20 Y24 Z35 E35
Recv: ok
Send: M913
Recv: tmc2130_print_currents()
Recv: H R
Recv: X 20 20
Recv: Y 24 24
Recv: Z 17 17
Recv: E 17 17
Recv: ok

ojrik:
It somehow doesn't work as it should for me. I'm a bit disappointed. But I like your improvements. Well thank you.

gudnimg:
@ojrik I’ll test some more on my printer over the next week. The changes are quite many so maybe I missed some details.

The results you showed with Gcodes 912 vs 911 seems wrong but I need to look at the code to see whats going on. 

Thanks for testing the changes 👍

gudnimg:
@ojrik You're sure your MK3S+ has 12V? I don't think Prusa Research sells MK3S+ units with 12V as changing 24V to 12V will cause problems on its own. We saw this before with the MMU when used on MK2.5 (which uses 12V) and it seems the current settings behave differently.

ojrik:
> @ojrik You're sure your MK3S+ has 12V? I don't think Prusa Research sells MK3S+ units with 12V as changing 24V to 12V will cause problems on its own. We saw this before with the MMU when used on MK2.5 (which uses 12V) and it seems the current settings behave differently.

I had a mk2.5s. I ordered a base plate and a 30x30 profile frame from China. So I made a 12v clone. But it worked for me until now and after this change. And MMU3 also works. Only this tmc2130 change doesn't work for me, not even the last fix.

ojrik:
i printed this on mk3s 12v MMU3 an hour ago. On the mk3s+ 24v I am printing new parts on the MMU3. There is also FW without tmc2130 mods.

PR #4192, opened 2023-05-06
PFW-1506 Load to nozzle: Fix too short purge distance

gudnimg:
In the test below I had orange filament before and then loaded yellow filament.
|  BEFORE  |  AFTER  |
|-----------|----------|
| IMG_3274 | IMG_3275 |

Change in memory:
Flash: +8 bytes
SRAM: 0 bytes


3d-gussner (approved):
Tested on MK404 and it extrudes extra 26ish mm of filament
Before E70
After E96
PFW-1506_E70
PFW-1506_E96

PR #4197, opened 2023-05-07
Updated the French language

Ilovemyhous:
Some French sentences/words were updated, to have a better meaning.
Changed only `lang/po/Firmware_fr.po`.

### DIFF
--- a/lang/po/Firmware_fr.po
+++ b/lang/po/Firmware_fr.po
@@ -306,12 +306,12 @@ msgstr "Verific. axe X"
 #: ../../Firmware/messages.cpp:22 ../../Firmware/ultralcd.cpp:6045
 #: ../../Firmware/ultralcd.cpp:7148
 msgid "Checking Y axis"
-msgstr "Verific. axe Y"
+msgstr "Verification axe Y"
 
 #. MSG_SELFTEST_CHECK_Z c=20
 #: ../../Firmware/ultralcd.cpp:7149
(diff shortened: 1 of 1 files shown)


gudnimg:
It seems most of the files in the repo are changed for some reason.
image

Ilovemyhous:
I'm trying to fix that ASAP, not sure why it edited everything.
Sorry for the inconvenience.

lang/po/Firmware_fr.po
```diff
@@ -2359,8 +2360,8 @@ msgstr "Correct-X"
 
 #. MSG_XFLASH c=18
 #: ../../Firmware/ultralcd.cpp:5462
-msgid "XFLASH init"
-msgstr "Init XFLASH"
+msgid "XFLASH init."
+msgstr "Init. XFLASH"
```
gudnimg, line 2364: @3d-gussner should we consider removing this string from translations? The string is essentially same in all current translations.

gudnimg: Actually on a second look, this string is no longer in the firmware 🤔 We need to re-sync the po files.

leptun: I removed it recently. It was from an unused test menu

3d-gussner:
@Ilovemyhous Thanks for the PR. Forwarded a review request to internal translators.

Ilovemyhous:
See #4441.

PR #4200, opened 2023-05-13
Fix stepper inactivity timeout

leptun:
Wrong order of if statements for `max_inactive_time` expired the `previous_millis_cmd` prematurely, so the stepper inactivity timeout would never end up being checked.


gudnimg (approved):

3d-gussner (approved):
Tested on MK404
Before the X Y axis stay enabled all the time after a `G28W`
   
XY_stay_enabled

After the timeout works
XY_timeout_enable

PR #4201, opened 2023-05-13
PFW-1508 Fix issue where Loading Test always starts with unload

gudnimg:
The condition where a unload is run on a toolchange should only be done when the FINDA detects filament (here we can safely assume the FINDA is calibrated correctly).
Change in memory:
Flash: +8 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/mmu2.cpp
+++ b/Firmware/mmu2.cpp
@@ -366,9 +366,7 @@ bool MMU2::tool_change(uint8_t slot) {
         return false;
 
     if (slot != extruder) {
-        if (/*FindaDetectsFilament()*/
-            /*!IS_SD_PRINTING && !usb_timer.running()*/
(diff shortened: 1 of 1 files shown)


DRracer (commented):
What if FINDA is faulty and filament is present in the extruder? The unconditional unload has been done exactly for this reason, even though it looks stupid.

gudnimg:
> What if FINDA is faulty and filament is present in the extruder? The unconditional unload has been done exactly for this reason, even though it looks stupid.

If another slot is selected I assume the selector will snap/break the filament. But due to the faulty FINDA the load will fail anyway. But it should have triggered correctly to load filament to the extruder before🤔

 I don’t see any other way to fix PFW-1508. Should we consider to close that ticket?

PR #4203, opened 2023-05-13
🚧 Optimise LCD update RAII

gudnimg:
Optimise and use the RAII for LCD update enable flag in more places in the firmware. This does increase memory usage somewhat but I think this is beneficial to add some "structure" to the LCD code.

Using `__attribute__((always_inline))` becomes critical when `LCDUpdateEnableRAII` is applied to the SD card menu items. For some reason, specifically the flash memory increases there significantly (I'm talking about ~100B)

Change in memory:
Flash: -14 bytes


gudnimg:
I'm a bit suspcious of `LCDUpdateEnableRAII()` may not call the deconstructor. Will verify tomorrow if a variable name is really needed. It increases code size so it would make sense.

gudnimg:
Synced with MK3 branch (~634 commits)

PR #4204, opened 2023-05-13
optimisation: always inline `ReportingRAII`

gudnimg:
Tested on MK3S+ using the LCD.

Change in memory:
Flash: -80 bytes
SRAM: 0 bytes


leptun (approved):
lgtm

DRracer (approved):

PR #4206, opened 2023-05-14
optimisation: `menu_item == menu_line` is always true in `menu_item_sddir()` and `menu_item_sdfile()`

gudnimg:
The if statement is checked in `lcd_sdcard_menu()` so checking it again in `menu_item_sddir()` and `menu_item_sdfile()` is
redundant since it's always going to be true.

Tested on MK3S+

Change in memory:
Flash: -22 bytes
SRAM: 0 bytes


3d-gussner (approved):
Tested on MK404 and SD card menu works as before.

PR #4207, opened 2023-05-14
`crashdet_cancel()` doesnt cleanup all variables when USB printing

### LINKED ISSUE #4324: [BUG] Mk3S+ with MMU3 not re-heating after safety timer
**Printer type**
Mk3S+
**Printer firmware version**
3.13.0

**MMU upgrade**
MMU3
**MMU upgrade firmware version** 
3.0.0

**SD card or USB/Octoprint** 
PrusaLink 0.7.0

**Describe the bug**
 The printer detected a crash, repeatedly, so it paused. After confirming I didn't want to continue with the print, the MMU tried to pull the filament out of the nozzle, as with any other end of the print, however, the nozzle was cold at this point (30 min safety timer) yet the printer ignored that and never reheated the nozzle so the MMU triggered an error - filament in Finda.
 The only way out of this was restarting the printer as the LCD menu is not accessible through the MMU's error message.
  
**To Reproduce**
  Trigger the 30-minute safety timer where the filament is still in the nozzle (so not after filament runout for example).
  
**Expected behaviour**
The printer should re-heat the nozzle before any following action.
  
**G-code**
The issue is not G-code relevant.

gudnimg:
This function doesn't look correct to me 🤔 I would think that this should behave similarly as when stopping the print via the LCD. This PR is my proposed improvement, un-tested at the moment.

Changed `UnconditionalStop()` to not close the SD card file if we're using Octoprint. Then there shouldnt be any file open.

Some of the variables which were not reset:
`isPrintPaused`
`pause_time`
`saved_start_position`
`saved_printing_type`

It seems like the bed heater could also be left enabled?

Fixes https://github.com/prusa3d/Prusa-Firmware/issues/4324

Change in memory:
Flash: -28 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -635,16 +635,9 @@ void crashdet_recover()
   if (lcd_crash_detect_enabled()) tmc2130_sg_stop_on_crash = true;
 }
 
-void crashdet_cancel()
-{
-	saved_printing = false;
-	tmc2130_sg_stop_on_crash = true;
-	if (saved_printing_type == PowerPanic::PRINT_TYPE_SD) {
-		print_stop();
-	}else if(saved_printing_type == PowerPanic::PRINT_TYPE_USB){
-		SERIAL_ECHOLNRPGM(MSG_OCTOPRINT_CANCEL); //for Octoprint: works the same as clicking "Abort" button in Octoprint GUI
-		cmdqueue_reset();
-	}
+/// Crash detection cancels the print
+void crashdet_cancel() {
+    print_stop();
 }
 
 #endif //TMC2130
@@ -9693,8 +9686,10 @@ void UnconditionalStop()
     cmdqueue_serial_disabled = false;
 
     // Reset the sd status
-    card.sdprinting = false;
-    card.closefile();
+    if (card.sdprinting) {
+        card.sdprinting = false;
+        card.closefile();
+    }
 
     st_reset_timer();
     CRITICAL_SECTION_END;
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -9685,12 +9685,6 @@ void UnconditionalStop()
     cmdqueue_reset();
     cmdqueue_serial_disabled = false;
 
-    // Reset the sd status
-    if (card.sdprinting) {
-        card.sdprinting = false;
-        card.closefile();
-    }
-
     st_reset_timer();
     CRITICAL_SECTION_END;
 }
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -631,12 +631,16 @@ void crashdet_detected(uint8_t mask)
 
 void crashdet_recover()
 {
-  if (!print_job_timer.isPaused()) crashdet_restore_print_and_continue();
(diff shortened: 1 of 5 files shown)


wavexx:
I generally agree here, I think this has always been overlooked recently

Firmware/Marlin_main.cpp
```diff
@@ -640,16 +640,9 @@ void crashdet_recover()
...
-	tmc2130_sg_stop_on_crash = true;
-	if (saved_printing_type == PRINTING_TYPE_SD) {
-		print_stop();
-	}else if(saved_printing_type == PRINTING_TYPE_USB){
-		SERIAL_ECHOLNRPGM(MSG_OCTOPRINT_CANCEL); //for Octoprint: works the same as clicking "Abort" button in Octoprint GUI
-		cmdqueue_reset();
-	}
+/// Crash detection cancels the print
+void crashdet_cancel() {
+    print_stop();
```
wavexx, line 644: The only difference I see here is that before we would send MSG_OCTOPRINT_CANCEL, but now we first call UnconditionalStop(), which clears card.sdprinting, so the condition in print_stop at ultralcd.cpp:5582 is always false...

this seems like a pre-existing bug?

wavexx (dismissed):
I agree here, I think this has always been overlooked

Maybe we can also call directly print_stop() when handling the CRASHDET_CANCEL command in Marlin_main.cpp:4035.

Firmware/Marlin_main.cpp
gudnimg: I just looked at it now and you're right!

image
image

gudnimg: I added another commit to resolve this hopefully.

gudnimg:
I pushed another commit which optimizes setting the crash detection on or off.

Firmware/Marlin_main.cpp
3d-gussner: @gudnimg @wavexx is this solved? If so please use Resolve conversation to make it easier to merge.

gudnimg:
Adding 🚧 to the title as this PR is not tested

Firmware/Marlin_main.cpp
```diff
@@ -631,20 +631,17 @@ void crashdet_detected(uint8_t mask)
 
 void crashdet_recover()
 {
-  if (!print_job_timer.isPaused()) crashdet_restore_print_and_continue();
-  if (lcd_crash_detect_enabled()) tmc2130_sg_stop_on_crash = true;
+	if (!print_job_timer.isPaused()) crashdet_restore_print_and_continue();
```
3d-gussner, line 634: Change the line 620 `uint8_t yesno = lcd_show_yes_no_and_wait(false);` to `uint8_t yesno = lcd_show_yes_no_and_wait(false, LCD_LEFT_BUTTON_CHOICE);`
so RECOVER CRASH default is YES.
I think an accidental knob click should try to recover the crash instead of killing it.

3d-gussner (changes requested):
Tested on MK404
Started USB print
- [X] Triggered crash recovers
- [X] Triggered crash stop, stops the print correctly on printer and host
- [X] Crash det. [OFF] does work as it ignores the crashes

Approved only change I would like to see is that Recover crash default is YES

3d-gussner (approved):

PR #4209, opened 2023-05-14
Fix `t_fan_rising_edge` compiler warning on MK25 builds

gudnimg:
This warning likely came when we made the variable `static` in 3.13.

* Add ifdef around `t_fan_rising_edge`
* Moved the relevant fan code to `fancheck.cpp` as cleanup

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -385,10 +385,6 @@ void uvlo_tiny();
 void recover_print(uint8_t automatic); 
 void setup_uvlo_interrupt();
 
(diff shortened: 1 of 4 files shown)


gudnimg:
@wavexx Do you agree with moving the code to `fancheck.cpp`? If not we can drop that commit.

DRracer (approved):

DRracer:
Moving the `setup_fan_interrupt` is IMO correct, it semantically belongs there. Approved

wavexx (approved):
Fully agree, much cleaner this way

PR #4212, opened 2023-05-17
Leave user's MMU-loaded filament loaded after first layer cal

mmalecki:
Presently, if user runs the first layer calibration with the filament that's been loaded into the nozzle by MMU, this filament will be unloaded after the calibration is complete.

This behavior seems unexpected from user experience point of view: if you load a filament into the nozzle and proceed to first layer calibration, you did so for a reason and probably don't want it unloaded after the first try (just like you wouldn't want it unloaded if you did so without the MMU).

Any other filament that's not been previously loaded into the nozzle is fair game to unload (as it was before).

I was, sadly, unable to test this. I successfully flashed this firmware onto my printer, but I was unable to find the applicable MMU firmware version/build process. (There used to be a `MMU2_dev` branch here that I'd hoped would contain the latest developments but it seems gone?)

### DIFF
--- a/Firmware/first_lay_cal.cpp
+++ b/Firmware/first_lay_cal.cpp
@@ -66,7 +66,7 @@ void lay1cal_wait_preheat()
 //! @brief Load filament
 //! @param cmd_buffer character buffer needed to format gcodes
 //! @param filament filament to use (applies for MMU only)
-//! @returns true if extra purge distance is needed in case of MMU prints (after a toolchange), otherwise false
+//! @returns true if filament has been changed (therefore an extra purge distance is needed in case of MMU prints), otherwise false
 bool lay1cal_load_filament(uint8_t filament)
 {
     if (MMU2::mmu2.Enabled())
@@ -225,7 +225,7 @@ void lay1cal_square(uint8_t step, float layer_height, float extrusion_width)
     }
 }
 
-void lay1cal_finish(bool mmu_enabled)
+void lay1cal_finish(bool unload_filament)
 {
     static const char cmd_cal_finish_1[] PROGMEM = "G1 E-0.075 F2100"; //retract
     static const char cmd_cal_finish_2[] PROGMEM = "M104 S0"; // turn off temperature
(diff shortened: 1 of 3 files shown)


DRracer:
@mmalecki Please note that a printer equipped with an MMU shall behave like one machine. That implies after each print the filament used gets unloaded. Also, any print is started without filament loaded and the loading is completely handled by the printer+MMU machine.

Therefore we unload filament after doing first layer calibration.

Regarding the latest development of the MMU - please hold on, we are polishing the code, it is coming really soon.

mmalecki:
I understand the semantics here and I'm a fan of it for multi-filament handling, however, doesn't "Load into nozzle" imply the user intending to switch to a "single filament mode", where, for example, they can run GCode sliced for a MK3s+ with no MMU?
We don't unload the filament after the print is finished in this case, and IMO, I don't think we should do it after the first layer calibration either.

Just to reiterate: this only prevents the unloading of filament explicitly loaded into nozzle by the user beforehand, if the user picks it for first layer cal. It will unload any other filament that's been picked for first layer cal and loaded by MMU. It uses the same logic as the purge line length determination.

> Regarding the latest development of the MMU - please hold on, we are polishing the code, it is coming really soon.

Gotcha, looking forward to it!

Prusa-Support:
Thank you for creating this pull request.

Based on our extensive tests, leaving the filament loaded into the nozzle at the end of a print may frequently cause problems that require major manual user interaction. According to our tests, the first layer calibration print is no exception.

Theoretically, the only truly safe condition for the printer and the MMU module to smoothly operate together is if the filament is always withdrawn: ideally, the filament should never be loaded into the nozzle unless the printer is actively busy in printing/extruding operations.
Otherwise, we would almost intentionally expose the user to known rather problematic scenarios.

As a result, our developers couldn't merge this pull request and sadly we have no option but to close it.
Thank you for your effort and for your understanding.

Michele Moramarco
Prusa Research

Ro3Deee:
I confirm, from my experience, that leaving the filament loaded causes issues. Better to withdraw it.

PR #4214, opened 2023-05-19
Fix IRsensor with MMU

3d-gussner:
saves 38bytes

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -5884,9 +5884,6 @@ bool lcd_selftest()
 		if (!MMU2::mmu2.Enabled()) {
 			lcd_detect_IRsensor();
 		}
-		else {
-			fsensor.setSensorRevision(IR_sensor_analog::SensorRevision::_Old, true);
-		}
 	}
 #endif
 	lcd_wait_for_cool_down();
@@ -6669,6 +6666,7 @@ static bool lcd_selftest_fsensor(void)
 //! @retval false failed
 static bool selftest_irsensor()
 {
+    set_extrude_min_temp(0);
     class TempBackup
     {
     public:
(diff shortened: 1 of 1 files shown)


Firmware/ultralcd.cpp
```diff
@@ -6679,13 +6677,10 @@ static bool selftest_irsensor()
...
-        progress = lcd_selftest_screen(TestScreen::Fsensor, 0, 1, true, 0);
-    }
-    progress = lcd_selftest_screen(TestScreen::Fsensor, progress, 1, true, 0);
-    MMU2::mmu2.unload();
+    progress = lcd_selftest_screen(TestScreen::Fsensor, 0, 1, true, 0);
+    MMU2::mmu2.tool_change(0); //Allow cold extrusion only loads to the gears not nozzle
+    MMU2::mmu2.unload(); //Unload filament
+    set_extrude_min_temp(EXTRUDE_MINTEMP);
 
     for(uint_least8_t i = 0; i < 200; ++i)
```
gudnimg: I feel the code below should be deleted as it is part of the old ancient MMU2 test.

Firmware/ultralcd.cpp
```diff
@@ -6669,6 +6666,7 @@ static bool lcd_selftest_fsensor(void)
 //! @retval false failed
 static bool selftest_irsensor()
 {
+    set_extrude_min_temp(0);
     class TempBackup
```
gudnimg: Please remove this class. We are no longer heating the extruder to some temperature.

gudnimg (changes requested):
Above are two comments. I feel we should delete the old code since we should really be re-writing the test for the new MMU implementation. The old code is for MMU FW 1.0.6 which we don't support anymore.


Changing `selftest_irsensor()` to this saves 68 bytes:

image

gudnimg:
Here's an easy way to ask the user to select which slot to use for the test:

LCD updates are disabled so the slot selection menu doesn't timeout 👍 

```cpp
static bool selftest_irsensor()
{
    // Ask user which slot to load filament from
    uint8_t slot = choose_menu_P(_T(MSG_SELECT_FILAMENT), _T(MSG_FILAMENT));

    // Render self-test screen
    lcd_selftest_screen(TestScreen::Fsensor, 0, 1, true, 0);

    // Run self-test
    set_extrude_min_temp(0);
    MMU2::mmu2.tool_change(slot);
    MMU2::mmu2.unload(); //Unload filament
    set_extrude_min_temp(EXTRUDE_MINTEMP);
    return true;
}
```

Tested on MK3S+

3d-gussner:
@gudnimg Please review the PR again. Thanks

gudnimg (approved):
Looks good to me👍

PR #4217, opened 2023-05-20
`errorCode` should reflect the state of the MMU

gudnimg:
If the protocol reports an error and assigns `errorCode` a value, the value is not reset to `OK` when the error is actually resolved.

If the MMU is not in an error state I would expect `errorCode` to reflect that since we use it to check what is the "current" error code being reported (if any).

This fixes `MMUCurrentErrorCode()`

**Steps to reproduce:**
1. Add a serial print to log the value from `MMUCurrentErrorCode()` ---> `MYSERIAL.println((uint16_t)MMUCurrentErrorCode());`
2. Eject a filament
3. Notice the value is now 32780
4. Click Done on the LCD
5. Notice the value is still 32780, even though the MMU is no longer reporting the error code.

Change in memory:
Flash: +12 bytes
SRAM: 0 bytes


DRracer:
Actually, you are right, I had to tweak the protocol in the Buddy FW in a similar way.

DRracer (approved):

PR #4219, opened 2023-05-21
Limit scope of `is_mmu_error_monitor_active`

gudnimg:
Move the variable to `mmu2_reporting.cpp` and make it `static`

Change in memory:
Flash: -6 bytes
SRAM: 0 bytes


DRracer (approved):
A little surprising, but correct IMO. I like the fact that the `is_mmu_error_monitor_active` has been removed from the MMU2 class where it semantially didn't belong.

PR #4226, opened 2023-05-25
Add E3D Revo variant

3d-gussner:


arekm:
There will be separate hex for revo users provided, not runtime setting?

darraghbr:
So does this mean that there will be a dedicated E3D Revo variant of the firmware? We will not have to set any of the variables for the Thermal Model manually through a serial connection? Sounds like a good implementation!

DRracer (approved):

DRracer:
@arekm yes, there will be a separate build.

Firmware/variants/1_75mm_MK3-EINSy10a-E3Dv6full.h
```diff
@@ -326,7 +331,11 @@
 #define  DEFAULT_Kp 21.70
 #define  DEFAULT_Ki 1.60
 #define  DEFAULT_Kd 73.76
-#else
+#elif HOTEND == 2 //E3D REVO PID values
```
vintagepc, line 334: Please don't do this and add clutter to the files for defines that are never used in this context.

It just adds either more churn when every single file needs updating or it will get out of sync and cause even more headaches..

No #define hotend, just set the constants.

PR #4228, opened 2023-05-26
Speed up acceleration limit checking

gudnimg:
This is how Marlin 2.1 implements it. I'll run some prints over the weekend, not sure how exactly to test this though, will think about it 🤔 

If the block acceleration is less than `max_acceleration_steps_per_s2`, then the firmware will no longer calculate a division on each respective axis.

The for-loop was essential to save memory, otherwise these changes would _add_ 100 bytes. I changed `dda_isteps_t steps_x, steps_y, steps_z, steps_e` to an array so it is easier to loop through.

Change in memory:
Flash: -314 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/planner.h
+++ b/Firmware/planner.h
@@ -71,7 +71,7 @@ union dda_usteps_t
 typedef struct {
   // Fields used by the bresenham algorithm for tracing the line
   // steps_x.y,z, step_event_count, acceleration_rate, direction_bits and active_extruder are set by plan_buffer_line().
-  dda_isteps_t steps[NUM_AXIS];  // Step count along each axis
+  dda_isteps_t steps[NUM_AXIS];             // Step count along each axis
   dda_usteps_t step_event_count;            // The number of step events required to complete this block
   uint32_t acceleration_rate;               // The acceleration rate used for acceleration calculation
   unsigned char direction_bits;             // The direction bit set for this block (refers to *_DIRECTION_BIT in config.h)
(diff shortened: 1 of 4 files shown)


gudnimg:
Synced with MK3 branch. Current memory usage (MK3S+ Multilang build): 
image

gudnimg:
For what its worth I printed one B1 battledroid using these changes. This doesnt ofcourse verify these changes but shows it’s safe to use. 
IMG_3364

gudnimg:
Removing the 🚧 from the title as this PR is safe to review now.

Firmware/planner.h
```diff
@@ -71,7 +71,7 @@ union dda_usteps_t
 typedef struct {
   // Fields used by the bresenham algorithm for tracing the line
   // steps_x.y,z, step_event_count, acceleration_rate, direction_bits and active_extruder are set by plan_buffer_line().
-  dda_isteps_t steps_x, steps_y, steps_z, steps_e;  // Step count along each axis
+  dda_isteps_t steps[NUM_AXIS];  // Step count along each axis
```
wavexx: Maybe align the comment here to keep it tidy

wavexx (approved):
Changing steps_? to an array is also much nicer!

Firmware/planner.h
gudnimg: Done

gudnimg:
@DRracer Can we consider merging this into 3.13.1?

DRracer (approved):

PR #4235, opened 2023-06-11
Log event when planner aborts individual moves

gudnimg:
It may be useful to see if any moves are lost. Maybe this can explain the lost Z-move in https://github.com/prusa3d/Prusa-Firmware/issues/4105 ?

Change in memory:
Flash: +32 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/planner.cpp
+++ b/Firmware/planner.cpp
@@ -718,6 +718,8 @@ void plan_buffer_line(float x, float y, float z, const float &e, float feed_rate
 #endif /* PLANNER_DIAGNOSTICS */
   if(planner_aborted) {
       // avoid planning the block early if aborted
+      SERIAL_ECHO_START;
+      SERIAL_ECHOLNRPGM(_n("Move aborted"));
       return;
   }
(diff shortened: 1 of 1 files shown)


Firmware/planner.cpp
```diff
@@ -718,6 +718,8 @@ void plan_buffer_line(float x, float y, float z, const float &e, float feed_rate
 #endif /* PLANNER_DIAGNOSTICS */
   if(planner_aborted) {
       // avoid planning the block early if aborted
+      SERIAL_ECHO_START;
+      SERIAL_ECHOLNRPGM(_n("Planner aborted"));
```
leptun: I find the message a bit confusing. I would expect this message when the planner is discarded, not when new moves are dropped due to a plan abort event. I would rename this to "Move aborted".

gudnimg: I agree, it's done.

leptun (approved):

DRracer (approved):

PR #4243, opened 2023-06-24
PFW-1510: Improve power panic when print is already saved in RAM

### LINKED ISSUE #905: Enable Power Panic in a paused print
Hi, I paused a print for a minute of two and then had a blackout for approx 3 minutes and after the blackout the printer turned on normally and forgot the past print so I lost the paused print

### Behavior
1. Start a print
2. Pause the print
3. Disconnect the printer, or suffer a blackout
4. Reconnect the printer
   - Results:
     - The printer turns on normally, forgetting the paused print

   - Expected Results
      - The printer will turn on, and display a message of the paused and unfinished print
      - Or, if the blackout was only a couple of seconds only wait to the user to resume the paused print, if the bed is already cool prompt the same message of a common long blackout.

Is this a new feature request?
Yes 
or maybe a bug I dont know

Thanks :)

### LINKED ISSUE #1695: Bug/Feature? - Resume after pause and power off like in power outage
When power goes off, printer will resume printing.

When I manually pause print and then power goes off, it DOES NOT resume. 

I wanted to move printer to another room. I know about power outage function so I didn't worry about my print. I paused printing (to be more gentle to my printer), powered off, moved printer, plugged in, but no resume poped up. Whole print was lost.

So now I know, that if I want to move printer, i can't pause print, I have to switch off power during print and them move it. It feels very illogical to do it this way, but seems to be only way possible today. Could you please fix this?

### LINKED ISSUE #1765: Bug - Print does not recover from power outage if paused

### LINKED ISSUE #1859: Recovering from power failure, extruder not heating
Latest installed as of this writing (3.7.1) in i3 MK3S with MMU2S. I pulled the plug as I needed the MMU2 loader to get out of the way of the filament as there is no way to move it left/right when filament fails to load (another issue). After restoring power it asked to resume the print, answered yes, bed heated but extruder never heated, print was unable to resume, I had to kill it I think by pulling power again.

### LINKED ISSUE #2001: [3.7.1] No recovery after Power outage if printer was in a filament change state
There was a power outage while the printer was waiting for a filament change (ran out of filament), there was no recovery option, the printer was sitting there Idle.

MK3S on 3.7.1 firmware printing from the SD card I think.

PS: Might do the same if on a M600?

### LINKED ISSUE #2013: Problems after Powerpanic with MMU2
I dicovered the following behavior (repeatable) at my MK3S with MMU2S:

If the power is cut off while there is a loading error at the MMU it is not possible to restart the print caused by two bugs:
1. The printer ask (after Power on) to resume printing, if confirmed he heats up the bed, but then nothing more happens. even some minutes after the bed reaches the set value the nozzle set temperature is still at 0°C. That means that the printer doesent set the temperature of the nozzle. As a workaround it helps to go to preheat or temp. Settings and select the temperature manually. When then tho nozzle hits the set point the printer try to resume the print and leds to the second issue:
2. After the printer reaches temperatures he try to load the filament (what failed just before the power has gone). The load of the filament had success and he moves on to the purge tower.
And here comes the main problem. The printer is not on the last printing hight because if there is a MMU load fail the printer rises the head some mm and then try to recover on this Z hight.
I think the code in the Powerpanic feature stores the last printing hight but not the "raised some mm because of MMU load error" height. After return of power the printer doesnt have the right Z height to successfull recover the print.

### LINKED ISSUE #2249: Power panic issue
When power panic occurs during printer pause (for example if there is an issue with MMU, the printer pauses and asks for user intervention), then after power is restored the following occurs:

1) hotend does not heat (can be solved manually by setting "load to nozzle" - when nozzle starts heating, it continues even after pressing cancel on the "load to nozzle" screen)
2) Printer considers Z axis to be on the last layer printed, whilest it is in fact higher (because when pausing for user intervention, z is lifted slightly and x is moved to 0). This causes print to fail, because even though X and Y are reset correctly, Z is too high when printing resumes.

### LINKED ISSUE #2325: Power Panic not working when print is paused
In the latest version when I pause a print and I unplug the printer from the wall, the power panic doesn't kick in. It wouldn't usually need to be in use when the printer is paused, but what would happen if you were to trip over the power cord or experience a power outage after you paused it. 

A simple fix *(from my point of view)* to this would be to save the current positions when the user tells the printer to pause. Then, if it were to become unplugged while paused, it would take that position and start from there when it turns back on. Once that all happens, the printer would continue as if a normal power panic just occurred.

### LINKED ISSUE #3241: HOTEND not preheating after power OFF recovery
the problem occurs on  MK3 printers (so far I have these ones only), the latest version of sw.

 How to reproduce the problem:
    1. Bring printing to a stage (by Gcode): change filament.
    2.  Let the printer be until it changes to "hotend down due to safety timmer" 
    3.  Simulate the electricity blackout
    4. confirm to recover the printing 
    
  After those steps, only the bed is preheated .. the hotend stays at room temperature.

gudnimg:
Change the power panic code such that it will also work when a print is paused or the printer is sitting in an MMU error screen. Basically all scenarios which have called `stop_and_save_print_to_ram()` and have parked the extruder. The parking move must not overwrite what has already been written to Flash/RAM.

The following scenarios now recover the print job after a power outage (previously the print would be lost or the printer would try printing in mid-air):
* Power outage during a paused print
* Power outage during an MMU error
* Power outage during filament runout event (when waiting in M600)
    * Both FINDA and Fsensor runouts
 * Power outage during the fullscreen prompt after a crash detection (the screen which asks if you want to resume the print)

Fixes https://github.com/prusa3d/Prusa-Firmware/issues/3241
Fixes https://github.com/prusa3d/Prusa-Firmware/issues/2325
Fixes https://github.com/prusa3d/Prusa-Firmware/issues/2249
Fixes https://github.com/prusa3d/Prusa-Firmware/issues/2013
Fixes https://github.com/prusa3d/Prusa-Firmware/issues/2001
Fixes https://github.com/prusa3d/Prusa-Firmware/issues/1859
Fixes https://github.com/prusa3d/Prusa-Firmware/issues/1765
Fixes https://github.com/prusa3d/Prusa-Firmware/issues/1695
Fixes https://github.com/prusa3d/Prusa-Firmware/issues/905

Change in memory:
Flash: -612 bytes
SRAM: -4 bytes


Firmware/Marlin.h
```diff
@@ -380,23 +382,25 @@ float temp_compensation_pinda_thermistor_offset(float temperature_pinda);
...
 extern void print_physical_coordinates();
 extern void print_mesh_bed_leveling_table();
 
+void save_print_file_state();
+void restore_print_file_state();
+void save_planner_global_state();
 extern void stop_and_save_print_to_ram(float z_move, float e_move);
 void restore_extruder_temperature_from_ram();
 extern void restore_print_from_ram_and_continue(float e_move);
 extern void cancel_saved_printing();
```
gudnimg: I wonder if we should create a separate file for these functions as well or rename `power_panic.cpp` to something else to fit this code? 🤔

gudnimg:
The PR is working well to resolve the issue with paused prints. Now my printer recovers the print :) 

Only thing left is to run a lot of tests to make sure everything which was working before still works.

The following scenarios should now recover a print after a power outage (there may be more):

- [x] Power outage during a paused print
- [x] Power outage during an MMU error
- [x] Power outage during filament runout event (when waiting in M600)
    - Both FINDA and Fsensor runouts

gudnimg:
Marking as ready for review. I'll try to spend the rest of my time this weekend on testing and verification. Leaving the 🚧 in the title until testing is done.

Firmware/power_panic.cpp
```diff
@@ -0,0 +1,490 @@
...
+    // are in action.
+    planner_abort_hard();
+
+    // Store the print logical Z position, which we need to recover (a slight error here would be
+    // recovered on the next Gcode instruction, while a physical location error would not)
+    float logical_z = saved_pos[Z_AXIS];
+    if(mbl_was_active) {
+        // Mesh bed leveling was being actively applied to the Z-position. Revert the
+        // mesh bed leveling offset value.
+        logical_z -= mbl.get_z(st_get_position_mm(X_AXIS), st_get_position_mm(Y_AXIS));
```
gudnimg: I suspect this should be `mbl.get_z(saved_pos[X_AXIS], saved_pos[Y_AXIS])` since if the extruder had been parked prior to the power panic, the X and Y machine positions are not correct (i.e. they are not equal to the saved print position)

3d-gussner:
Tested on real MK3S+MMU
M155 S1 C5
1. SD print :heavy_check_mark: 
1.a. feedrate 100% :heavy_check_mark:
1.b. feedrate 90% :heavy_check_mark:
1.c. wait hotend/bed cool down to confirm :heavy_check_mark:
2. SD print paused :heavy_check_mark:
3. SD print M600 :heavy_check_mark:
4. MMU SD print :heavy_check_mark:
5. MMU SD print paused :heavy_check_mark:
6. MMU SD print during tool change :heavy_check_mark:
7. MMU SD print Error screen hot :heavy_check_mark:
8. MMU SD print Error screen cold / timed-out :heavy_check_mark:

Normally I would approve this but I want to wait for Prusa testing approval.

gudnimg:
@DRracer @3d-gussner [5d19198](https://github.com/prusa3d/Prusa-Firmware/pull/4243/commits/5d191986081cd1cac9a1a8aa3975257933302d9f) fixes the "extrusion" issue :) It was a regression in my PR and was working fine before.

Firmware/power_panic.cpp
```diff
@@ -163,10 +163,8 @@ void uvlo_() {
         eeprom_update_word((uint16_t*)(EEPROM_UVLO_MESH_BED_LEVELING_FULL +2*mesh_point), *reinterpret_cast<uint16_t*>(&v));
     }
 
-    // Write the _final_ Z position and motor microstep counter (unused).
+    // Write the _final_ Z position
     eeprom_update_float((float*)EEPROM_UVLO_TINY_CURRENT_POSITION_Z, current_position[Z_AXIS]);
-    z_microsteps = tmc2130_rd_MSCNT(Z_AXIS);
-    eeprom_update_word((uint16_t*)(EEPROM_UVLO_Z_MICROSTEPS), z_microsteps);
```
DRracer: Are you sure this part was never used? In the past, there were issues with PP not recovering the Z-height correctly in some cases. @wavexx spent a lot of time debugging that.

gudnimg: Yes at least not by the firmware.  We are not reading from this EEPROM address anywhere. It was only being written to.

DRracer (approved):

PR #4244, opened 2023-06-24
Optimize `st_set_position()` for speed and size

leptun:
`st_set_position()` generated very bad code because `count_position` is volatile. By using a memcpy, the code is much shorter and even faster actually since it doesn't fiddle with pointers for each coordinate as much as before. I now also provide the coordinates via a pointer instead of passing each coordinate by reference.
I'm not really sure how correct the original time estimate was, but the new code was calculated by me by hand and it looks much better now.

flash: -186
SRAM: 0
time: x0.45

### DIFF
--- a/Firmware/stepper.cpp
+++ b/Firmware/stepper.cpp
@@ -1321,7 +1321,7 @@ void st_set_position(const long *pos)
   // which corresponds to a maximum repeat frequency of ~484kHz.
   // This blocking is safe in the context of a 10kHz stepper driver interrupt
   // or a 115200 Bd serial line receive interrupt, which will not trigger faster than 12kHz.
-  memcpy((uint8_t *)count_position, pos, sizeof(count_position));
+  memcpy((long *)count_position, pos, sizeof(count_position));
   CRITICAL_SECTION_END;
 }
 
--- a/Firmware/stepper.cpp
+++ b/Firmware/stepper.cpp
@@ -1313,18 +1313,15 @@ void st_synchronize()
 	}
 }
 
-void st_set_position(const long &x, const long &y, const long &z, const long &e)
+void st_set_position(const long *pos)
 {
   CRITICAL_SECTION_START;
   // Copy 4x4B.
-  // This block locks the interrupts globally for 4.56 us,
-  // which corresponds to a maximum repeat frequency of 219.18 kHz.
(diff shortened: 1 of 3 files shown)


Firmware/stepper.cpp
```diff
@@ -1313,18 +1313,15 @@ void st_synchronize()
...
-  // which corresponds to a maximum repeat frequency of 219.18 kHz.
+  // This block locks the interrupts globally for 2.06 us,
+  // which corresponds to a maximum repeat frequency of ~484kHz.
   // This blocking is safe in the context of a 10kHz stepper driver interrupt
   // or a 115200 Bd serial line receive interrupt, which will not trigger faster than 12kHz.
-  count_position[X_AXIS] = x;
-  count_position[Y_AXIS] = y;
-  count_position[Z_AXIS] = z;
-  count_position[E_AXIS] = e;
+  memcpy((uint8_t *)count_position, pos, sizeof(count_position));
```
gudnimg: Shouldn't the casting here be `long *` instead of `uint8_t *`

gudnimg: We don't do this casting in other places when copying the `current_position` vector.

leptun: I cast it to uint8_t* since memcpy expects a void* argument. With pointers you can always go to a lower alignment variant (say from 32 bit to 8 bit addressing), but not the other way around. Since memcpy copies byte by byte, I think uint_8t* makes the most sense here.

I don't remember exactly why I had to cast it, but it might be related to the fact that the original array was declared volatile and the compiler complained. I don't care that it was volatile since we are in an atomic block and are completely overwriting count_position.

gudnimg: The `void *` in `memcpy` is just an address in memory. I feel `long *` would be more correct as the variable is pointing to a `long` type in memory and not `uint8_t`.

The cast is needed to change `volatile long *` to `long *` (tested it on my end)

gudnimg (approved):

Firmware/stepper.cpp
leptun: I pushed a commit with the change as requested.

gudnimg:
Ran a simple print with the MMU and with feedrate multiplier at 150%. Finished without issues.

IMG_3487

DRracer (approved):

PR #4261, opened 2023-07-06
PFW-1528 Intercept MMU register `0x14`

gudnimg:
This commit syncs the 8-bit firmware with the 32-bit

Test procedure:
1. Read the value of the register by sending `M707 A0x14`
2. Run a toolchange and print out the feedrate being used onto serial.
   * Value should be `20` be default.
4. Write the value of the register by sending `M708 A0x14 X40`
5. Read the value of the register by sending `M707 A0x14`
   * Value read should be `40`
6. Run a toolchange and print out the feedrate being used onto serial. It should be `40`

Change in memory:
Flash: +56 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/mmu2.cpp
+++ b/Firmware/mmu2.cpp
@@ -141,9 +141,16 @@ bool MMU2::WriteRegister(uint8_t address, uint16_t data) {
     if (!WaitForMMUReady())
         return false;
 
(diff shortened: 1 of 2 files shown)


gudnimg:
Tested this on my machine and it works well. Opening PR for review.

DRracer (approved):

PR #4274, opened 2023-07-21
Fixed printer_smodel_check for MK3/S and possible older MMU machines

DRracer:
Cherry-pick of PR#4265 into 3.13.0

Fixed printer_smodel_check for non MMU machines
Commit 136ef96 broke the compatibility check for MK3S without MMU.

May have fixed bug for older MMU machines.
Only comparing up to the length of the value from the g-code, would return equal on older MMU machines trying to run g-code sliced without the MMU. Unfortunately if that is a feature, it will cause the different printer warning.

### DIFF
--- a/Firmware/util.cpp
+++ b/Firmware/util.cpp
@@ -407,16 +407,19 @@ return pStrBegin;
 }
 
 void printer_smodel_check(const char *pStrPos, const char *actualPrinterSModel) {
-char* pResult;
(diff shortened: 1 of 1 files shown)


gudnimg (approved):
Approving since the code was unit tested :)

DRracer:
... but it would be nice to have the unit test actually included in this repo ... but tbd. later

PR #4279, opened 2023-07-25
Delay Z-Axis on Autoload

### LINKED ISSUE #4115: [ENHANCEMENT] Do not move printhead up when loading filament is triggered by sensor
**Printer type** - MK3S

Currently the printhead is moved up when you insert filament and the filament sensor is active. I get that this can be an advantage if the menu "Load Filament" is used. But when you insert it by hand and it's triggered by the filament sensor this doesn't bring any advantage; The filament is already in the printhead, better clearance is not needed anymore.

The printhead moving on it's own while someones hand is touching it, scared anyone I introduced to this printer so far. They don't like that parts are moving without warning while their hands are in "harms way".

I deactivated it myself and compiled my own firmware, but I think it's probably worth discussing if this should happen or if it could only be triggered when "Load Filament" is used via the LCD menu.

sarusani:
Delay z-axis movement when the operators hand is close to the printhead.

This is my proposed solution to resolve #4115.

Case 1: Temp is already >= target temp
Instead of moving z-axis immediately when the sensor is triggered, the movement is executed after the fast load sequence. (Just before filament extrusion is expected)

Case 2: Temp is < target temp and preheat is in progress
Instead of moving z-axis immediately when the target temp is reached, the movement is executed after the operator confirms the "Press the knob to load filament" message. (To prevent movement if the temp was just a few degrees under the target temp when the sensor was triggered)

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -1756,66 +1756,65 @@ void lcd_print_target_temps_first_line(){
     }
 }
 
-static void mFilamentPrompt()
-{
-uint8_t nLevel;
+static void mFilamentPrompt() {
+    uint8_t nLevel;
 
-lcd_print_target_temps_first_line();
-lcd_puts_at_P(0,1, _i("Press the knob"));                 ////MSG_PRESS_KNOB c=20
-lcd_set_cursor(0,2);
-switch(eFilamentAction)
-     {
-     case FilamentAction::Load:
-     case FilamentAction::AutoLoad:
-     case FilamentAction::MmuLoad:
-     case FilamentAction::MmuLoadingTest:
-          lcd_puts_P(_i("to load filament"));     ////MSG_TO_LOAD_FIL c=20
-          break;
-     case FilamentAction::UnLoad:
-     case FilamentAction::MmuUnLoad:
-          lcd_puts_P(_i("to unload filament"));   ////MSG_TO_UNLOAD_FIL c=20
-          break;
-     case FilamentAction::MmuEject:
-     case FilamentAction::MmuCut:
-     case FilamentAction::None:
-     case FilamentAction::Preheat:
-     case FilamentAction::Lay1Cal:
-          break;
-     }
+    lcd_print_target_temps_first_line();
+    lcd_puts_at_P(0,1, _i("Press the knob"));                 ////MSG_PRESS_KNOB c=20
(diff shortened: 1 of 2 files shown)


Firmware/ultralcd.cpp
```diff
@@ -1756,66 +1756,65 @@ void lcd_print_target_temps_first_line(){
...
+            lcd_puts_P(_i("to unload filament"));   ////MSG_TO_UNLOAD_FIL c=20
+            break;
+        case FilamentAction::MmuEject:
+        case FilamentAction::MmuCut:
+        case FilamentAction::None:
+        case FilamentAction::Preheat:
+        case FilamentAction::Lay1Cal:
+            break;
+    }
     if(lcd_clicked()
```
sarusani, line 1783: All the changes above are just indentation fixes.

Firmware/ultralcd.cpp
```diff
@@ -1933,12 +1934,10 @@ void mFilamentItem(uint16_t nTemp, uint16_t nTempBed)
             case FilamentAction::MmuLoad:
             case FilamentAction::MmuLoadingTest:
                 lcd_puts_P(_i("Preheating to load")); ////MSG_PREHEATING_TO_LOAD c=20
-                if (once) raise_z_above(MIN_Z_FOR_LOAD);
                 break;
             case FilamentAction::UnLoad:
             case FilamentAction::MmuUnLoad:
                 lcd_puts_P(_i("Preheating to unload")); ////MSG_PREHEATING_TO_UNLOAD c=20
-                if (once) raise_z_above(MIN_Z_FOR_UNLOAD);
```
sarusani, line 1941: I also moved the unload raise, so I could get rid of the "once" var.

Firmware/ultralcd.cpp
```diff
@@ -1756,66 +1756,65 @@ void lcd_print_target_temps_first_line(){
...
+            case FilamentAction::MmuUnLoad:
+            case FilamentAction::MmuEject:
+            case FilamentAction::MmuCut:
+            case FilamentAction::None:
+            case FilamentAction::Preheat:
+            case FilamentAction::Lay1Cal:
+                break;
+        }
+    }
 }
```
sarusani, line 1818: Again, indentation was messed up.
Only line 1800 & 1804 are actual changes

PR #4280, opened 2023-07-25
Delay Z-Axis on Autoload

### LINKED ISSUE #4115: [ENHANCEMENT] Do not move printhead up when loading filament is triggered by sensor

sarusani:
Delay z-axis movement when the operators hand is close to the printhead.

This is my proposed solution to resolve https://github.com/prusa3d/Prusa-Firmware/issues/4115.

The basic idea is to wait with z-axis movement until the user removed his hand from the printhead.
Instead of moving as soon as the filament sensor is triggered, the filament is first grabbed by "fast load" and after that the z-axis is moved up.
 
How to reproduce:
1. Unload filament if loaded.
2. Make sure filament sensor is activated.
3. Make sure the printed does not know the current z position of the extruder or the z position is below MIN_Z_FOR_LOAD (Currently 50mm).
4. Insert filament.

Original behavior: Printhead lifts Z position to MIN_Z_FOR_LOAD (50mm) as soon as filament sensor is triggered and then starts fast loading the filament.

New behavior: Printhead grabs filament with fast load and then moves Z axis up to MIN_Z_FOR_LOAD (50mm).

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3576,7 +3576,7 @@ static void gcode_M600(const bool automatic, const float x_position, const float
     custom_message_type = CustomMsg::Status;
 }
 
-void gcode_M701(float fastLoadLength, uint8_t mmuSlotIndex){
+void gcode_M701(float fastLoadLength, uint8_t mmuSlotIndex, bool raise_z_axis = false){
     FSensorBlockRunout fsBlockRunout;
 
     prusa_statistics(22);
@@ -3591,6 +3591,10 @@ void gcode_M701(float fastLoadLength, uint8_t mmuSlotIndex){
 
         current_position[E_AXIS] += fastLoadLength;
         plan_buffer_line_curposXYZE(FILAMENTCHANGE_EFEED_FIRST); //fast sequence
+
+        if (raise_z_axis) { // backwards compatibility for 3.12 and older FW
+          raise_z_above(MIN_Z_FOR_LOAD);
+        }
 
         load_filament_final_feed(); // slow sequence
         st_synchronize();
@@ -8769,13 +8773,12 @@ Sigma_Exit:
 
         // Z lift. For safety only allow positive values
         if (code_seen('Z')) z_target = fabs(code_value());
-        else raise_z_above(MIN_Z_FOR_LOAD); // backwards compatibility for 3.12 and older FW
-
+        
         // Raise the Z axis
         float delta = raise_z(z_target);
 
         // Load filament
-        gcode_M701(fastLoadLength, mmuSlotIndex);
+        gcode_M701(fastLoadLength, mmuSlotIndex, !code_seen('Z')); // if no z -> trigger MIN_Z_FOR_LOAD for backwards compatibility on 3.12 and older FW 
 
         // Restore Z axis
         raise_z(-delta);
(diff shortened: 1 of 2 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -3575,6 +3575,10 @@ void gcode_M701(float fastLoadLength, uint8_t mmuSlotIndex){
         current_position[E_AXIS] += fastLoadLength;
         plan_buffer_line_curposXYZE(FILAMENTCHANGE_EFEED_FIRST); //fast sequence
 
+        if (eFilamentAction == FilamentAction::AutoLoad || raise_z_axis) { // backwards compatibility for 3.12 and older FW
+          raise_z_above(MIN_Z_FOR_LOAD);
```
gudnimg, line 3596: I'm not sure if this is the right place for `raise_z_above()` since the M701 gcode has a Z parameter now, can we use it instead?

```
#### Usage

        M701 [ P | T | L | Z ]

    #### Parameters
    - `P` - n index of MMU slot (zero based, so 0-4 like T0 and T4)
    - `T` - Alias of `P`. Used for compatibility with Marlin
    - `L` - Extrude distance for insertion (positive value)(manual reload)
    - `Z` - Move the Z axis by this distance. Default value is 0 to maintain backwards compatibility with older gcodes.
    */
```

sarusani: Z parameter in M701 uses raise_z not raise_z_above which behaves different.
It also moves the head too early because M701 is used when the filament sensor is triggered and temp is >= target temp.

sarusani:
Commit 26369ac  made some code obsolete.
raise_z_above was now called twice during load and unload when the preheat screen was involved.
Cleaning it up saved some bytes.

The behavior is now consistent in all load cases (autoload, manual, gcode, mmu). If MIN_Z_FOR_LOAD is set, it will always raise the z-axis after the fast load sequence.

Firmware/Marlin_main.cpp
sarusani: It also makes the code easier to understand, before raise_z_above was scattered around in two different places in the code. This way it's only used once.

3d-gussner:
@gudnimg @sarusani is the https://github.com/prusa3d/Prusa-Firmware/pull/4280#pullrequestreview-1544900195 resolved if so please mark it as "resolved".

gudnimg:
> @gudnimg @sarusani is the [#4280 (review)](https://github.com/prusa3d/Prusa-Firmware/pull/4280#pullrequestreview-1544900195) resolved if so please mark it as "resolved".

Looked over the changes again and the PR looks good to me. I haven't tested the changes though.

gudnimg (approved):
Looks good to me. It would be nice to get another approval as well (@3d-gussner ?)

3d-gussner (approved):
Tested with MK404

- [X] MK3S
- [X] MK3

Issue #4304, opened 2023-08-03
Full Setup Wizard XYZ calibration error


jeepac:
**Printer type** - MK3S
**Printer firmware version** - 3.13.0

**MMU upgrade** - N/A
**MMU upgrade firmware version** - N/A
**SD card or USB/Octoprint** -print via SD card

**Describe the bug**
 XYZ calibration error
  
**To Reproduce**
  when run full setup wizard, however at the end of XYZ calibration the whole Z assembly rams in to the bed
  
**Expected behavior**
 perform 9 point z calibration
  
**G-code**
  n/A
**Video**

https://forum.prusa3d.com/wp-content/uploads/2023/08/XYZ-calibration-failure.mp4

D-an-W:
I have also had this error.

At the point where the calibration with the paper finishes the Z axis should raise 150mm, if it doesn't and you continue then you get the result as in your video.

It was briefly discussed here [https://github.com/prusa3d/Prusa-Firmware/issues/4105](url) as two of us testing the firmware had the problem but it then started working ok for me.

sdh2:
I had that issue too. The only workaround I found is to shut off the machine during the end of XYZ calibration (which seems enough to save the calibration), then move the Z up and run the Z calibration separately. Even then it only sometimes works, other times it fails immediately.

sarusani:
When you move the z axis up and down manually (by long press of the knob). Does it move freely or does it also sometimes stop when it shouldn't?

If so, it's indication that the resistance for the motor is too high (binding) and the motor thinks it reached the top/bottom. That would explain why it's not moving up the full 15cm...

I looked at the code and when it shows the "place steel sheet" message it definitely issued the command to raise 15cm. The question is, what are possible reasons for this command to fail?

@gudnimg 
As I understand raise_z should return the actual traveled distance. So maybe something like this could prevent (not fix) this?
```
//make space
if (raise_z(150) == 150) {
  lcd_show_fullscreen_message_and_wait_P(_T(MSG_PLACE_STEEL_SHEET));
}
else {
  return false;
}
```

gudnimg:
@sarusani that's correct. But I suspect something set `planner_aborted` to `true` because it looks like the planned move is just ignored completely.

I added the following log into 3.13.0, can anyone check and see if it appears in the logs?
![image](https://github.com/prusa3d/Prusa-Firmware/assets/8218499/5e80aabb-14c0-4fb9-8afb-542b8237ee0e)

[...]

PR #4306, opened 2023-08-03
Fix Calibration XYZ for MK3S

ojrik:
This is the solution for XYZ Calibration. For MK3S and MK3S+. There is no height defined here for the MK3S machine, only for the MK2 MK2.5 MK3.

MK3S.h Configuration
// Printer revision
#define PRINTER_TYPE PRINTER_MK3S
#define PRINTER_NAME PRINTER_MK3S_NAME
#define PRINTER_NAME_ALTERNATE PRINTER_MK3_NAME //the other similar printer to this. #define PRINTER_MMU_TYPE PRINTER_MK3S_MMU3
#define PRINTER_MMU_NAME PRINTER_MK3S_MMU3_NAME

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -2174,7 +2174,7 @@ bool calibrate_z_auto()
 	plan_buffer_line_destinationXYZE(feedrate / 60);
 	st_synchronize();
 	enable_endstops(endstops_enabled);
-	if (PRINTER_TYPE == PRINTER_MK3) {
+	if((PRINTER_TYPE == PRINTER_MK3) || (PRINTER_TYPE == PRINTER_MK3S)){
 		current_position[Z_AXIS] = Z_MAX_POS + 2.0;
 	}
 	else {


Firmware/Marlin_main.cpp
```diff
@@ -2174,7 +2174,7 @@ bool calibrate_z_auto()
 	plan_buffer_line_destinationXYZE(feedrate / 60);
 	st_synchronize();
 	enable_endstops(endstops_enabled);
-	if (PRINTER_TYPE == PRINTER_MK3) {
+	if((PRINTER_TYPE == PRINTER_MK3) || (PRINTER_TYPE == PRINTER_MK3S)){
```
gudnimg, line 2177: I'm not sure if this is correct. For MK3S the `else` statement is run `current_position[Z_AXIS] = Z_MAX_POS + 9.0;`

ojrik: > I'm not sure if this is correct. For MK3S the `else` statement is run `current_position[Z_AXIS] = Z_MAX_POS + 9.0;`
I recently had to hard reset the MK3S 24v to modify the TMC2130.
The X Axis went all the way to the top stop and then the Axis stopped completely on the steel plate. The machine reported failed calibration. If I inserted the PINDA sensor more, it performed the calibration, but the nozzle scratched the heated bed. I started to find out the cause and noticed that MK3S is not defined in the same way as MK3. And the MK3 has the same height XYZ calibration as the MK3S. After adjusting the sensor, I returned it to its original place, and the calibration is without an alarm.

ojrik: > I'm not sure if this is correct. For MK3S the `else` statement is run `current_position[Z_AXIS] = Z_MAX_POS + 9.0;`

I recently had to hard reset the MK3S 24v to modify the TMC2130.
The X Axis went all the way to the top stop and then the Axis stopped completely on the steel plate. The machine reported failed calibration. If I inserted the PINDA sensor more, it performed the calibration, but the nozzle scratched the heated bed. I started to find out the cause and noticed that MK3S is not defined in the same way as MK3. And the MK3 has the same height XYZ calibration as the MK3S. After adjusting the sensor, I returned it to its original place, and the calibration is without an alarm.

gudnimg: Any chance you are seeing this issue? https://github.com/prusa3d/Prusa-Firmware/issues/4304 It seems to happen unpredictably. For example I don't see it on my printer. There is a 150mm Z move that should happen but for some it seems to be ignored completely and afterwards the Z-axis will move the nozzle into the bed

ojrik: > Any chance you are seeing this issue? #4304 It seems to happen unpredictably. For example I don't see it on my printer. There is a 150mm Z move that should happen but for some it seems to be ignored completely and afterwards the Z-axis will move the nozzle into the bed


Yes, I saw it, and the comments. maybe that would solve that too. But I can't replicate the same error here.

PR #4307, opened 2023-08-04
PFW-1531 Fix a possible recursion in fullscreen code

### LINKED ISSUE #4300: [BUG] FW 3.13.0 Cannot perform first layer calibration
**Printer type** - MK3S+
**Printer firmware version**: 3.13.0

**SD card**

**Describe the bug**
I have just upgraded from 3.12.2 to 3.13.0 for my MK3S+ and found that I cannot perform first layer calibration. I have filament sensor disabled. On LCD screen, the printer gets stuck on the "do you have filament loaded?" choice question. No matter what I choose yes/no, it always shows the same question and choices again.

I have a few modifications:
 - Revo hotend
 - Custom printed parts for Z axis and X axis.
  
**To Reproduce**
 - My initial setup is with Revo and on FW 3.12.2. PID calibrated for Revo.
 - Turn off filament sensor.
 - Upgrade to FW 3.13.0.
 - Rerun self-test.
 - Load PLA.
 - On LCD, choose to calibrate -> first layer calibration.
 - Get to the question "do you have filament loaded?"
 - Choose yes/no.
 - It beeps and repeats the same question "do you have filament loaded?". No matter if I choose yes or no.
  
**Expected behavior**
The first layer calibration process will continue after I choose yes/no for "do you have filament loaded?".
  
**G-code**
N/A

**Video**

https://github.com/prusa3d/Prusa-Firmware/assets/2630468/b3952149-2c8e-4685-a1e5-c1ab19e226ac

gudnimg:
We may not call `lcd_update_enable(true)` in `lcd_show_multiscreen_message_with_choices_and_wait_P` when running a menu item. Because this can cause a recursion scenario such as in `lcd_v2_calibration` menu.

Fixes #4300

No change in memory

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -3095,7 +3095,11 @@ uint8_t lcd_show_multiscreen_message_with_choices_and_wait_P(
 exit:
     KEEPALIVE_STATE(IN_HANDLER);
     lcd_set_custom_characters();
-    lcd_update_enable(true);
+    // Enable LCD updates again. We may not call lcd_update_enable(true)
+    // because it may create a recursion scenario when the caller of lcd_show_multiscreen_message_with_choices_and_wait_P
+    // is a submenu lcd_update_enable(true) will cause another call to the submenu immediately
+    // and so won't allow the user to exit the submenu
+    lcd_update_enabled = true;
     return current_selection;
 }
 


wavexx (approved):

3d-gussner (approved):
Tested on MK404

sarusani:
This changed the behavior of the "Filament extruding & correct color?" msg. (Possibly others, but that's the one I noticed)

Before this change, when you selected "no", the menu exited and the screen immediately changed to "filament loading".
Now the yes/no selection stays on the LCD even after selecting the "no" option. (It works fine if you select "yes")

gudnimg:
> This changed the behavior of the "Filament extruding & correct color?" msg. (Possibly others, but that's the one I noticed)
> 
> Before this change, when you selected "no", the menu exited and the screen immediately changed to "filament loading". Now the yes/no selection stays on the LCD even after selecting the "no" option. (It works fine if you select "yes")

Thanks for reporting, I’ll look into it 👍

gudnimg:
@sarusani it seems to be `lcd_load_filament_color_check()` that is causing issues. It's used in M701 and also when using "Load to nozzle" menu with MMU.

I don't see how it can happen in M701 since it calls `lcd_update(2)`, but in the MMU menu maybe this improves it (I haven't tested it yet):

image

sarusani:
I don't have a MMU. I noticed it during "normal" filament loading. (Preheat & trigger sensor by inserting filament)

The answers (yes/no) stays on screen for a few seconds after selecting "no", then it switches to "loading filament" for a very short time before it will go back to the "correct color?" msg as expected.
So it does clear the answers eventually (or the "load filament" overwrites it), just to late.

sarusani:
I submitted a [PR](https://github.com/prusa3d/Prusa-Firmware/pull/4327) to fix it.

PR #4308, opened 2023-08-04
Fix a bug where Load filament menu disappears

gudnimg:
Kudos to @3d-gussner for finding the issue

Steps to reproduce:
1. Boot up printer
2. Select "Load filament" menu
3. Select "Back" to leave the menu
4. **Observe issue** => "Load filament" menu and other MMU menus are gone!
5. **Workaround**: Got the "Preheat" menu, and Select "Back".

The issue also appears when a filament action actually completes running. For example by unloading filament.

This commit is my proposed fix.

When `eFilamentAction` is equal to
`FilamentAction::Load` we must reset it to `FilamentAction::None` when the Back button in Load Filament is selected

Issue was likely introduced with https://github.com/prusa3d/Prusa-Firmware/pull/3564

Change in memory:
Flash: -82 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -1833,13 +1833,8 @@ switch(eFilamentAction)
 
 void mFilamentBack()
 {
-    if (eFilamentAction == FilamentAction::AutoLoad ||
-        eFilamentAction == FilamentAction::Preheat ||
-        eFilamentAction == FilamentAction::Lay1Cal)
-    {
-        // filament action has been cancelled
-        eFilamentAction = FilamentAction::None;
-    }
+    // filament action has been cancelled
+    eFilamentAction = FilamentAction::None;
 }
 
 void mFilamentItem(uint16_t nTemp, uint16_t nTempBed)
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -1837,10 +1837,15 @@ void mFilamentBack()
     eFilamentAction = FilamentAction::None;
 }
 
+/// Reset the menu stack and clear the planned filament action flag
+static void mFilamentDone()
+{
+    menu_back(bFilamentPreheatState ? 1 : 2);
+    mFilamentBack();
+}
+
 void mFilamentItem(uint16_t nTemp, uint16_t nTempBed)
 {
-    uint8_t nLevel;
-
     setTargetHotend((float)nTemp);
     if (!shouldPreheatOnlyNozzle()) setTargetBed((float)nTempBed);
 
@@ -1869,6 +1874,7 @@ void mFilamentItem(uint16_t nTemp, uint16_t nTempBed)
     // then continue with the filament action if any is set
     if (bFilamentSkipPreheat || abs((int)current_temperature[0] - nTemp) < TEMP_HYSTERESIS)
     {
+        menu_func_t filamentActionMenu = nullptr;
         switch (eFilamentAction)
         {
         case FilamentAction::Load:
@@ -1877,9 +1883,7 @@ void mFilamentItem(uint16_t nTemp, uint16_t nTempBed)
             if (bFilamentWaitingFlag) menu_submenu(mFilamentPrompt, true);
             else
             {
-                nLevel = bFilamentPreheatState ? 1 : 2;
-                menu_back(nLevel);
-
+                mFilamentDone();
                 if (eFilamentAction == FilamentAction::AutoLoad) {
                     // loading no longer cancellable
                     eFilamentAction = FilamentAction::Load;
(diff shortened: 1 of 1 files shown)


gudnimg:
There are more issues like this that need fixing, adding 🚧

gudnimg:
I've fixed all scenarios I could find but I haven't tested Autoloading since I'm using the MMU Fsensor setup

Firmware/ultralcd.cpp
```diff
@@ -1833,19 +1833,19 @@ switch(eFilamentAction)
...
 {
-    if (eFilamentAction == FilamentAction::AutoLoad ||
-        eFilamentAction == FilamentAction::Preheat ||
-        eFilamentAction == FilamentAction::Lay1Cal)
-    {
-        // filament action has been cancelled
-        eFilamentAction = FilamentAction::None;
-    }
+    // filament action has been cancelled
+    eFilamentAction = FilamentAction::None;
```
wavexx, line 1837: The original idea was that if the current action is "Load", and you exit the menu, it stays "Load" until it is finally cleared by unload_filament().

This is why this checks for AutoLoad/Preheat/Lay1Cal: in the comments these are the states which can be "aborted" in the preheat menu.

wavexx: The menu should return to full automatically as soon as the load is complete.

gudnimg:

Firmware/ultralcd.cpp
```diff
@@ -1929,8 +1919,16 @@ void mFilamentItem(uint16_t nTemp, uint16_t nTempBed)
             // handled earlier
             break;
         }
-        if (bFilamentWaitingFlag) Sound_MakeSound(e_SOUND_TYPE_StandardPrompt);
-        bFilamentWaitingFlag = false;
+        if (bFilamentWaitingFlag) {
```
DRracer, line 1922: :thinking: how was the previous code indented? There has been no `{`, it looks like only `Sound_MakeSound` was executed `if (bFilamentWaitingFlag)`

It actually makes sense moving the line:
`bFilamentWaitingFlag = false;` into the `if` clause

Firmware/ultralcd.cpp
DRracer: @gudnimg would it make sense to keep `mFilamentBack()` as it has been before and just set `eFilamentAction = FilamentAction::None;` in `mFilamentDone()`?

I haven't tried myself, I'm just trying to understand what's going on.

DRracer (approved):

Firmware/ultralcd.cpp
wavexx: I still think this was incorrect. The filament action should only be cleared by the actual move, not by the menu, otherwise we re-introduce the issue fixed in #3564 where exiting the menu will allow actions such as autoload while autoload is still taking place.

gudnimg: Ah that make sense to me. I agree with @wavexx I can try find an alternative fix.

sarusani:
Since this was merged the printer fails performing autoload & unload after the first time.

Steps to reproduce:
1) Trigger autoload, select material -> wait for temp & load filament. (For testing i don't let the gear grab the filament, so I can perform step 3) later...)
2) Select unload filament
Instead of unloading  the filament, the screen immediately returns to the main menu
3) Remove filament from sensor & trigger autoload again.
Instead of loading the filament, the screen immediately returns to the main menu

gudnimg:
@sarusani it's similar to the issue @wavexx pointed out. Simply an oversight on my part. I am planning on fixing it tomorrow. Should be easy to fix, just need to think how to minimize memory usage :)

sarusani:
@gudnimg No stress 😃 I noticed it during testing the TEMP_HYSTERESIS, so I thought I'll document it.

gudnimg:
@sarusani any chance you could test my changes to fix this? Specifically Autoloading scenario. I can invite you to my fork so you can checkout the branch directly.

Or I can send you a hex build instead if that suits better

sarusani:
@gudnimg Yes, I have time to test. I think invite is easier, in case there's some additional changes.

sarusani:
@gudnimg got it. What branch should I test?
(Just noticed, I don't need invitation for checking out, only for push. So you can revoke it if you prefer)

gudnimg:
@sarusani I pushed the branch here: https://github.com/gudnimg/Prusa-Firmware/tree/fix-filament-actions so far it works well on MMU. If you can verify Autoload that would be awesome ❤️ 

I added temporary logging to see when the filament action variable is changed. Will be removed in the final PR (to save memory)

gudnimg:
This conveniently fixes one bug we have logged internally where you could trigger "Unload filament" with the MMU while the first unload is still in progress :)

sarusani:
Everything works as expected. 👍 

Logs:

first autoload:
```
Recv: echo:FilamentAction updated: 0 -> 2
Recv: echo:Warning: FilamentAction is already: 2
...
Recv: echo:enqueing "M701"
Recv: LCD status changed
```
2nd autoload:
```
Recv: echo:FilamentAction updated: 0 -> 2
Recv: echo:Warning: FilamentAction is already: 2
...
Recv: echo:enqueing "M701"
Recv: LCD status changed
```
unload:
```
echo:FilamentAction updated: 0 -> 3
...
Recv: echo:enqueing "M702"
Recv: LCD status changed
...
Recv: echo:FilamentAction updated: 3 -> 0
```
3rd autoload:
```
Recv: echo:FilamentAction updated: 0 -> 2
Recv: echo:enqueing "M701"
Recv: LCD status changed
```

PR #4316, opened 2023-08-04
Fix and optimise VERBOSE_CHECK_HIT_ENDSTOPS

gudnimg:
Make `VERBOSE_CHECK_HIT_ENDSTOPS` code compile again

### DIFF
--- a/Firmware/stepper.cpp
+++ b/Firmware/stepper.cpp
@@ -189,20 +189,17 @@ void checkHitEndstops()
 {
  if(endstop_hit) {
 #ifdef VERBOSE_CHECK_HIT_ENDSTOPS
-   SERIAL_ECHO_START;
-   SERIAL_ECHORPGM(MSG_ENDSTOPS_HIT);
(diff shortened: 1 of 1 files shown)


wavexx (approved):

gudnimg:
Testing this now, and for some reason I don't get these logs 🤔 I've tried Auto homing and Calibrating Z. Maybe this is only for physical endstop switches?

Looks like `endstop_hit` is not set at all for TMC2130

gudnimg:
Found a fix. We need to call `checkHitEndstops()` in `manage_inactivity()` and not `loop()`
image

gudnimg:
Dropped the last commit because we cannot call `checkHitEndstops()` in `manage_inactivity()`. The reason is `checkHitEndstops()` clears the variable `endstop_hit`.

`endstop_hit` is used by functions `endstops_hit_on_purpose`, `endstop_z_hit_on_purpose`, `enable_z_endstop`, and allowing `checkHitEndstops()` to clear it creates more bugs.

PR #4318, opened 2023-08-05
PFW-1530 Add "Tune" button to IDLER CANNOT HOME screen

gudnimg:
TODOs:
- [x] Change "Idler" text to something more descriptive while also being less than 17 characters
- [x] Add similar options to SELECTOR CANNOT HOME screen ( with range 1 to 4)
- [x] Try to change the screen such that it reads the current value from MMU EEPROM instead of showing a hard-coded value on the printer side.

----

Instead of adding a new Stallguard Threshold editing menu to the LCD Experimental menu, I propose to add a new _Tune_ button option to the MMU error screen. The Experimental menu is not accessible at all when the MMU error screen is active and so I think having the options available during the MMU error UI itself is more user friendly. Which ever way we go there is a lot of memory consumed by this since the menu code is very memory (flash) hungry.

Below are some images to show the new UI. Note that "Idler" text will be changed, I only took these photos to show the concept.

When the _Done_ button is clicked, the MMU error screen will disappear and the Idler will attempt to re-home with the new setting.

| Error screen with new option| Tune menu |
|------|------|
| IMG_3793    |   IMG_3794   |

### DIFF
--- a/Firmware/mmu2.h
+++ b/Firmware/mmu2.h
@@ -88,12 +88,6 @@ public:
     /// @returns true upon success
     bool ReadRegister(uint8_t address);
 
-    /// Variant of ReadRegister which runs in blocking context such as manage_response()
-    /// Be careful of using this as it is not recursion protected!
-    /// @param address Address of register in hexidecimal
-    /// @return true upon success
-    bool ReadRegisterInner(uint8_t address);
-
     /// Write from a MMU register (See gcode M708)
     /// @param address Address of register in hexidecimal
     /// @param data Data to write to register
--- a/Firmware/mmu2.h
+++ b/Firmware/mmu2.h
@@ -69,6 +69,10 @@ public:
         ErrorSourceNone = 0xFF,
     };
 
+    /// Tune value in MMU registers as a way to recover from errors
+    /// e.g. Idler Stallguard threshold
+    void Tune();
+
     /// Perform a reset of the MMU
     /// @param level physical form of the reset
     void Reset(ResetForm level);
--- a/Firmware/mmu2.h
+++ b/Firmware/mmu2.h
@@ -88,6 +88,12 @@ public:
     /// @returns true upon success
     bool ReadRegister(uint8_t address);
 
+    /// Variant of ReadRegister which runs in blocking context such as manage_response()
+    /// Be careful of using this as it is not recursion protected!
+    /// @param address Address of register in hexidecimal
+    /// @return true upon success
+    bool ReadRegisterInner(uint8_t address);
+
     /// Write from a MMU register (See gcode M708)
     /// @param address Address of register in hexidecimal
     /// @param data Data to write to register
@@ -193,6 +199,12 @@ public:
     inline uint16_t TMCFailures() const { return tmcFailures; }
     inline void IncrementTMCFailures() { ++tmcFailures; }
     inline void ClearTMCFailures() { tmcFailures = 0; }
+
+    /// Retrieve cached value parsed from ReadRegister()
+    /// or using M707
+    inline uint16_t GetLastReadRegisterValue() const {
+        return lastReadRegisterValue;
+    };
 
 private:
     /// Perform software self-reset of the MMU (sends an X0 command)
@@ -301,6 +313,7 @@ private:
     ErrorCode lastErrorCode = ErrorCode::MMU_NOT_RESPONDING;
     ErrorSource lastErrorSource = ErrorSource::ErrorSourceNone;
     Buttons lastButton = Buttons::NoButton;
+    uint16_t lastReadRegisterValue = 0;
 
     StepStatus logicStepLastStatus;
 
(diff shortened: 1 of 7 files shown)


gudnimg:
Synced the branch with MK3, hopefully the reported PR size is now more correct.

Ro3Deee:
I reduced the stall guard even if it was working ok with the default value,because it works smoother with a lower value. So i think it should be available always.
 We should also be able to able to have an option to "stop"/cancel the error message , than change the value and then restart the mmu.

gudnimg:
> We should also be able to able to have an option to "stop"/cancel the error message , than change the value and then restart the mmu.

I agree but currently the way we designed the protocol interface it's not possible. The MMU will report an error every 1 second (for every `Q0` Query). And the printer simply checks => if there is no error screen, then show it.

I'd also like to see some way to Cancel MMU operations, but we're not there yet (not highest priority).

gudnimg:
The point with this is to at least get the homing working so there is no error screen reported. Maybe we can re-use the same code in the Experimental menu without using more Flash. But as this is still a draft, we'll see. I have a lot of cleanup to do yet.

Ro3Deee:
Another option would be to be able to send configuration g codes from Prusa slicer and also configure the printer using a visual interface with menus and windows there, as in PS there is no memory limitation for menus. I didn't noticed that pronterface upcased x letter in hexadecimal register number (until you told me).

Sending the config g codes seems to work even if error message is active on lcd.

SNDR0:
Idler ----> Engager, coupler, REF(Rotary Engager Filament)?

gudnimg:
@DRracer I wonder if we can improve `ReadRegisterInner()` 🤔 It now adds ~100B

gudnimg:
Opening PR for review... I have fixed all bugs I've found so far and tried to minimize flash usage, but this could use some more testing.

gudnimg:
Video of the menu currently. I trigger Selector Cannot Home by holding it in place with my hand. Hopefully Github doesnt compress the video too much 👀

https://github.com/prusa3d/Prusa-Firmware/assets/8218499/4e41fa78-9df1-4e32-9674-5401b4f8c102

Firmware/mmu2.h
```diff
@@ -297,6 +307,7 @@ class MMU2 {
     ErrorCode lastErrorCode = ErrorCode::MMU_NOT_RESPONDING;
     ErrorSource lastErrorSource = ErrorSource::ErrorSourceNone;
     Buttons lastButton = Buttons::NoButton;
+    uint16_t lastReadRegisterValue = 0;
```
DRracer, line 310: Not important now, but may be make a pair<uint8_t register index, uint16_t value> - to make sure the last read value is related to a specific register index. This will become important when we start reading multiple registers on demand.

Firmware/mmu2_reporting.cpp
```diff
@@ -216,11 +218,22 @@ enum ReportErrorHookStates ReportErrorHookState = ReportErrorHookStates::RENDER_
...
+// Helper variable to stop rendering the error screen when the firmware is rendering complementary
+// UI to resolve the error screen, for example tuning Idler Stallguard Threshold
+// Set to false to allow the error screen to render again.
+static bool putErrorScreenToSleep;
+
 bool isErrorScreenRunning() {
     return is_mmu_error_monitor_active;
 }
 
+bool isErrorScreenSleeping() {
```
DRracer: May be we could call it "TuneMenuEntered"

Firmware/mmu2_reporting.cpp
```diff
@@ -359,4 +372,68 @@ void ScreenClear(){
...
+struct TuneItem {
+    uint8_t address;
+    uint8_t minValue;
+    uint8_t maxValue;
+} __attribute__((packed));
+
+static const TuneItem TuneItems[] PROGMEM = {
+  { (uint8_t)Register::Selector_sg_thrs_R, 1, 4},
+  { (uint8_t)Register::Idler_sg_thrs_R, 4, 7},
+};
```
DRracer, line 384: :+1:

Firmware/mmu2_reporting.cpp
```diff
@@ -359,4 +372,68 @@ void ScreenClear(){
...
+
+static_assert(sizeof(TuneItems)/sizeof(TuneItem) == 2);
+
+typedef struct
```
DRracer: let's make it a C++ struct

DRracer (commented):
Actually, the code is quite clean, I don't see any significant design flaw. I added a few minor comments.

Firmware/mmu2_reporting.cpp
gudnimg: Do I understand correctly you mean to drop `typedef`?

Firmware/mmu2_reporting.cpp
gudnimg: Done

Firmware/mmu2_reporting.cpp
gudnimg: Done

DRracer (approved):

Issue #4323, opened 2023-08-12
MK3S/+ MMU2S start gcode issue with T[initial_tool] command


mmoraes80:
**Printer type** - [MK3S Stock]
**Printer firmware version** -  [3.13.0]

**MMU upgrade** - [MMU2S]
**MMU upgrade firmware version** - [3.0.0]

**SD card or USB/Octoprint** - [PrusaLink 0.7.0]

**Describe the bug**
 When sliced with the MMU2S profile, the print job starts and the printer runs the start gcode normally. However, right after the bed leveling is complete and the print head moves to the origin (G1 Y-3 F1000, G1 Z0.4 F1000), the T[initial_tool] command makes the print head to quickly lift and move all the way to right as in a filament change procedure, it then loads the filament to the nozzle, purges the filament, asks the user to confirm the color change and moves back to origin before starting a new bed leveling instead of printing the purge line, which only happens after the second bed leveling. 

So the issue is that no filament change nor the second bed leveling is supposed to happen at this time since the print is just starting. The issue has been noticed since 3.12.0
  
**To Reproduce**
Slice the model with the Stock MK3s MMU2S profile.
  
**Expected behavior**
 Correct filament to load to nozzle after the bed leveling and start printing the purge line.
  
**G-code**
M862.3 P "[printer_model]" ; printer model check
M862.1 P[nozzle_diameter] ; nozzle diameter check
M115 U3.12.2 ; tell printer latest fw version
G90 ; use absolute coordinates
M83 ; extruder relative mode
M104 S[first_layer_temperature] ; set extruder temp
M140 S[first_layer_bed_temperature] ; set bed temp
M190 S[first_layer_bed_temperature] ; wait for bed temp
M109 S[first_layer_temperature] ; wait for extruder temp
G28 W ; home all without mesh bed level
G80 ; mesh bed leveling

; Send the filament type to the MMU2.0 unit.
; E stands for extruder number, F stands for filament type (0: default; 1:flex; 2: PVA)
M403 E0 F{"" + ((filament_type[0]=="FLEX") ? 1 : ((filament_type[0]=="PVA") ? 2 : 0))}
M403 E1 F{"" + ((filament_type[1]=="FLEX") ? 1 : ((filament_type[1]=="PVA") ? 2 : 0))}
M403 E2 F{"" + ((filament_type[2]=="FLEX") ? 1 : ((filament_type[2]=="PVA") ? 2 : 0))}
[...]

PR #4325, opened 2023-08-13
🚧 filament runout: only allowed if E-axis is active

gudnimg:
Only allow filament runouts when the extruder motor is moving. We do this for FINDA runouts, so why not Fsensor as well? 

I think this may improve situations like https://github.com/prusa3d/Prusa-Firmware/issues/4323

Change in memory:
Flash: -6 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Filament_sensor.cpp
+++ b/Firmware/Filament_sensor.cpp
@@ -128,7 +128,7 @@ void Filament_sensor::triggerFilamentRemoved() {
     if (runoutEnabled
         && (eFilamentAction == FilamentAction::None)
         && (
-            moves_planned() != 0
+            e_active()
             || printJobOngoing()
         )
         && !(


leptun (dismissed):
This sounds like a good idea, but I have a few notes:

What about triggerFilamentInserted()? Do we change to e_active() also there? I'm not sure what such a change would have as a consequence. Do we want to allow filament autoloads when axes are moving, but E is idle?

Also, e_active() can be sped up by breaking out of the loop early when an E move is found in the first block.

leptun:
https://github.com/prusa3d/Prusa-Firmware/pull/3885 will also need rebasing and rechecking after this is merged

gudnimg:
> What about triggerFilamentInserted()? Do we change to e_active() also there? I'm not sure what such a change would have as a consequence. Do we want to allow filament autoloads when axes are moving, but E is idle?

For this scenario I think it makes sense to not allow autoloading if the axis are moving. I don't see why a user would need to load while the motors are moving 🤔 

> Also, e_active() can be sped up by breaking out of the loop early when an E move is found in the first block.

I tried it and saw having two `return`'s seems to increase code size. But perhaps the speed improvement is worth it.

Firmware/Filament_sensor.cpp
```diff
@@ -128,7 +128,7 @@ void Filament_sensor::triggerFilamentRemoved() {
     if (runoutEnabled
         && (eFilamentAction == FilamentAction::None)
         && (
-            moves_planned() != 0
+            e_active()
             || printJobOngoing()
```
gudnimg, line 132: Looking back at this... shouldn't this be `e_active() && printJobOngoing()`? I don't know of any situation where we can recover from filament runout except when printing.

DRracer: Is first layer calibration a `printJobOngoing()`?

gudnimg: No it's only true if `card.sdprinting = true`, which happens only when a gcode file is printing. Or when `usb_timer.running()` is `true`, which only happens if the user is sending commands via Serial

gudnimg: Filament runout during First Layer Calibration is ignored explicitly. Similar to how we disable it when MMU is enabled.
image

gudnimg:
Rebased to sync with MK3 branch (123 commits)

DRracer (commented):
I have mixed feelings about the `e_active()`. Technically, it seems to be correct, but I don't see all the consequences.

This part has been a bit sketchy in the past...

gudnimg:
Probably not going to finish this PR.

PR #4327, opened 2023-08-15
Fix lcd update after PFW-1531

sarusani:
Before #4307, when you selected "no" on the "Filament extruding & correct color?" msg., the menu exited and the screen immediately changed to "filament loading".
Now the yes/no selection stays on the LCD even after selecting the "no" option.

This restores the original behavior.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -3111,6 +3111,7 @@ exit:
     // is a submenu lcd_update_enable(true) will cause another call to the submenu immediately
     // and so won't allow the user to exit the submenu
     lcd_update_enabled = true;
+    lcd_draw_update = 2;
     return current_selection;
 }
 


Firmware/ultralcd.cpp
```diff
@@ -2234,6 +2234,7 @@ void lcd_load_filament_color_check()
 {
 	uint8_t clean = lcd_show_fullscreen_message_yes_no_and_wait_P(_T(MSG_FILAMENT_CLEAN), false, LCD_LEFT_BUTTON_CHOICE);
 	while (clean == LCD_MIDDLE_BUTTON_CHOICE) {
+		lcd_update(2);
```
gudnimg: I suspect this will break the MMU menu. I can test it on Friday/Saturday latest.

sarusani: The call that was removed in PFW-1531 also did an lcd_update(2), so I don't think this should break anything.
Unfortunately I don't have a MMU to test myself.

gudnimg: Regarding MMU I can test it or @3d-gussner if he has time. 

lcd_update() should be used with caution because it can create a recursion scenario as in PFW-1531. When lcd_update() is called it forces a re-entry (recursively) into whatever the current LCD menu is rendering. In some situations as PFW-1531 was supposed to fix it can make it seem like the fullscreen buttons (yes/no) dont do anything. When actually the firmware is stuck in menu as it keeps re-entering it. Eventually it will crash the firmware when the stack is full.

sarusani: Thanks for the details!

I went through the code again and I believe forcing the re-entry is what should happen in this particular situation.

gudnimg: I have an alternative proposal to fix the issue which is much more general. It seems to work on my end but does this also work on your end? There should be max ~100ms delay in re-drawing the LCD.

Note: I'm testing with MMU
image

sarusani: Yes, works fine 👍 And I also think this more general approach is the much better solution.

Strange thing is, I'm sure I tried `lcd_draw_update = 2` inside the `while` and I had a much longer delay in screen update. But I'm unable to reproduce this. 🤷

gudnimg (approved):
Tested on MK3S+ with MMU, PFW-1531 scenario is still fixed, and also `lcd_load_filament_color_check()` 👍

DRracer (approved):

Issue #3610, opened 2022-09-19
  [ENHANCEMENT]<Allow for steel sheet profile change and babystep info gathering by gCode>


dawidpieper:
Enter what type of printer or upgrade the enhancement applies to.
**Printer type** - [-]
**MMU Upgrade** - [-]

**Is your enhancement related to a problem? Please describe.**
The only way blind users can use Prusa printers, as there's no screenreader and its implementation is not possible due to firmware limitations, is via Octoprint, Pronterface or other external software.
Almost any operation is possible. The biggest problem I encounter is the inability to use the steel sheets profiles and check the value of the Live Adjust Z parameter. It can be set with ```G83``` but there is no way to read the current or use profiles.
  
**Describe the solution you'd like**
* When ```G83``` called with no parameters, return information about current babystep and selected profile
* Add option to change profile without overriding its value

If this issue gets accepted, I will be happy to work on this. In fact, I made some ugly and temporary solutions for personal use, so I have a start base.

In hope that you will look favorably on it,
Kind regards,
Dawid Pieper

Prusa-Support:
Hello! Thank you for your message. After checking with our devs, unfortunately, they informed me that this specific way, with GCode possibly manipulating sheet profiles, won't be implemented. Closing this specific issue as not planned, but in case of more ideas for enhancements do not hesitate to open new threads with more ideas. 

Alessandro Pantaleo
Prusa Research

wavexx:
@dawidpieper We have a PR related to this: #2201 
It looks like that should do everything you need, and if not that would be the perfect time to have some feedback.

PR #2201, opened 2019-09-12
Add M850 code for setting sheet label and offsets

vintagepc:
Per #2091 and #2153 
This adds M-code support for reading and setting sheet labels and offsets to satisfy advanced user needs. Logic being, if you know how to manually send g-codes to your printer, you're advanced enough to handle the responsibility.
I chose 850 because it's near other first-layer/levelling parameters and appears to be unused in official marlin as well.

Usage: 
- M850 S[sheet] Z[offset] L[label] B[Bed temp] P[PINDA temp]
- Z and L are optional, if one or both are missing the current values are reported instead.
- If L and/or Z are specified, the sheet's values are updated.
- Z range is validated
- Sheet index is validated
- Requesting info (no L or Z) on an uncalibrated sheet reports as such.

Note: Pronterface capitalizes anything sent. To use lowercase in your sheet names you'll need to use a different terminal program.

Closes #2091
Closes #2153
Closes #3534

PR #4328, opened 2023-08-15
M850 add possibility to set sheet as active

dawidpieper:
#2201 added possibility to set sheet parameters using new M850 gCode, but there still lacks possibility to set or check what sheet is active.
Added ```A``` parameter to set sheet as Active.
Related to #3610

New usage

```M850 S[sheet] Z[offset] L[label] B[Bed temp] P[PINDA temp] A[is active]```

* ```A1``` enables the selected sheet
* * If ```A``` set to 0 or missing, current status is reported

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -7795,15 +7795,17 @@ Sigma_Exit:
 		iPindaC = code_value_uint8();
 	}
 	else
+	{
 		iPindaC = eeprom_read_byte(&EEPROM_Sheets_base->s[iSel].pinda_temp);
+	}
 	if (code_seen('A'))
 	{
 		bHasIsActive = true;
 		bIsActive = code_value_uint8() || (eeprom_read_byte(&(EEPROM_Sheets_base->active_sheet)) == iSel);
 	}
 	else
-		bIsActive = (eeprom_read_byte(&(EEPROM_Sheets_base->active_sheet)) == iSel);
 	{
+		bIsActive = (eeprom_read_byte(&(EEPROM_Sheets_base->active_sheet)) == iSel);
 	}
 	
 	SERIAL_PROTOCOLPGM("Sheet ");
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -7724,14 +7724,15 @@ Sigma_Exit:
   
   case 850: {
 	//! ### M850 - set sheet parameters
-	//! //!@n M850 - Set sheet data S[id] Z[offset] L[label] B[bed_temp] P[PINDA_TEMP]
-	bool bHasZ = false, bHasLabel = false, bHasBed = false, bHasPinda = false;
+	//! //!@n M850 - Set sheet data S[id] Z[offset] L[label] B[bed_temp] P[PINDA_TEMP] A[IS_ACTIVE]
+	bool bHasZ = false, bHasLabel = false, bHasBed = false, bHasPinda = false, bHasIsActive = false;
 	uint8_t iSel = 0;
 	int16_t zraw = 0;
 	float z_val = 0;
 	char strLabel[8];
 	uint8_t iBedC = 0;
 	uint8_t iPindaC = 0;
+	bool bIsActive=false;
 	strLabel[7] = '\0'; // null terminate.
 	size_t max_sheets = sizeof(EEPROM_Sheets_base->s)/sizeof(EEPROM_Sheets_base->s[0]);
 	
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: Firmware/Marlin_main.cpp [hunk 1]
```diff
@@ -7795,6 +7796,13 @@ SERIAL_PROTOCOLPGM("\n\n");
...
 	else
 		iPindaC = eeprom_read_byte(&EEPROM_Sheets_base->s[iSel].pinda_temp);
+	if (code_seen('A'))
+	{
+		bHasIsActive = true;
+		bIsActive = code_value_uint8() || (eeprom_read_byte(&(EEPROM_Sheets_base->active_sheet)) == iSel);
+	}
+	else
+		bIsActive = (eeprom_read_byte(&(EEPROM_Sheets_base->active_sheet)) == iSel);
 	{
```
wavexx, line 7808: @vintagepc this random {} block looks like a typo and should have been wrapping iPindaC I guess?

dawidpieper:
My oversight. Now I standardized the formatting in all the code responsible for this gCode.

Firmware/Marlin_main.cpp
```diff
@@ -7794,8 +7795,17 @@ SERIAL_PROTOCOLPGM("\n\n");
 		iPindaC = code_value_uint8();
 	}
 	else
+	{
 		iPindaC = eeprom_read_byte(&EEPROM_Sheets_base->s[iSel].pinda_temp);
+	}
```
wavexx, line 7800: would be nice to have an empty line like the other block

wavexx (approved):
Looks very sensible to me. @gudnimg what do you think?

Firmware/Marlin_main.cpp
```diff
@@ -7794,8 +7796,18 @@ SERIAL_PROTOCOLPGM("\n\n");
...
+	
+	if (code_seen('A'))
+	{
+		bHasIsActive = true;
+		bIsActive = code_value_uint8() || (eeprom_read_byte(&(EEPROM_Sheets_base->active_sheet)) == iSel);
+	}
+	else
 	{
+		bIsActive = (eeprom_read_byte(&(EEPROM_Sheets_base->active_sheet)) == iSel);
```
gudnimg, line 7809: I can't help feel that this code can be optimised more for code size. `(eeprom_read_byte(&(EEPROM_Sheets_base->active_sheet)) == iSel);` is duplicated in both `if` and `else`.

I am nitpicking here, I don't see any issue.

Firmware/Marlin_main.cpp
```diff
@@ -7794,8 +7796,18 @@ SERIAL_PROTOCOLPGM("\n\n");
 		iPindaC = code_value_uint8();
 	}
 	else
+	{
 		iPindaC = eeprom_read_byte(&EEPROM_Sheets_base->s[iSel].pinda_temp);
+	}
+	
+	if (code_seen('A'))
```
gudnimg, line 7802: Does it make sense to change this code into this?:

```
bIsActive = (eeprom_read_byte(&(EEPROM_Sheets_base->active_sheet)) == iSel);
if (code_seen('A'))
{
    bHasIsActive = true;
    bIsActive |= code_value_uint8()
}
```

And skip the `else`? If it uses less memory it would be great :)

Firmware/Marlin_main.cpp
```diff
@@ -7819,6 +7831,10 @@ SERIAL_PROTOCOLPGM("\n\n");
 	{
 		eeprom_update_byte(&EEPROM_Sheets_base->s[iSel].pinda_temp, iPindaC);
 	}
+	if (bHasIsActive)
+	{
+		if(bIsActive) eeprom_update_byte(&EEPROM_Sheets_base->active_sheet, iSel);
+	}
```
gudnimg: This can be optimised into one `if` statement

`if (bHasIsActive && bIsActive) eeprom_update_byte(&EEPROM_Sheets_base->active_sheet, iSel);`

I don't trust the compiler to do this optimisation everytime.

dawidpieper:
I made suggested optimizations, but as for the last if statement, I think the code may be a bit less readable now. Previously only conditions for bHasXXX were used and I think the compiler optimizes it well. But these are only my feelings. :)

gudnimg:
Code size seems to have increased according to our Github bot 🤔 I wonder why
image

dawidpieper:
I don't know AVR architecture as well. In fact I'm not sure if statement
bIsActive |= code_value_uint8() || (eeprom_read_byte(&(EEPROM_Sheets_base->active_sheet)) == iSel);
is calculated in RAM or CPU registers.
May it be the issue here?

gudnimg:
It's measured in Flash memory and SRAM. The if statement is in Flash memory, while global variables and `static` variables are in SRAM.

I usually stash my changes and check the current build stats:

Then re-apply my stash and re-build to check if the optimisation reduces code size (memory). We try to save memory whenever we can as we don't have much left of Flash (94% full).
image

gudnimg (approved):
Approving, but I have not tested the changes.

dawidpieper:
Thanks, original commit seemed to work on MK3S+. I have no access to printer now to check the last one, if no one overtakes me, I'll do it tomorrow or the day after tomorrow.

dawidpieper:
We can reduce flash usage by 14 bytes (MK3S) that way. Is it okay?

gudnimg:
> We can reduce flash usage by 14 bytes (MK3S) that way. Is it okay?

Sounds good 💪

dawidpieper:
Seems to work on MK3S plus with MMU2S

PR #4330, opened 2023-08-16
cmake: Set source epoch from the HEAD commit

wavexx:
Allow to set the source epoch of the build with PROJECT_VERSION_TIMESTAMP.

When unset fetch the commit timestamp of the current HEAD automatically instead of leaving the source as 0. This fixes the firmware DATE information in the support menu.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,10 @@ set(BUILD_NUMBER
     ""
     CACHE STRING "Build number of the firmware. Resolved automatically if not specified."
     )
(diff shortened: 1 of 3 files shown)


wavexx:
@gudnimg can you test this on vscode if everything works as expected?

gudnimg (approved):
Works well on my MK3S+ when using VScode and the CMake plugin.

I tried setting the value to exactly one year before: 1660670429 and the printer showed the correct date
image

BEFORE:
IMG_3829



AFTER:
IMG_3831

DRracer (approved):

PR #4332, opened 2023-08-16
G80: Fix false FINDA runout event for MK3S users

### LINKED ISSUE #1905: 3.8.1: MK3S/MMU2S randomly (?) ejects the filament at start of a print
Firmware: 3.7.0 bundle

1) print something on filament 1
2) load filament in 4 (not load to nozzle)
3) print something with filament 4
4) printer ejects filament **4** and asks me to replace it
5) it loads the filament and now has a blob around the extruder, ruining the print

and this case

1) print something on filament 1
2) load filament 5 (not load to nozzle)
3) print something with filament 5
4) printer ejects filament **1** and asks me to replace it
5) it loads the filament and now has a blob around the extruder, ruining the print

I have no idea why it does this. It loads the filament perfectly without cutting it. 

image
image

### LINKED ISSUE #1948: [3.7.1 MMU2S] Failed initial load causes mesh bed leveling to repeat
When starting a print sometimes the MMU2S will fail to detect filament and will attempt a spool join during the initial loading sequence. When the printer does not get a successful load, it will move the print head away from the starting point of the priming line (to the position of an M600?).

The MMU2S will select the next filament and feed it until it has purged the nozzle, leaving a pile of filament in the path of the purge line.

Then the MK3S will position the nozzle for the priming line, but redo the 7x7 mesh bed leveling. Once this is complete it will dump a ton of filament at the start, as if filament isn't already loaded.

This process is all done automatically, and if no one removes the pile of filament will cause a large glob stuck to the nozzle and heater block. This also causes grinding of the filament during the initial back and forth of the priming line where the filament is supposed to be making its way from the Bondtech gears to the nozzle since filament is already extruding and is blocking the path.

### LINKED ISSUE #2579: [QUESTION] Code weirdness in G80

### LINKED ISSUE #2602: [BUG] G28 or G80 moves E when temp is >=180

### LINKED ISSUE #3281: i3 MK3s+ &MMU2s wrong Filament Selection
PrusaSlicer
Version:   2.3.3+win64
Build:     PrusaSlicer-2.3.3+win64-202107161027

i3 MK3s+ & MMU2s
* <small>Prusa MK3s+ 3.9.3</small>
* <small>MMU2s 1.0.6</small>
* <small>Rpi 4B</small>
* <small>OctoPrint 1.6.1</small>
* <small>Python 3.7.3</small>
* <small>OctoPi 0.18.0</small>

This is really confusing Trouble I am facing. Regularly By downloading Gcode file over Octopi and/or also loading directly over SD Card, The Gcode Tool selection is correct but Physically it selects random slot of an other Filament of MMU2.
This morning finally I could catch the case as follow:

Example: 
> IR_Sensor_State_Window_0.15mm_PETG_MK3SMMU2S_2m.gcode shows 
`; generated by PrusaSlicer 2.3.3+win64 on 2021-09-25 at 06:13:20 UTC
; 
; external perimeters extrusion width = 0.45mm
; perimeters extrusion width = 0.45mm
; infill extrusion width = 0.45mm
; solid infill extrusion width = 0.45mm
; top infill extrusion width = 0.40mm
; first layer extrusion width = 0.42mm

M73 P0 R2
M73 Q0 S2
M201 X1000 Y1000 Z200 E8000 ; sets maximum accelerations, mm/sec^2
M203 X200 Y200 Z12 E120 ; sets maximum feedrates, mm/sec
M204 P1250 R1250 T1250 ; sets acceleration (P, T) and retract acceleration (R), mm/sec^2
M205 X8.00 Y8.00 Z0.40 E4.50 ; sets the jerk limits, mm/sec
M205 S0 T0 ; sets the minimum extruding and travel feed rate, mm/sec
M107
;TYPE:Custom
M862.3 P "MK3SMMU2S" ; printer model check
M862.1 P0.4 ; nozzle diameter check
M115 U3.10.0 ; tell printer latest fw version
M83 ; extruder relative mode
M104 S240 ; set extruder temp
M140 S70 ; set bed temp
M190 S70 ; wait for bed temp
M109 S240 ; wait for extruder temp
G28 W ; home all without mesh bed level
G80 ; mesh bed leveling

; Send the filament type to the MMU2.0 unit.
; E stands for extruder number, F stands for filament type (0: default; 1:flex; 2: PVA)
M403 E0 F0
M403 E1 F0
M403 E2 F0
M403 E3 F0
M403 E4 F0

;go outside print area
G1 Y-3.0 F1000.0
M73 P14 R2
M73 Q14 S2
G1 Z0.4 F1000.0
; select extruder
T3
; initial load
.......

`
> `T3` is correct, but Physically is Loading T1 on LCD Display showing F2 ` I made the test as well downloading from Octopi as well loading over SD Card. Behave in both cases the same way. It seems that somehow ensy or mmu board hangs. So thats the reason why I do not think it is octopi related. 

>  Over octopi once startet job tried to enter in therminal window T0 with no reaction then T2 right away changed filament to T0 tried as well  T3 and it changed back again to T1 confusing.... 

> After Restarting (Power OFF / On cycle ) everythings works again as expected. 

> checked on octopi that 5 Extruder with shared nozzle is configured. 

> modified filament sensor with LED stat to exclude Loading Filament Issues 

I have noticed that behave since really a long time and my turn arround was always before restarting a job to consequently Power Off/ On cycle, and up to know it always worked. 

My question it is permanently behaving again that way, but I could not reproduce it up to know. There somebody got me a tip where I might be look at? 

thanks a lot

### LINKED ISSUE #4323: MK3S/+ MMU2S start gcode issue with T[initial_tool] command

gudnimg:
Fixes a long-standing issue for MK3S (not MK3S+) users who use MMU. Also applies to MK3S printer which may have been upgraded to MK3S+. There was a short extrusion move at the end of G80 which may cause a FINDA filament runout if the EEPROM has PINDA temperature calibration set.

FINDA runout should now be completely ignored during Homing and Mesh Bed Leveling. Similarly to how the extruder filament sensor runout is ignored.

⚠️ A stock MK3S+ will not be able to reproduce the issue.

* Fixes https://github.com/prusa3d/Prusa-Firmware/issues/4323
* Fixes https://github.com/prusa3d/Prusa-Firmware/issues/3281
* Fixes https://github.com/prusa3d/Prusa-Firmware/issues/2602
* Fixes https://github.com/prusa3d/Prusa-Firmware/issues/2579
* Fixes https://github.com/prusa3d/Prusa-Firmware/issues/1948
* Fixes https://github.com/prusa3d/Prusa-Firmware/issues/1905

Change in memory:
Flash: -114 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -228,6 +228,10 @@ uint8_t newFanSpeed = 0;
   float retract_recover_length_swap = RETRACT_RECOVER_LENGTH_SWAP;
 #endif
 
+#ifndef PINDA_THERMISTOR
+static bool temp_compensation_retracted = false;
+#endif // !PINDA_THERMISTOR
+
   #ifdef PS_DEFAULT_OFF
     bool powersupply = false;
   #else
@@ -3159,11 +3163,14 @@ static void gcode_G80()
     //		SERIAL_ECHOLNPGM("Mesh bed leveling activated");
     go_home_with_z_lift();
     //		SERIAL_ECHOLNPGM("Go home finished");
+#ifndef PINDA_THERMISTOR
     //unretract (after PINDA preheat retraction)
-    if (((int)degHotend(active_extruder) > extrude_min_temp) && eeprom_read_byte((unsigned char *)EEPROM_TEMP_CAL_ACTIVE) && calibration_status_pinda() && (target_temperature_bed >= 50)) {
+    if (temp_compensation_retracted) {
+        temp_compensation_retracted = false;
         current_position[E_AXIS] += default_retraction;
         plan_buffer_line_curposXYZE(400);
     }
+#endif // !PINDA_THERMISTOR
     KEEPALIVE_STATE(NOT_BUSY);
     // Restore custom message state
     lcd_setstatuspgm(MSG_WELCOME);
@@ -10237,11 +10244,11 @@ void bed_analysis(float x_dimension, float y_dimension, int x_points_num, int y_
 
 #ifndef PINDA_THERMISTOR
 static void temp_compensation_start() {
-	
 	custom_message_type = CustomMsg::TempCompPreheat;
 	custom_message_state = PINDA_HEAT_T + 1;
 	lcd_update(2);
-	if ((int)degHotend(active_extruder) > extrude_min_temp) {
+	if (!temp_compensation_retracted && (int)degHotend(active_extruder) > extrude_min_temp) {
+		temp_compensation_retracted = true;
 		current_position[E_AXIS] -= default_retraction;
 	}
 	plan_buffer_line_curposXYZE(400);
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -528,6 +528,8 @@ bool check_fsensor() {
     return printJobOngoing()
         && mcode_in_progress != 600
         && !saved_printing
+        && !mesh_bed_leveling_flag
+        && !homing_flag
         && e_active();
 }
 


Firmware/Marlin_main.cpp
```diff
@@ -10263,6 +10264,15 @@ static void temp_compensation_start() {
 	custom_message_state = 0;
 }
 
+static void temp_compensation_stop()
+{
+    //unretract (after PINDA preheat retraction)
+    if (((int)degHotend(active_extruder) > extrude_min_temp) && eeprom_read_byte((unsigned char *)EEPROM_TEMP_CAL_ACTIVE) && calibration_status_pinda() && (target_temperature_bed >= 50)) {
+        current_position[E_AXIS] += default_retraction;
```
wavexx: There is still another potential race here if the extruder wasn't hot enough during temp_compensation_start() then the retraction is never done. Also similarly, the conditions done here do not match the conditions of the retract anyway.

I would change temp_compensation_start() to return a boolean indicating if the retract is done, then just unretract based on that instead of doing all this checking.

wavexx (changes requested):

Firmware/Marlin_main.cpp
gudnimg: I added another commit. How does it look now?

Firmware/Marlin_main.cpp
```diff
@@ -10236,13 +10242,14 @@ void bed_analysis(float x_dimension, float y_dimension, int x_points_num, int y_
...
 
 #ifndef PINDA_THERMISTOR
-static void temp_compensation_start() {
-	
+static bool temp_compensation_start() {
+	bool retracted = false;
 	custom_message_type = CustomMsg::TempCompPreheat;
 	custom_message_state = PINDA_HEAT_T + 1;
 	lcd_update(2);
 	if ((int)degHotend(active_extruder) > extrude_min_temp) {
```
wavexx: If we want to keep this global, then maybe also check for  temp_compensation_retracted here, so that if MBL fails we don't do it again..

gudnimg: Done. It is now global.

wavexx (approved):

gudnimg:
Rebased to merge 3 commits into 1. No code change.

wavexx (approved):

DRracer (approved):

PR #4342, opened 2023-08-20
optimisation: Timer constructor can be `constexpr`

gudnimg:
There could be more classes which can be optimised this way.

Change in memory:
Flash: -206 bytes
SRAM: -16 bytes


leptun (approved):
Looks like a sensible optimization to me. Don't see how this could go wrong in any way.

PR #4348, opened 2023-08-27
🚧 Improve handling of stepper inactivity timeout in `manage_inactivity()`

gudnimg:
If steppers are not allowed to be disabled when `no_stepper_sleep = true` then `manage_inactivity()` should enforce this. This wasn't being done in Marlin 1, but I noticed this is now done in Marlin 2.1 and I think it is a good improvement.

I also added a log for the event when the steppers are disabled.

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -123,7 +123,7 @@ bool is_buffer_empty();
 void process_commands();
 void ramming();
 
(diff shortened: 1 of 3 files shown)


gudnimg:
We can probably remove a lot of the calls to `refresh_cmd_timeout()` with this change because `st_synchronize()` will make sure the steppers are not disabled and refresh the timer.

PR #4355, opened 2023-08-30
lang: Do not hardcode python executable name

wavexx:
Use sys.executable instead to ensure the executable used by "env" is kept.

This fixes update-pot.py when run within a virtualenv and/or if "python" is not available (python3 is the default)


3d-gussner (approved):
Working on Linux and Windows

DRracer (approved):

PR #4356, opened 2023-08-30
lang/mmu: Translate MMU menu strings within caller

wavexx:
This fixes the undefined _T(label) reference, at the expense of a few extra bytes.

I would argue this is worth the cost for the ability to check translation references for the future. The warning happens because `lang-check` cannot check a reference which is not _directly_ a catalog entry.

We could introduce a method to suppress this warning (either a new macro or some ///IGNORE comment), but that would mean that the additional translation check is completely bypassed, defeating the purpose.


3d-gussner (approved):
Tested on MK404 and the menu is working as before also tested different language.

DRracer (approved):

PR #4372, opened 2023-09-05
Fix build issues when filament sensor is not defined

GilesBathgate:
In Firmware/variants/M*-RAMBo*.h there is a `#define FILAMENT_SENSOR` however the code doesn't compile when this define is removed. Maybe all printers have a filament sensor now but mine doesn't, So I would like to compile the firmware without it.

This PR fixes the compile issues in the way that seems to make sense.

### DIFF
--- a/Firmware/Filament_sensor.cpp
+++ b/Firmware/Filament_sensor.cpp
@@ -40,6 +40,7 @@ FSensorBlockRunout::FSensorBlockRunout() { }
 FSensorBlockRunout::~FSensorBlockRunout() { }
 #endif // FILAMENT_SENSOR
 
+#ifdef FILAMENT_SENSOR
 void Filament_sensor::setEnabled(bool enabled) {
     eeprom_update_byte((uint8_t *)EEPROM_FSENSOR, enabled);
     if (enabled) {
@@ -520,3 +521,4 @@ bool PAT9125_sensor::updatePAT9125() {
     return (filter == 0); // return stability
 }
 #endif // #if (FILAMENT_SENSOR_TYPE == FSENSOR_PAT9125)
+#endif // FILAMENT_SENSOR
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -5360,7 +5360,11 @@ void process_commands()
       starttime=_millis();
       if (MMU2::mmu2.Enabled())
       {
+#ifdef FILAMENT_SENSOR
         if (MMU2::mmu2.FindaDetectsFilament() && !fsensor.getFilamentPresent())
+#else
+        if (MMU2::mmu2.FindaDetectsFilament())
+#endif
         { // Filament only half way into the PTFE. Unload the filament.
           MMU2::mmu2.unload();
           // Tx and Tc gcodes take care of loading the filament to the nozzle.
--- a/Firmware/mmu2_fsensor.cpp
+++ b/Firmware/mmu2_fsensor.cpp
@@ -4,7 +4,11 @@
 namespace MMU2 {
 
 FilamentState WhereIsFilament(){
+#ifdef FILAMENT_SENSOR
     return fsensor.getFilamentPresent() ? FilamentState::AT_FSENSOR : FilamentState::NOT_PRESENT;
+#else
+    return FilamentState::NOT_PRESENT;
+#endif
 }
 
 } // namespace MMU2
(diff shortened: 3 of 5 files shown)


3d-gussner:
@GilesBathgate Thanks for the PR. I have seen you have a MK2 branch with FW 3.13.2. Is it working?

GilesBathgate:
> @GilesBathgate Thanks for the PR. I have seen you have a MK2 branch with FW 3.13.2. Is it working?

Yes but my MK2 seems to have deviated from stock somewhat 😄

3d-gussner:
> Yes but my MK2 seems to have deviated from stock somewhat

We have the MK2/S as obsolete, but I wonder if you could prepare a PR for these, so users that still have a MK2 (I have and will keep it as it is [with fsensor] see [here](https://github.com/3d-gussner/Prusa-Firmware/releases/tag/v3.1.0-001-HFM) ) to use it with newer custom build firmware?

3d-gussner:
@GilesBathgate Thanks, indeed it fixes the compile issue BUT there are also `FILAMENT_SENSOR_TYPE` without being within `#ifdef FILAMENT_SENSOR` . Also `#include "Filament_sensor.h"` may cause issues, especially with the MMU3.

So conclusion would be that disabling the `FILAMENT_SENSOR` also disables all related code like MMU and other code using fsensor functions.

I think there is way more work to do to find all fsensor related functions to prevent issues in custom firmware.

Removed milestone as it is not complete, sorry.

GilesBathgate:
> We have the MK2/S as obsolete, but I wonder if you could prepare a PR for these

The main commit (which gets my mk2 passed selftest at least) is https://github.com/prusa3d/Prusa-Firmware/commit/9a9d19686ca6890401e29f74d8035294575ef07e

GilesBathgate:
@3d-gussner Can you help me with my MK2 backport? I tried to fix the probe positions (old firmware uses diamond, new firmware needs the square) https://github.com/GilesBathgate/Prusa-Firmware/commit/9a9d19686ca6890401e29f74d8035294575ef07e

I get the selftest to pass, but I have to add: `current_position[Y_AXIS] += 2.2;` :man_shrugging: 

I am trying to get Calibrate XYZ to pass, currently I get 'compromised left front is unreachable', and my Y offset is -1.5 or so :disappointed:

GilesBathgate:
@3d-gussner So I think the 2.2 comes from Y_MIN_POS, in MK3 its -4, in MK2 its 2.2

3d-gussner:
@GilesBathgate I will close this PR as `FILAMENT_SENSOR` is important for the MMU3, which can be de- and activated while running.
Cleaning up the `FILAMENT_SENSOR` would take way more time to also disable the MMU3 functions.

The easiest way to have a MK2.0 version is to disable the FSensor in the menu or EEPROM.

GilesBathgate:
@3d-gussner I think i couldn't get through the self-test. So applying this to my specific MK2 firmware was the solution that worked for me. The reason I created the pull request was simply to contribute it back to the community. This is because the `FILAMENT_SENSOR` define is clearly broken. Its misleading and should be removed or fixed. However its up to you and the community to decide, and I am happy with the decision.

3d-gussner:
@GilesBathgate I agree that the `FILAMENT_SENSOR` define isn't anymore valid everywhere. But when I reviewed your PR I have found so many MMU3 functions that may break trying fixing. It will take a lot of effort and testing to get it working.

As an alternative it would be easier to "fix" selftest to skip the FSensor check when it is disabled.

For factory reset and wizard that would mean we would need to have a define `FILAMENT_SENSOR_ACTIVE` or so to turn it on or off during this reset function and passing the Wizard.

Nowadays a 3D printer without a Filament Sensor is "outdated" and adding a FSensor to the MK2 isn't that hard. I did that few years ago in the MK2 branch.

I still hope to find few volunteers to help to get MK2.1 to MK2.4 working with MK3 branch, see https://github.com/prusa3d/Prusa-Firmware/issues/4604

GilesBathgate:
@3d-gussner Yes, a `FILAMENT_SENSOR_ACTIVE` that just sets the default as off, would be a much better approach. Then just add a comment to `FILAMENT_SENSOR` that indicates to use the other.

3d-gussner:
> @3d-gussner Yes, a `FILAMENT_SENSOR_ACTIVE` that just sets the default as off, would be a much better approach. Then just add a comment to `FILAMENT_SENSOR` that indicates to use the other.

Glad you agree on that. Again thanks for the idea and your work on the MK2.

PR #4381, opened 2023-09-09
Refine CMake version handling based on tags

vintagepc:
- Sets version information from git tags, if available. Otherwise, falls back to Configuration.h
- Attempts to detect repository owner
- add option `ENFORCE_VERSION_MATCH` to cause a failure if Configuration.h and the git tag do not match.  Default: Off
- Add option `NO_TAG_IS_FATAL` to cause an error if no git tag could be properly parsed Default: Off


3d-gussner (changes requested):
To fix a download zip without a git clone we need to fix the `CMake Error: Cannot parse SOURCE_DATE_EPOCH as **integer** ` error 
```
     git_head_commit_timestamp(timestamp)
     set(ERRORS "GIT-NOTFOUND" "HEAD-FORMAT-NOTFOUND" "HEAD-HASH-NOTFOUND")
     if(timestamp IN_LIST ERRORS)
       # git not availibe, set fallback values
       set(timestamp 1456063320)
     endif()
     set(PROJECT_VERSION_TIMESTAMP
```

sarusani:
A lot of the hex files info is still based on configuration.h, even if git & tag is available.

configuration.h:
```
#define FW_MAJOR 3
#define FW_MINOR 13
#define FW_REVISION 0
#define FW_COMMITNR 6853
#define FW_FLAVOR RC
#define FW_FLAVERSION 1
```
git tag: `v3.13.3`
git commit nr: `7696`
git sha-short: `363c53af6`

The build info from cmake looks correct:
```
[build] -- Found a defined FW_FLAVOR
[build] -- Project version (Configuration.h): 3.13.0.25
[build] -- Project version suffix ..........: 363c53af6
[build] -- Project version description......: v3.13.3-7696
[build] -- Configuration.h and tag (not enforced): (3.13.0.25/3.13.3.64)
[build] -- Commit Nr: Configuration.h: 6853 Tag: 7696
[build] -- These tag values will override Configuration.h
[build] remote URL is https://github.com/sarusani/Prusa-Firmware.git
[build] Found repository name sarusani
```

But the generated hex file doesn't only use the git & tag info, it's a mixed result.

Filename: (uses configuration.h version)
FW3.13.***0***.***25***+363c53af6-MK3S_MULTILANG
Expected:
FW3.13.***3***.***64***+363c53af6-MK3S_MULTILANG

LCD ->Bootup Firmware Version:  (uses configuration.h version)
3.13.***0***
Expected:
3.13.***3***

LCD ->Support: (uses git & tag info)
***v***3.13.3-7696
repo sarusani
Expected: (not sure about this, but the leading v from the tag was removed in other places, so maybe it should be also removed on this screen)
3.13.3-7696
repo sarusan

vintagepc:
Filenames are deliberately not yet supported due to further planned changes there. 

The leading v  is an oversight as it's part of the tag and the bootup version is probably doing something silly/different or is not properly inheriting its #includes, because the CMake #defines have the same names as those they override in Configuration.h 🤔

vintagepc:
Don't see anything obvious on the boot screen issue. It is using the same FW_VERSION macro every other place does that doesn't include the commit number :shrug:

sarusani:
Sorry to be so annoying! 😄
Latest commit breaks build if FW_FLAVOR is not set in configuration.h.

vintagepc:
That's what I get for coding at late hours 😛

3d-gussner (approved):
Fixing some minor issues in next PR

PR #4383, opened 2023-09-11
Sync MK3<->MK4 MMU2

DRracer:
This is a collection of commits towards unifying the code base on MK3 and MK4.

Best reviewed per commit as there are many small changes scattered around the MMU-related code base.

### DIFF
--- a/Firmware/mmu2.cpp
+++ b/Firmware/mmu2.cpp
@@ -115,7 +115,7 @@ void MMU2::ResetX0() {
     logic.ResetMMU(); // Send soft reset
 }
 
-void MMU2::ResetX42(){
+void MMU2::ResetX42() {
     logic.ResetMMU(42);
 }
 
@@ -453,7 +453,7 @@ uint8_t MMU2::get_tool_change_tool() const {
 bool MMU2::set_filament_type(uint8_t /*slot*/, uint8_t /*type*/) {
     if (!WaitForMMUReady())
         return false;
-    
+
     // @@TODO - this is not supported in the new MMU yet
     //    slot = slot; // @@TODO
     //    type = type; // @@TODO
@@ -927,7 +927,7 @@ void MMU2::execute_extruder_sequence(const E_Step *sequence, uint8_t steps) {
     planner_synchronize();
 
     const E_Step *step = sequence;
-    for (uint8_t i = steps; i ; --i) {
+    for (uint8_t i = steps; i > 0; --i) {
         extruder_move(pgm_read_float(&(step->extrude)), pgm_read_float(&(step->feedRate)));
         step++;
     }
@@ -980,6 +980,7 @@ void MMU2::ReportError(ErrorCode ec, ErrorSource res) {
             IncrementMMUFails();
 
             // check if it is a "power" failure - we consider TMC-related errors as power failures
+            // clang-format off
             static constexpr uint16_t tmcMask =
                 ( (uint16_t)ErrorCode::TMC_IOIN_MISMATCH
                 | (uint16_t)ErrorCode::TMC_RESET
@@ -988,6 +989,7 @@ void MMU2::ReportError(ErrorCode ec, ErrorSource res) {
                 | (uint16_t)ErrorCode::TMC_OVER_TEMPERATURE_WARN
                 | (uint16_t)ErrorCode::TMC_OVER_TEMPERATURE_ERROR
                 | (uint16_t)ErrorCode::MMU_SOLDERING_NEEDS_ATTENTION ) & 0x7fffU; // skip the top bit
+            // clang-format on
             static_assert(tmcMask == 0x7e00); // just make sure we fail compilation if any of the TMC error codes change
 
             if ((uint16_t)ec & tmcMask) { // @@TODO can be optimized to uint8_t operation
@@ -1000,7 +1002,7 @@ void MMU2::ReportError(ErrorCode ec, ErrorSource res) {
     if (!mmu2.RetryIfPossible(ec)) {
         // If retry attempts are all used up
         // or if 'Retry' operation is not available
-        // raise the MMU error sceen and wait for user input
+        // raise the MMU error screen and wait for user input
         ReportErrorHook((CommandInProgress)logic.CommandInProgress(), ec, uint8_t(lastErrorSource));
     }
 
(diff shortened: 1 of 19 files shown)


Firmware/mmu2.cpp
```diff
@@ -295,42 +285,23 @@ bool MMU2::VerifyFilamentEnteredPTFE() {
...
-            }
+            filament_inserted = filament_inserted && (WhereIsFilament() == FilamentState::AT_FSENSOR);
+            tlur.Progress(filament_inserted);
             safe_delay_keep_alive(0);
         }
     }
-
-    Disable_E0();
-    TryLoadUnloadProgressbarEcho();
-    TryLoadUnloadProgressbarDeinit();
```
gudnimg, line 325: I think these two lines are removed by mistake. 

Echo will log the result onto serial. 

Deinit will signal to the LCD code that the next status message should be rendered from the beginning.

gudnimg (changes requested):

Firmware/mmu2.cpp
DRracer: Good point, you are right, I missed the serial dump... I'll probably push it into the destructor of the `TryLoadUnloadReporter`

DRracer:
rebased onto MK3

Firmware/mmu2_protocol_logic.cpp
```diff
@@ -9,36 +9,43 @@
...
+/// Beware - on AVR/MK3S:
 /// Changing the supportedMmuVersion numbers requires patching MSG_DESC_FW_UPDATE_NEEDED and all its related translations by hand.
 ///
 /// The message reads:
 ///   "MMU FW version is incompatible with printer FW.Update to version 2.1.9."
 ///
 /// Currently, this is not possible to perform automatically at compile time with the existing languages/translations infrastructure.
 /// To save space a "dumb" solution was chosen + a few static_assert checks in errors_list.h preventing the code from compiling when the string doesn't match.
+/// -----
+/// On Buddy FW we should improve the error screen to be able to print formatted strings
```
gudnimg: Does this TODO need to be in the code? I feel it would be better to create a JIRA ticket for this as it only relates to the 32-bit.

Firmware/mmu2_reporting.h
```diff
@@ -2,12 +2,19 @@
...
+    #include "../../../../../../Prusa-Firmware-MMU/src/logic/error_codes.h"
+    #include "../../../../../../Prusa-Firmware-MMU/src/logic/progress_codes.h"
+#endif
 
 namespace MMU2 {
 
 enum CommandInProgress : uint8_t {
     NoCommand = 0,
-    CutFilament = 'C',
+    CutFilament = 'K',
```
gudnimg, line 17: Nice 👍

Firmware/mmu2.cpp
```diff
@@ -295,42 +284,23 @@ bool MMU2::VerifyFilamentEnteredPTFE() {
...
-                TryLoadUnloadProgressbar(lcd_cursor_col++, fsensorStateLCD);
-                fsensorStateLCD = 0;      // Clear temporary bit
-            }
+            filament_inserted = filament_inserted && (WhereIsFilament() == FilamentState::AT_FSENSOR);
+            tlur.Progress(filament_inserted);
             safe_delay_keep_alive(0);
         }
     }
-
-    Disable_E0();
```
gudnimg, line 323: It would be nice to keep this `Disable_E0();` It is relevant for single color prints. Tx/Tc will execute this code before heating is complete. So disabling the E-motor sooner will be a better user experience. The printer is more silent while heating.

gudnimg:
Code changes look great, just minor comments above.
Will test this PR on hardware tomorrow before approving👍 


**EDIT:**

I didn't notice before but it seems this PR is reverting the improvements in this PR: https://github.com/prusa3d/Prusa-Firmware/pull/4080

Note the changes I made to `TryLoadUnloadProgressbarDeinit`, `TryLoadUnloadProgressbarInit` and `TryLoadUnloadProgressbar` in 3.13.2 to keep the status screen rendering without problems. This solved a few problems with the LCD rendering.

It looks like this PR code is based on the code before PR 4080

Firmware/mmu2_reporting.cpp
```diff
@@ -295,30 +295,32 @@ void ReportProgressHook(CommandInProgress cip, ProgressCode ec) {
...
 }
 
-void TryLoadUnloadProgressbarDeinit() {
-    // Delay the next status message just so
-    // the user can see the results clearly
+void TryLoadUnloadReporter::Render(uint8_t col, bool sensorState) {
+    // Set the cursor position each time in case some other
+    // part of the firmware changes the cursor position
+    lcd_putc_at(col, 3, sensorState ? '-' : LCD_STR_SOLID_BLOCK[0]);
     lcd_reset_status_message_timeout();
```
gudnimg, line 314: Note this line is still missing `lcd_reset_status_message_timeout()`

I'm not sure if this should be used in `TryLoadUnloadReporter::Render()`

gudnimg, line 314: Here we should use `if (!lcd_update_enabled) lcdui_print_status_line();` this is hack to allow showing the progress bar in fullscreens like "Loading Filament 1".

gudnimg:
Compiler warnings
image

This below fixes it:

image

Firmware/mmu2.cpp
```diff
@@ -295,42 +284,23 @@ bool MMU2::VerifyFilamentEnteredPTFE() {
...
-
-            // Always round up, you can only have 'whole' pixels. (floor is also an option)
-            dpixel1 = ceil((stepper_get_machine_position_E_mm() - planner_get_current_position_E()) * pixel_per_mm);
-            if (dpixel1 - dpixel0) {
-                dpixel0 = dpixel1;
-                if (lcd_cursor_col > (LCD_WIDTH - 1)) lcd_cursor_col = LCD_WIDTH - 1;
-                TryLoadUnloadProgressbar(lcd_cursor_col++, fsensorStateLCD);
-                fsensorStateLCD = 0;      // Clear temporary bit
-            }
+            filament_inserted = filament_inserted && (WhereIsFilament() == FilamentState::AT_FSENSOR);
```
gudnimg, line 294: The logic is flipped compared to previous solution in 3.13.2.

This may be a good change in scenario when the filament sensor is reporting a false positive 0 value. 3.13.2 would in that case always fail the test (from one sample). But with this change, all the samples for a given LCD pixel must now see 0 values from the filament sensor. Could be useful for users who have the filament sensor flickering occasionally if sampled very fast.

gudnimg:
Just for reference: a PR with my proposed changes https://github.com/DRracer/Prusa-Firmware/pull/80

gudnimg (approved):

PR #4387, opened 2023-09-13
MBL: Rename repeat param from R to C to avoid overlap

sarusani:
The param R was used twice which meant that setting the amounts of probe repeats clashed with the setting of the manual bed level correction. This renames the param for probe retries to C.

- [ ] Update RepRap wiki after release

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -2825,9 +2825,9 @@ static void gcode_G80()
         nMeasPoints = 3;
     }
 
-    uint8_t nProbeRetry = code_seen('C') ? code_value_uint8() : eeprom_read_byte((uint8_t*)EEPROM_MBL_PROBE_NR);
-    if (nProbeRetry > 10) {
-        nProbeRetry = 10;
+    uint8_t nProbeRetryCount  = code_seen('C') ? code_value_uint8() : eeprom_read_byte((uint8_t*)EEPROM_MBL_PROBE_NR);
+    if (nProbeRetryCount  > 10) {
+        nProbeRetryCount  = 10;
     }
 
     const float area_min_x = code_seen('X') ? code_value() - x_mesh_density - X_PROBE_OFFSET_FROM_EXTRUDER : -INFINITY;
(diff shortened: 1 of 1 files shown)


leptun:
Damn. I didn't notice this overlap with the parameters. Maybe we should change one of them to some other letter now that we are making many MBL changes. I propose we change the retry count argument from `R` to `C` (from count) and leave the clamping to 10 as it is now. @3d-gussner what do you think about this?

sarusani:
~~That was actually my first change and I almost created a PR for it, but then I saw that Marlin already uses R for repeat in G29. So it would be better to change the R param for bed level correction. But that would break backwards compatibility...~~

~~That's way I just opted for this small change 😺~~

PS: After reading G29 documentation again... It's a complete different meaning from this, so I don't think it has to be considered. I'll adapt the PR

leptun:
Can you link where marlin handles G80 R parameter? Is this a prusa thing or mainline marlin?

Firmware/Marlin_main.cpp
```diff
@@ -2825,7 +2825,7 @@ static void gcode_G80()
         nMeasPoints = 3;
     }
 
-    uint8_t nProbeRetry = code_seen('R') ? code_value_uint8() : eeprom_read_byte((uint8_t*)EEPROM_MBL_PROBE_NR);
+    uint8_t nProbeRetry = code_seen('C') ? code_value_uint8() : eeprom_read_byte((uint8_t*)EEPROM_MBL_PROBE_NR);
```
3d-gussner: Maybe also change `nProbeRetry` to `nProbeCount` as well

Firmware/Marlin_main.cpp
```diff
@@ -4882,11 +4882,11 @@ void process_commands()
     Default 3x3 grid can be changed on MK2.5/s and MK3/s to 7x7 grid.
     #### Usage
 	  
-          G80 [ N | R | L | R | F | B | X | Y | W | H ]
+          G80 [ N | C | O | L | R | F | B | X | Y | W | H ]
       
 	#### Parameters
       - `N` - Number of mesh points on x axis. Default is 3. Valid values are 3 and 7.
-      - `R` - Probe retries. Default 3 max. 10
+      - `C` - Probe retries. Default 3 max. 10
```
3d-gussner: Probe counts

leptun (approved):
Lgtm

3d-gussner (changes requested):

leptun:
I agree with @3d-gussner's requests

Firmware/Marlin_main.cpp
sarusani: Maybe nProbeRetryCount would be better? It will still do 1 Probe if set to 0. (2 if set to 1 etc.)

Firmware/Marlin_main.cpp
sarusani: Same as above: "Probe retry counts"?

3d-gussner (approved):

3d-gussner:
@sarusani Thanks :medal_sports:

Issue #2708, opened 2020-06-02
[QUESTION] M-Code to query stealth/normal mode?


hergtoler:
I've searched, and the closest thing I can find is M914/M915 to set the mode.  I only want to query what the current mode is.  Is there a code that allows this?

leptun:
No, there isn’t a gcode for that.
If you understand the firmware, you can easily mod this in to some unused Mcode value.
The variable that says what mode is selected is tmc2130_mode
You implement the command in process_commands()

hergtoler:
Yeah, easy enough.
#2709

3d-gussner:
Duplicate of https://github.com/prusa3d/Prusa-Firmware/issues/2477 or not?

hergtoler:
The question has been answered, and options for implementation are being discussed.

PR #4388, opened 2023-09-13
Add Q parameter to M91[4/5]

vintagepc:
This adds a Q option to `M91[4|5]` to facilitate reporting the current mode. 

Note it does _not_ check whether an M91x is in effect (i.e. eeprom value != SRAM). It only reports according to the effective value in SRAM. 
This obsoletes #2709 and implements #2708

Strings are deliberately NOT translated in the serial output to facilitate easier parsing of these values. They are always in English. 

- [ ] Update RepRap wiki after release

Example test cases (with printer starting in normal mode:

```
M915
M915 Q
ok
Mode: Silent
ok

M914
M914 Q
ok
Mode: Normal
ok

<Normal on display>
M914 Q
Mode: Normal
ok

<Change to Stealth on display>
M914 Q
Mode: Silent
ok
```


3d-gussner (approved):
Tested on MK404 and works

PR #4390, opened 2023-09-14
MBL: Add magnet compensation parameter

sarusani:
If the user selected 3x3 in the LCD menu he can't set if he wants to use magnet compensation or not. (It's only available for 7x7)
So I added a magnet compensation parameter (M). It is only in use if a 7x7 mesh bed leveling is performed.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3053,7 +3053,8 @@ static void gcode_G80()
 
     mbl.upsample_3x3(); //interpolation from 3x3 to 7x7 points using largrangian polynomials while using the same array z_values[iy][ix] for storing (just coppying measured data to new destination and interpolating between them)
 
-    if (nMeasPoints == 7 && eeprom_read_byte((uint8_t*)EEPROM_MBL_MAGNET_ELIMINATION)) {
+    uint8_t useMagnetCompensation = code_seen('M') ? code_value_uint8() : eeprom_read_byte((uint8_t*)EEPROM_MBL_MAGNET_ELIMINATION);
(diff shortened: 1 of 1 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -4882,12 +4883,13 @@ void process_commands()
...
     #### Usage
 	  
-          G80 [ N | C | O | L | R | F | B | X | Y | W | H ]
+          G80 [ N | C | O | M | L | R | F | B | X | Y | W | H ]
       
 	#### Parameters
       - `N` - Number of mesh points on x axis. Default is 3. Valid values are 3 and 7.
       - `C` - Probe retry counts. Default 3 max. 10
       - `O` - Return to origin. Default 1 (true)
+      - `M` - Use magnet compensation. Will only be used if number of mesh points is set to 7. Default is value stored in EEPROM. Vaild values are 0 (false) and 1 (ture).
```
gudnimg: Minor nitpick:  `Vaild` => `Valid`

gudnimg (approved):

gudnimg (approved):

leptun (approved):
Lgtm, but I'm a bit on the fence about choosing `M` as the argument letter. Don't really like that part, but it should work now that only the first letter in the command is interpreted as the "command", so `M` is properly identified as an argument.

sarusani:
I feel the same way. I just wasn't able to come up with something else that remotely makes sense and doesn't clash with the other arguments. I was thinking about E for "elimination". But outside of the code, the feature is always referred to as "magnet compensation".

3d-gussner:
@sarusani Had to fix the merge conflict as I merged 4391 first.

PR #4391, opened 2023-09-14
MBL: Default to EEPROM values

sarusani:
"Probe retry counts" didn't default to 3, it defaulted to the value stored in EEPROM. (Just like bed leveling corrections)
This is now also true for "Number of mesh points" and stated in the usage doc.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -2821,12 +2821,14 @@ static void gcode_G80()
 #endif //PINDA_THERMISTOR
 
     uint8_t nMeasPoints = eeprom_read_byte((uint8_t*)EEPROM_MBL_POINTS_NR);
-    if (code_seen('N') && (code_value_uint8() == 7 || code_value_uint8() == 3))
-      nMeasPoints = code_value_uint8();
-    
+    bool bCodeSeen = code_seen('N');
+    if (uint8_t value = code_value_uint8(); bCodeSeen && (value == 7 || value == 3))
+      nMeasPoints = value;
+
     uint8_t nProbeRetryCount = eeprom_read_byte((uint8_t*)EEPROM_MBL_PROBE_NR);
-    if (code_seen('C') && code_value_uint8() >= 1 && code_value_uint8() <= 10)
-      nProbeRetryCount = code_value_uint8();
+    bCodeSeen = code_seen('C');
+    if (uint8_t value = code_value_uint8(); bCodeSeen && value >= 1 && value <= 10)
+      nProbeRetryCount = value;
 
     const float area_min_x = code_seen('X') ? code_value() - x_mesh_density - X_PROBE_OFFSET_FROM_EXTRUDER : -INFINITY;
     const float area_min_y = code_seen('Y') ? code_value() - y_mesh_density - Y_PROBE_OFFSET_FROM_EXTRUDER : -INFINITY;
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -2821,13 +2821,11 @@ static void gcode_G80()
 #endif //PINDA_THERMISTOR
 
     uint8_t nMeasPoints = eeprom_read_byte((uint8_t*)EEPROM_MBL_POINTS_NR);
-    bool bCodeSeen = code_seen('N');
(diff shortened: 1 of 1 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -2820,15 +2820,13 @@ static void gcode_G80()
...
-    if (nProbeRetryCount > 10) {
-        nProbeRetryCount = 10;
-    }
+    uint8_t nMeasPoints = eeprom_read_byte((uint8_t*)EEPROM_MBL_POINTS_NR);
+    if (code_seen('N') && (code_value_uint8() == 7 || code_value_uint8() == 3))
+      nMeasPoints = code_value_uint8();
+    
+    uint8_t nProbeRetryCount = eeprom_read_byte((uint8_t*)EEPROM_MBL_PROBE_NR);
+    if (code_seen('C') && code_value_uint8() >= 1 && code_value_uint8() <= 10)
+      nProbeRetryCount = code_value_uint8();
```
gudnimg: This code can be optimised more :) 

This below shaves off 14 bytes, so the PR only adds 6 bytes instead of 20

image

gudnimg: This way `code_value_uint8()` is called less frequently and value scope is limited to the if statement

gudnimg (approved):
Approving as I don't see anything wrong. Though would be nice to optimise the code a bit more.

Firmware/Marlin_main.cpp
sarusani: code_seen needs to be called before code_value. It sets the pointer. So I had to change it up a bit. Still saves 12 bytes.

gudnimg (approved):

Firmware/Marlin_main.cpp
```diff
@@ -2820,15 +2820,15 @@ static void gcode_G80()
     run = false;
 #endif //PINDA_THERMISTOR
 
-    uint8_t nMeasPoints = code_seen('N') ? code_value_uint8() : eeprom_read_byte((uint8_t*)EEPROM_MBL_POINTS_NR);
-    if (nMeasPoints != 7) {
-        nMeasPoints = 3;
-    }
+    uint8_t nMeasPoints = eeprom_read_byte((uint8_t*)EEPROM_MBL_POINTS_NR);
+    bool bCodeSeen = code_seen('N');
```
leptun: Hmm, don't really like that this variable is shared between the two arguments parsing. Could the code_seen be used directly in the if statement and avoid declaring bCodeSeen in the first place?

Firmware/Marlin_main.cpp
```diff
@@ -2820,15 +2820,15 @@ static void gcode_G80()
...
+    if (uint8_t value = code_value_uint8(); bCodeSeen && (value == 7 || value == 3))
```
leptun: ngl, this is some weird syntax, but if it results in more optimal code, no problem with this :)

leptun (approved):

Firmware/Marlin_main.cpp
sarusani: If it's used in the if statement, it's has to be executed after the value declaration. (You can only do one declaration inside the if statement and it has to be the first thing you do). That won't work because code_value_uint8 has to be called after code_seen so the pointer is set beforehand.

I would have preferred to do:
```
bool bCodeSeen;

uint8_t nMeasPoints = eeprom_read_byte((uint8_t*)EEPROM_MBL_POINTS_NR);
bCodeSeen = code_seen('N');
...
```
To make it look proper, but somehow that leads to higher mem usage too... 🤷 😄

Firmware/Marlin_main.cpp
gudnimg: It should be legal in C++17 https://en.cppreference.com/w/cpp/language/if

Firmware/Marlin_main.cpp
sarusani: I could do:
```
uint8_t nMeasPoints = eeprom_read_byte((uint8_t*)EEPROM_MBL_POINTS_NR);
if (uint8_t bCodeSeen = code_seen('N'), value = code_value_uint8(); bCodeSeen && (value == 7 || value == 3))
  nMeasPoints = value;
```
Also not super sexy, but same mem usage.

PR #4396, opened 2023-09-18
PFW-1538: Add M420

3d-gussner:


gudnimg:
Looking at https://marlinfw.org/docs/gcode/M420.html should we use `V` parameter to print this data? 

`M420 V` 

Not sure if the reprap wiki is up to date for Marlin. 

Marlin source: https://github.com/MarlinFirmware/Marlin/blob/bugfix-2.1.x/Marlin/src/gcode/bedlevel/M420.cpp

DRracer (approved):

PR #4399, opened 2023-09-20
Add M125

3d-gussner:
Added M125 with few parameters to set XYZ positions.

There have been requests by the community to allow to change the pause/park position.

During testing the pausing a XY crash would recover the print without reseting the pause state.
Fixed it by checking if printer is in pausing/paused state before trying the recovery.
If the hotend crashes during pausing it will try to home X Y only and not pause at PAUSE POS but recovers correctly.
Tested on MK404 with `lite with Quad_HR`
- Normal pause :heavy_check_mark:
- `M125 X254 Y212.5 Z210` :heavy_check_mark:
- `M125 X254 Y212.5 Z210` and causing a X crash during printer pauses :heavy_check_mark:

The print always continues where it paused in first place, even with the forced X crash, 

Todo
- [X] Check XYZ don't exceed MIN MAX XY positions and limits
- [ ] Update RepRap Wiki

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -232,7 +232,6 @@ extern int feedmultiply;
 extern int extrudemultiply; // Sets extrude multiply factor (in percent) for all extruders
 extern float extruder_multiplier[EXTRUDERS]; // reciprocal of cross-sectional area of filament (in square millimeters), stored this way to reduce computational burden in planner
 extern float current_position[NUM_AXIS] ;
-extern float pause_position[3];
 extern float destination[NUM_AXIS] ;
 extern float min_pos[3];
 extern float max_pos[3];
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -232,6 +232,7 @@ extern int feedmultiply;
 extern int extrudemultiply; // Sets extrude multiply factor (in percent) for all extruders
 extern float extruder_multiplier[EXTRUDERS]; // reciprocal of cross-sectional area of filament (in square millimeters), stored this way to reduce computational burden in planner
 extern float current_position[NUM_AXIS] ;
+extern float pause_position[3];
 extern float destination[NUM_AXIS] ;
 extern float min_pos[3];
 extern float max_pos[3];
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -7705,21 +7705,15 @@ Sigma_Exit:
     case 125:
     case 601:
     {
-        float temp_pause_position[3] = { X_PAUSE_POS, Y_PAUSE_POS, Z_PAUSE_LIFT };
+        // Set default values
+        pause_position[X_AXIS] = X_PAUSE_POS;
+        pause_position[Y_AXIS] = Y_PAUSE_POS;
+        pause_position[Z_AXIS] = Z_PAUSE_LIFT;
+
         for (uint8_t axis = 0; axis < E_AXIS; axis++) {
-          if (code_seen(axis_codes[axis])) temp_pause_position[axis] = code_value();
-        }
-        //Check that the X pause position is within printer limits
-        if ((temp_pause_position[X_AXIS] >= X_MIN_POS) && (temp_pause_position[X_AXIS] <= X_MAX_POS-1)) {
-            pause_position[X_AXIS] = temp_pause_position[X_AXIS];
+          if (code_seen(axis_codes[axis])) {
+            pause_position[axis] = constrain(code_value(), min_pos[axis], max_pos[axis]);
           }
-        //Check that the Y pause position is within printer limits
-        if ((temp_pause_position[Y_AXIS] >= Y_MIN_POS) && (temp_pause_position[Y_AXIS] <= Y_MAX_POS-1)) {
-          pause_position[Y_AXIS] = temp_pause_position[Y_AXIS];
-        }
-        //Check that the Z lift position is within printer limits
-        if (temp_pause_position[Z_AXIS] <=Z_MAX_POS) {
-          pause_position[Z_AXIS] = temp_pause_position[Z_AXIS];
         }
 
         if (code_seen('S')) {
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -200,6 +200,8 @@ float min_pos[3] = { X_MIN_POS, Y_MIN_POS, Z_MIN_POS };
 float max_pos[3] = { X_MAX_POS, Y_MAX_POS, Z_MAX_POS };
 bool axis_known_position[3] = {false, false, false};
 
+float pause_position[3] = { X_PAUSE_POS, Y_PAUSE_POS, Z_PAUSE_LIFT };
+
 uint8_t fanSpeed = 0;
 uint8_t newFanSpeed = 0;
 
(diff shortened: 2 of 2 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -7601,16 +7603,50 @@ SERIAL_PROTOCOLPGM("\n\n");
...
+     - `Z` - Z raise before park (otherwise Z_PAUSE_LIFT 20) these are saved until change or reset.
+     - `S` - Set values [S0 = set to default values | S1 = sset values only without pausing]
     */
     /*!
     ### M25 - Pause SD print <a href="https://reprap.org/wiki/G-code#M25:_Pause_SD_print">M25: Pause SD print</a>
     */
     case 25:
+    case 125:
     case 601:
     {
```
gudnimg, line 7720: Assigning `uint8_t` values to `float` array? This seems counterintuitive to me.

Firmware/Marlin_main.cpp
```diff
@@ -7601,16 +7603,50 @@ SERIAL_PROTOCOLPGM("\n\n");
...
+        if (code_seen('X')) pause_position[1] = code_value();
+        if (code_seen('Y')) pause_position[2] = code_value();
+        if (code_seen('Z')) pause_position[3] = code_value();
```
gudnimg: This saves 22 bytes:

```c
for (uint8_t axis = 0; axis < E_AXIS; axis++)
{
      if (code_seen(axis_codes[axis])) pause_position[axis] = code_value();
}
```

Firmware/Marlin_main.cpp
```diff
@@ -7601,16 +7603,50 @@ SERIAL_PROTOCOLPGM("\n\n");
...
+        if (code_seen('S')) {
+            if ( code_value_uint8() == 0 ) {
+                pause_position[1] = X_PAUSE_POS;
+                pause_position[2] = Y_PAUSE_POS;
+                pause_position[3] = Z_PAUSE_LIFT;
```
gudnimg: `pause_position[3]` is accessing memory outside the array. I'm a bit surprised there is no compiler warning raised.

Code probably should be:

```c
pause_position[X_AXIS] = X_PAUSE_POS;
pause_position[Y_AXIS] = Y_PAUSE_POS;
pause_position[Z_AXIS] = Z_PAUSE_LIFT;
```

Firmware/Marlin_main.cpp
3d-gussner: Can't do that as I now check the limits of the printer

Firmware/Marlin_main.cpp
```diff
@@ -7638,16 +7640,72 @@ SERIAL_PROTOCOLPGM("\n\n");
...
     /*!
     ### M25 - Pause SD print <a href="https://reprap.org/wiki/G-code#M25:_Pause_SD_print">M25: Pause SD print</a>
     */
     case 25:
+    case 125:
     case 601:
     {
+        if (code_seen('X')) {
+          //Check that the X pause position is within printer limits
+          if ((code_value() >= X_MIN_POS) && (code_value() <= X_MAX_POS-1)) {
```
gudnimg: Why -1?

Firmware/Marlin_main.cpp
```diff
@@ -7638,16 +7640,72 @@ SERIAL_PROTOCOLPGM("\n\n");
...
+            pause_position[X_AXIS] = code_value();
+          } else pause_position[X_AXIS] = X_PAUSE_POS;
```
gudnimg: The code can be simplified a lot by defining a temporary `float temp_pause_position[3] = { X_PAUSE_POS, Y_PAUSE_POS, Z_PAUSE_LIFT };` then the `else` statements are not needed. 

And finally call `memcpy(pause_position, temp_pause_position, sizeof(pause_position);` to modify `pause_position` only once. 

We can also save some space by calling `code_value()` less.  You call it 3 times for each axis, I suspect its better to save it in a variable.

Firmware/Marlin_main.cpp
3d-gussner: In MK404 the `M125 X255` causes a crash and I just wanted to eliminate that may happen on real hardware as well.
Selftest on my MK3S shows
```
Recv: Measured axis length:255.570
Recv: Measured axis length:255.570
Recv: Axis length difference:0.000

Recv: Measured axis length:220.370
Recv: Measured axis length:220.690
Recv: Axis length difference:0.320
```

So my printer would not crash with `M125 X255` but others may

Firmware/Marlin_main.cpp
```diff
@@ -7672,24 +7672,23 @@ SERIAL_PROTOCOLPGM("\n\n");
...
-        if (code_seen('Z')) {
-          //Check that the Z pause lift is within printer limits
-          if (code_value() <=Z_MAX_POS) {
-            pause_position[Z_AXIS] = code_value();
-          } else pause_position[Z_AXIS] = Z_PAUSE_LIFT;
+        //Check that the Z lift position is within printer limits
+        if (temp_pause_position[Z_AXIS] <=Z_MAX_POS) {
+          pause_position[Z_AXIS] = temp_pause_position[Z_AXIS];
         }
+
```
gudnimg: @3d-gussner 
Replacing all this, with this:

```cpp
        // Set default values
        pause_position[X_AXIS] = X_PAUSE_POS;
        pause_position[Y_AXIS] = Y_PAUSE_POS;
        pause_position[Z_AXIS] = Z_PAUSE_LIFT;

        for (uint8_t axis = 0; axis < E_AXIS; axis++) {
          if (code_seen(axis_codes[axis])) {
            pause_position[axis] = constrain(code_value(), min_pos[axis], max_pos[axis]);
          }
        }
```

Saves 66 bytes of Flash, and 12 bytes of SRAM

I'm assuming here we can use `min_pos` and `max_pos` arrays.

Firmware/Marlin.h
```diff
@@ -232,6 +232,7 @@ extern int feedmultiply;
 extern int extrudemultiply; // Sets extrude multiply factor (in percent) for all extruders
 extern float extruder_multiplier[EXTRUDERS]; // reciprocal of cross-sectional area of filament (in square millimeters), stored this way to reduce computational burden in planner
 extern float current_position[NUM_AXIS] ;
+extern float pause_position[3];
```
gudnimg: Please remove this. `pause_position` is not used in any other files than `Marlin_main.cpp`

The `extern` is used for variables here only if you want to use the variable in other `.cpp` files. In my opinion it is generally a bad practice and should be avoided if possible.

Firmware/Marlin_main.cpp
```diff
@@ -200,6 +200,8 @@ float min_pos[3] = { X_MIN_POS, Y_MIN_POS, Z_MIN_POS };
 float max_pos[3] = { X_MAX_POS, Y_MAX_POS, Z_MAX_POS };
 bool axis_known_position[3] = {false, false, false};
 
+float pause_position[3] = { X_PAUSE_POS, Y_PAUSE_POS, Z_PAUSE_LIFT };
```
gudnimg: Make this `static float` 

The `static` will tell the compiler the `pause_position` symbol only exists within `Marlin_main` object.

gudnimg (changes requested):

3d-gussner:
@gudnimg Thanks for the help and code optimization. Please review the PR again.
Added some documentation and neither `S0` or `S1` pause the printer.

gudnimg (approved):

PR #4403, opened 2023-09-23
PFW-1542 Fix an issue where print can't be stopped under specific circumstances

gudnimg:
ℹ️ This has been an issue since at least FW 3.9

There is a bug where if the printer is recovering a print, it will execute a blocking loop to restore the extruder and bed temperatures. But if a Fan error is triggered in this loop, then the user can't abort the print via LCD.

If the fan error resolves on its own the 'Resume print' menu will appear in a few seconds. But if not, then the user can't resume the print (which is normal). But with the bug above the user can't abort the print either!

The problem is essentially `isPrintPaused` variable is cleared too early. We should wait until the print is completely restored first.

Steps to reproduce:
1. Start a print
2. Pause the print
3. Wait for extruder temperature to fall at lest 180°C
4. Click 'Resume' print
5. While heating, stop the hotend fan and wait for a few seconds until an error is raised
6. Observe issue => 'Stop print' menu item is gone!

TODO:
- [x] Unload filament, and similar 'filament' menus should not be visible in this state.

PFW-1542

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -5009,7 +5009,6 @@ void lcd_resume_print()
     lcd_setstatuspgm(_T(MSG_FINISHING_MOVEMENTS));
     st_synchronize();
     custom_message_type = CustomMsg::Resuming;
-    isPrintPaused = false;
 
     // resume processing USB commands again and restore hotend fan state (in case the print was
     // stopped due to a thermal error)
@@ -5017,6 +5016,7 @@ void lcd_resume_print()
     Stopped = false;
 
     restore_print_from_ram_and_continue(default_retraction);
+    isPrintPaused = false;
     pause_time += (_millis() - start_pause_print); //accumulate time when print is paused for correct statistics calculation
     refresh_cmd_timeout();
     SERIAL_PROTOCOLLNRPGM(MSG_OCTOPRINT_RESUMED); //resume octoprint
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -99,7 +99,6 @@ static void lcd_language_menu();
 static void lcd_main_menu();
 static void lcd_tune_menu();
 static void lcd_settings_menu();
-static void lcd_calibration_menu();
 static void lcd_control_temperature_menu();
 #ifdef TMC2130
 static void lcd_settings_linearity_correction_menu_save();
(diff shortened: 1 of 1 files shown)


gudnimg:
@3d-gussner I think this fixes a lot of inconsistencies with the menus. For example I suspect we don't need this change anymore: https://github.com/prusa3d/Prusa-Firmware/commit/77a4b99f5e83dd94dd5d311108bf124e01518dd1

Firmware/ultralcd.cpp
```diff
@@ -5253,17 +5250,15 @@ static void lcd_main_menu()
...
 #endif //SDSUPPORT
 
-    if(!isPrintPaused && !printJobOngoing() && (lcd_commands_type == LcdCommands::Idle)) {
-        if (!farm_mode) {
-            const int8_t sheet = eeprom_read_byte(&(EEPROM_Sheets_base->active_sheet));
-            const int8_t nextSheet = eeprom_next_initialized_sheet(sheet);
-            if ((nextSheet >= 0) && (sheet != nextSheet)) { // show menu only if we have 2 or more sheets initialized
-                MENU_ITEM_FUNCTION_E(EEPROM_Sheets_base->s[sheet], eeprom_switch_to_next_sheet);
-            }
+    if(!printer_active() && !farm_mode) {
```
gudnimg, line 5253: This change hides the Sheet menu item for more scenarios:

* If the printer is Homing or running Mesh Bed Leveling
* A print is saved in RAM

Firmware/ultralcd.cpp
```diff
@@ -5253,17 +5250,15 @@ static void lcd_main_menu()
...
+        const int8_t sheet = eeprom_read_byte(&(EEPROM_Sheets_base->active_sheet));
+        const int8_t nextSheet = eeprom_next_initialized_sheet(sheet);
+        if ((nextSheet >= 0) && (sheet != nextSheet)) { // show menu only if we have 2 or more sheets initialized
+            MENU_ITEM_FUNCTION_E(EEPROM_Sheets_base->s[sheet], eeprom_switch_to_next_sheet);
         }
     }
 
-    if ( ! ( printJobOngoing() || (lcd_commands_type != LcdCommands::Idle) || (eFilamentAction != FilamentAction::None) || Stopped ) ) {
+    if ( ! ( printer_active() || (eFilamentAction != FilamentAction::None) || Stopped ) ) {
```
gudnimg, line 5261: This change hides the menu items for more scenarios:

* If the printer is Homing or running Mesh Bed Leveling
* A print is saved in RAM
* A print is paused

gudnimg:
@3d-gussner let me know if I should remove the menu changes (they are technically unrelated to the core issue blocking final 3.13.2 release). I only added them since you mentioned these menus shouldn't be available when resuming a print or when it's paused.

I did try unloading filament during a pause and it behaves strangely. (With MMU btw)

3d-gussner (approved):
Works well, tried really hard to BREAK it with multiple stall/resume stall while resuming multiple times. Always recovered correctly.
@gudnimg Please prepare a cherry-pick PR for MK_3.13.2 branch. Maybe only the fan fix without the 2nd part and menus.

PR #4405, opened 2023-09-23
MMU: always wait for a request's expected response

gudnimg:
If a planned request is activated when the protocol status is `Finished` then the firmware must wait for the response when the respective request is expecting it.

`manage_response` shouldn't return `Finished` unless the response is received/parsed because the printer needs to be able to act on it in a deterministic way. In other words always make sure a response is followed by its respective request before leaving `manage_response`.

An example is a register read using `ReadRegister()`.

Step to reproduce issue:
1. Fail selector homing
2. Select 'Tune' item
3. Observe issue. In this situation you can see the value for the previous register read is shown. Which is 0. This is very timing dependent and does not always happen.
4. Repeat step 2 until the issue appears. It may take a few times.

Some serial logs I had locally:

| BEFORE | AFTER |
|------|------|
|  mmu-bug-log    |   mmu-fix-log   |

I've observed this issue on Write Register requests and MMU Button requests. Probably applies to all requests which expect a response.

### DIFF
--- a/Firmware/mmu2_protocol_logic.cpp
+++ b/Firmware/mmu2_protocol_logic.cpp
@@ -809,14 +809,11 @@ StepStatus ProtocolLogic::Step() {
         // We are ok, switching to Idle if there is no potential next request planned.
         // But the trouble is we must report a finished command if the previous command has just been finished
         // i.e. only try to find some planned command if we just finished the Idle cycle
-        bool previousCommandFinished = currentScope == Scope::Command; // @@TODO this is a nasty hack :(
         if (!ActivatePlannedRequest()) {                               // if nothing is planned, switch to Idle
             SwitchToIdle();
-        } else {
+        } else if (ExpectsResponse()) {
             // if the previous cycle was Idle and now we have planned a new command -> avoid returning Finished
-            if (!previousCommandFinished && currentScope == Scope::Command) {
-                currentStatus = Processing;
-            }
+            currentStatus = Processing;
         }
     } break;
     case CommandRejected:


gudnimg:
I'm tempted to add this to 3.13.2 milestone but will let you guys decide @DRracer @3d-gussner

Firmware/mmu2_protocol_logic.cpp
```diff
@@ -818,6 +818,11 @@ StepStatus ProtocolLogic::Step() {
                 currentStatus = Processing;
             }
         }
+
+        if (ExpectsResponse()) {
```
DRracer: Is it really relevant at this level of code? Shouldn't it be in the `else` block?

IMO if we activated a planned request (which is the case of ReadRegister), it should step into the `else` block ... may be the condition:
```
if (!previousCommandFinished && currentScope == Scope::Command) {
```
should be changed to
```
if(ExpectsResponse()){
```
What do you think?

DRracer (changes requested):
Generally I agree, but may be the added code doesn't fix the real problem - please see my other comment.

Ideally, I'd like to see this scenario unit tested in MK4

Firmware/mmu2_protocol_logic.cpp
3d-gussner: @gudnimg Is this solved?

gudnimg: We cant use the else block because the if statement will always be true in this scenario: `if (!ActivatePlannedRequest())`, where there is no planned request waiting but we are expecting a response from a previously “activated planned request”.

The bug here may be a planned request is always activated regardless if we’ve not received the response yet. For example if 3 requests were theoretically planned, they’d all be activated after the other and the responses are “ignored”. 

I didnt touch the if-else since I was worried about breaking things, but I suspect it is wrong.

gudnimg: @DRracer I've got the unit tests working on the MK4 side now. I tried your suggestion where

```c
if (!previousCommandFinished && currentScope == Scope::Command) {
```

is changed to:

```c
if(ExpectsResponse()){
```

And it also passes the unit test! 👍 I'll test this on the hardware now, and if it works then I'll submit the changes. If it doesn't work, then I guess I'll need to update the unit tests ;)

gudnimg: I can confirm the fix works.
image

gudnimg: Here is an example where the fix is working for register writes. The "Expecting response!" log is in the same line as we have `currentStatus = Processing;`
image

DRracer:
This change definitely break the unit tests
`Marlin::MMU2::ProtocolLogic AsyncCommandRqCommandReset`
Not sure yet, if the unit tests are correct in this regard.

DRracer (approved):

PR #4410, opened 2023-09-24
🚧 `stop_and_save_print_to_ram` should not overwrite valid partial backup

gudnimg:
Thermal errors need to take a backup of the target temperatures before disabling the heaters. When the print is then later paused, `stop_and_save_print_to_ram` should use the partial backup if it is still valid (`isPartialBackupAvailable=true`). After `stop_and_save_print_to_ram` is done running and we have a 'full' backup, `isPartialBackupAvailable` is cleared in favor of `saved_printing`

Change in memory:
Flash: -156 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -9735,16 +9735,8 @@ void ThermalStop(bool allow_recovery)
 
                 // we cannot make a distinction for the host here, the pause must be instantaneous
                 // so we call the lcd_pause_print to save the print state internally. Thermal errors
-                // disable heaters and save the original temperatures to saved_*, which will get
-                // overwritten by stop_and_save_print_to_ram. For this corner-case, re-instate the
-                // original values after the pause handler is called.
-                uint8_t bed_temp = saved_bed_temperature;
-                uint16_t ext_temp = saved_extruder_temperature;
-                uint8_t fan_speed = saved_fan_speed;
+                // disable heaters and save the original temperatures to saved_*
                 lcd_pause_print();
-                saved_bed_temperature = bed_temp;
-                saved_extruder_temperature = ext_temp;
-                saved_fan_speed = fan_speed;
             }
         } else {
             // We got a hard thermal error and/or there is no print going on. Just stop.
@@ -10792,14 +10784,15 @@ void stop_and_save_print_to_ram(float z_move, float e_move)
 
 	planner_abort_hard(); //abort printing
 
-	memcpy(saved_pos, current_position, sizeof(saved_pos));
+    if (!isPartialBackupAvailable) {
+        refresh_print_state_in_ram();
+    }
+
     if (pos_invalid) saved_pos[X_AXIS] = X_COORD_INVALID;
 
-    saved_feedmultiply2 = feedmultiply; //save feedmultiply
-	saved_extruder_temperature = (uint16_t)degTargetHotend(active_extruder);
-	saved_bed_temperature = (uint8_t)degTargetBed();
-	saved_extruder_relative_mode = axis_relative_modes & E_AXIS_MASK;
-	saved_fan_speed = fanSpeed;
+    // Reset partial backup flag, we have a full back-up now
+    clear_print_state_in_ram();
+
 	cmdqueue_reset(); //empty cmdqueue
 	card.sdprinting = false;
 //	card.closefile();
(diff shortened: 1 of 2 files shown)


gudnimg:
Rebased to sync with MK3 branch (105 commits)

3d-gussner (approved):
Looks good to me as the removed `saved...` values are refreshed using `refresh_print_state_in_ram` function.

Firmware/Marlin_main.cpp
```diff
@@ -9512,16 +9512,8 @@ void ThermalStop(bool allow_recovery)
...
-                // original values after the pause handler is called.
-                uint8_t bed_temp = saved_bed_temperature;
-                uint16_t ext_temp = saved_extruder_temperature;
-                uint8_t fan_speed = saved_fan_speed;
+                // disable heaters and save the original temperatures to saved_*
                 lcd_pause_print();
-                saved_bed_temperature = bed_temp;
-                saved_extruder_temperature = ext_temp;
-                saved_fan_speed = fan_speed;
             }
```
wavexx, line 9740: Doesn't think break exactly the scenario written in the comment?

A thermal error will save temperatures into saved_* (see set_temp_error), then disable heaters and call ThermalStop. Calling lcd_pause_print will then save "0" into saved_, so resuming from a thermal error will restore broken temperatures?

3d-gussner: @gudnimg ???

3d-gussner: @wavexx Thanks for looking into this

gudnimg: @wavexx @3d-gussner 

I don't see how "0" is saved with these changes. This is how I see the code running:

First `set_temp_error()` is called and it calls `refresh_print_state_in_ram();` This will set `isPartialBackupAvailable = true`.

Then `lcd_pause_print();` is executed and calls `stop_and_save_print_to_ram()`. Since `isPartialBackupAvailable = true`, we do not refresh the print state in RAM (i.e. overwrite `saved_extruder_temperature` since we already have a valid partial backup).

The "partial backup" was added very recently to support Power Panic when the extruder is parked for example in M600, Paused print, MMU error etc.

Does this make sense?

wavexx: Am I looking at the wrong piece of code maybe?

https://github.com/prusa3d/Prusa-Firmware/blob/ff42510a5c9a97601e6a912495694796f9d053ae/Firmware/temperature.cpp#L511

there's no call to refresh_print_state_in_ram(). It sets saved_, then disables the heater.

wavexx: Duh.. looks like I was. I didn't realize set_temp_error was changed within the PR itself.

wavexx: Re-looking at the new lcd pause and set_temp_error code, my point is moot. I don't see a problem with this. It's definitely much better compared to what we had before where we would always overwrite saved_ values.

Firmware/Marlin_main.cpp
```diff
@@ -10559,14 +10551,15 @@ void stop_and_save_print_to_ram(float z_move, float e_move)
 
 	planner_abort_hard(); //abort printing
 
-	memcpy(saved_pos, current_position, sizeof(saved_pos));
+    if (!isPartialBackupAvailable) {
+        refresh_print_state_in_ram();
+    }
```
wavexx, line 10789: A problem I see here with this approach, is that now set_temp_error saves current_position before planner_abort_hard() is called, so current_position is not correct during a thermal error.

3d-gussner: @gudnimg Is this solved?

gudnimg: No I havent had a good opportunity to look into this. I was too sick to do any productive things previous weekend 

I will hopefully look at this again soon (and other PRs)

gudnimg:
Synced PR with latest MK3 branch (~126 commits)

Firmware/Marlin_main.cpp
3d-gussner: @gudnimg is this solved?

gudnimg: No the PR needs to be reworked to address this. I think we should postpone this after 3.14

3d-gussner:
@gudnimg Is it still in progress or can that be reviewed?

gudnimg:
> @gudnimg Is it still in progress or can that be reviewed?

@3d-gussner This pull request needs more work, the comment from @wavexx is still valid and I haven't looked more into this. Closing PR for now since I'm unlikely to finish it soon.

PR #4419, opened 2023-10-01
M600: reduce stack usage

gudnimg:
To handle power panic in M600 we started saving relevant data at the start of M600 gcode in RAM (starting with 3.13.2). We are currently also saving the same data within `gcode_M600()` which is saved on the stack. I propose we just use the data already saved in SRAM to reduce stack usage.

Change in memory:
Flash: -64 bytes
SRAM: 0 bytes


3d-gussner (approved):
Looks good to mea and MK404 Test passed :heavy_check_mark:

PR #4420, opened 2023-10-01
PFW-1543 Save print type in EEPROM

gudnimg:
Small step forward to support USB prints.

### DIFF
--- a/Firmware/eeprom.h
+++ b/Firmware/eeprom.h
@@ -134,7 +134,9 @@ static_assert(sizeof(Sheets) == EEPROM_SHEETS_SIZEOF, "Sizeof(Sheets) is not EEP
 | 0x0F95 3989 | char    | EEPROM_FILENAME                       | ???          | ffh 255               | Power Panic Filename                              | ???          | D3 Ax0f95 C8
 | 0x0F91 3985 | unint32 | EEPROM_FILE_POSITION                  | ???          | ff ff ff ffh          | Power Panic File Position                         | ???          | D3 Ax0f91 C4
 | 0x0F8D 3981 | float   | EEPROM_UVLO_CURRENT_POSITION_Z        | ???          | ff ff ff ffh          | Power Panic Z Position                            | ^            | D3 Ax0f8d C4
-| 0x0F8C 3980 | ???     | EEPROM_UVLO_UNUSED_001                | ???          | ffh 255               | Power Panic _unused_                              | ^            | D3 Ax0f8c C1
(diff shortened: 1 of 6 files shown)


Firmware/eeprom.h
```diff
@@ -134,7 +134,9 @@ static_assert(sizeof(Sheets) == EEPROM_SHEETS_SIZEOF, "Sizeof(Sheets) is not EEP
 | 0x0F95 3989 | char    | EEPROM_FILENAME                       | ???          | ffh 255               | Power Panic Filename                              | ???          | D3 Ax0f95 C8
 | 0x0F91 3985 | unint32 | EEPROM_FILE_POSITION                  | ???          | ff ff ff ffh          | Power Panic File Position                         | ???          | D3 Ax0f91 C4
 | 0x0F8D 3981 | float   | EEPROM_UVLO_CURRENT_POSITION_Z        | ???          | ff ff ff ffh          | Power Panic Z Position                            | ^            | D3 Ax0f8d C4
-| 0x0F8C 3980 | ???     | EEPROM_UVLO_UNUSED_001                | ???          | ffh 255               | Power Panic _unused_                              | ^            | D3 Ax0f8c C1
+| 0x0F8C 3980 | uint8   | EEPROM_UVLO_PRINT_TYPE                | 00h 0        | ffh 255               | Power Panic print type: SD                        | ???          | D3 Ax0f8c C1
+| ^           | ^       | ^                                     | 01h 1        | ffh 255               | Power Panic print type: USB                       | ^            | ^
```
3d-gussner: replace `ffh 255`  with `^` here and below.

3d-gussner (changes requested):
Minor change in eeprom documentation.
Renaming some functions to be more consistent.
Maybe add `M118 A1 uvlo_triggered` when power panic triggers replace `MSG_INT4[] PROGMEM = "INT4";` ? Or something else that is faster?
Also add `M118 A1 uvlo_recover`or similar to notify host, when its connects to the printer?
Tested on MK404 and it looks correct.

Firmware/eeprom.h
gudnimg: Done

gudnimg:
> Maybe add `M118 A1 uvlo_triggered` when power panic triggers replace `MSG_INT4[] PROGMEM = "INT4";` ? Or something else that is faster?

Printing onto serial directly in the interrupt handler is the fastest way. I would want to avoid executing gcodes in power panic if possible.

What's wrong with using `SERIAL_ECHOLNRPGM(MSG_INT4);`? Do we want something like `"//INT4"` with the `"//"`? If so I propose we check if it's a USB print and add the `"//"`.

3d-gussner:
No need to change "INT4" anymore.

3d-gussner (approved):
Tested on MK404
- SD print :heavy_check_mark: 
- USB print :heavy_check_mark:

PR #4427, opened 2023-10-05
optimisation: stepper: make `dda_isteps_t` an array

gudnimg:
This way we can loop through it with a for-loop
and save some code size

### DIFF
--- a/Firmware/stepper.cpp
+++ b/Firmware/stepper.cpp
@@ -110,12 +110,8 @@ block_t *current_block;  // A pointer to the block currently being traced
 //static makes it inpossible to be called from outside of this file by extern.!
 
 // Variables used by The Stepper Driver Interrupt
-static unsigned char out_bits;        // The next stepping-bits to be output
-static dda_isteps_t
-               counter_x,       // Counter variables for the bresenham line tracer
-               counter_y,
-               counter_z,
-               counter_e;
+static unsigned char out_bits;         // The next stepping-bits to be output
+static dda_isteps_t counter[NUM_AXIS]; // Counter variables for the bresenham line tracer
 volatile dda_usteps_t step_events_completed; // The number of step events executed in the current block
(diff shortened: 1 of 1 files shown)


Firmware/stepper.cpp
```diff
@@ -393,18 +389,20 @@ FORCE_INLINE void stepper_next_block()
     }
 
     if (current_block->flag & BLOCK_FLAG_DDA_LOWRES) {
-      counter_x.lo = -(current_block->step_event_count.lo >> 1);
-      counter_y.lo = counter_x.lo;
-      counter_z.lo = counter_x.lo;
-      counter_e.lo = counter_x.lo;
+      const int16_t value = -(current_block->step_event_count.lo >> 1);
+      for (uint8_t axis = 0; axis < NUM_AXIS; axis++)
```
DRracer, line 393: you can even iterate from the back which saves 2-4 bytes more
```
for (uint8_t axis = NUM_AXIS-1; axis >=0 ; --axis){
    counter[axis].lo = value;
}
```
Caches and memory ordering is not an issue on the AVR IMHO

DRracer (approved):
:thinking: most of the savings in this PR come from setting `counter_*` variables in loops. The rest is just a mechanical rewrite to match the proposed "array" principle.

I'm surprised that the compiler understood the intent and actually saved something while not breaking the rest of the code.

Firmware/stepper.cpp
Ro3Deee: what about this, using postdecrementation :
`for (uint8_t axis = NUM_AXIS; axis-- ; ){`
`    counter[axis].lo = value;`
`}` ?

3d-gussner: @petrubecheru Thanks but compiling with `PF-build.sh` the usage of Flash and RAM are the same before and after.

PR #4429, opened 2023-10-05
PFW-1525 Add `Set Ready`menu

### LINKED ISSUE #4412: "Set Ready" Prusa Connect button in menu of MK3
A button to pass "set ready" to PrusaConnect in order to continue on the print queue would be needed on the menu of the physical printer if the printer and the PC are separated by each other. In my case the printers are in another room. So, if I wated to see if everthing was alright I would clean up the printer, run back to the PC, set the printer ready in connect, back to the printer, have a look and be back again. 
A simple "set ready" in the info screen where eg "unload filament"resides would perfectly do.

**MMU Upgrade** - [e.g. MMU2S, MMU2, MMU1]

3d-gussner:
Translation in 2nd part to prevent conflicts with other PRs

Menu `Set Ready`
Shown
- [X] After reboot or start
- [X] After finished print
- [X] After print stop

Hidden
- [X] While SD printing
- [X] While USB printing

When menu selected a serial message `// action:printer_ready` is sent and Menu is hidden :heavy_check_mark: 

- [X] `Set Ready` via gcode `M72 S[1|0]` from PrusaConnect or PrusaLink

- [ ]  Update RepRap Wiki `M72`


Firmware/cmdqueue.cpp
```diff
@@ -479,9 +479,10 @@ void get_command()
             allow_when_stopped = true;
 
         // Handle the USB timer
-        if ((*cmd_start == 'G') && !(IS_SD_PRINTING))
+        if ((*cmd_start == 'G') && !(IS_SD_PRINTING)) {
             usb_timer.start();
-
+            printer_state = 3; //set printer state busy printing to hide LCD menu and report correctly M862.7 Q while USB printing
```
gudnimg: I think it would be nice to use `enum` or` enum class` instead of magic numbers.

Firmware/ultralcd.cpp
```diff
@@ -5191,7 +5200,9 @@ static void lcd_main_menu()
     } else if (!Stopped) {
         MENU_ITEM_SUBMENU_P(_i("Preheat"), lcd_preheat_menu);////MSG_PREHEAT c=18
     }
-
+    if (printer_state == 0) {
```
gudnimg: I feel this menu item should be hidden if there is no host connected/enabled. Can we check that somehow?

If user has no remote host and uses only SD card this menu won’t do anything. In that case I think we should try minimize amount of menu items to simplify the user interface.

Firmware/cmdqueue.cpp
3d-gussner: Agreed, but i would like to change that later to cover more.

Firmware/ultralcd.cpp
3d-gussner: Right, there will be a new USB PrusaLink connected state to be checked. It have been better to start with that.
BTW the `Set Ready`on MK4 is atm also shown even without LAN/WiFi connected.

Firmware/Marlin.h
```diff
@@ -319,6 +319,8 @@ bool printJobOngoing();
 
 bool printer_active();
 
+extern uint8_t printer_state;
```
DRracer: please avoid externs, make Setter and Getter functions

Firmware/Marlin_main.cpp
```diff
@@ -519,6 +519,15 @@ bool __attribute__((noinline)) printer_active() {
...
 }
 
+//!@brief printer_state
+//!printer state state
+//!0 = unknown or finished print
+//!1 = printer ready for next print
+//!2 = printer SD printing
+//!3 = printer USB printing
+//!@endcode
+uint8_t printer_state=0;
```
DRracer: make it an enum

Firmware/Marlin_main.cpp
```diff
@@ -7875,6 +7884,8 @@ SERIAL_PROTOCOLPGM("\n\n");
       - M862.3 { P"<model_name>" | Q }
       - M862.4 { P<fw_version> | Q }
       - M862.5 { P<gcode_level> | Q }
+      - M862.6 Not used but reserved
+      - M862.7 { P<0|1> | Q } 0 = Printer not ready for next print (default), 1 = Printer is ready for next print , 2 = Printer busy printing
```
DRracer: please check Buddy FW if M862.7 is not used for anything else

DRracer: I don't like using M862.x for setting stuff

3d-gussner:
@DRracer @gudnimg Please review again. I don't know if the TOGGLE menu is a good idea as in Buddy firmware it is one menu but that is grayed out or not

Firmware/Marlin.h
```diff
@@ -316,6 +316,19 @@ bool printJobOngoing();
...
+    NotReady = 0,
+    IsReady = 1,
+    Idle = 2,
+    SDPrintingFinished = 3,
+    HostPrintingFinished = 4,
+    IsSDPrinting = 5,
+    IsHostPrinting = 6,
+};
+
+extern PrinterStatus printer_status;
```
DRracer: please do not add new extern variables :pray:

DRracer (dismissed):
- please don't add new extern variables
- I'm not sure about using M862.7 for querying the state, I'd advise using `M72 Q` for reporting the state

other than that, this PR looks ok

Firmware/ultralcd.cpp
```diff
@@ -5121,6 +5121,16 @@ static void lcd_sheet_menu()
...
 }
 
+//! @brief Set printer state
+//! Sets the printer state for next print via LCD menu
+//! @endcode
+static void lcd_printer_status_toggle()
+{
+    if (printer_status == PrinterStatus::NotReady) printer_status = PrinterStatus::IsReady;
+    else printer_status = PrinterStatus::NotReady;
+    enquecommandf_P(PSTR("M118 A1 action:%s"), (printer_status == PrinterStatus::NotReady) ? "not_ready" : "ready");
```
gudnimg: Optimisation idea:

Please add to `messages.cpp/h`:

```c
const char MSG_OCTOPRINT_NOT_READY[] PROGMEM_N1 = "// action:not_ready"; ////
const char MSG_OCTOPRINT_READY[] PROGMEM_N1 = "// action:ready"; ////
```

And change:
`enquecommandf_P(PSTR("M118 A1 action:%s"), (printer_status == PrinterStatus::NotReady) ? "not_ready" : "ready");` 
into:
`enquecommandf_P(PSTR("M118 %S"), (printer_status == PrinterStatus::NotReady) ? MSG_OCTOPRINT_NOT_READY : MSG_OCTOPRINT_READY);`

Change in memory:
Flash: +16 bytes
SRAM: -10 bytes

Note that `"not_ready" : "ready"` without `PSTR()` will use up SRAM instead of Flash. Note that for PROGMEM strings you must use `%S` instead of `%s`

Firmware/cmdqueue.cpp
```diff
@@ -479,9 +479,10 @@ void get_command()
             allow_when_stopped = true;
 
         // Handle the USB timer
-        if ((*cmd_start == 'G') && !(IS_SD_PRINTING))
+        if ((*cmd_start == 'G') && !(IS_SD_PRINTING)) {
```
gudnimg: should `!(IS_SD_PRINTING)`  be `printer_status != PrinterStatus::IsSDPrinting`?

Firmware/ultralcd.cpp
```diff
@@ -5193,7 +5203,9 @@ static void lcd_main_menu()
     } else if (!Stopped) {
         MENU_ITEM_SUBMENU_P(_i("Preheat"), lcd_preheat_menu);////MSG_PREHEAT c=18
     }
-
+    if (printer_status < PrinterStatus::IsSDPrinting) {
```
gudnimg: ~~Shouldn't this be `printer_status != PrinterStatus::IsSDPrinting`? I think that will read more clearly what the intention is with the code.~~

Another solution is to update `printJobOngoing()` to check for `PrinterStatus::IsSDPrinting` and `PrinterStatus::IsHostPrinting`. But then again, maybe `PrinterStatus::IsSDPrinting` can replace `IS_SD_PRINTING`?

This is something we also discussed before. Why have both `IS_SD_PRINTING` and `PrinterStatus::IsSDPrinting`?

gudnimg: Also how does `M79` come into this now that we have it?

Firmware/Marlin.h
gudnimg: Adding here the discussion we had two weeks ago:

Do not add new `extern` variables.

Instead create "get" and "set" functions. Preferably in a new file like `printer_status.cpp/h` or somewhere else than `Marlin.h`.

Then define `PrinterStatus printer_status` in the cpp file and use `static` keyword. I have something like this in mind:

```cpp
// cpp file
static PrinterStatus printer_status;

PrinterStatus GetPrinterStatus() {
    return printer_status;
}

PrinterStatus SetPrinterStatus(PrinterStatus status) {
    printer_status = status;
}
```

```cpp
// header file
enum class PrinterStatus : uint8_t
{
    NotReady = 0,
    IsReady = 1,
    Idle = 2,
    SDPrintingFinished = 3,
    HostPrintingFinished = 4,
    IsSDPrinting = 5,
    IsHostPrinting = 6,
};

PrinterStatus GetPrinterStatus();
PrinterStatus SetPrinterStatus(PrinterStatus status);
```

Firmware/ultralcd.cpp
3d-gussner: > Also how does M79 come into this now that we have it?

The LCD menu Set Ready [Yes|No]  is only shown when the M79 host timer is running

3d-gussner:
@gudnimg please review again

Firmware/ultralcd.cpp
```diff
@@ -5194,7 +5205,13 @@ static void lcd_main_menu()
...
         MENU_ITEM_SUBMENU_P(_i("Preheat"), lcd_preheat_menu);////MSG_PREHEAT c=18
     }
-
+    if (GetPrinterState() < PrinterState::IsSDPrinting && M79_timer_get_status()) {
+        if(GetPrinterState() == PrinterState::IsReady) {
+            MENU_ITEM_TOGGLE_P(_T(MSG_SET_READY), _T(MSG_NO), lcd_printer_status_toggle);
+        } else {
+            MENU_ITEM_TOGGLE_P(_T(MSG_SET_READY), _T(MSG_YES), lcd_printer_status_toggle);
+        }
+    }
```
gudnimg: Nitpick, this comment doesnt block this PR... but I think this would be more readable: (it doesnt cost any extra memory)

```cpp
if (M79_timer_get_status()) {
        switch (GetPrinterState())
        {
        case PrinterState::IsReady:
            MENU_ITEM_TOGGLE_P(_T(MSG_SET_READY), _T(MSG_NO), lcd_printer_status_toggle);
            break;
        case PrinterState::NotReady:
        case PrinterState::Idle:
        case PrinterState::SDPrintingFinished:
        case PrinterState::HostPrintingFinished:
            MENU_ITEM_TOGGLE_P(_T(MSG_SET_READY), _T(MSG_YES), lcd_printer_status_toggle);
            break;
        default:
            break;
        }
    }
```

Firmware/ultralcd.cpp
```diff
@@ -901,6 +901,7 @@ void lcd_commands()
 			pid_temp = DEFAULT_PID_TEMP;
 			lcd_commands_step = 0;
 			lcd_commands_type = LcdCommands::Idle;
+            SetPrinterState(PrinterState::Idle);
```
gudnimg: Isn't this missing in more places? For example after first layer calibration is done, thermal model calibration etc.

`lcd_commands_type = LcdCommands::Idle;` is used in more places.

gudnimg (approved):
I find the new print states a bit questionable since they are not used everywhere for example `card.sdprinting` is still being used.

But I will approve the PR continue working on this with new PRs.

gudnimg:
@3d-gussner Looking at the translation PR, it seems `M862.7` is missing? Not sure if that was intentional.

3d-gussner:
The state can be checked with M72 without parameters.

Firmware/ultralcd.cpp
```diff
@@ -5191,7 +5206,13 @@ static void lcd_main_menu()
     } else if (!Stopped) {
         MENU_ITEM_SUBMENU_P(_i("Preheat"), lcd_preheat_menu);////MSG_PREHEAT c=18
     }
-
+    if (GetPrinterState() < PrinterState::IsSDPrinting && M79_timer_get_status()) {
```
DRracer, line 5208: Based on this statement, the order / numerical values of PrinterState members is important. A big warning should be added to the structure so that no one tries to reorder the items arbitrarily.

DRracer (changes requested):
just document the importance of PrinterState items' ordering, otherwise ok

DRracer (approved):

PR #4432, opened 2023-10-07
Fix `FW_COMMIT_HASH` compiler warning

gudnimg:
On my end, the default length for the abbreviated commit hash is 9 characters and won't fit into `uint32_t`.

To fix this I propose to change `FW_COMMIT_HASH` into a string and create preprocessor symbol for the string length so that it's known at compile time.

If the string should be longer or shorter then only `FW_COMMIT_HASH_LENGTH` needs to be configured on the CMake side

- [ ] Cherry-pick PR to MK3_3.13.2

### DIFF
--- a/cmake/ProjectVersion.cmake
+++ b/cmake/ProjectVersion.cmake
@@ -87,6 +87,12 @@ ENDIF()
 
 set(PROJECT_VERSION "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_REV}.${PROJECT_VERSION_TWEAK}")
 
+# Define a constant length for the commit hash
+set(FW_COMMIT_HASH_LENGTH 9)
+
+# Create fallback value with constant length
+string(REPEAT "0" ${FW_COMMIT_HASH_LENGTH} FW_COMMIT_HASH_UNKNOWN)
+
 function(resolve_version_variables)
   if(FW_COMMIT_DSC)
     return()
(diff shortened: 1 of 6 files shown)


gudnimg:
The function `FW_VERSION_HASH_P()` isn't called anywhere yet. Is it supposed to be in M115?

cmake/ProjectVersion.cmake
```diff
@@ -87,18 +87,30 @@ ENDIF()
 
 set(PROJECT_VERSION "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_REV}.${PROJECT_VERSION_TWEAK}")
 
+# Define a constant length for the commit hash
+set(FW_COMMIT_HASH_LENGTH 9)
+
+# Create fallback value with constant length
+string(REPEAT "0" ${FW_COMMIT_HASH_LENGTH} FW_COMMIT_HASH_UNKNOWN)
```
vintagepc, line 94: This is at odds with/will get clobbered by line 94-ish of CMakeLists.txt.

gudnimg: You're right, I missed it. Will look into it.

gudnimg:
@vintagepc the last commit seems to have resolved the static assert :) 

There is still an issue with the Github Action. The runner doesn't detect `git`. It's unrelated to this PR but I don't see why this is happening. The Ubuntu version we're using should have Git included according to Github.

I've tried to run the Workflow via Docker on Windows using https://github.com/nektos/act but no luck so far.

image

gudnimg:
Fixed the issue with the git description command. The problem was not that git command wasn't found, our script is just interpreting the error that way :) 
image

3d-gussner:
We can shorten the hash to 8 chars with `--abbrev=8` in https://github.com/gudnimg/Prusa-Firmware/blob/fix-compiler-warning-sha/cmake/GetGitRevisionDescription.cmake#L353.

gudnimg:
> We can shorten the hash to 8 chars with `--abbrev=8` in https://github.com/gudnimg/Prusa-Firmware/blob/fix-compiler-warning-sha/cmake/GetGitRevisionDescription.cmake#L353.

Note that `git_head_commit_data(FW_COMMIT_DATE "%ct")` is using the same function. We would need to create a new function.

I think having the data as a string (`char *`) in the firmware is better than `uint32_t`, because then we don't need to convert the `uint32_t` to hexidecimal at runtime.

3d-gussner:
Okay. Is the other comment resolved?

gudnimg:
> Okay. Is the other comment resolved?

Yeah https://github.com/prusa3d/Prusa-Firmware/pull/4432/commits/69e06bac8416f620763b2499c2cffcb94c35c0a5 resolved it. Github always failed to build before that fix because I added a new `static_assert` which checks the string length is correct.

3d-gussner (approved):
Tested and works well. No warnings.

3d-gussner:
@vintagepc Can we merge it?

PR #4442, opened 2023-10-11
Update Italian

3d-gussner:


wavexx (approved):

leptun:
@3d-gussner Is there any other change necessary for IT or is this all that was needed?

3d-gussner:
I don't think so. If we can always update it later

leptun:
Oke. Marking IT as complete in the diacritics tasks table. If we make any other PR, please add it to that table as well for reference

PR #4452, opened 2023-10-17
PFW-1460 Part1: Fix PF-build.sh with new cmake

3d-gussner:
Same as MK3_3.13.2 branch


Firmware/Configuration.h
```diff
@@ -43,7 +42,9 @@ extern const char _sPrinterMmuName[] PROGMEM;
 #define FW_COMMIT_HASH_LENGTH 1
 #define FW_COMMIT_HASH "0"
 #define FW_REPOSITORY "Unknown"
-#define FW_VERSION_FULL FW_VERSION "-unknown"
+#ifndef FW_VERSION_FULL
+#define FW_VERSION_FULL FW_VERSION
```
vintagepc: This entire ifndef can be removed. it's impossible for FW_VERSION_FULL to not be defined in the if block above.

vintagepc:
seems reasonable apart from the unnecessary code. That's probably leftover from a previous iteration that didn't define FW_VERSION_FULL above

vintagepc (approved):

PR #4454, opened 2023-10-18
MMU: Backport code changes from 32-bit firmware

gudnimg:
Relevant fix description from the MK4 side:

> * On MMU power cycle, some attempt counters were not properly reset, resulting in the MMU not being able to reconnect after power cycle
> * Power cycle was also not restarting stuff like error reported, causing the MMU FW upgrade error not being reported after restart.

⚠️ Keep in mind the MK4 can actually turn the MMU off (electrically). While the MK3S+ cannot do that, so in this PR the issues related to power cycling are not really relevant since the printer will be power cycled as well.

### DIFF
--- a/Firmware/mmu2_reporting.cpp
+++ b/Firmware/mmu2_reporting.cpp
@@ -1,4 +1,5 @@
 #include <avr/pgmspace.h>
+#include "eeprom.h"
 #include "mmu2.h"
 #include "mmu2_log.h"
 #include "mmu2_reporting.h"
@@ -342,6 +343,11 @@ void TryLoadUnloadReporter::DumpToSerial(){
     }
     buf[LCD_WIDTH] = 0;
     MMU2_ECHO_MSGLN(buf);
+}
+
+/// Disables MMU in EEPROM
+void DisableMMUInSettings() {
+    eeprom_update_byte((uint8_t *)EEPROM_MMU_ENABLED, false);
 }
 
 void IncrementLoadFails(){
--- a/Firmware/mmu2_reporting.cpp
+++ b/Firmware/mmu2_reporting.cpp
@@ -223,15 +223,11 @@ static bool is_mmu_error_monitor_active;
 static bool putErrorScreenToSleep;
 
 void CheckErrorScreenUserInput() {
-    if (isErrorScreenRunning()) {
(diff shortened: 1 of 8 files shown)


Firmware/mmu2_reporting.cpp
```diff
@@ -221,6 +222,14 @@ static bool is_mmu_error_monitor_active;
...
 
+void CheckErrorScreenUserInput() {
+    if (isErrorScreenRunning()) {
+        // Call this every iteration to keep the knob rotation responsive
+        // This includes when mmu_loop is called within manage_response
+        ReportErrorHook((CommandInProgress)mmu2.GetCommandInProgress(), mmu2.GetLastErrorCode(), mmu2.MMULastErrorSource());
+    }
+}
+
 bool isErrorScreenRunning() {
```
DRracer: Do we still need this function? May be it can be removed

DRracer (approved):
Generally, I can see all the relevant changes. Only please consider if we still need to keep `isErrorScreenRunning`.

Firmware/mmu2_reporting.cpp
gudnimg: I suspect you’re right. I’ll remove it. That way the code is a little bit more in sync with MK4.

gudnimg: Done

gudnimg:
I havent tested the changes yet on real hardware, please do not merge unless testing it first. :) I can run tests over the weekend.

The main thing to verify in my opinion is `DisableMMUInSettings()`, and the removal of `eeprom_update_byte((uint8_t *)EEPROM_MMU_ENABLED, true);` and `eeprom_update_byte((uint8_t *)EEPROM_MMU_ENABLED, false);` in `Firmware/mmu2_power.cpp`

gudnimg:
Tests (10/21/2023)

1. Trigger `ERR_SYSTEM_FW_UPDATE_NEEDED` and Disable the MMU via the MMU error screen.
   * ✅ **Test:** The MMU is disabled
   * ✅ **Test:** The MMU stays disabled even after a complete power cycle

3d-gussner (commented):
Edited wrong PR

PR #4455, opened 2023-10-19
mmu: Don't log Query (Q0) requests/responses when MMU is Idle

### LINKED ISSUE #4414: [BUG] Too much MK3S<->MMU2 communication logging on serial port
**Printer type** - MK3S
**Printer firmware version** -  3.13.1+

**MMU upgrade** - MMU2S
**MMU upgrade firmware version** - 3.0.0, 3.0.1

**SD card or USB/Octoprint**
  Mainly USB/Octoprint

**Describe the bug**
 Serial communications between the printer and MMU are continually output on the USB port.
  
**To Reproduce**
  Connect to USB serial and notice continual chatter.
  
**Expected behavior**
 We need a way to suppress these messages from log output either by:
1. Providing menu option to disable (LogRequestMsg, LogResponse)
2. Conditional DEBUG build
3. Remove altogether
 
I suppose one could write an Octoprint serial filter to remove the transactions from the console but I would prefer not to have the excess data always clogging the serial communications path.

gudnimg:
An idea I have to resolve https://github.com/prusa3d/Prusa-Firmware/issues/4414

If a Query is running in Idle state machine, then don't log anything but still allow asynchronous requests/responses to be logged. For example a register read via M707 will still show up fine.

image

This makes the serial log muuch cleaner when the MMU is idle. When running a printjob, it is only when a command such as toolchange is running that everything will be logged again.

### DIFF
--- a/Firmware/mmu2_protocol_logic.cpp
+++ b/Firmware/mmu2_protocol_logic.cpp
@@ -57,8 +57,8 @@ void ProtocolLogic::CheckAndReportAsyncEvents() {
 }
 
 void ProtocolLogic::SendQuery() {
-    SendMsg(RequestMsg(RequestMsgCodes::Query, 0));
     scopeState = ScopeState::QuerySent;
+    SendMsg(RequestMsg(RequestMsgCodes::Query, 0));
 }
 
 void ProtocolLogic::StartReading8bitRegisters() {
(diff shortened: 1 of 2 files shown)


gudnimg:
Closing this PR since it won't be merged. I will keep the branch open though since it may be useful.

PR #4456, opened 2023-10-19
Update enabled diacritics

leptun:
Updated the enabled diacritics after the PO files were finalized.

Rule 1: Only in po files used diacritics are activated. Unused will be replaced by "aA-zZ" chars by default.
Rule 2: Uppercase: C I O S U Z are replaced by lower case version if only used ONCE


lang/lib/FontGen.py
```diff
@@ -34,45 +36,97 @@ def __init__(self, utf, charListIndex, alternate = None):
...
+    # CustomCharacter('ŕ', 100, 'r'),
+    # CustomCharacter('Ŕ', 104, 'R'),
     CustomCharacter('ř', 101, 'r'),
+    CustomCharacter('Ř', 105, 'R'),
     CustomCharacter('ś', 108, 's'),
+    CustomCharacter('Ś', 112, 'S'),
     CustomCharacter('š', 109, 's'),
+    CustomCharacter('Š', 113, 'S'),
     CustomCharacter('ș', 110, 's'),
+    CustomCharacter('Ș', 114, 'S'),
```
3d-gussner, line 100: Where are the `ş  Ş` ?

lang/lib/FontGen.py
```diff
@@ -34,45 +36,97 @@ def __init__(self, utf, charListIndex, alternate = None):
...
     CustomCharacter('ß', 111, '\\xe2'),
+    # CustomCharacter('ẞ', 115, '\\xe2'),
```
3d-gussner, line 102: Why is `ß` 111 and 115 and one active and the other inactive?

lang/lib/FontGen.py
```diff
@@ -34,45 +36,97 @@ def __init__(self, utf, charListIndex, alternate = None):
...
     CustomCharacter('ť', 116, 't'),
+    # CustomCharacter('Ť', 118, 'T'),
     CustomCharacter('ț', 117, 't'),
-    CustomCharacter('Ț', 119, 'T'),
+    # CustomCharacter('Ț', 119, 'T'),
     CustomCharacter('ú', 120, 'u'),
```
3d-gussner, line 107: Where are `ţ Ţ` ?

lang/lib/FontGen.py
```diff
@@ -34,45 +36,97 @@ def __init__(self, utf, charListIndex, alternate = None):
...
+    CustomCharacter('Ý', 138, 'Y'),
+    # CustomCharacter('ÿ', 137, 'y'),
+    # CustomCharacter('Ÿ', 139, 'Y'),
     CustomCharacter('ź', 140, 'z'),
+    # CustomCharacter('Ź', 144, 'Z'),
     CustomCharacter('ž', 141, 'z'),
+    CustomCharacter('Ž', 145, 'Z'),
     CustomCharacter('ż', 142, 'z'),
+    # CustomCharacter('Ż', 146, 'Z'),
+    # CustomCharacter('¿', 254, ' '),
```
3d-gussner, line 128: Shouldn't the  `¿ ¡` be replaced with anything?

3d-gussner (commented):
Tested on MK404 and looks correct.

lang/lib/FontGen.py
leptun: We only implement s-comma. s-cedilla gets preprocessed to s-comma
https://github.com/prusa3d/Prusa-Firmware/pull/3383#issuecomment-1730923150

lang/lib/FontGen.py
leptun: uppercase version of ß. Should only ever be used in all-caps words such as MMU errors, but it isn't used anywhere. But since I have it in the font file, I added it to FontGen.py for future reference.

lang/lib/FontGen.py
leptun: Same as for s-comma vs s-cedilla

lang/lib/FontGen.py
3d-gussner: Never saw an upper case ß does this exists?

lang/lib/FontGen.py
leptun: I added these here since I had them in my font. We don't use them since these unicode characters get preprocessed to an empty string, so effectively removed. I have them here just for reference. We can't use an empty character sadly as the alternate form, so this is the best I could make with the available codebase. In any case, it's disabled, so not something to worry about.

lang/lib/FontGen.py
leptun: yes

leptun: https://en.wikipedia.org/wiki/%C3%9F#:~:text=A%20second%20proposal,orthography%20in%202017.

lang/lib/FontGen.py
3d-gussner: Can confirm that the Spanish wizard only has one Space in 2nd part of the message which is a question.

3d-gussner (approved):

PR #4460, opened 2023-10-21
PFW-1523 Add a new timer for Prusa Link

gudnimg:
The idea is to have the host ping the printer periodically with a M79 to enable certain features/UI. Using the `usb_timer` is not a good solution for this as it depends on seeing a 'G' character

The LCD code, or whatever code is implementing the new functionality will need to include `host.h` and check whether `M79_timer_get_status()` returns `0` (timer not running) or `1` (timer is running).

I created a new file for the code `host.cpp` which we can use to expand host related functionality and not clutter `Marlin_main.cpp` further.

**Some ideas how this could be used (nothing set in stone):**
* Show IP address in support menu
* Show "Printer Ready" menu
* Disable or Enable MMU logs or filter them even (unlikely to happen though)
* Different Power Panic behavior
* Lots of ways to render the UI differently e.g. show "PL" instead of "USB" on the status screen.
* Custom Attention messages which should not be used with Octoprint. Messages specific to Prusa Link.
* And so on :) 

----
TODO

- [x] Implement M79, revert changes in M113
- [x] Implement `manage_timeouts()` or something in `manage_inactivity()` which will periodically check if the timer is expired.


3d-gussner (approved):
Tested on MK404
To display the state I modified the `void lcdui_print_percent_done(void)` and added 
if (M79_timer_get_status()) src = _N("HO");

PR #4461, opened 2023-10-21
optimisation: simplify two LCD messages

gudnimg:
There is no need to restrict how often the message is rendered. It was being restricted to render every 1 second. We don't do this for most of the static menus. So I propose this 1 second period be removed.

Tested on MK3S+

Change in memory:
Flash: -168 bytes
SRAM: -8 bytes

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -3340,27 +3340,15 @@ static void lcd_sort_type_set() {
 #ifdef TMC2130
 static void lcd_crash_mode_info()
 {
-	lcd_update_enable(true);
-	static uint32_t tim = 0;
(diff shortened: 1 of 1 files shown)


Firmware/ultralcd.cpp
```diff
@@ -3340,27 +3340,15 @@ static void lcd_sort_type_set() {
 #ifdef TMC2130
 static void lcd_crash_mode_info()
 {
-	lcd_update_enable(true);
```
gudnimg, line 3343: **Note:** This function is always called as a “menu” and with all menus the LCD updates are always enabled. The `lcd_update_enable()` here is redundant/not useful.

3d-gussner (approved):
Tested on MK404. works well and saves resources

PR #4466, opened 2023-10-25
Optimise EEPROM access in a few places by reading/writing larger blocks

gudnimg:
Saves a bit of flash memory


3d-gussner (approved):
Looks good to me.

PR #4471, opened 2023-10-28
MMU: Set PAUSED_FOR_USER state on MMU error screen

### LINKED ISSUE #4470: [BUG] MMU error screen doesn't trigger "attention" in PrusaLink/Prusa Connect (never sets `PAUSED_FOR_USER` state)
**Printer type** - MK3S+

**MMU upgrade** - MMU3
**MMU upgrade firmware version** - 3.0.1
**SD card or USB/Octoprint**
  Both are affected. However, a connected PrusaLink Pi is needed to actually see the issue. In my case, I'm using a Pi Zero and the RPi port.

**Describe the bug**
  PrusaLink and Prusa Connect don't signal that the printer needs attention when the printer ends up on the MMU error screen. I've tried to navigate the code, and can't find anywhere in the MMU error path (`MMU2::ReportErrorHook()`) that sets the `PAUSED_FOR_USER` state (admittedly, I've only looked quite coarsely), which I think _might_ be the issue. Therefore I'm posting the issue here rather than in PrusaLink/Connect channels.
  
**To Reproduce**
  * Add a Pi with PrusaLink to the printer and connect it to Prusa Connect.
  * Trigger an MMU error during a print.
  * Notice that Prusa Connect still think the printer is printing fine.
  
**Expected behavior**
  Prusa Connect should react to the MMU error and send "Printer requires your attention" notifications.

gudnimg:
Setting this state notifies PrusaLink/PrusaConnect that the printer is waiting for user input (attention).

Resolves https://github.com/prusa3d/Prusa-Firmware/issues/4470

Change in memory:
Flash: +12 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/mmu2_reporting.cpp
+++ b/Firmware/mmu2_reporting.cpp
@@ -248,6 +248,7 @@ void ReportErrorHook(CommandInProgress /*cip*/, ErrorCode ec, uint8_t /*es*/) {
 
     switch ((uint8_t)ReportErrorHookState) {
     case (uint8_t)ReportErrorHookStates::RENDER_ERROR_SCREEN:
+        KEEPALIVE_STATE(PAUSED_FOR_USER);
         ReportErrorHookStaticRender(ei);
         ReportErrorHookState = ReportErrorHookStates::MONITOR_SELECTION;
         [[fallthrough]];
@@ -270,6 +271,7 @@ void ReportErrorHook(CommandInProgress /*cip*/, ErrorCode ec, uint8_t /*es*/) {
                 sound_wait_for_user_reset();
                 // Reset the state in case a new error is reported
                 is_mmu_error_monitor_active = false;
+                KEEPALIVE_STATE(IN_HANDLER);
                 ReportErrorHookState = ReportErrorHookStates::RENDER_ERROR_SCREEN;
                 break;
             default:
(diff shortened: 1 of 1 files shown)


firetech:
I built a test firmware with this cherry-picked on top of 3.13.2 (7aff0db6), and did some quick tests with induced failures, both from idle state (Load to nozzle) and from within a print. Seems to work. :)
image

gudnimg:
I did a few tests on my side:

- ✅ M600 with MMU (Tested the new filament change screen in 3.13.2)
- :x: `SELECTOR CANNOT HOME` when Idle. When the firmware is running the top-level `loop()` the status is reset, I rather not touch this code.

3d-gussner (approved):
Tested with MK404, PrusaLink 0.72.0 and PrusaConnect

Test Steps:

1. Start MK404 with MMU
2. Start `prusalink -s /tmp/simavr-uart0 -l prusa.link.serial.serial_adapter=DEBUG -i -f`
3. In PrusaConnect start SD print
4. During print press "F" in MK404 to manually set FINDA
5. Wait for MMU error
6. Check in PrusaLink and PrusaConnect the printer state

Before:
PrusaLink and PrusaConnect don't detect any issue and printer state says printing

With this PR:
PruasLink and PrusaConncet change state to Attention

PR #4472, opened 2023-10-29
Fix regression in MMU enable menu

gudnimg:
Fixes a regression introduced in https://github.com/prusa3d/Prusa-Firmware/pull/4454


DRracer (approved):

3d-gussner (approved):
Tested on MK404

Issue #4473, opened 2023-10-30
  [ENHANCEMENT] [Prusa-Link_Connect] G-code for MMU status for Prusalink


Ro3Deee:
**Printer type** - MK3S
**MMU Upgrade** - MMU2S (with FW 3.0.1)

**Is your enhancement related to a problem?**
With FW 3.13.0 and MMU FW 3.0.0, an option to disable/enable the MMU2 was implemented in MK3 firmware and menu (see https://github.com/prusa3d/Prusa-Firmware/issues/1876#issuecomment-1658145989).

This option or, at least, the viewing of **status of the MMU (enabled/disabled)** should be available in Prusa-Link/Connect.
  
**Describe the solution you'd like**
  @TojikCZ  says here https://github.com/prusa3d/Prusa-Link/issues/872#issuecomment-1785157285 that a gcode is necessary.

3d-gussner:
@petrubecheru Closing this issue as we implemented `M709` with https://github.com/prusa3d/Prusa-Firmware/pull/4474 

Forgot to Thank you for the idea and interacting on github. Very appreciated.

PR #4474, opened 2023-10-30
Add en- and disable MMU to M709

3d-gussner:
Add X42 to erase the MMU eeprom

@gudnimg Any optimization possible?

Related to https://github.com/prusa3d/Prusa-Firmware/issues/4473 and https://github.com/prusa3d/Prusa-Link/issues/872

- [X] Update Doxygen
- [x] Update RepRap wiki
- [x] Update `M709` for MK4
   - [X]  `S0|0` and `X0|X1` already implemented
   - [x] Add `X42` to erase the MMU eeprom

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -8581,14 +8581,15 @@ Sigma_Exit:
     } break;
 
     /*!
-    ### M709 - MMU reset <a href="https://reprap.org/wiki/G-code#M709:_MMU_reset">M709: MMU reset</a>
-    The MK3S cannot not power off the MMU, for that reason the functionality is not supported.
+    ### M709 - MMU power & reset <a href="https://reprap.org/wiki/G-code#M709:_MMU_power_&_reset">M709: MMU power & reset</a>
+    The MK3S cannot not power off the MMU, but we can en- and disable the MMU.
     #### Usage
 
-        M709 [ X ]
+        M709 [ S | X ]
 
     #### Parameters
-    - `X` - Reset MMU (0:soft reset | 1:hardware reset)
+    - `X` - Reset MMU (0:soft reset | 1:hardware reset | 42: erease MMU eeprom)
+    - `S` - En-/disable the MMU (0:off | 1:on)
 
     #### Example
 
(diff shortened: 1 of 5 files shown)


Firmware/Marlin_main.cpp
```diff
@@ -8581,24 +8581,44 @@ SERIAL_PROTOCOLPGM("\n\n");
...
     {
+        if (code_seen('S'))
+        {
+            switch (code_value_uint8())
+            {
+            case 0:
+                MMU2::mmu2.Stop();
+                break;
+            case 1:
+                MMU2::mmu2.Start();
```
gudnimg: The EEPROM must be set accordingly here, otherwise the menus may show the wrong value e.g. `MMU [ON]` when it is really off.

gudnimg (changes requested):

Firmware/Marlin_main.cpp
```diff
@@ -8581,24 +8581,44 @@ SERIAL_PROTOCOLPGM("\n\n");
...
+    ### M709 - MMU power & reset <a href="https://reprap.org/wiki/G-code#M709:_MMU_power_&_reset">M709: MMU power & reset</a>
+    The MK3S cannot not power off the MMU, but we can en- and disable the MMU.
     #### Usage
 
-        M709 [ X ]
+        M709 [ S | X ]
 
     #### Parameters
-    - `X` - Reset MMU (0:soft reset | 1:hardware reset)
+    - `X` - Reset MMU (0:soft reset | 1:hardware reset | 42: erease MMU eeprom)
```
gudnimg: erease => erase

DRracer: Also, it should be noted, that the new state of the MMU is stored in printer's EEPROM

Firmware/Marlin_main.cpp
3d-gussner: @gudnimg Added eeprom update

Firmware/Marlin_main.cpp
3d-gussner: @gudnimg fixed typo and @DRracer updated documentation

DRracer: @3d-gussner : with the EEPROM comment I meant something like this:
"The new state of the MMU is stored in printer's EEPROM - i.e. if you disable the MMU via M709, it will not be activated after the printer resets."

3d-gussner: @DRracer Better now? See https://github.com/prusa3d/Prusa-Firmware/pull/4474/commits/5cdb1916102fbac9bd5bace0509367911a6175e7

Firmware/Marlin_main.cpp
```diff
@@ -8609,10 +8633,14 @@ SERIAL_PROTOCOLPGM("\n\n");
...
                 MMU2::mmu2.Reset(MMU2::MMU2::ResetPin);
                 break;
+            case 42:
+                MMU2::mmu2.Reset(MMU2::MMU2::EraseEEPROM);
+                break;
             default:
                 break;
             }
         }
+        printf_P(_n("MMU state:%d\n"), MMU2::mmu2.Enabled());
```
gudnimg: @3d-gussner the PR looks good to me now. But I’m not sure about this log.

I feel it is not descriptive enough.

Maybe `MMU On` and `MMU Off` is better? Then it would be similar to the log I added for SpoolJoin when the printer boots up.

gudnimg (approved):
Tested on MK3S+ with MMU3.

PR #4494, opened 2023-11-12
Fix #3122

### LINKED ISSUE #3122: [BUG] Changing feedrate during pause should affect the paused print *only*
**Printer type** - MK2.5+
**Printer firmware version**-  3.9+

**MMU Upgrade** - n/a

**SD card or USB/Octoprint**
n/a

**Describe the bug**
When the print is paused, changing the feedrate (either through the Tune menu, or by rotating the encoder from the status screen) should change the feedrate of the saved print, not the feedrate of the paused state.
  
**To Reproduce**
- Start print
- Pause print
- Change the feedrate (rotate the encoder)
- Resume print: resume moves are *incorrectly* affected (bug)
- New feedrate is lost as soon as printing resumes (bug)

**Expected behavior**
Feedrate set while paused should set the feedrate of the saved print _only_.

gudnimg:
If the feedrate multiplier is changed while the print is paused update the saved feedrate multiplier in RAM

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -378,6 +378,10 @@ void save_print_file_state();
 void restore_print_file_state();
 void save_planner_global_state();
 void refresh_print_state_in_ram();
+
+/// Updates the feedrate multiplier when a print is saved such that
+/// it is not overwritten when the print is later resumed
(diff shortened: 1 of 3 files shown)


3d-gussner (approved):
Tested on MK404

PR #4514, opened 2023-11-27
MMU2: Report current tool onto the serial

DRracer:
PFW-1546

### DIFF
--- a/Firmware/mmu2.cpp
+++ b/Firmware/mmu2.cpp
@@ -72,7 +72,7 @@ void MMU2::Start() {
     PowerOn();
     mmu2Serial.flush(); // make sure the UART buffer is clear before starting communication
 
-    extruder = MMU2_NO_TOOL;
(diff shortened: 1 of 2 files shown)


3d-gussner (approved):
Tested on MK404 with MMU FW 3.0.1

- [X] Normal MMU print reports correctly
- [X] Spooljoin  works as well

End of print reports `MMU2:MMU2tool=99` which is  MMU2_NO_TOOL

DRracer:
squashed individual dev commits into one

TojikCZ:
Looks good, the only problem i see is that I cannot ask for the value to be written out and it does not get written out regularly. So I'm going to assume no tool until I see a tool getting selected.

I think that this can only happen when PrusaLink connected to the Einsy pins boots up during a single tool SD print, anything else will eventually change the active tool. So if we agree that that is fine and we don't find any other scenario, where it's a problem, I am okay with it as is. Thank you for making it

PR #4519, opened 2023-11-30
Fix M600 initial loading showing status screen

Panayiotis-git:
At the M600 sequence, after unloading the filament and selecting to start the next filament loading, only for the first time, instead of showing the message "loading filament" the status screen is shown and the progress bar at the bottom.

This PR fixes the wrong behavior.


Panayiotis-git:
The fix was  moved to the https://github.com/prusa3d/Prusa-Firmware/pull/2423

PR #4525, opened 2023-12-03
🚧 Don't show  "Set ready" & "Set not ready" if host is not PrusaLink

sarusani:
I believe the LCD menus "Set ready" & "Set not ready" are only intended for PrusaLink and not OctoPrint. (Maybe I'm wrong?)

So I added a function (M79_is_host_name_pl()) to check if M79 was set with S"PL".
If not, it will replace the menu entry "Set ready" & "Set not ready" with "Print from host" wich sends an action:start to the host.

ToDo: Translations for "Print from host"

### DIFF
--- a/Firmware/host.cpp
+++ b/Firmware/host.cpp
@@ -30,3 +30,10 @@ bool M79_timer_get_status() {
 void M79_timer_update_status() {
     M79_timer.expired(M79_TIMEOUT);
 }
+
+bool M79_is_host_name_pl() {
+    if(strcmp(GetHostStatusScreenName(), "PL") == 0) {
+        return true;
+    }
+    return false;
+}
--- a/Firmware/host.h
+++ b/Firmware/host.h
@@ -20,3 +20,7 @@ bool M79_timer_get_status();
 /// Checks if the timer period has expired. If the timer
 /// has expired, the timer is stopped
 void M79_timer_update_status();
+
+// Check if set host name is PrusaLink
+/// @returns true if the set hostname is PrusaLink ("PL"), false otherwise
+bool M79_is_host_name_pl();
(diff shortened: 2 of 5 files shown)


gudnimg:
Setting the hostname like “PL” is not required for the Set Ready feature. Hostname was only meant to be optional to change the UI.
PrusaConnect or Octoprint may use something else than “PL” or just use default.

For Set Ready and other Host specific functionality should only be active when the M79 timer is running. Setting the hostname isnt required

sarusani:
Are all features enabled by M79 intended for all host or would it make sense to have a way to differentiate between PrusaLink and others?

gudnimg:
> Are all features enabled by M79 intended for all host or would it make sense to have a way to differentiate between PrusaLink and others?

Its my understanding that it is mainly for PrusaLink/PrusaConnect at the moment but it can be used on any host if they can use the new M79 gcode which must be sent periodically to keep the timer running.

sarusani:
I made a OctoPrint plugin just for that. (https://plugins.octoprint.org/plugins/ActivatePrusaHostTimer/)

But I don't think "Set ready" makes sense for OctoPrint... So I think it would be useful to have a way to know which one is connected.

Currently I use another plugin to "convert" the "action: ready" into a "action:start". 

On the other hand, the reprint function makes perfect sense for both hosts.

sarusani:
For now I added action: interception to the OctoPrint plugin. So if the firmware sends action commands  that are unknown to OctoPrint (like action:ready), the plugin will intercept it and trigger a command that mirrors PrusaLink as close as possible.
For example: In the case of action:ready it will trigger the currently loaded print job.

So if you don't want to implement different behaviours for OctoPrint in the firmware, the plugin will try to take care of it. 😄

Issue #4531, opened 2023-12-06
[FEATURE REQUEST]<Move "Disable steppers" to main menu while not printing>


Pitkox:
**Printer type** -MK3S
  
When we change the filament, there is always a problem with leaking residue (pressure in the hotend). there are many memes where people stand over the printer at the beginning of printing with tweezers to remove this "hair" off of the nozzle. You can get rid of it by slightly pulling the filament away from the hotend. I always do this and it helps a lot to keep the nozzle clean before printing. To do this, you need to turn off the stepper motors that hold the filament after it is loaded. Wouldn't it be easier if we gave this option at the top, to the main menu only when we prepare the printer for printing, or automatically turn off the stepper motors if the printer is not printing. I understand that you can't give this option while printing, but it's different, when printing it's also not in the settings. Please consider this, I've written about this before, but the problem hasn't magically disappeared, and I don't believe I'm the only one with this problem...

sarusani:
I used to be one of the people standing over the printer with tweezers at the beginning of every print. :)

I made a change to my firmware to retract 8mm after the filament loading. (Instead of doing it manually)
Solved this issue completely. I could make a PR if that's something people are interested in... But it's not on me to decide if it will be included.

I also added "G1 E-8" to the start of my "end G-code" in the slicer to prevent oozing after a print is finished.

Pitkox:
Yeah this idea might be helpful too but I think my case will be more universal, Maybe good idea is to implement both and this retraction thing set like option (ON-OFF)?

3d-gussner:
@sarusani If you create a PR with good arguments and some test scenarios and results we will gladly look into this and consider to approve the PR in future releases (not in FW 3.14.0)
@Pitkox Moving the "Disable Steppers" to main menu will probably not approved as it changes the main menu and most users don't use "Disable Steppers" at all.

Pitkox:
It's not about using or not. It's about the problem that this could fix. I'm one and only with this leak problem? I don't think so....
At least, think about @sarusani idea but as I mentioned this approche is not universal.

PR #4534, opened 2023-12-09
Prevent oozing after filament load

sarusani:
This is my attempt to fix leaking filament issues like it's described in #4531.

After a filament load & canceled/ended prints, the filament tends to leak extensively, which will mess up the first layer of a new print if not properly cleaned.

By retracting the filament by 10mm & then extracting 2mm (to make sure it's close enough to do a successful "prime line"). I eliminated this problem on my printer completely. I use this modification for at least a year and never had any issues.
(It tends to still leak a bit when I just do a single 8mm retraction...)

This results in the same behaviour like leaving the printer preheated for about 10min. after loading filament. -> The "prime line" is not printed 100% but it starts after about 1/3 of the total distance.

I did't touch M600, so filament change during print is not affected.

This change combined with adding
```
G1 E-10
G1 E2
```
to the start of my "end G-code" in the slicer, fixed the issue for me and I don't have to hover over my printer with tweezers at the beginning of every print anymore.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3545,7 +3545,10 @@ void gcode_M701(float fastLoadLength, uint8_t mmuSlotIndex){
             lcd_load_filament_color_check();
         }
 
-        retract_for_ooze_prevention(); // Retract a little of filament to prevent oozing
+        #ifdef COMMUNITY_PREVENT_OOZE
+        // Retract filament to prevent oozing
+        retract_for_ooze_prevention();
+        #endif //COMMUNITY_PREVENT_OOZE
 
         lcd_update_enable(true);
         lcd_update(2);
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3544,6 +3544,9 @@ void gcode_M701(float fastLoadLength, uint8_t mmuSlotIndex){
         if (!farm_mode && (eFilamentAction != FilamentAction::None)) {
             lcd_load_filament_color_check();
         }
+
+        retract_for_ooze_prevention(); // Retract a little of filament to prevent oozing
+
         lcd_update_enable(true);
         lcd_update(2);
         lcd_setstatuspgm(MSG_WELCOME);
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -5603,6 +5603,15 @@ static void lcd_sd_updir()
   menu_data_reset(); //Forces reloading of cached variables.
 }
 
+/// @brief retract a little of filament to prevent oozing
+void retract_for_ooze_prevention() {
+        current_position[E_AXIS] += FILAMENTCHANGE_ROOZEFEED;
+        plan_buffer_line_curposXYZE(FILAMENTCHANGE_RFEED);
+        current_position[E_AXIS] += FILAMENTCHANGE_EOOZEFEED;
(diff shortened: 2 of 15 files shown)


3d-gussner:
@sarusani Thanks for the PR. Please provide some more information how we could test this.
As far I see you retract quite a bit and the purge line isn't extruding completely, which makes me nervous that we may end-up with some under extrusions.
"Loosing" 8 mm during filament changes needs some good arguments before considering the change.

sarusani:
I found the easiest way to test this is to:
1.  increase the nozzle temp to 5 degrees more than you would normally use for that filament. (To "simulate" the worst case oozing).
2. Load some filament (wait a couple of minutes to check for oozing)
3. Start any print, then cancel/stop it
4. Restart print.

Check if enough of the "intro/prime line" is printed to build up proper nozzle pressure in both cases.
<br />
You can also change the retraction/extrude params in the variants config file to check if lower values give you similar results with lower impact on the intro line.

In my experience only the first part of the intro line is not extruding in full. The second part, that actually builds up the nozzle pressure, always prints without problem. (This also makes sense, since it's extruding two times 9mm, so the second part should always be fine)

If I'm using the original code (without my modification) and just let the printer ooze for a couple of minutes I also loose about 5-8mm of filament to oozing. (Depends on the filament) So I think my code is not changing the outcome of the intro line too much.

I've been using this with PLA, PETG, PVB, ASA, ABS, PC Blend, PC Blend CF, Woodfill (LayWood) & Stonefill (Laybrick) filaments.

sarusani:
I did a lot more testing during the last few days and I get almost identical results by retracting 10mm and then extruding 4mm. Retracting less than 10mm always gives me worse results and by extruding more than 4mm, it immediately starts to ooze again.

In my opinion 4mm is a good compromise between ooze prevention and not compromising the "intro line".

I updated the variants configs accordingly.

3d-gussner:
@sarusani Thanks for the update. I will discuss this PR internally and try to get some answer from our PrusaSlicer profile gurus.

3d-gussner:
@sarusani I got some feedback:

> No one can tell with certainty if it will possibly break something. Also, I guess he only played with 0.4mm nozzle. The missing 8mm will create different scenario with different nozzle diameters. There is also some non-zero possibility of clogging the hotend, loads are often performed with much higher temps than target temps (for example if you unload high temp material and then load low temp material immediately).

Maybe as a compromise create a `//#define COMMUNITY_PREVENT_OOZE` in the variants and encapsulate your changes with `#idef COMMUNITY_PREVENT_OOZE` and `#endif //End COMMUNITY_PREVENT_OOZE`.

This would allow us to merge it while keeping on "Stock" printers the known settings and more experienced users can compile the firmware with this change.
With more community tests and different nozzle sizes we may consider again to make it default in the future.

sarusani:
I totally agree. I only have a very limited test set. (nozzle sizes/materials, filament types/brands).

I added #ifdef with the latest commit.

3d-gussner:
With `#define COMMUNITY_PREVENT_OOZE`
 
All values in bytes. Δ Delta to base

| Target| ΔFlash | ΔSRAM | Used Flash | Used SRAM | Free Flash | Free SRAM |
| ------| ------ | ----- | -----------| --------- | ---------- | --------- |
| `MK3S_MULTILANG` | 104 | 0 | 247192 | 5656 | 6760 | 2536 |


Cmake compiling off all variants also works. Didn't test it but I fine to approve it so the community can use it.

3d-gussner (approved):
Doesn't change anything on stock printers but can be used by the community.

PR #4541, opened 2023-12-16
Remove leading M118 from "action:ready" & "action:not_ready"

sarusani:
There should't be a leading "M118" in the sent "action:ready" & "action:not_ready" command.

M118 is a command for the printer, not the host.


3d-gussner (approved):
Tested on MK404 and works same as before but saves also 30bytes

Issue #4424, opened 2023-10-02
Bad translation or bad display of the message


LoakAudio:
**Printer type** - MK3S+ no MMU
**Printer firmware version** -  3.13.1- French language selected

**Describe the bug**
 During a filament change, when it's time to introduce the new filament the message showed is not clear, I think last letters of the message are mixed at the start of it ( see photo)

![Message Prusa](https://github.com/prusa3d/Prusa-Firmware/assets/39970118/da9567fe-8b92-44b3-a00e-70978f72d81a)

In addition, when introduced, the filament loads automatically, no need to push a button as it seems to be said on the screen
  
**To Reproduce**
Try a Print with firmware 3.13.1 in french, no MMU, with a filament change 

  
**Expected behavior**
 A showed message describing which action is needed
  
Don't hesitate if you need more information
Thanks !

3d-gussner:
@LoakAudio Thanks for reporting the issue. Yes in French the line is too long `et appuyez sur le bouton` and isn't shown correctly on LCD screen
```
 |01234567890123456789|
0|                    |
1|Inserez le filament |
2|et appuyez sur le bo|uton
```
And the `uton` overwrite part of the `Inserez`

3d-gussner:
@LoakAudio Merci beaucoup again I created a fix in #4543 and it looks like this
![Fix_4424](https://github.com/prusa3d/Prusa-Firmware/assets/25530011/74d0a7ef-0a2b-49f6-b63a-df61f94bee0d)

I am closing this issue. If you disagree please reopen it again.

Issue #3361, opened 2022-01-23
[BUG] Settings/Calibrate/Calibrate Z should not have a "right arrow"


gudnimg:
**Printer type** - [e.g. MK3S, MK3, MK2.5S, MK2.5, MK2S, MK2]
**Printer firmware version** -  3.10.1

**MMU upgrade** - N/A
**MMU upgrade firmware version** -  N/A

**SD card or USB/Octoprint**
 N/A

**Describe the bug**
I've interpreted menus that **don't** have the right arrow to mean that if you select that menu then there is immediate action from the printer i.e. the printer will do something. And when a right arrow is present then that means there is an additional submenu below. This is not the case with Calibrate Z.
  
![271940400_660531225301754_8810136371854693825_n](https://user-images.githubusercontent.com/8218499/150676548-718b2160-d9cb-48eb-9761-25d93d893b30.jpg)

**To Reproduce**
N/A
  
**Expected behavior**
I would expect there to not be a right arrow next to "Calibrate Z" just like "Calibrate XYZ"

leptun:
Noticed this issue as well.
Not sure why MENU_ITEM_SUBMENU_P was used in place of MENU_ITEM_FUNCTION_P. I think that is a bug or it was done this way to hide some other issue. This should be investigated.
If you see any other inconsistencies like this, please add them to the issue report.

Panayiotis-git:
Is there a reason that the Move X, Move Y and Move Z follows the exact same logic? They all are a single field form that require the use of the rotation wheel to change a value.

leptun:
The "Move" menu items are in fact submenus. Even though it doesn't look like it, they are.
Think of it like this: Does clicking the item change the screen to something else entirely? Then it's a submenu. Otherwise, the view will stay right there (like with a toggle or when pressing Auto Home), so it's just a menu item (not a submenu), hence why it should have no arrow in that case.

I think all menu items should have standardized rules. The current situation is suboptimal, especially considering that the older menu items seem not to confirm to how menus look now.

Here is another example of an inconsistency:
In some menus the "back -^" menu text field actually says the name of the menu above, not the generic "Back". Not all menus do this and some have nasty hacks so that the correct back message is displayed. If there are going to be any changes to the UI, I think it might be worth making everything consistent (so a standard must be decided on) and some of these hacks need to be toned down a bit since they make the UI code really ugly and don't really add anything to the look of the UI.

Panayiotis-git:
[...]

PR #4543, opened 2023-12-18
MK3 cleanup menu

3d-gussner:
Cleanup few menus
- Calibrate XYZ
- Calibrate Z
   - Add Sheet messages
- Mesh Bed Leveling
- Reset XYZ calibr.
   - Add Yes/No menu with `No` time out
- Thermal Model cal.
- Add Yes/No menu with `No` time out to Reset in HW Setup -> Sheets ->
- Change Stop print menu
   - Add timeout to Stop print menu
- Preheat without 

See also https://github.com/prusa3d/Prusa-Firmware/issues/3361

Todo:

- [x] https://github.com/prusa3d/Prusa-Firmware/issues/4424


Firmware/ultralcd.cpp
```diff
@@ -2097,16 +2097,16 @@ void lcd_generic_preheat_menu()
...
-        MENU_ITEM_SUBMENU_P(bPreheatOnlyNozzle ? PSTR("ABS  -  " STRINGIFY(ABS_PREHEAT_HOTEND_TEMP)) : PSTR("ABS  -  " STRINGIFY(ABS_PREHEAT_HOTEND_TEMP)  "/" STRINGIFY(ABS_PREHEAT_HPB_TEMP)) , mFilamentItem_ABS);
-        MENU_ITEM_SUBMENU_P(bPreheatOnlyNozzle ? PSTR("HIPS -  " STRINGIFY(HIPS_PREHEAT_HOTEND_TEMP)): PSTR("HIPS -  " STRINGIFY(HIPS_PREHEAT_HOTEND_TEMP) "/" STRINGIFY(HIPS_PREHEAT_HPB_TEMP)), mFilamentItem_HIPS);
-        MENU_ITEM_SUBMENU_P(bPreheatOnlyNozzle ? PSTR("PP   -  " STRINGIFY(PP_PREHEAT_HOTEND_TEMP))  : PSTR("PP   -  " STRINGIFY(PP_PREHEAT_HOTEND_TEMP)   "/" STRINGIFY(PP_PREHEAT_HPB_TEMP))  , mFilamentItem_PP);
-        MENU_ITEM_SUBMENU_P(bPreheatOnlyNozzle ? PSTR("FLEX -  " STRINGIFY(FLEX_PREHEAT_HOTEND_TEMP)): PSTR("FLEX -  " STRINGIFY(FLEX_PREHEAT_HOTEND_TEMP) "/" STRINGIFY(FLEX_PREHEAT_HPB_TEMP)), mFilamentItem_FLEX);
+        MENU_ITEM_FUNCTION_P(bPreheatOnlyNozzle ? PSTR("PLA  -  " STRINGIFY(PLA_PREHEAT_HOTEND_TEMP)) : PSTR("PLA  -  " STRINGIFY(PLA_PREHEAT_HOTEND_TEMP)  "/" STRINGIFY(PLA_PREHEAT_HPB_TEMP)) , mFilamentItem_PLA);
```
leptun, line 2100: This change is not ok. MENU_ITEM_SUBMENU_P works in a different way to MENU_ITEM_FUNCTION_P. With the item function, the function that is provided is simply executed on click and the "current" menu stays the same as before. With item submenu (as it was before), the current menu is also switched to the function that is provided, so subsequent lcd updates execute that function again instead of the parent menu. So while this change does fix how the menu item looks, it is fundamentally changing how the action is performed, and in this case it is incorrect.

Firmware/ultralcd.cpp
```diff
@@ -4562,25 +4544,26 @@ static void lcd_calibration_menu()
 #endif //TMC2130
     MENU_ITEM_FUNCTION_P(_i("Selftest"), lcd_selftest_v);////MSG_SELFTEST c=18
     // MK2
-    MENU_ITEM_FUNCTION_P(_i("Calibrate XYZ"), lcd_mesh_calibration);////MSG_CALIBRATE_BED c=18
+    MENU_ITEM_GCODE_P(_i("Calibrate XYZ"), PSTR("M45"));////MSG_CALIBRATE_BED c=18
     // "Calibrate Z" with storing the reference values to EEPROM.
-    MENU_ITEM_SUBMENU_P(_T(MSG_HOMEYZ), lcd_mesh_calibration_z);
+    MENU_ITEM_GCODE_P(_T(MSG_HOMEYZ), PSTR("M45 Z"));
```
leptun, line 4549: unintended behaviour change. The code change makes it so clicking these menu items doesn't go back to the status screen. It also makes it possible to spam click this

Firmware/ultralcd.cpp
```diff
@@ -5118,7 +5103,7 @@ static void lcd_sheet_menu()
         MENU_ITEM_SUBMENU_P(_T(MSG_V2_CALIBRATION), activate_calibrate_sheet);
     }
     MENU_ITEM_SUBMENU_P(_i("Rename"), lcd_rename_sheet_menu); ////MSG_RENAME c=18
-	MENU_ITEM_FUNCTION_P(_T(MSG_RESET), lcd_reset_sheet);
+	MENU_ITEM_SUBMENU_P(_T(MSG_RESET), lcd_reset_sheet);
```
leptun, line 5106: Wrong behaviour. see other comment.

Firmware/ultralcd.cpp
```diff
@@ -5695,18 +5680,21 @@ void lcd_thermal_model_cal()
...
+}
+}
 
-    // Show No options first, the default selection
-    MENU_ITEM_FUNCTION_P(_T(MSG_NO), lcd_return_to_status);
-    MENU_ITEM_FUNCTION_P(_T(MSG_YES), lcd_print_stop);
-    MENU_END();
+void lcd_calibration_bed_reset()
+{
+uint8_t result = lcd_show_fullscreen_message_yes_no_and_wait_P(_T(MSG_CALIBRATE_BED_RESET), true, LCD_MIDDLE_BUTTON_CHOICE);
```
leptun, line 5693: this isn't really a question message. Looks a bit weird to me

Firmware/ultralcd.cpp
```diff
@@ -4562,25 +4544,26 @@ static void lcd_calibration_menu()
...
+    MENU_ITEM_SUBMENU_P(_T(MSG_CALIBRATE_BED_RESET), lcd_calibration_bed_reset);
 #ifdef PINDA_THERMISTOR
     if(has_temperature_compensation())
+        //Keep as function menu to show status screen and process of the calibration
         MENU_ITEM_FUNCTION_P(_T(MSG_PINDA_CALIBRATION), lcd_calibrate_pinda);
 #endif
 
 #ifdef THERMAL_MODEL
-    MENU_ITEM_SUBMENU_P(_n("Thermal Model cal."), lcd_thermal_model_cal);
+    MENU_ITEM_FUNCTION_P(_n("Thermal Model cal."), lcd_thermal_model_cal);
```
leptun, line 4566: This is a good change. The behaviour was wrong before, but it was saved by the lcd_return_to_status() in the function.

Firmware/ultralcd.cpp
```diff
@@ -4562,25 +4544,26 @@ static void lcd_calibration_menu()
 #endif //TMC2130
     MENU_ITEM_FUNCTION_P(_i("Selftest"), lcd_selftest_v);////MSG_SELFTEST c=18
     // MK2
-    MENU_ITEM_FUNCTION_P(_i("Calibrate XYZ"), lcd_mesh_calibration);////MSG_CALIBRATE_BED c=18
+    MENU_ITEM_GCODE_P(_i("Calibrate XYZ"), PSTR("M45"));////MSG_CALIBRATE_BED c=18
```
leptun, line 4547: see other comment

Firmware/ultralcd.cpp
```diff
@@ -4562,25 +4544,26 @@ static void lcd_calibration_menu()
...
     // "Calibrate Z" with storing the reference values to EEPROM.
-    MENU_ITEM_SUBMENU_P(_T(MSG_HOMEYZ), lcd_mesh_calibration_z);
+    MENU_ITEM_GCODE_P(_T(MSG_HOMEYZ), PSTR("M45 Z"));
 
-    MENU_ITEM_SUBMENU_P(_T(MSG_MESH_BED_LEVELING), lcd_mesh_bedleveling); ////MSG_MESH_BED_LEVELING c=18
+    MENU_ITEM_GCODE_P(_T(MSG_MESH_BED_LEVELING), PSTR("G80")); ////MSG_MESH_BED_LEVELING c=18
```
leptun, line 4551: see other comment

Firmware/ultralcd.cpp
leptun: Moreover, this will appear as if no problem is there, but it is there. It only manifests itself next time lcd_update gets called, which might be quite a bit further in the future, but it will still happen. I think I was able to make it render the wrong screen by loading to nozzle with the MMU. The parent menu got rendered during the purging moves (before the Is color clear prompt)

Firmware/ultralcd.cpp
```diff
@@ -5695,18 +5680,21 @@ void lcd_thermal_model_cal()
...
 void lcd_sdcard_stop()
 {
-    // Show static message
-    lcd_puts_at_P(0, 0, _T(MSG_STOP_PRINT));
-    lcd_putc_at(0, 1, '\n');
-
-    MENU_BEGIN();
-    // Skip first two LCD rows used by static message
-    if(menu_row == 0) menu_row = 2;
+uint8_t result = lcd_show_fullscreen_message_yes_no_and_wait_P(_T(MSG_STOP_PRINT), true, LCD_MIDDLE_BUTTON_CHOICE);
```
leptun, line 5683: This change is not ok. It might save resources since it reuses code, but it breaks one of the quirks of the old implementation: gcodes are no longer queued while the message is displayed. This is in contrast to the old implementation which behaved like a menu that gets periodically re-rendered by lcd_update. As a result, with this change when you click stop print, the printer just drains all the queued moves in the planner queue and then the planner stalls since gcode are not processed during lcd_show_fullscreen_message_yes_no_and_wait_P

leptun (changes requested):

Firmware/ultralcd.cpp
3d-gussner: Just want a Yes No confirmation instead just to reset the XYZ and Z offset values

Firmware/ultralcd.cpp
gudnimg: I agree with @leptun this change is not OK as is.

Firmware/ultralcd.cpp
gudnimg: I agree with @leptun here. This menu must be “non-blocking” so it doesnt halt the print.

PR #4547, opened 2023-12-21
Fix few PrusaLink/PrusaConnect reported issue

3d-gussner:
- [X] Don't show `Set Ready` when not idle
- [X] Don't show `Print from SD` when host print is paused

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -979,7 +979,7 @@ static void fw_crash_init()
     if(xfdump_check_state(&crash_reason))
     {
         // always signal to the host that a dump is available for retrieval
-        puts_P(_N("// action:dump_available"));
+        puts_P(_N("//action:dump_available"));
 
 #ifdef EMERGENCY_DUMP
         if(crash_reason != dump_crash_reason::manual &&
@@ -6003,7 +6003,7 @@ Sigma_Exit:
             // but up to this point the printer has been waiting for the M79 from the host
             // Send action to the host, so the host can resume the print. It is up to the host
             // to resume the print correctly.
-            SERIAL_ECHOLNRPGM(MSG_OCTOPRINT_UVLO_RECOVERY_READY);
+            SERIAL_ECHOLNRPGM(MSG_HOST_ACTION_UVLO_RECOVERY_READY);
         }
 
         break;
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -522,7 +522,8 @@ bool __attribute__((noinline)) printer_active() {
         || (lcd_commands_type != LcdCommands::Idle)
         || MMU2::mmu2.MMU_PRINT_SAVED()
         || homing_flag
-        || mesh_bed_leveling_flag;
+        || mesh_bed_leveling_flag
+        || (eeprom_read_byte((uint8_t*)EEPROM_UVLO) != PowerPanic::NO_PENDING_RECOVERY);
 }
 
 // Currently only used in one place, allowed to be inlined
@@ -4126,8 +4127,8 @@ void process_commands()
                 // M24 - Start SD print
                 enquecommand_P(MSG_M24);
 
-            // Print is recovered, clear the recovery flag
-            eeprom_update_byte((uint8_t*)EEPROM_UVLO, PowerPanic::NO_PENDING_RECOVERY);
+                // Print is recovered, clear the recovery flag
+                eeprom_update_byte((uint8_t*)EEPROM_UVLO, PowerPanic::NO_PENDING_RECOVERY);
             }
             else if (eeprom_read_byte((uint8_t*)EEPROM_UVLO_PRINT_TYPE) == PowerPanic::PRINT_TYPE_USB)
             {
@@ -5998,7 +5999,9 @@ Sigma_Exit:
             }
         }
 
-        if (eeprom_read_byte((uint8_t*)EEPROM_UVLO_PRINT_TYPE) == PowerPanic::PRINT_TYPE_USB && printingIsPaused()) {
+        if (eeprom_read_byte((uint8_t*)EEPROM_UVLO_PRINT_TYPE) == PowerPanic::PRINT_TYPE_USB
+           && eeprom_read_byte((uint8_t*)EEPROM_UVLO) != PowerPanic::NO_PENDING_RECOVERY
+           && printingIsPaused()) {
             // The print is in a paused state. The print was recovered following a power panic
             // but up to this point the printer has been waiting for the M79 from the host
             // Send action to the host, so the host can resume the print. It is up to the host
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -321,6 +321,7 @@ uint8_t saved_bed_temperature; //!< Bed temperature
 bool saved_extruder_relative_mode;
 uint8_t saved_fan_speed = 0; //!< Print fan speed
 //! @}
+bool uvlo_auto_recovery_ready = false;
 
 static int saved_feedmultiply_mm = 100;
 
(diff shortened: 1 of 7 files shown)


gudnimg:
Could you please somehow remove commits [ecd75e9](https://github.com/prusa3d/Prusa-Firmware/pull/4547/commits/ecd75e933aa8c8e3bb7a92cc6573ce5deef5e333) and [3bc4aba](https://github.com/prusa3d/Prusa-Firmware/pull/4547/commits/3bc4aba266336d40683edab6865a0c1c603bde37) from the PR since they have nothing to do with the proposed changes.

3d-gussner:
> Could you please somehow remove commits [ecd75e9](https://github.com/prusa3d/Prusa-Firmware/pull/4547/commits/ecd75e933aa8c8e3bb7a92cc6573ce5deef5e333) and [3bc4aba](https://github.com/prusa3d/Prusa-Firmware/pull/4547/commits/3bc4aba266336d40683edab6865a0c1c603bde37) from the PR since they have nothing to do with the proposed changes.

Done, i did cherry-picks

Firmware/ultralcd.cpp
```diff
@@ -5272,6 +5272,7 @@ static void lcd_main_menu()
 
     // only allow starting SD print if hardware errors (temperature or fan) are cleared
     if(!get_temp_error()
+            && !printer_active()
```
gudnimg, line 5301: I feel `printingIsPaused()` is more appropriate here given the commit message. `printer_active` checks for a lot more than just paused print state.

Firmware/ultralcd.cpp
```diff
@@ -5261,7 +5263,11 @@ static void lcd_main_menu()
             }
         }
     }
-    if((printJobOngoing() || printingIsPaused()) && (custom_message_type != CustomMsg::MeshBedLeveling) && !processing_tcode) {
+    if((printJobOngoing()
+        || printingIsPaused()
+        || (eeprom_read_byte((uint8_t*)EEPROM_UVLO) != PowerPanic::NO_PENDING_RECOVERY))
```
gudnimg, line 5288: Not sure about this `eeprom_read_byte((uint8_t*)EEPROM_UVLO) != PowerPanic::NO_PENDING_RECOVERY`.

Doesn't this mean you can _stop_ a print while power panic recovery is in progress? Maybe I'm missing something here. What does this fix?

In my opinion a print job can only be stopped if it is _ongoing_ or _paused_.

Firmware/Marlin_main.cpp
```diff
@@ -321,6 +321,7 @@ uint8_t saved_bed_temperature; //!< Bed temperature
 bool saved_extruder_relative_mode;
 uint8_t saved_fan_speed = 0; //!< Print fan speed
 //! @}
+bool uvlo_auto_recovery_ready = false;
```
gudnimg: Nitpick: Please make this `static bool` :)

gudnimg (changes requested):
I went through each commit one by one and have a few comments.

Firmware/Marlin_main.cpp
```diff
@@ -979,7 +979,7 @@ static void fw_crash_init()
     if(xfdump_check_state(&crash_reason))
     {
         // always signal to the host that a dump is available for retrieval
-        puts_P(_N("// action:dump_available"));
+        puts_P(_N("//action:dump_available"));
```
DRracer, line 984: Is it really compatible with Octoprint? Just making sure...

Firmware/messages.cpp
```diff
@@ -228,15 +228,15 @@ const char MSG_ENDSTOP_OPEN[] PROGMEM_N1 = "open"; ////
...
-const char MSG_OCTOPRINT_ASK_PAUSE[] PROGMEM_N1 = "// action:pause"; ////
-const char MSG_OCTOPRINT_PAUSED[] PROGMEM_N1 = "// action:paused"; ////
-const char MSG_OCTOPRINT_ASK_RESUME[] PROGMEM_N1 = "// action:resume"; ////
-const char MSG_OCTOPRINT_RESUMED[] PROGMEM_N1 = "// action:resumed"; ////
-const char MSG_OCTOPRINT_CANCEL[] PROGMEM_N1 = "// action:cancel"; ////
-const char MSG_OCTOPRINT_READY[] PROGMEM_N1 = "// action:ready"; ////
-const char MSG_OCTOPRINT_NOT_READY[] PROGMEM_N1 = "// action:not_ready"; ////
-const char MSG_OCTOPRINT_START[] PROGMEM_N1 = "// action:start"; ////
-const char MSG_OCTOPRINT_UVLO_RECOVERY_READY[] PROGMEM_N1 = "// action:uvlo_recovery_ready"; ////
+const char MSG_HOST_ACTION_ASK_PAUSE[] PROGMEM_N1 = "//action:pause"; ////
```
DRracer, line 231: Is it really compatible with Octoprint? Just making sure...

Firmware/Marlin_main.cpp
```diff
@@ -522,7 +522,8 @@ bool __attribute__((noinline)) printer_active() {
         || (lcd_commands_type != LcdCommands::Idle)
         || MMU2::mmu2.MMU_PRINT_SAVED()
         || homing_flag
-        || mesh_bed_leveling_flag;
+        || mesh_bed_leveling_flag
+        || (eeprom_read_byte((uint8_t*)EEPROM_UVLO) != PowerPanic::NO_PENDING_RECOVERY);
```
DRracer, line 527: Reading the eeprom takes some time, so any call to `printer_active` will get slower. Have you considered that?

Firmware/messages.cpp
3d-gussner: Yes I tested Octoprint and PrusaLink.
Only change is the naming of the message to HOST as we have now PrusaLink, PrusaConnect, Octoprint and other host system

Firmware/ultralcd.cpp
3d-gussner: The `Print from SD` is shown during the power panic recovery while heating.
These menus should only be shown when the printer is "idle" or not active. This is working well while the it is printing and some other conditions now. Lot of these menus below check `!printer_active()` to NOT be shown.
   
Print from SD
Auto/Load filament
Preload to MMU
Load to nozzle
Unload filament
Eject from MMU
Cut filament
Settings
Calibration

Firmware/ultralcd.cpp
3d-gussner: You can't get out of the power panic recovery endless loop in case the print has failed and you DON'T want to continue.

Firmware/Marlin_main.cpp
3d-gussner: Done

Firmware/Marlin_main.cpp
3d-gussner: Only time critical usage of `printer_active` I can see is in `handleSafetyTimer`.

I could move the this to the each ultralcd.cpp menu to avoid being shown during power panic recovery. This may cost bit more but would not "slow down" other functions checking the `printer_active` state.

3d-gussner: Any other suggestions?

DRracer: in such a case it is probably ok and no changes are really necessary :+1:

Firmware/ultralcd.cpp
gudnimg: Note this will also affect SD prints (no Host involved). When you click "Stop print" while the printer is still recovering the print, does the printer stop correctly? If so then we can resolve this thread.

Does `EEPROM_UVLO` get reset to `NO_PENDING_RECOVERY`?

gudnimg: Looked at the code in more detail and I can see that `cancel_saved_printing()` resets `EEPROM_UVLO` when the print is stopped. 👍

gudnimg (approved):
Looks good to me.

PR #4548, opened 2023-12-21
Use M117 to set label for the filament to load at the M600 Insert filament screen

Panayiotis-git:
The M600 G-code command is often used during a print job, to trigger a filament change to a different material/color, without the need of additional specialized hardware.
This pull request (PR) extends the command's functionality, by allowing the users to specify which filament to load, adding more flexibility to the multi-color printing process.

In contrast with the solution implemented at the https://github.com/prusa3d/Prusa-Firmware/pull/2380, this PR tries to not modify the command's parameters, but rather use the M117 command to set the next filament's label:

```
M117 Deep Blue
M600
```
The Insert filament screen will then show the custom text defined with the M117 command:
```
Insert filament
Deep Blue
and press the knob
```

With this PR, the M117 content is also stored in a new text buffer that is updated only by the M117 command. The content remains in the buffer until either of the following happens:
- a subsequent M117 stores a new content to the buffer, or
- a M600 command is issued.  The M600 command checks this buffer and if it finds any content, it prints it at the insert filament screen and afterwards it clears the buffer.

There are two drawbacks to this solution:
1. The M117 command will immediately update the status screen with the custom message. It will remain visible during the M600's initial printing head movement. This is a rather cosmetic side-effect and it will not affect the command's functionality.
2. The M600 will consume any message that was previously set by an M117 command. To clear unwanted messages, an empty M117 command should be issued before the M600:
```
M117
M600
```

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -3456,6 +3456,8 @@ static void mmu_M600_load_filament(bool automatic) {
 static void gcode_M600(const bool automatic, const float x_position, const float y_position, const float z_shift, const float e_shift, const float e_shift_late) {
     st_synchronize();
 
+    lcd_setstatuspgm(MSG_FILAMENTCHANGE);
+
     // When using an MMU, save the currently use slot number
     // so the firmware can know which slot to eject after the filament
     // is unloaded.
@@ -3527,6 +3529,8 @@ static void gcode_M600(const bool automatic, const float x_position, const float
             repeat = M600_check_state_and_repeat();
     }
     while (repeat);
+
+    lcd_clear_generic_use_text();
 
     lcd_update_enable(true);
 
@@ -6618,8 +6622,11 @@ Sigma_Exit:
     */
     case 117: {
         const char *src = strchr_pointer + 4; // "M117"
-        lcd_setstatus(*src == ' '? src + 1: src);
+        while (*src == ' ') src++;
+        lcd_setstatus(src);
         custom_message_type = CustomMsg::M117;
+
+        lcd_set_generic_use_text(src);
     }
     break;
 
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -81,6 +81,9 @@ static uint8_t lcd_status_message_level;
 static uint8_t lcd_status_message_idx = 0;
 static char lcd_status_message[LCD_WIDTH + 1];
 
+/* Buffer for a generic LCD text */
+static char lcd_generic_use_text[LCD_WIDTH + 1];
+
 /* !Configuration settings */
 
 static uint8_t lay1cal_filament = 0;
@@ -2160,11 +2163,17 @@ void lcd_wait_interact() {
   lcd_clear();
 
   lcd_puts_at_P(0, 0, _T(MSG_INSERT_FILAMENT));
+  lcd_set_cursor(0, 1);
+  if (lcd_generic_use_text[0]) {
+    lcd_print(lcd_generic_use_text);
+    lcd_set_cursor(0, 2);
+  }
+
 #ifdef FILAMENT_SENSOR
   if (!fsensor.getAutoLoadEnabled())
 #endif //FILAMENT_SENSOR
   {
-    lcd_puts_at_P(0, 1, _T(MSG_PRESS));
+    lcd_puts_P(_T(MSG_PRESS));
   }
 }
 
@@ -2279,11 +2288,11 @@ void show_preheat_nozzle_warning()
 void lcd_load_filament_color_check()
 {
     uint8_t clean = lcd_show_multiscreen_message_with_choices_and_wait_P(_T(MSG_FILAMENT_CLEAN), false, LCD_LEFT_BUTTON_CHOICE, _T(MSG_YES), _T(MSG_NO), _T(MSG_EJECT), 8);
-    while (clean == LCD_MIDDLE_BUTTON_CHOICE) {
-        load_filament_final_feed();
-        st_synchronize();
+	while (clean == LCD_MIDDLE_BUTTON_CHOICE) {
+		load_filament_final_feed();
+		st_synchronize();
         clean = lcd_show_multiscreen_message_with_choices_and_wait_P(_T(MSG_FILAMENT_CLEAN), false, LCD_LEFT_BUTTON_CHOICE, _T(MSG_YES), _T(MSG_NO), _T(MSG_EJECT), 8);
-    }
+	}
     if (clean == LCD_RIGHT_BUTTON_CHOICE) {
         unload_filament(FILAMENTCHANGE_FINALRETRACT);
     }
@@ -3882,16 +3891,16 @@ void lcd_wizard(WizState state)
 			} else {
 				raise_z_above(MIN_Z_FOR_SWAP);
 				if(!MMU2::mmu2.Enabled()) {
-					//current filament needs to be unloaded and then new filament should be loaded
-					//start to preheat nozzle for unloading remaining PLA filament
-					setTargetHotend(PLA_PREHEAT_HOTEND_TEMP);
+				//current filament needs to be unloaded and then new filament should be loaded
+				//start to preheat nozzle for unloading remaining PLA filament
+				setTargetHotend(PLA_PREHEAT_HOTEND_TEMP);
 					lcd_display_message_fullscreen_P(_T(MSG_WIZARD_WILL_PREHEAT));
-					wait_preheat();
-					unload_filament(FILAMENTCHANGE_FINALRETRACT); // unload current filament
-					lcd_wizard_load(); // load filament
-					setTargetHotend(0); //we are finished, cooldown nozzle
+				wait_preheat();
+				unload_filament(FILAMENTCHANGE_FINALRETRACT); // unload current filament
+				lcd_wizard_load(); // load filament
+				setTargetHotend(0); //we are finished, cooldown nozzle
 				}
-			state = S::Restore;
+				state = S::Restore;
 			}
 			break;
 #ifdef THERMAL_MODEL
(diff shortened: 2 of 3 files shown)


3d-gussner:
@Panayiotis-git First test looks good. The M117 message is shown correctly during M600. :tada: 

- [X] MMU3 print :heavy_check_mark: It behaves the same as before
- [X] MMU3 print with manually injected M600 :heavy_check_mark: It behaves the same as before

Firmware/ultralcd.cpp
```diff
@@ -2121,7 +2124,12 @@ void lcd_wait_interact() {
 
   lcd_clear();
 
-  lcd_puts_at_P(0, 1, _i("Insert filament"));////MSG_INSERT_FILAMENT c=20
+  lcd_puts_at_P(0, 0, _i("Insert filament"));////MSG_INSERT_FILAMENT c=20
+  if (lcd_generic_use_text) {
```
3d-gussner: Please change to `if (lcd_generic_use_text[0]) {

Firmware/ultralcd.cpp
```diff
@@ -2121,7 +2124,12 @@ void lcd_wait_interact() {
...
+      lcd_set_cursor(0, 1);
+      lcd_print(lcd_generic_use_text);
+  }
```
3d-gussner, line 2170: Changing to
```
  lcd_set_cursor(0, 2);
+  } else {
+    lcd_set_cursor(0, 1);
   }
   
 #ifdef FILAMENT_SENSOR
   if (!fsensor.getAutoLoadEnabled())
 #endif //FILAMENT_SENSOR
   {
-    lcd_puts_at_P(0, 2, _i("and press the knob"));////MSG_PRESS c=20 r=2
+    lcd_puts_P(_i("and press the knob"));////MSG_PRESS c=20 r=2
   }
```
keeps the message same as before

3d-gussner (changes requested):
Please have a look at the changes i made
It also fixes issue #4424
[Fix_PR4548.diff.txt](https://github.com/prusa3d/Prusa-Firmware/files/13751551/Fix_PR4548.diff.txt)

3d-gussner:
Here some test pictures with my changes in French as it is one of the messages that "normally" exceeds the limit of 20 char. 
First Change with `M117 Deep Blue` followed by `M600`
Fri Dec 22 11:16:26 2023
Second `M600` Change without `M117`. This is the same as before.
Fri Dec 22 11:16:58 2023

Only minor issue I see is if someone sends a different `M117` long time before the `M600`it will also be shown, but that is something we can't prevent from happening without an extra parameter to `M117` or `M600`

Firmware/Marlin_main.cpp
```diff
@@ -3471,7 +3471,9 @@ static void gcode_M600(const bool automatic, const float x_position, const float
             repeat = M600_check_state_and_repeat();
     }
     while (repeat);
-
+
```
3d-gussner: Remove white space

Firmware/Marlin_main.cpp
```diff
@@ -3471,7 +3471,9 @@ static void gcode_M600(const bool automatic, const float x_position, const float
             repeat = M600_check_state_and_repeat();
     }
     while (repeat);
-
+    
+    lcd_clear_generic_use_text();
+
```
3d-gussner: Remove white space

3d-gussner (changes requested):
Checked for white spaces with `git diff upstream/MK3` and found few.
Please remove these.

Firmware/ultralcd.cpp
```diff
@@ -7232,6 +7240,17 @@ uint8_t get_message_level()
 {
 	return lcd_status_message_level;
 }
+/**/
```
3d-gussner: Did you plan to add some comments  here?

3d-gussner:
@Panayiotis-git Thank you so much to provide a new PR for your initial awesome idea from 19 December 2019. Impressive that you are still willing to work on this 4 years later :hugs:

3d-gussner:
@gudnimg Do you have an idea how we can not spend 21 bytes of RAM?

sarusani:
Two (maybe stupid) ideas to mitigate the mentioned drawbacks. Feel free to ignore :)

1) Use a dedicated prefix in the M117 msg when it's intended for M600. Have M117 not print that msg but store it for later use in M600.

2) Change the order and have M600 only consume the M117 msg when it's received between the start of M600 and the rendering of the msg. (Basically the time it takes to do the initial printing head movement)
I don't know at all how the receive buffer works or if a host would even send commands after a M600. So this might not be technically possible....

Firmware/ultralcd.cpp
Panayiotis-git: You are right, I was trying something and forgot to remove my marks

Panayiotis-git:
@3d-gussner, big thanks for your review! I've integrated the patch with your changes and fixes.

After a bit of tweaking, the PR now only needs 62 bytes of ROM.

Hey @sarusani, love your suggestions! Sadly, the second one can't be done, but the first one could be a great solution. @3d-gussner, could you share your thoughts on whether this change is desirable, and if so, what m117 prefix should we use?

3d-gussner:
@Panayiotis-git and @sarusani The previous PR #2380 has been rejected as it would need changes in PrusaSlicer by adding a parameter to `M600` . So not sure if a change to `M117` with a parameter will be approved.

Firmware/Marlin_main.cpp
```diff
@@ -6531,8 +6533,11 @@ SERIAL_PROTOCOLPGM("\n\n");
     */
     case 117: {
         const char *src = strchr_pointer + 4; // "M117"
-        lcd_setstatus(*src == ' '? src + 1: src);
+        while (isspace(*src)) src++;
```
gudnimg: Potential code size optimisation: `while (*src == ' ') src++;` only needs 10 bytes of flash but `while (isspace(*src)) src++;` needs 36 bytes.

Panayiotis-git: @gudnimg, thank you for reviewing this PR. Indeed, your suggestion dramatically reduce the required storage in bytes.

Panayiotis-git:
In order to mitigate the first drawback mentioned at the opening post, the M600, when issued, initializes the status message to the MSG_FILAMENTCHANGE.

3d-gussner:
@gudnimg Please have a look if we can save any RAM. I guess not possible without modifying M600 or M117

gudnimg:
I don't see any good way to free up the RAM. I think the 20 bytes is OK since we have plenty available (2500 bytes). 👍

3d-gussner (approved):

3d-gussner:
@Panayiotis-git The PR looks good but i found while testing again a minor issue.
Steps:
1. Connect with a serial terminal
2. Start a print
3. Send `M117 Deep Blue`
4. Send `M600`
   1. The Status line shows for a moment a `N                       ` 
   2. This only happens when a `M117`has been sent before. A `M600` on its own works as before and has no weird `N` Status line.

Please check your PR. I would guess that 


Here a picture:
Selection_016

3d-gussner:
Issue with PrusaLink active and this PR.
Mon Jan  8 16:53:47 2024

3d-gussner:
We need to rethink this PR as `M117` is used by PrusaLink to show the status. At this moment I don't see any other option than adding a parameter to `M117` or `M600`. I will discuss this within our team. Sorry for the delay @Panayiotis-git as I really like your idea.

Panayiotis-git:
> @Panayiotis-git The PR looks good but i found while testing again a minor issue. Steps:
> 
>     1. Connect with a serial terminal
> 
>     2. Start a print
> 
>     3. Send `M117 Deep Blue`
> 
>     4. Send `M600`
>        
>        1. The Status line shows for a moment a `N                       `

@3d-gussner, I was not able to reproduce this behavior.
Whenever I issue the `M600` command, the status line changes immediately to `Change filament`. Either, I've sent an `M117` previously or not.

3d-gussner:
@Panayiotis-git 
> Whenever I issue the M600 command, the status line changes immediately to Change filament. Either, I've sent an M117 previously or not.

The MK404 shows quite often LCD issues which aren't seen in real hardware. But as you can see there is a `N` in the status line running this PR test. It is repeatable so there must be somewhere a minor issue.

3d-gussner:
My biggest issue with this PR now is that the combination of PrusaLink, which uses the Status line `M117` very often to show the state of PrusaLink and PrusaConnect "break" the `M600` gcode.
Sorry have to move it to FW3.14.1 until we decide if we can use some parameters or find a different solution.

Panayiotis-git:
While testing with an actual SD card print, I've noticed something about the command `M117`. Unlike the 'M600' command, the `M117` does not wait for the planner to complete the pending moves. It immediately sets the status message with the filament name, even if the printer still executes earlier move commands. On the other hand, the command `M600` waits for the planner to complete the moves and then sets the status to `Change filament` and initiates the change filament sequence. This results in the filament name being visible at the status line for a much longer time.

To address this, we can move the `lcd_setstatuspgm(MSG_FILAMENTCHANGE);` as the first line of code after the `case 600:` statement. However, this will cause the 'Change filament' message to be shown much sooner, potentially before the planner completes the moves.  This may affect the user experience by displaying the message prematurely.

One solution to this situation could be to add a `G4` or a `M400` command before the `M117`. Something like this:
```
M400
M117 Deep Blue
M600
```

Another solution could be to set the filament name using some specific `M117` parameter.

Not changing anything in the code, until a decision is made.

3d-gussner:
@Panayiotis-git Thanks looking into this deeper. Yes please wait until we decide how to continue.

3d-gussner:
@Panayiotis-git Your PR isn't forgotten and we will look into it again in few weeks.

3d-gussner:
@Panayiotis-git Sorry for the late response.
We discussed this PR and the issue that PrusaLink uses M117 to update the LCD info screen. So we gonna need some other buffered message for this feature. Try to make it "open" so other features can use the 2nd buffered message as well. Yes it will cost at least 20 bytes (please limit it to one LCD line) but that is okay (approved by the team).

There are also few other Gcodes in RepRap wiki we could re-use for that
- https://reprap.org/wiki/G-code#M70:_Display_message
- https://reprap.org/wiki/G-code#M291:_Display_message_and_optionally_wait_for_response

If you find the time please implement something and I promise to review it asap and finally merge it.

Panayiotis-git:
Hey @3d-gussner, no worries!

I propose creating a command to store a generic text that can be utilized later by another command. For G-code, we could either define a new command or repurpose an existing one, such as M70 or M291. These commands are currently not implemented in the Prusa firmware, making them available for this purpose.

3d-gussner:
> Hey @3d-gussner, no worries!
> 
> I propose creating a command to store a generic text that can be utilized later by another command. For G-code, we could either define a new command or repurpose an existing one, such as M70 or M291. These commands are currently not implemented in the Prusa firmware, making them available for this purpose.

Thanks, I try to avoid to create new Gcodes and reuse existing one. For me the [RepRap Gcode Wiki ](https://reprap.org/wiki/G-code) covers most platforms and use it as a reference.

3d-gussner:
@Panayiotis-git Please contact me via e-mail.

3d-gussner:
@Panayiotis-git Any update on this PR using a different Gcode `M70` for the message?

Panayiotis-git:
@3d-gussner , I opened a new PR (https://github.com/prusa3d/Prusa-Firmware/pull/4731)  for the implementation of the feature using the M70 command.

Panayiotis-git:
Closed in favor of https://github.com/prusa3d/Prusa-Firmware/pull/4731

PR #4552, opened 2023-12-30
reprint: reduce duplicate code

gudnimg:
`restore_file_from_sd` function now supports any filename extension. This was not the case when the reprint feature was merged.

If a print was finished (to allow reprinting) then the EEPROM filename and extension should always be valid.

Change in memory:
Flash: -312 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -380,6 +380,7 @@ void refresh_print_state_in_ram();
 void refresh_saved_feedrate_multiplier_in_ram();
 void clear_print_state_in_ram();
 extern void stop_and_save_print_to_ram(float z_move, float e_move);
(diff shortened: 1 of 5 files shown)


gudnimg:
Pinging @jfestrada author of the reprint function.

3d-gussner (approved):
Tested on MK404.

- [X] SD Reprint
- [X] SD print with Power panic

3d-gussner:
@jfestrada Please have a look at this PR and let us know if you approve it or not.

jfestrada:
It's OK, and it make sense, this change will solve the issue with the filename extensions and will save some bytes and CPU usage will be lower.

PR #4553, opened 2023-12-30
Add Shutdown host to LCD Menu

sarusani:
My Raspberry Pi is powered by my printer power supply, but I was never a fan of powering it off while the OS wasn't shutdown properly. (Despite the fact that it never gave me any problems).

I'm also too lazy to go to my browser and shut the OS down before I power off the printer... So I added a "Shutdown host" menu entry. It's only showing if the printer is not active and a host is connected (M79).
I placed it second last in the menu order to make it less likely to get selected by accident.

It will send the action command action:shutdown to the host. This is supported by [OctoPrint](https://docs.octoprint.org/en/master/features/action_commands.html?highlight=shutdown) (But disabled by default!)

I don't think PrusaLink currently supports it. So it's probably not something you want to add right now. But I just wanted to offer the idea/code anyway. 😸


3d-gussner:
@sarusani Thanks for the PR and I think shutting down the RPi properly is a very good thing. Can't promise if this will be something we will consider and merge. Need to discuss that in the team and with PrusaLink.

3d-gussner:
@sarusani Please enclose this feature in `#define HOST_SHUTDOWN` in variants files and `#ifdef HOST_SHUTDOWN ... #endif //End HOST_SHUTDOWN` to make it an option for custom variant.
If you need some help please let me know. Did this so many times.

3d-gussner:
Also as you have the LCD message `Shutdown host` as `_I` you gonna need to update the /lang/Firmware.pot and lang/Firmware_<languges>.po files and translate the message.

3d-gussner:
Here a diff file you can use as an idea. I translated few languages     
[MK3_shutdownHostLCDEntry.diff.txt](https://github.com/prusa3d/Prusa-Firmware/files/13816132/MK3_shutdownHostLCDEntry.diff.txt)

sarusani:
@3d-gussner 
Thanks for the diff! I added the changes.

Some translations are still missing (hu, ro, sk) and the ones I added (cs, hr, pl) should be checked.

Ro3Deee:
> Some translations are still missing (hu, ro, sk) and the ones I added (cs, hr, pl) should be checked.

For Romanian: _Shutdown host_ is _Oprește gazda_

But I think that _Shutdown RPi_ (en) [_Oprește RPi_ (ro)] is a better message

3d-gussner:
@petrubecheru Thanks for the update. We should keep host as there are more options to run the host. PC, old notebooks, SBCs ...

Firmware/ultralcd.cpp
```diff
@@ -5331,10 +5339,16 @@ static void lcd_main_menu()
     if (MMU2::mmu2.Enabled()) {
         MENU_ITEM_SUBMENU_P(_i("Fail stats MMU"), lcd_menu_fails_stats_mmu);////MSG_MMU_FAIL_STATS c=18
     }
+
+#ifdef HOST_SHUTDOWN
+    if (!printer_active() && M79_timer_get_status()) {
+        MENU_ITEM_FUNCTION_P(_T(MSG_SHUTDOWN_HOST), lcd_shutdown_menu);////MSG_SHUTDOWN_HOST c=18
```
3d-gussner: Please remove `////MSG_SHUTDOWN_HOST c=18`as it this is already in `messages.cpp`

lang/po/Firmware_sk.po
```diff
@@ -2575,6 +2575,11 @@ msgstr "Nie je pripravené"
 msgid "Reprint"
 msgstr "Vytlačiť znova"
 
+#. MSG_SHUTDOWN_HOST c=18
+#: ../../Firmware/messages.cpp:108 ../../Firmware/ultralcd.cpp:5342
+msgid "Shutdown host"
+msgstr ""
```
3d-gussner: @ingbrzy @shatter136 Please provide a translation

lang/po/Firmware_hu.po
```diff
@@ -2592,6 +2592,11 @@ msgstr "Készen nem áll"
 msgid "Reprint"
 msgstr "Újranyomtatás"
 
+#. MSG_SHUTDOWN_HOST c=18
+#: ../../Firmware/messages.cpp:108 ../../Firmware/ultralcd.cpp:5342
+msgid "Shutdown host"
+msgstr ""
```
3d-gussner: @AttilaSVK @Hauzman Please provide a translation.

Firmware/ultralcd.cpp
sarusani: Done

gudnimg (approved):
Code changes look good to me 👍

AttilaSVK:
Hi! The correct translation should be
Gazdagép leállítása

but that’s too long, so either
Gazdagép leállítás

or
Gazdagép leállít.

are usable shorter alternatives.

Cheers,

Attila

lang/po/Firmware_sk.po
shatter136: "Vypnutie hostiteľa"

Ro3Deee:
>  Need to discuss that in the team and with PrusaLink.

prusalink has shutdown command:
https://github.com/prusa3d/Prusa-Link-Web/blob/68fb6d0aca7fd56bb1f5390e253c8df287a320b1/spec/openapi-legacy.yaml#L537

PR #4556, opened 2024-01-02
Restart USB timer during long moves

### LINKED ISSUE #4551: [BUG] Gcode sliced with PrusaSlicer 2.7.1 causes single beep when switching filament with MMU3
**Printer type** - MK3S+

**MMU upgrade** - MMU3
**MMU upgrade firmware version** - 3.0.1

**SD card or USB/Octoprint** - Octoprint

**Describe the bug**
  When a filament change occurs while printing a multimaterial print using the MMU3, the printer makes a single loud beep right before changing the filament. This doesn't happen when the same model is sliced with Prusaslicer 2.6.1, and only started happening when I downloaded 2.7.1. I've read online that this also happens with 2.7

  This may seem like an issue with PrusaSlicer, and not the firmware, but there is only a single difference in generated gcode for the two files, and no lines that indicate something that would cause the beep. In any case, there's also an issue open on the PrusaSlicer page as well: https://github.com/prusa3d/PrusaSlicer/issues/11786

  Ignoring progress messages, the only difference between the filament change sections of the gcode files is that in PrusaSlicer 2.6.1, this single line precedes the T[0-4] command:

  `G1 Z1.8 F720`

  The Z value varies, depending on the nozzle height. PrusaSlicer 2.7.1 does not add this extra nozzle hop line before the T[0-4] command. Seemingly the absence of this extra line causes the firmware to beep for a filament change, even though it should be silent and automatic.
  
**To Reproduce**
  Slice a multimaterial print for MMU3 on PrusaSlicer 2.7.1. The printer will beep on every automatic filament change. Slice on 2.6.1 to see the difference in behavior.
  
**Expected behavior**
  A loud beep should only happen in the case of a manual filament change, to get the user's attention. MMU changes should be silent
  
**G-code**
  The only gcode I've sliced with the new PrusaSlicer doesn't have it's first filament change until an hour in. If needed, I can make some quick test prints to show off the issue.
  
  Once my current print is done, I plan on manually sending gcode commands to try and replicate the issue.

gudnimg:
One idea to resolve #4551.

Keep in mind it only affects host prints. Printing with SD card will not reproduce the problem. 

G-code to reproduce the problem:

```gcode
G1 E15 F60 ; plan 15mm extrusion at 1mm/s
G4 S0 ; wait for all planned moves to finish
T1 ; MMU filament change
```

Make sure to set the Sound settings to _Loud_ to observe the buzzer beeping. Or simply observe that T1 will trigger manual unload. 

With this PR, manual unload is not triggered and there is no buzzer beep. 

Change in memory:
Flash: +48 bytes
SRAM: 0 bytes


gudnimg:
@3d-gussner We may see the same problem with M79 timer if for some reason someone changes `M79_TIMEOUT` to a much lower value like 10 seconds. But since it only affect UI for now, its low risk.

3d-gussner (approved):
Tested on MK404 and the `HO`st print stays active during long running Gcode. 
```
M117 Line 1
G1 E15 F60
M117 Line 2
G4 S0
M117 Line 3
T1
M117 Done
```

PR #4561, opened 2024-01-06
:construction: WIP: MMU: Allow reading/writing registers while handling a non-idle error

### LINKED ISSUE #4558: [BUG] MMU Idler cannot home message leads to unrecoverable state
**Printer type** - MK3S

**MMU upgrade** - MMU2S
**MMU upgrade firmware version** - 3.13.2

**SD card or USB/Octoprint**
 SD

**Describe the bug**
Upon receiving an “IDLER CANNOT HOME” error from the mmu there is an option to adjust the sensitivity. Going into this menu then back by selecting “Done” returns the printer to the main info screen with “OK” on the bottom line and the print does not resume. Both temperatures remain set but otherwise the printer does not appear to be actively doing anything. Pause / Stop print options do not appear in the menu and there’s no apparent way to resume or recover the print. I have experienced this more than once, with different gcodes. Not sure how to specifically reproduce as I haven’t tried to artificially trigger the idler error message.

gudnimg:
Draft for now... this uses a bit more Flash than I'd hoped.

Fixes #4558

Steps to reproduce a problem:

1. Send toolchange command to the printer e.g. T0
2. Hold the Idler in place until you get a "Idler cannot home" error.
3. Select 'Tune' button
4. Select 'Done' button (you need to click it 3 times to exit the menu)
5. Printer is in a 'frozen' state.

Change in memory:
Flash: +136 bytes
SRAM: +1 byte

### DIFF
--- a/Firmware/mmu2.cpp
+++ b/Firmware/mmu2.cpp
@@ -155,9 +155,11 @@ void MMU2::PowerOn() {
 bool MMU2::ReadRegister(uint8_t address) {
     if (!WaitForMMUReady())
         return false;
+
+    RequestMsg msg = RequestMsg(RequestMsgCodes::Read, address);
     do {
         logic.ReadRegister(address); // we may signal the accepted/rejected status of the response as return value of this function
-    } while (!manage_response(false, false));
+    } while (!manage_response(false, false, &msg));
 
     // Update cached value
     lastReadRegisterValue = logic.rsp.paramValue;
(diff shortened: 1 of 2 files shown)


gudnimg:
It would be nice to have this in 3.14.0 but we should **not** rush this.

DRracer (commented):
We need a unit test for that on MK4, if it goes ok, we can merge it.

gudnimg:
I don't see this being easy to unit test. Next weekend, I'll try find time to look into what the changes look like to disable the button during printing as a workaround. If I recall correctly... it won't work since the buttons are set at compile time but I may be remembering wrong.

I'd really like to mitigate this in 3.14 so some prints are saved.

3d-gussner:
@DRracer @gudnimg What is the status here?

gudnimg:
> @DRracer @gudnimg What is the status here?

Currently the PR is on hold until the changes are unit tested on the MK4 side. I’m not sure if I will be able to finish that any time soon :(

gudnimg:
Closing this as updating the MK4 unit tests will just take me too much time at the moment

PR #4562, opened 2024-01-09
PFW-1551: Fix Pause->Tune->Auto Home layer shift

3d-gussner:
Before the `Pause -> Tune -> Auto Home` repeatability causes a layer shift.

Before_Pause_Tune_Auto_home

With this PR the layer shift is gone.
After_Pause_Tune_Auto_home

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -238,8 +238,6 @@ static void lcd_sheet_menu();
 static void menu_action_sdfile(const char* filename);
 static void menu_action_sddirectory(const char* filename);
 
-static void lcd_rehome_xy();
-
 #define ENCODER_FEEDRATE_DEADZONE 10
 
 #define STATE_NA 255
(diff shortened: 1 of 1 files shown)


Firmware/ultralcd.cpp
```diff
@@ -5477,7 +5465,7 @@ static void lcd_tune_menu()
         MENU_ITEM_FUNCTION_P(_T(MSG_FILAMENTCHANGE), lcd_colorprint_change);
 #endif
     if (print_job_timer.isPaused()) {// Don't allow rehome if actively printing. Maaaaybe it could work to insert on the fly, seems too risky.
-        MENU_ITEM_FUNCTION_P(_T(MSG_AUTO_HOME), lcd_rehome_xy);
+        MENU_ITEM_GCODE_P(_T(MSG_AUTO_HOME),PSTR("G28 XY"));
```
gudnimg, line 5468: > Do home directly, G28 X Y resets MBL

If this is still true I think we should keep that comment to explain why the menu isnt shown when printing. 

Isnt this also bad if the print is paused?

gudnimg (changes requested):

Firmware/ultralcd.cpp
3d-gussner: In case the user has to move the extruder or bed when paused for some maintenance, the Auto home of XY is the only way to be able to resume the print.
This is only for the worst case scenario when you have/had to move the extruder or bed for some reason.

gudnimg (approved):

Firmware/ultralcd.cpp
gudnimg: Approved. Losing the MBL is not good but it wont ruin the print if its not the first layer.

PR #4569, opened 2024-01-14
Steel sheet alert before start the print 

jfestrada:
One of the usual problems with the MK3S and the different steel sheets is to forget to change the sheet selection resulting  on a PEI damage. So as I have been dealing with that issue I developed this alert. 
In my case I extended the **M115** G-Code in order to perform the check before to start the print and print a message on the LCD alerting the selected sheet, raise a beep and wait for the user input.
The check will work only if the check is enabled in the settings menu which use a eeprom variable, if the printer has more than one steel sheet registered and if is a local print.

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -240,6 +240,7 @@ extern bool axis_known_position[3];
 extern uint8_t fanSpeed; //!< Print fan speed, ranges from 0 to 255
 extern uint8_t newFanSpeed;
 extern float default_retraction;
+extern bool sheet_alert_enabled;
 
 void get_coordinates();
 void prepare_move(uint16_t start_segment_idx = 0);
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -146,6 +146,8 @@
 //===========================================================================
 //=============================public variables=============================
 //===========================================================================
+bool sheet_alert_enabled = true;
+
 #ifdef SDSUPPORT
 CardReader card;
 #endif
@@ -437,6 +439,10 @@ void serialprintlnPGM(const char *str) {
     }
   }
 #endif //!SDSUPPORT
+
+void SetSheetAlert() {
+	sheet_alert_enabled = eeprom_read_byte((uint8_t *)EEPROM_ED_SHEET_ALERT);
+}
 
 void setup_killpin()
 {
@@ -1608,6 +1614,7 @@ void setup()
 #endif //UVLO_SUPPORT
 
   fCheckModeInit();
+  SetSheetAlert();
   KEEPALIVE_STATE(NOT_BUSY);
 #ifdef WATCHDOG
   wdt_enable(WDTO_4S);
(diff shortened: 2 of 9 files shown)


3d-gussner:
@jfestrada This PR is quite resource hungry. Not sure how many users will use it.

Have you seen https://github.com/prusa3d/Prusa-Firmware/pull/3948 ?

Do I understand it correctly:
- You want to be informed on LCD which sheet is active and you have to confirm the selection
- This option can be en-/disabled in LCD Settings menu

When multiple sheets are calibrated the active sheet is shown on LCD Info screen when idle.

By adding two lines to Start gcode you can get same result:
``` 
M300 P50 S500 ; beep
M1 Check sheet correct ;pause until click
```
Check sheet correct

Example:
[Check_sheet_correct.3mf.txt](https://github.com/prusa3d/Prusa-Firmware/files/13938355/Check_sheet_correct.3mf.txt)
[Check_sheet_correct.gcode.txt](https://github.com/prusa3d/Prusa-Firmware/files/13938356/Check_sheet_correct.gcode.txt)

3d-gussner:
@jfestrada Does this workaround work for you?

jfestrada:
> @jfestrada This PR is quite resource hungry. Not sure how many users will use it.
> 
> Have you seen #3948 ?
> 
> Do I understand it correctly:
> 
>     * You want to be informed on LCD which sheet is active and you have to confirm the selection
> 
>     * This option can be en-/disabled in LCD Settings menu
> 
> 
> When multiple sheets are calibrated the active sheet is shown on LCD Info screen when idle.
> 
> By adding two lines to Start gcode you can get same result:
> 
> ```
> M300 P50 S500 ; beep
> M1 Check sheet correct ;pause until click
> ```
> 
> Check sheet correct
> 
> Example: [Check_sheet_correct.3mf.txt](https://github.com/prusa3d/Prusa-Firmware/files/13938355/Check_sheet_correct.3mf.txt) [Check_sheet_correct.gcode.txt](https://github.com/prusa3d/Prusa-Firmware/files/13938356/Check_sheet_correct.gcode.txt)

@3d-gussner you are right in your approach, in my point of view your solution is valid for "advanced users" and of course is efficient, but the reality is different, because the average user will not change the profile. And finally if the user is in a hurry maybe the user will forget about to change of the steel sheet selection and then is when the shinny new PEI steel sheet will have a new tattoo because the user was printing PETG on the textured sheet and now this user is going to print with PLA on the PEI sheet and the user forgot about to change the sheet selection. 
Maybe the situation descrived avobe will sound a little bit dramatic from my side, but it's happening with many people since years. Obviously this situation will not happen with the MK4 but for the MK3S+, MK3.5 and the Mini is something that will be there. 

This is the workflow implemented:

* (If there are more than 1 steel sheet saved) and (the check is enabled in the settings menu) and (is not a streamed print)
  * The printer will ask for confirmation and raise a beep
  *  If after 30 seconds there is no confirmation the print will not be started and the printer will abort the print job.
* The option can be enabled/disabled from the settings menu

By the way I perfectly understand your point of view, and in this FW the first thing is to think about the efficiency. But this modification will save some €€€ to the end user if their steel sheet is not damaged to be in a hurry.

3d-gussner:
@jfestrada Thanks for the comment and arguments for your PR. Please keep this PR open, for now I would say we won't merge it, but things can change.

Ro3Deee:
the problem of wrong steel sheet happened to me. I wa slucky because Z value was different between the two plates and the current one (the wrong one) needed a lower nozzle, so the PETG didn't stick and I catched the problem and saved the plate.

3d-gussner:
@jfestrada I am closing this PR as https://github.com/prusa3d/Prusa-Firmware/pull/4832 does improve the sheet types even more. Thanks again for the idea and PR.

jfestrada:
@3d-gussner you are welcome, for me was nice to contribute in someway.

PR #4582, opened 2024-01-27
PFW-1552 and PFW-1544: Fix issue with MMU unload when a print is stopped with a cold nozzle

gudnimg:
Steps to reproduce PFW-1552:

1. Start MMU print
2. Wait until it is printing and stall the hotend fan
3. Wait 30 minutes 
4. Stop print
5. Printer tries to unload the filament without heated nozzle

Steps to reproduce PFW-1544:

1. Start a print
6. Motor crash => until the screen shows message to abort print or resume
7. Wait for nozzle to cool down
8. Click 'Abort print'
9. Observe issue => Printer tries to unload filament while nozzle is cold

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -197,6 +197,7 @@ void kill(const char *full_screen_message = NULL);
 void finishAndDisableSteppers();
 
 void UnconditionalStop();                   // Stop heaters, motion and clear current print status
+void ConditionalStop();                     // Similar to UnconditionalStop, but doesn't disable heaters
 void ThermalStop(bool allow_pause = false); // Emergency stop used by overtemp functions which allows
                                             // recovery (with pause=true)
 bool IsStopped();                           // Returns true if the print has been stopped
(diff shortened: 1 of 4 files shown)


gudnimg:
Tests results:

**Stopping a print without pausing:**
* ✅ Target temperature stays at 215°C (PLA) while the MMU unloads the filament. Before this PR it would be set to 0°C while unloading.

**Stopping a paused print:**
* ✅ Target temperature stays at 215°C (PLA) while the MMU unloads the filament.

**Stopping a print after a fan error paused the print:**
* ✅ Target temperature stays at 215°C (PLA) while the MMU unloads the filament.

**Stopping a print after TMC2130 Crash detection:**
* ✅ Target temperature stays at 0°C (PLA) and the MMU shall not unload the filament. This happens because the print is not 'paused'. Before this PR the MMU would attempt to unload the filament while it was cold, potentially causing damage.

----

Items I haven't tested and the expected results
* Thermal Runway: It should stop the print and the MMU shall not unload the filament.
* Min Temp, Max Temp etc: It should stop the print and the MMU shall not unload the filament.

3d-gussner (approved):
Tested on MK404 and works well.

PR #4586, opened 2024-02-02
Fix 3x3 mbl condition

leptun:
Fix an edge case when running 3x3 mbl with region provided in gcode

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -2902,10 +2902,12 @@ static void gcode_G80()
         float x_pos = BED_X(ix);
         float y_pos = BED_Y(iy);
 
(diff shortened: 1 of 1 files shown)


3d-gussner (approved):
Tested with MK404

- [X] 3x3 only `G80` with 9 points
- [X] 3x3 using `G80 X109.557 Y89.5573 W30.8854 H30.8854 ; mesh bed leveling` with 9 points
- [X] 3x3 after factory reset
- [X] 7x7 only `G80` with 49 points
- [X] 7x7 using `G80 X109.557 Y89.5573 W30.8854 H30.8854 ; mesh bed leveling` with 17 points
- [X] 7x7 after factory reset

PR #4589, opened 2024-02-03
:construction: WIP: Improve Power panic

3d-gussner:
Add DEBUG_PRINTER_STATE
Fix LCD menus shown or hidden
- Resume for SD and host prints

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -5287,7 +5287,7 @@ static void lcd_main_menu()
     }
     if((printJobOngoing()
         || printingIsPaused()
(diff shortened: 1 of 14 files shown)


Firmware/ultralcd.cpp
```diff
@@ -5731,7 +5731,7 @@ void print_stop(bool interactive, bool unconditional_stop)
         ConditionalStop();
     }
 
-    if (card.mounted) {
+    if (card.isFileOpen()) {
```
gudnimg: I agree with this change. Could you create a separate PR for it?

PR #4590, opened 2024-02-04
PFW-1552 Fix a regression introduced in f022567

gudnimg:
Kudos to @3d-gussner for spotting the issue :) 

I have re-tested the items here: https://github.com/prusa3d/Prusa-Firmware/pull/4582#issuecomment-1913586179 and they are all behaving as expected.

With this PR, now the heaters are disabled correctly if the print is stopped before heating is completed.

Steps to reproduce issue:
1. Start a print
2. While heating, stop the print
3. Observe issue: Hotend/Bed continue to heat up to the target temperature
4. Expected result: The Hotend/Bed should stop heating

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -9776,6 +9776,15 @@ void ConditionalStop()
 {
     CRITICAL_SECTION_START;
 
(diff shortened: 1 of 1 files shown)


3d-gussner (approved):
Works the same a before.

PR #4591, opened 2024-02-04
Remove redundant ` cancel_heatup` assignment

gudnimg:
`cancel_heatup` is set to `false` inside `wait_for_heater()` there is no need to do it before the function call

See: https://github.com/prusa3d/Prusa-Firmware/blob/c999c2948e39bc52d20d558b1686f5ab1f1b4f32/Firmware/Marlin_main.cpp#L10009

Change in memory:
Flash: -4 bytes
SRAM: 0 bytes


3d-gussner:
I think we should review this in 3.14.1 and see if we can implement `M108`

3d-gussner (approved):
Looks good to me
Tested on MK404

PR #4592, opened 2024-02-04
Fix Stop print to send host action when host print is active

3d-gussner:
See https://github.com/prusa3d/Prusa-Firmware/pull/4589/commits/2b4f92ddc2b0e9d800b3ae5d8d8e1d6836385487#r1477083462


Firmware/ultralcd.cpp
```diff
@@ -5731,7 +5731,7 @@ void print_stop(bool interactive, bool unconditional_stop)
         ConditionalStop();
     }
 
-    if (card.mounted) {
+    if (card.isFileOpen) {
```
gudnimg: Missing `()` 😊

PR #4594, opened 2024-02-08
Update action as v3 is deprecated

3d-gussner:
See https://github.com/prusa3d/Prusa-Firmware/actions/runs/7829191740 and https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
 
     # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
     - name: Checkout ${{ github.event.pull_request.head.ref }}
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
       if: ${{ github.event.pull_request }}
       with:
         ref: ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -22,15 +22,15 @@ jobs:
 
     # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
     - name: Checkout ${{ github.event.pull_request.head.ref }}
-      uses: actions/checkout@v3
+      uses: actions/cache@v3.0.11
```
gudnimg: Looking at https://github.com/actions/checkout/blob/main/CHANGELOG.md

This should probably be: `actions/checkout@v4`

gudnimg: cache can also be updated to v4 [`https://github.com/actions/cache`](https://github.com/actions/cache/releases/tag/v4.0.0) :) But you need to do that at the relevant line.

3d-gussner:
@gudnimg Thanks, please review again

gudnimg (approved):

PR #4599, opened 2024-02-19
TMC Restore default XY currents after homing

leptun:
The currents were not properly restored at the end of homing, so the XY axes would remain with the lower homing current, resulting in more skipped steps on the Y axis and worse crash detection.

@3d-gussner please test.

- [x] Create a new branch MK3_3.13.3 based on v3.13.2
- [x] Create PR to MK3_3.13.3
- [x] Create PR with a cherry-pick to MK3 branch
  - [ ] Create tag v3.13.3
  - [ ] Ask for PrusaSlicer-settings PR

### DIFF
--- a/Firmware/tmc2130.cpp
+++ b/Firmware/tmc2130.cpp
@@ -439,6 +439,9 @@ void tmc2130_home_exit()
 		{
 			if (tmc2130_sg_homing_axes_mask & mask) {
 				tmc2130_XYZ_reg_init(axis);
+				currents[axis].setiRun(tmc2130_current_r[axis]);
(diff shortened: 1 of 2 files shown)


3d-gussner:
@leptun I can confirm the default running currents are restored

leptun:
With this latest commit the homing currents are only temporarely applied, so if the user sets custom currents, the custom currents are restored at homing end or after a crash is detected.

3d-gussner (approved):
Tested with MK404
Stock build
- [X] Default currents are restored after homing

Custom build with M910-M918 active
- [X] Changed current values with M912 are restored correctly after homing

PR #4600, opened 2024-02-19
Adjust Nozzle Change Menu For Quick Change Nozzles

### LINKED ISSUE #4595: [ENHANCEMENT]  Revo Firmware:  Do Not Heat Extruder When Initiating Nozzle Change
**Printer type** - MK3S/+ with REVO Hotend
**MMU Upgrade** - N/A

The MK3S/+ firmware includes an option under Settings/HW Setup/Nozzle Change that centers the extruder on the X axis, raises it to the maximum height, and heats the nozzle to 280C.  Revo hotends are designed to be swapped cold and with bare fingers, so heating the nozzle prior to a nozzle change can actually result in injury to the user.

  
The requested enhancement is to modify the Revo firmware behavior to set the hotend temperature to 0 during a nozzle change, rather than setting the extruder temperature to 280C.

sarusani:
Changes the nozzle change procedure for variants that have QUICK_NOZZLE_CHANGE defined.
Resolves https://github.com/prusa3d/Prusa-Firmware/issues/4595

If extruder temp is > 40°C
1) Show warning that extruder is hot.
2) Turn on fan, home, move to top center, wait for extruder to cooldown to 40°C.
3) Show msg that nozzle is ready to be changed, if confirmed turn off fan.
4) Switch to the nozzle diameter selection after nozzle change has been confirmed. (This change affects all variants)

If extruder temp is < 40°C
1) Home and move to top center
2) Show msg that nozzle is ready to be changed
4) Switch to the nozzle diameter selection after nozzle change has been confirmed. (This change affects all variants)

### DIFF
--- a/Firmware/messages.cpp
+++ b/Firmware/messages.cpp
@@ -192,7 +192,12 @@ extern const char MSG_TM_ACK_ERROR[] PROGMEM_I1 = ISTR("Clear TM error");////MSG
 extern const char MSG_LOAD_ALL[] PROGMEM_I1 = ISTR("Load All"); ////MSG_LOAD_ALL c=18
 extern const char MSG_NOZZLE_CNG_MENU [] PROGMEM_I1 = ISTR("Nozzle change");////MSG_NOZZLE_CNG_MENU c=18
 extern const char MSG_NOZZLE_CNG_READ_HELP [] PROGMEM_I1 = ISTR("For a Nozzle change please read\nprusa.io/nozzle-mk3s");////MSG_NOZZLE_CNG_READ_HELP c=20 r=4
+#ifdef QUICK_NOZZLE_CHANGE
+extern const char MSG_NOZZLE_CNG_COOLDOWN [] PROGMEM_I1 = ISTR("Nozzle is hot! Wait for cooldown.");////MSG_NOZZLE_CNG_COOLDOWN c=20 r=3
+extern const char MSG_NOZZLE_CNG_CHANGED [] PROGMEM_I1 = ISTR("Nozzle changed?");////MSG_NOZZLE_CNG_CHANGED_QUICK c=20 r=3
+#else
 extern const char MSG_NOZZLE_CNG_CHANGED [] PROGMEM_I1 = ISTR("Hotend at 280C! Nozzle changed and tightened to specs?");////MSG_NOZZLE_CNG_CHANGED c=20 r=6
+#endif //QUICK_NOZZLE_CHANGE
 extern const char MSG_REPRINT [] PROGMEM_I1 = ISTR("Reprint"); ////MSG_REPRINT c=18
 extern const char MSG_FILE_CNT [] PROGMEM_I1 = ISTR("Some files will not be sorted. Max. No. of files in 1 folder for sorting is 100."); ////MSG_FILE_CNT c=20 r=6
 extern const char MSG_CHANGED_MOTHERBOARD [] PROGMEM_I1 = ISTR("Warning: motherboard type changed."); ////MSG_CHANGED_MOTHERBOARD c=20 r=4
--- a/Firmware/messages.h
+++ b/Firmware/messages.h
@@ -194,6 +194,9 @@ extern const char MSG_LOAD_ALL[];
 extern const char MSG_NOZZLE_CNG_MENU [];
 extern const char MSG_NOZZLE_CNG_READ_HELP [];
 extern const char MSG_NOZZLE_CNG_CHANGED [];
+#ifdef QUICK_NOZZLE_CHANGE
+extern const char MSG_NOZZLE_CNG_COOLDOWN [];
+#endif //QUICK_NOZZLE_CHANGE
 extern const char MSG_REPRINT [];
 extern const char MSG_FILE_CNT [];
 extern const char MSG_CHANGED_MOTHERBOARD [];
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -972,15 +972,18 @@ void lcd_commands()
 	{
         if (!blocks_queued() && cmd_buffer_empty() && !saved_printing)
         {
+#ifndef QUICK_NOZZLE_CHANGE //thermal model can be ignored if a quickchange nozzle is in use, no heatup necessary
 #ifdef THERMAL_MODEL
             static bool was_enabled;
 #endif //THERMAL_MODEL
+#endif //QUICK_NOZZLE_CHANGE
             switch(lcd_commands_step)
             {
             case 0:
                 lcd_commands_step = 3;
                 break;
             case 3:
+#ifndef QUICK_NOZZLE_CHANGE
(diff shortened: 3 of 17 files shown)


3d-gussner:
@sarusani Thanks for contributing another PR very appreciated. I just wonder why the PR costs 44 additional bytes.

3d-gussner:
> I didn't include a warning that the nozzle still could be hot. Can be added if preferred.

Very good point. Maybe a good idea would be to check that the REVO hotend is "cold" to touch before showing the LCD message to change the nozzle.

1. LCD Settings -> HW Setup -> Nozzle Change
2. G28W
3. Turn off hotend 
4. Raise to change position
5. Wait until hotend <= 30°C
  b. Turn on Part cooling fan to boost cool down
6. Turn off Part cooling fan
7. Show LCD message to confirm

3d-gussner:
@sarusani As you are working on this. Do you think you could have a look at a request I have seen somewhere?
User asked to go to the nozzle diameter menu to check/change nozzle size after a nozzle change.
This would be useful for REVO and E3D v6 nozzles.

sarusani:
To be honest I was a wondering about the 44 bytes myself... The only thing that changed for non Revo builds is that I moved "lcd_commands_step = 2" a few lines up. Maybe @gudnimg or @leptun have some more insight?

(Edit: Moving the line back down really makes the difference... I update the commit. Why this makes a difference is not clear to me at all 🤷 )

I will add the additional changes as a separate commit. So you'll be able to cherry pick the solution if needed.

Firmware/ultralcd.cpp
```diff
@@ -993,27 +995,36 @@ void lcd_commands()
                 was_enabled = thermal_model_enabled();
                 thermal_model_set_enabled(false);
 #endif //THERMAL_MODEL
-                lcd_commands_step = 2;
 #else //nozzle change without heating
-                enquecommand_P(G28W);
-                enquecommand_P(PSTR("G1 X125 Z200 F1000"));
-                enquecommand_P(PSTR("M109 S0"));
-                lcd_commands_step = 1;
+                if((int)degHotend(active_extruder)>40) { //check temp
```
3d-gussner: I think 40°C is too hot to touch.

Firmware/ultralcd.cpp
```diff
@@ -992,21 +995,36 @@ void lcd_commands()
...
+#else //nozzle change without heating
+                if((int)degHotend(active_extruder)>40) { //check temp
+                    lcd_update_enabled = false; //hack to avoid lcd_update recursion.
+                    lcd_show_fullscreen_message_and_wait_P(_T(MSG_NOZZLE_CNG_COOLDOWN));
+                    lcd_update_enabled = true;
+                    lcd_draw_update = 2; //force lcd clear and update after the stack unwinds.
+                    enquecommand_P(PSTR("M106")); //turn on fan
+                    enquecommand_P(G28W); //home
+                    enquecommand_P(PSTR("G1 X125 Z200 F1000")); //move to top center
+                    enquecommand_P(PSTR("M109 S40")); //set extruder temp and wait
```
3d-gussner: See above

Firmware/ultralcd.cpp
sarusani: Body temp is 37°. 40° is safe and saves time.

3d-gussner:
@sarusani Thanks a lot looks good. I made few changes you may look at:
- Disable heater instead of setting it to 40°C
- Shorter messages and translations
- while loop with lcd yes no message which times out when temp is below 40°C
Here my diff file
[MK3_Change_Nozzle.diff.txt](https://github.com/prusa3d/Prusa-Firmware/files/14346837/MK3_Change_Nozzle.diff.txt)

sarusani:
@3d-gussner 
Looks good! I added your changes.

The only thing I think is a bit confusing is the yes/no option for the cooldown message. The no option is great, but I'm not sure if the users would think they're stuck when pressing yes and nothing happens. Since it's not totally clear that you are already in the cooldown phase and how long it will take.

I think a single option "ignore" would be much clearer, but that needs additional work because currently the messages need at least two options... Or maybe instead of yes/no we could use wait/ignore?

(I made a version with the [one option menu](https://github.com/prusa3d/Prusa-Firmware/commit/d6e4c6bcb3fa25e6bdacd8c45e45d44601146e7d), the changes on the existing menu use 24 bytes of additional space)

3d-gussner:
@sarusani I have seen you Ignore option, only thing I less like about it is that buy accidental double click you ignore it.
Hope to give you more feedback next week.
As always very appreciated your engagement and PRs.

Firmware/messages.cpp
```diff
@@ -192,7 +192,12 @@ extern const char MSG_TM_ACK_ERROR[] PROGMEM_I1 = ISTR("Clear TM error");////MSG
 extern const char MSG_LOAD_ALL[] PROGMEM_I1 = ISTR("Load All"); ////MSG_LOAD_ALL c=18
 extern const char MSG_NOZZLE_CNG_MENU [] PROGMEM_I1 = ISTR("Nozzle change");////MSG_NOZZLE_CNG_MENU c=18
 extern const char MSG_NOZZLE_CNG_READ_HELP [] PROGMEM_I1 = ISTR("For a Nozzle change please read\nprusa.io/nozzle-mk3s");////MSG_NOZZLE_CNG_READ_HELP c=20 r=4
+#ifndef QUICK_NOZZLE_CHANGE
```
3d-gussner: Please use same `#ifdef` or `#ifndef` in `messages.cpp/.h`

Firmware/messages.h
```diff
@@ -194,6 +194,9 @@ extern const char MSG_LOAD_ALL[];
 extern const char MSG_NOZZLE_CNG_MENU [];
 extern const char MSG_NOZZLE_CNG_READ_HELP [];
 extern const char MSG_NOZZLE_CNG_CHANGED [];
+#ifdef QUICK_NOZZLE_CHANGE
```
3d-gussner, line 197: See above

3d-gussner (changes requested):

3d-gussner:
@sarusani Please rebase the PR and solve the conflicts.

Firmware/messages.cpp
sarusani: Now using `#ifdef` in both files.

3d-gussner (approved):
Tested on MK404

- [X] MK3S works as before but switches to Nozzle diameter menu

MK3S_Nozzle_change_1

MK3S_Nozzle_change_2

MK3S_Nozzle_change_3

- [X] REVO nozzle change doesn't heatup and also switches to Nozzle diameter menu

REVO_Nozzle_change_1

REVO_Nozzle_change_2

REVO_Nozzle_change_3

REVO_Nozzle_change_4

REVO_Nozzle_change_4a


REVO_Nozzle_change_5

3d-gussner:
@gudnimg Please have a look as well.

PR #4608, opened 2024-02-23
Xyzcal use stallguard during z cal

leptun:
Properly enable the stallguard during the Z calibration part of the XYZ calibration.
Also added some debug code for the currents which can be enabled

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -262,6 +262,7 @@ uint16_t restore_interrupted_gcode();
 float __attribute__((noinline)) get_feedrate_mm_s(const float feedrate_mm_min);
 
 #ifdef TMC2130
+void check_Z_crash(void);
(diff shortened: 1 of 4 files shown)


gudnimg:
I looked at the code and see the following things which need to be tested:

1. `M45 Z` (to set `onlyZ = true`)
2. XYZ calibration => Specifically the Z-axis lift just before the user is asked to place the steel sheet on again
3. Calibration Wizard calls `gcode_M45(true, 0)`. I would think `M45 Z` should cover this, but maybe @3d-gussner can double check

I'll try find some time to test at least 1) and 2) and report here later.

gudnimg (approved):
Code looks good to me:

I tested the following on MK3S+ hardware:

* ✅ Z calibration from LCD
* ✅ XYZ calibration from LCD

PR #4613, opened 2024-02-24
optimisation: initialise `axis_ustep_resolution` in EEPROM

gudnimg:
A minor optimisation I found.

When the `cs` variable is loaded into RAM, `cs.axis_ustep_resolution` will always have some value and none of the `0xFF` values.

Change in memory:
Flash: -40 bytes
SRAM: 0 bytes


gudnimg:
For some reason omitting `sudo apt-get update` started being a problem: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/customizing-github-hosted-runners#installing-software-on-ubuntu-runners

> Note: Always run `sudo apt-get update` before installing a package. In case the `apt` index is stale, this command fetches and re-indexes any available packages, which helps prevent package installation failures.

I created a separate PR to resolve this since the change has nothing to do with firmware code.

3d-gussner (approved):
Looks good to me.
Tested with MK404

PR #4617, opened 2024-02-26
Cherrypick xyzcal use stallguard during z cal

leptun:
Cherrypick of #4608

### DIFF
--- a/Firmware/Marlin.h
+++ b/Firmware/Marlin.h
@@ -253,6 +253,7 @@ uint16_t restore_interrupted_gcode();
 float __attribute__((noinline)) get_feedrate_mm_s(const float feedrate_mm_min);
 
 #ifdef TMC2130
+void check_Z_crash(void);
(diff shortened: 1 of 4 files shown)


3d-gussner (approved):
Tested in MK404 and real hardware

1. logging via USB
2. Send M45 Z to start Z calibration
3. After LCD screen "Please clean the nozzle..." click the X gantry moves down
4. Triggering the PINDA with some metal stops it really fast. :heavy_check_mark: 
   a. This definitely prevents nozzle dives when the PINDA is working and no big debris is on the nozzle
5. Repeat steps 2+3
6. Hold both Z motor lead screws with hands is easy to trigger stall guard and it stops :heavy_check_mark: 
7.  Repeat steps 2+3
8. Put a cube under the nozzle it stops okay-ish
    a. it bends the bend a bit
    b. BUT NOT as hard as without it. The previous test was quite valiant and bend the bed hard and the x gantry got really skewed

PR #4618, opened 2024-02-26
:construction: WIP: Relieve pressure after a nozzle crash

leptun:


3d-gussner:
@leptun Is this still in progress and draft?

leptun:
I didn't have time to look into this. The part that is still left to be done is checking other places in code where we relieve pressure after a nozzle crash and use nozzle_crash_handler() so that code is deduplicated

PR #4619, opened 2024-02-27
Add DEBUG_EEPROM_CHANGES

3d-gussner:
Very useful to debug multiple issue.
Thanks @DRracer for the idea https://github.com/prusa3d/Prusa-Firmware/pull/2842 and help to get this working.

### DIFF
--- a/Firmware/ConfigurationStore.cpp
+++ b/Firmware/ConfigurationStore.cpp
@@ -181,7 +181,7 @@ static const M500_conf default_conf PROGMEM =
 void Config_StoreSettings()
 {
   strcpy_P(cs.version, default_conf.version);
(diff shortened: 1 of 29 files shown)


DRracer (approved):
:thinking: where are the +2 bytes coming from? It looks like something didn't get optimized correctly... may be adding some `__attribute__((always_inline))` might push the compiler the right way.

Anyway, I'm ok with this PR as it is now.

3d-gussner:
> 🤔 where are the +2 bytes coming from? It looks like something didn't get optimized correctly... may be adding some `__attribute__((always_inline))` might push the compiler the right way.
> 
> Anyway, I'm ok with this PR as it is now.

Maybe @gudnimg can use some :magic_wand:

PR #4635, opened 2024-03-13
Power Panic and LCD menu

3d-gussner:
Improved Power Panic
- Added Debug for printer states
- SD power panic
  - Recovers from multiple power panic and even resets (as long it is in recovering)
- Host power panic
- LCD menu
  - During recovering all menus are hidden except
    - Stop print (to be able to exit)
    - Statistics
    - Fail Stats
    - Support   
  - Reprint hidden
    - Host reprint only shown when host sends `M79` to ensure that the host is ready to reprint
    - SD Reprint only shown when previous print was finished successfully.
      - ToDo for next release: Don't show Reprint when SD card has been changed
  - Resume hidden
    - When resuming
    - With host prints until host sends `M79`
  - `Autoload filament` menu only shown with active FSensor and NO filament detected
  - `Load filament` menu only shown with active FSensor and NO filament detected OR FSensor disabled
  - `Unload filament` menu only shown with active FSensor and filament detected OR FSensor disabled
 
 
This Firmware doesn't recover correctly with Prusa Link 0.8.0-alpha4 at this moment.
I was able to recover host prints with a modified OctoPrint + two Plugins and additional code (which isn't in this PR as this code is a workaround of existing OctoPrint+Plugins)

Update 27-03-2024

Added `EEPROM_UVLO_Z_LIFTED` to prevent multiple Z lifts in case of a power panic while the print is paused.
A host print recovery it doesn't wait for hotend temp to be stable as it will go into pause state anyway and the hotend will be disabled while paused anyway. 

Tested with MK404

- [X] SD print + power panic
  - For some unknown reason it doesn't recover correctly SD print + power panic + reset while recovering, while a double power panic does recover correctly even with a reset.   
- [X] SD print + power panic and while lifting Z another power panic
- [X] SD print + power panic and while lifting Z another power panic + reset while recovering
- [X] SD print + pause + power panic
- [X] SD print + pause + power panic and while lifting Z another power panic
- [X] SD print + pause + power panic and while lifting Z another power panic + reset while recovering
- [X] Host print same as above for SD

Difference between SD and Host print power panic:
- SD recovers automatically when bed temp isn't lower than 5°C to target bed temperature
- Host recovers the print settings and goes to pause state.
  - Then it waits for the host sending "M79" before showing Resume to ensure that the host is "there" to receive any messages from the printer.  
   
The print can't be recovered if a reset happens at the last step of the recovery as at this moment the power panic recovery is "completed" and the printer is back in normal mode.

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -312,7 +312,7 @@ static bool chdkActive = false;
 //! @{
 bool saved_printing = false; //!< Print is paused and saved in RAM
 uint32_t saved_sdpos = 0; //!< SD card position, or line number in case of USB printing
-uint8_t saved_printing_type = PowerPanic::PRINT_TYPE_SD;
+uint8_t saved_printing_type = PowerPanic::PRINT_TYPE_NONE;
 float saved_pos[NUM_AXIS] = { X_COORD_INVALID, 0, 0, 0 };
 uint16_t saved_feedrate2 = 0; //!< Default feedrate (truncated from float)
 static int saved_feedmultiply2 = 0;
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -1638,31 +1638,30 @@ void setup()
 #ifdef UVLO_SUPPORT
   if (printer_recovering()) { //previous print was terminated by UVLO
       manage_heater(); // Update temperatures 
-#ifdef DEBUG_UVLO_AUTOMATIC_RECOVER 
-		printf_P(_N("Power panic detected!\nCurrent bed temp:%d\nSaved bed temp:%d\n"), (int)degBed(), eeprom_read_byte((uint8_t*)EEPROM_UVLO_TARGET_BED));
-#endif
-     uvlo_auto_recovery_ready = (degBed() > ( (float)eeprom_read_byte((uint8_t*)EEPROM_UVLO_TARGET_BED) - AUTOMATIC_UVLO_BED_TEMP_OFFSET));
-     if (uvlo_auto_recovery_ready){
-          #ifdef DEBUG_UVLO_AUTOMATIC_RECOVER 
+      //Restore printing type
+      saved_printing_type = eeprom_read_byte((uint8_t*)EEPROM_UVLO_PRINT_TYPE);
+#ifdef DEBUG_UVLO_AUTOMATIC_RECOVER
+      printf_P(_N("Power panic detected!\nCurrent bed temp:%d\nSaved bed temp:%d\n"), (int)degBed(), eeprom_read_byte((uint8_t*)EEPROM_UVLO_TARGET_BED));
+      uvlo_auto_recovery_ready = (degBed() > ( (float)eeprom_read_byte((uint8_t*)EEPROM_UVLO_TARGET_BED) - AUTOMATIC_UVLO_BED_TEMP_OFFSET));
+      if (uvlo_auto_recovery_ready){
         puts_P(_N("Automatic recovery!")); 
-          #endif
-         recover_print(1); 
-      } 
-      else{ 
-          #ifdef DEBUG_UVLO_AUTOMATIC_RECOVER 
+        recover_print(1);
+      } else {
+#endif //DEBUG_UVLO_AUTOMATIC_RECOVER
         puts_P(_N("Normal recovery!")); 
-          #endif
-          if (eeprom_read_byte((uint8_t*)EEPROM_UVLO_PRINT_TYPE) == PowerPanic::PRINT_TYPE_HOST) {
+        if (saved_printing_type == PowerPanic::PRINT_TYPE_HOST) {
+          recover_print(0);
+        } else {
+          const uint8_t btn = lcd_show_fullscreen_message_yes_no_and_wait_P(_T(MSG_RECOVER_PRINT), false);
+          if ( btn == LCD_LEFT_BUTTON_CHOICE) {
             recover_print(0);
-          } else {
-              const uint8_t btn = lcd_show_fullscreen_message_yes_no_and_wait_P(_T(MSG_RECOVER_PRINT), false);
-              if ( btn == LCD_LEFT_BUTTON_CHOICE) {
-                  recover_print(0);
-            } else { // LCD_MIDDLE_BUTTON_CHOICE
-                  cancel_saved_printing();
-            }
+          } else { // LCD_MIDDLE_BUTTON_CHOICE
+            cancel_saved_printing();
           }
+        }
+#ifdef DEBUG_UVLO_AUTOMATIC_RECOVER
       }
+#endif //DEBUG_UVLO_AUTOMATIC_RECOVER
   }
 
   // Only arm the uvlo interrupt _after_ a recovering print has been initialized and
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -4171,62 +4171,52 @@ void process_commands()
         if (farm_prusa_code_seen()) {}
         else if(code_seen_P(PSTR("FANPINTST"))) {
             gcode_PRUSA_BadRAMBoFanTest();
-        }
-        else if (code_seen_P(PSTR("FAN"))) { // PRUSA FAN
+        } else if (code_seen_P(PSTR("FAN"))) { // PRUSA FAN
             printf_P(_N("E0:%d RPM\nPRN0:%d RPM\n"), 60*fan_speed[0], 60*fan_speed[1]);
-        }
-        else if (code_seen_P(PSTR("uvlo"))) // PRUSA uvlo
-        {
-            if (eeprom_read_byte((uint8_t*)EEPROM_UVLO_PRINT_TYPE) == PowerPanic::PRINT_TYPE_SD)
-            {
+        } else if (code_seen_P(PSTR("uvlo"))) { // PRUSA uvlo
+            if (eeprom_read_byte((uint8_t*)EEPROM_UVLO_PRINT_TYPE) == PowerPanic::PRINT_TYPE_SD) {
                 // M24 - Start SD print
                 enquecommand_P(MSG_M24);
 
                 // Print is recovered, clear the recovery flag
                 eeprom_update_byte_notify((uint8_t*)EEPROM_UVLO, PowerPanic::NO_PENDING_RECOVERY);
-            }
-            else if (eeprom_read_byte((uint8_t*)EEPROM_UVLO_PRINT_TYPE) == PowerPanic::PRINT_TYPE_HOST)
-            {
+            } else if (eeprom_read_byte((uint8_t*)EEPROM_UVLO_PRINT_TYPE) == PowerPanic::PRINT_TYPE_HOST) {
                 // For Host prints we need to start the timer so that the pause has any effect
                 // this will allow g-codes to be processed while in the paused state
                 // For SD prints, M24 starts the timer
                 print_job_timer.start();
+                usb_timer.start();
 
                 // Park the extruder to the side and don't resume the print
                 // we must assume that the host as not fully booted up at this point
                 lcd_pause_print();
             }
-        }
-		else if (code_seen_P(PSTR("MMURES"))) // PRUSA MMURES
-		{
-			MMU2::mmu2.Reset(MMU2::MMU2::Software);
-		}
-		else if (code_seen_P(PSTR("RESET"))) { // PRUSA RESET
+        } else if (code_seen_P(PSTR("MMURES"))) { // PRUSA MMURES
+            MMU2::mmu2.Reset(MMU2::MMU2::Software);
+        } else if (code_seen_P(PSTR("RESET"))) { // PRUSA RESET
 #if defined(XFLASH) && defined(BOOTAPP)
             boot_app_magic = 0;
 #endif //defined(XFLASH) && defined(BOOTAPP)
             softReset();
-    }
-    else if (code_seen_P(PSTR("SN"))) { // PRUSA SN
-        char SN[20];
-        eeprom_read_block(SN, (uint8_t*)EEPROM_PRUSA_SN, 20);
-        if (SN[19])
-            puts_P(PSTR("SN invalid"));
-        else
-            puts(SN);
-    }
-    else if(code_seen_P(PSTR("Fir"))){ // PRUSA Fir
+        } else if (code_seen_P(PSTR("SN"))) { // PRUSA SN
+            char SN[20];
+            eeprom_read_block(SN, (uint8_t*)EEPROM_PRUSA_SN, 20);
+            if (SN[19])
+                puts_P(PSTR("SN invalid"));
+            else
+                puts(SN);
+        } else if(code_seen_P(PSTR("Fir"))){ // PRUSA Fir
 
-      SERIAL_PROTOCOLLNPGM(FW_VERSION_FULL);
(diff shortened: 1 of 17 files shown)


Firmware/ultralcd.cpp
```diff
@@ -5228,6 +5228,10 @@ static void lcd_main_menu()
         MENU_ITEM_FUNCTION_P(PSTR("tst - Restore"), lcd_menu_test_restore);
 #endif //RESUME_DEBUG
 
+#ifdef DEBUG_PRINTER_STATES
+    debug_printer_states();
```
gudnimg, line 5232: I have a bad feeling about this. `lcd_main_menu()` is called very frequently and dumping this much stuff on the serial stream will likely do more harm than good. Does the printer actually keep up with this much serial prints?

gudnimg: Another point is, what happens when the printer is not rendering the main menu? Then these debug prints are ignored.

Firmware/Marlin_main.cpp
```diff
@@ -4117,62 +4173,52 @@ void process_commands()
...
-        else if (code_seen_P(PSTR("FAN"))) { // PRUSA FAN
+        } else if (code_seen_P(PSTR("FAN"))) { // PRUSA FAN
             printf_P(_N("E0:%d RPM\nPRN0:%d RPM\n"), 60*fan_speed[0], 60*fan_speed[1]);
-        }
-        else if (code_seen_P(PSTR("uvlo"))) // PRUSA uvlo
-        {
-            if (eeprom_read_byte((uint8_t*)EEPROM_UVLO_PRINT_TYPE) == PowerPanic::PRINT_TYPE_SD)
-            {
+        } else if (code_seen_P(PSTR("uvlo"))) { // PRUSA uvlo
+            if (eeprom_read_byte((uint8_t*)EEPROM_UVLO_PRINT_TYPE) == PowerPanic::PRINT_TYPE_SD) {
```
gudnimg, line 4180: Can `saved_printing_type` be used here?

`if (saved_printing_type == PowerPanic::PRINT_TYPE_SD)`

Firmware/Marlin_main.cpp
```diff
@@ -4117,62 +4173,52 @@ void process_commands()
...
                 // M24 - Start SD print
                 enquecommand_P(MSG_M24);
 
                 // Print is recovered, clear the recovery flag
                 eeprom_update_byte_notify((uint8_t*)EEPROM_UVLO, PowerPanic::NO_PENDING_RECOVERY);
-            }
-            else if (eeprom_read_byte((uint8_t*)EEPROM_UVLO_PRINT_TYPE) == PowerPanic::PRINT_TYPE_HOST)
-            {
+            } else if (eeprom_read_byte((uint8_t*)EEPROM_UVLO_PRINT_TYPE) == PowerPanic::PRINT_TYPE_HOST) {
```
gudnimg, line 4187: Can `saved_printing_type` be used here?

`if (saved_printing_type == PowerPanic::PRINT_TYPE_HOST)`

Firmware/Marlin_main.cpp
```diff
@@ -11024,7 +11069,7 @@ void restore_print_from_ram_and_continue(float e_move)
 	set_destination_to_current();
 
     restore_print_file_state();
-
+    eeprom_update_byte((uint8_t*)EEPROM_UVLO, PowerPanic::NO_PENDING_RECOVERY);
 	lcd_setstatuspgm(MSG_WELCOME);
     saved_printing_type = PowerPanic::PRINT_TYPE_NONE;
 	saved_printing = false;
```
gudnimg: Optimisation idea: Can we use `cancel_saved_printing()` here?

So change:

```c
    eeprom_update_byte((uint8_t*)EEPROM_UVLO, PowerPanic::NO_PENDING_RECOVERY);
	lcd_setstatuspgm(MSG_WELCOME);
    saved_printing_type = PowerPanic::PRINT_TYPE_NONE;
	saved_printing = false;
```

to this:

```c
    cancel_saved_printing();
	lcd_setstatuspgm(MSG_WELCOME);
```

Firmware/Marlin_main.cpp
```diff
@@ -1581,32 +1636,33 @@ void setup()
...
       manage_heater(); // Update temperatures 
-#ifdef DEBUG_UVLO_AUTOMATIC_RECOVER 
-		printf_P(_N("Power panic detected!\nCurrent bed temp:%d\nSaved bed temp:%d\n"), (int)degBed(), eeprom_read_byte((uint8_t*)EEPROM_UVLO_TARGET_BED));
-#endif
-     uvlo_auto_recovery_ready = (degBed() > ( (float)eeprom_read_byte((uint8_t*)EEPROM_UVLO_TARGET_BED) - AUTOMATIC_UVLO_BED_TEMP_OFFSET));
-     if (uvlo_auto_recovery_ready){
-          #ifdef DEBUG_UVLO_AUTOMATIC_RECOVER 
+      //Restore printing type
+      saved_printing_type = eeprom_read_byte((uint8_t*)EEPROM_UVLO_PRINT_TYPE);
+#ifdef DEBUG_UVLO_AUTOMATIC_RECOVER
```
gudnimg, line 1644: Personally I really don't like this `DEBUG_UVLO_AUTOMATIC_RECOVER` code. It makes the code very unreadable. We can probably move all the serial print messages into `recover_print()` function.

Just an observation/comment. Not a blocker.

Firmware/power_panic.cpp
```diff
@@ -315,11 +315,17 @@ void recover_print(uint8_t automatic) {
...
     bool mbl_was_active = recover_machine_state_after_power_panic();
 
-    // Lift the print head 25mm, first to avoid collisions with oozed material with the print,
-    // and second also so one may remove the excess priming material.
+    // Undo PP Z Lift by setting current Z pos to + Z_PAUSE_LIFT
+    // After a reboot the printer doesn't know the Z height and we have to set its previous value
+    if(eeprom_read_byte((uint8_t*)EEPROM_UVLO) == PowerPanic::PENDING_RECOVERY_RETRY) {
+        current_position[Z_AXIS] += Z_PAUSE_LIFT;
+    }
+
```
gudnimg: I'm not sure about this change. `current_position` is likely not modified unless the g-code was executed.

`enquecommandf_P(PSTR("G1 Z%.3f F800"), current_position[Z_AXIS] + Z_PAUSE_LIFT);`

Doesn't the if statement below make sure this lift isn't executed again?:

`if(eeprom_read_byte((uint8_t*)EEPROM_UVLO) == PowerPanic::PENDING_RECOVERY)`

There shouldn't be another lift when the flag is set to `PowerPanic::PENDING_RECOVERY_RETRY`. Am I missing something?

gudnimg (changes requested):
See my comments

Firmware/ultralcd.cpp
```diff
@@ -5297,81 +5314,96 @@ static void lcd_main_menu()
 #endif
 
     // only allow starting SD print if hardware errors (temperature or fan) are cleared
-    if(!get_temp_error()
-            && !printer_active()
+    if (!printer_recovering()
+        && !printer_active()
+        && !get_temp_error()
+        && !printer_active()
```
gudnimg: Redundant call to `!printer_active()` here

Firmware/ultralcd.cpp
```diff
@@ -5297,81 +5314,96 @@ static void lcd_main_menu()
...
-    }
-
-    if(!printer_active() && !farm_mode) {
-        const int8_t sheet = eeprom_read_byte(&(EEPROM_Sheets_base->active_sheet));
-        const int8_t nextSheet = eeprom_next_initialized_sheet(sheet);
-        if ((nextSheet >= 0) && (sheet != nextSheet)) { // show menu only if we have 2 or more sheets initialized
-            MENU_ITEM_FUNCTION_E(EEPROM_Sheets_base->s[sheet], eeprom_switch_to_next_sheet);
-        }
+        if(!printer_active()
+            && !farm_mode) {
```
gudnimg: Why is this split into two lines? Can easily fit one line and is more readable IMO

Firmware/ultralcd.cpp
```diff
@@ -5297,81 +5314,96 @@ static void lcd_main_menu()
...
+            const int8_t sheet = eeprom_read_byte(&(EEPROM_Sheets_base->active_sheet));
+            const int8_t nextSheet = eeprom_next_initialized_sheet(sheet);
+            if ((nextSheet >= 0)
+                && (sheet != nextSheet)) { // show menu only if we have 2 or more sheets initialized
```
gudnimg: Similar here. Why split this into two lines. I feel it is very unusual.
The comment can be either above the if statement or inside it.

Firmware/ultralcd.cpp
3d-gussner: You are correct, this menu will dump lot of information over serial and should only be used with the MK404 or with a subset of the messages.
I added this to be able to troubleshoot some LCD menu's.

Firmware/Marlin_main.cpp
3d-gussner: When you assume that `PRUSA uvlo` is only used by the "normal" Power Panic recovery functions then the `saved_printing_type` is already read before the `PRUSA uvlo`  is called.
But as this is a gcode that can be executed any time the `saved_printing_type` may not be correct.
I have no idea why this gcode even exists.

Firmware/Marlin_main.cpp
3d-gussner: No as `cancle_saved_printing()` sets `saved_start_position[0] = SAVED_START_POSITION_UNSET;` and the function 
`void restore_print_from_ram_and_continue(float e_move)` only resets some states BUT continues to print.

Firmware/Marlin_main.cpp
3d-gussner: It is 6 years old and as DEBUG code has been placed close to the operation you are trying to debug.
In VScode and an existing `Configuration_prusa.h` it is "grayed out" and so not that disturbing to read at least for me.

Firmware/power_panic.cpp
3d-gussner: Fixed with EEPROM_UVLO_Z_LIFTED

Firmware/ultralcd.cpp
```diff
@@ -5297,81 +5299,87 @@ static void lcd_main_menu()
...
+             MENU_ITEM_GCODE_P(_i("Change SD card"), PSTR("M21"));  // SD-card changed by user ////MSG_CNG_SDCARD c=18
 #endif //SDCARDDETECT
-            }
-        } else {
-            bMain=true;                                   // flag (i.e. 'fake parameter') for 'lcd_sdcard_menu()' function
-            MENU_ITEM_SUBMENU_P(_i("No SD card"), lcd_sdcard_menu); ////MSG_NO_CARD c=18
+                }
+            } else {
+                bMain=true;                                   // flag (i.e. 'fake parameter') for 'lcd_sdcard_menu()' function
+                MENU_ITEM_BACK_P(_i("No SD card")); ////MSG_NO_CARD c=18
```
gudnimg, line 5321: This changes the UI, is this change needed?

Firmware/ultralcd.cpp
```diff
@@ -5297,81 +5299,87 @@ static void lcd_main_menu()
 #endif
 
     // only allow starting SD print if hardware errors (temperature or fan) are cleared
-    if(!get_temp_error()
-            && !printer_active()
+    if (!printer_recovering() && !printer_active() && !get_temp_error()
```
gudnimg, line 5302: Potential optimisation: there are a few if statements inside this top-level if statements which check for `!printer_active()`. That shouldn't be needed anymore since this top-level if statement checks it.

Firmware/ultralcd.cpp
```diff
@@ -5297,81 +5299,87 @@ static void lcd_main_menu()
...
 #endif //SDSUPPORT
-    }
-
-    if(!printer_active() && !farm_mode) {
-        const int8_t sheet = eeprom_read_byte(&(EEPROM_Sheets_base->active_sheet));
-        const int8_t nextSheet = eeprom_next_initialized_sheet(sheet);
-        if ((nextSheet >= 0) && (sheet != nextSheet)) { // show menu only if we have 2 or more sheets initialized
-            MENU_ITEM_FUNCTION_E(EEPROM_Sheets_base->s[sheet], eeprom_switch_to_next_sheet);
-        }
+        if(!printer_active() && !farm_mode) {
```
gudnimg: `!printer_active()` should always be true here since top-level if statement checks for it

Firmware/ultralcd.cpp
```diff
@@ -5297,81 +5299,87 @@ static void lcd_main_menu()
...
-    }
+        }
 
-    if ( ! ( printer_active() || (eFilamentAction != FilamentAction::None) || Stopped ) ) {
-        if (MMU2::mmu2.Enabled()) {
-            if(!MMU2::mmu2.FindaDetectsFilament() && !fsensor.getFilamentPresent()) {
-                // The MMU 'Load filament' state machine will reject the command if any 
-                // filament sensor is reporting a detected filament
-                MENU_ITEM_SUBMENU_P(_T(MSG_PRELOAD_TO_MMU), mmu_preload_filament_menu);
+        if (!( printer_active() || (eFilamentAction != FilamentAction::None) || Stopped )) {
```
gudnimg: `!printer_active()` should always be true here since top-level if statement checks for it

Firmware/ultralcd.cpp
```diff
@@ -5249,45 +5253,43 @@ static void lcd_main_menu()
...
+        && fan_check_error != EFCE_REPORTED
 #endif //FANCHECK
-           ) {
-            if (saved_printing) {
+        && (saved_printing_type != PowerPanic::PRINT_TYPE_NONE || saved_printing)
+        && custom_message_type != CustomMsg::Resuming) {
+        if (saved_printing_type == PowerPanic::PRINT_TYPE_SD) {
                 MENU_ITEM_SUBMENU_P(_T(MSG_RESUME_PRINT), lcd_resume_print);
-            } else {
+        } else if ((saved_printing_type == PowerPanic::PRINT_TYPE_HOST) && (M79_timer_get_status())) {
```
gudnimg, line 5287: hmm, isn't this a regression for users which **don't** use the M79 timer? If a user doesn't use M79, this change means they can no longer resume the print from the LCD. But still they can pause the print from LCD.

See `lcd_pause_usb_print` which doesn't have this restriction.

gudnimg (changes requested):
See comments

Firmware/ultralcd.cpp
3d-gussner: Right, thanks

Firmware/ultralcd.cpp
3d-gussner: Right, thanks

Firmware/ultralcd.cpp
3d-gussner: Right, thanks

gudnimg (approved):

Firmware/ultralcd.cpp
3d-gussner: It has been pointed out that the menu shouldn't be shown as long the host isn't connected.
Testing selected Resume too early while the RPi Zero was still booting up after a power panic. Nothing happened as the `//action:resume` never reached the host.
Using the `M79` ensures that the host is connected and ready to receive instructions like `//action:resume` and prevent users to select the menu too early.

PR #4642, opened 2024-03-20
Fix checking file for PS2.7.3 additional comments from 30Kb to 50Kb

3d-gussner:


3d-gussner:
Thanks to PrusaSlicer team fixing it with 2.7.3-alpha1. No need for this PR anymore

PR #4644, opened 2024-03-22
Add Eject Option To Filament Load Message

### LINKED ISSUE #4643: [ENHANCEMENT]Filament load needs to be 3 state
Enter what type of printer or upgrade the enhancement applies to.
**Printer type** - MK3S, MK3,

  Load filament has 2 options was a filament loaded correctly Yes / No.

It is not sufficient enough. Many times I get into a situation when fillament stuck and both answers are irelevant.
You will need to immediatelly eject.

The situation might be bad cut fillament and stuck, wrong temperature and so on.

The answer Yes is the lie, but it only allows to further fillament eject.

The correct answer No makes a situation even worse.

There is a neccessity to add 3rd option **eject**.

sarusani:
Add a third option ("eject") to the "Filament extruding & with correct color?" message.

Closes https://github.com/prusa3d/Prusa-Firmware/issues/4643

The third_choice text was rendered incorrectly. (Only showed the first letter of the text because the position was hardcoded.)
I made sure the third_choice text will be offset to the left, but not overlap with the second_choice text.

This uses quite a lot of mem, any help in optimizing is appreciated 😸

### DIFF
--- a/Firmware/messages.cpp
+++ b/Firmware/messages.cpp
@@ -18,6 +18,7 @@ const char MSG_BED_HEATING[] PROGMEM_I1 = ISTR("Bed Heating"); ////MSG_BED_HEATI
 const char MSG_BED_LEVELING_FAILED_POINT_LOW[] PROGMEM_I1 = ISTR("Bed leveling failed. Sensor didn't trigger. Debris on nozzle? Waiting for reset."); ////MSG_BED_LEVELING_FAILED_POINT_LOW c=20 r=6
 const char MSG_BED_SKEW_OFFSET_DETECTION_FITTING_FAILED[] PROGMEM_I1 = ISTR("XYZ calibration failed. Please consult the manual."); ////MSG_BED_SKEW_OFFSET_DETECTION_FITTING_FAILED c=20 r=8
 const char MSG_BELT_STATUS[] PROGMEM_I1 = ISTR("Belt status");////MSG_BELT_STATUS c=18
+const char MSG_EJECT[] PROGMEM_I1 = ISTR("Eject"); ////MSG_EJECT c=8
 const char MSG_CANCEL[] PROGMEM_I1 = ISTR(">Cancel");////MSG_CANCEL c=10
 const char MSG_CALIBRATE_Z_AUTO[] PROGMEM_I1 = ISTR("Calibrating Z"); ////MSG_CALIBRATE_Z_AUTO c=20 r=2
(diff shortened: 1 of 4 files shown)


gudnimg:
I wonder if we can save memory by converting the following arguments into a single `struct` and adding `third_col` parameter. And that way force the caller to specify the column. 🤔

`void lcd_show_choices_prompt_P(uint8_t selected, const char *first_choice, const char *second_choice, uint8_t second_col, const char *third_choice)`

3d-gussner:
> I wonder if we can save memory by converting the following arguments into a single `struct` and adding `third_col` parameter. And that way force the caller to specify the column. 🤔
> 
> `void lcd_show_choices_prompt_P(uint8_t selected, const char *first_choice, const char *second_choice, uint8_t second_col, const char *third_choice)`

The biggest cost +196 bytes is https://github.com/prusa3d/Prusa-Firmware/pull/4644/files#diff-5fbfdca43c192573e31d8e78d255c510fb6e213b118cf38173277ec2a2163880R2261-R2269

The last part only costs additional +32 bytes https://github.com/prusa3d/Prusa-Firmware/pull/4644/files#diff-5fbfdca43c192573e31d8e78d255c510fb6e213b118cf38173277ec2a2163880R3047-R3053

gudnimg:
> The biggest cost +196 bytes is https://github.com/prusa3d/Prusa-Firmware/pull/4644/files#diff-5fbfdca43c192573e31d8e78d255c510fb6e213b118cf38173277ec2a2163880R2261-R2269

In that case I wonder if some of the function calls are being inlined. You can try adding `__attribute__((noinline))` to the definition and see if that helps. None of these are time critical.

gudnimg:
I looked a bit into this flash usage, and I suspect this is due to the compiler not able to optimise the code like it could before. `third_choice` was always set to `nullptr` before, but now you give it a value which complicates things. I've pretty much ruled out inlining.

3d-gussner:
@gudnimg Any idea if that can be optimized?

3d-gussner:
@sarusani Please rebase the PR and solve the conflicts.

3d-gussner (approved):
Tested with MK404 :heavy_check_mark: 

Tue Apr  9 18:50:31 2024

3d-gussner:
@gudnimg Do you think we can save any bytes here? If not please approve.

PR #4647, opened 2024-03-28
MMU maintenance: Sync with 32-bit FW a little bit

gudnimg:
The changes in this commit are mainly formatting, but there are some changes like adding a new function `planner_draining()`.

There are more differences between 8-bit and 32-bit, but it's nice to get most of the formatting out of the way to make future PRs easier to review.

### DIFF
--- a/Firmware/mmu2.cpp
+++ b/Firmware/mmu2.cpp
@@ -493,9 +493,10 @@ bool MMU2::unload() {
         return false;
     }
 
+    WaitForHotendTargetTempBeep();
+
     {
         ReportingRAII rep(CommandInProgress::UnloadFilament);
-        WaitForHotendTargetTempBeep();
         UnloadInner();
     }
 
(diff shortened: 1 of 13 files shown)


3d-gussner (approved):
Looks good to me

Firmware/mmu2.cpp
```diff
@@ -860,7 +885,7 @@ StepStatus MMU2::LogicStep(bool reportErrors) {
     case Finished:
         // At this point it is safe to trigger a runout and not interrupt the MMU protocol
         CheckFINDARunout();
-        break;
+        [[fallthrough]]; // let Finished be reported the same way like Processing
```
DRracer: Please verify this change on MK3S. It may break current implementation on this platform. MK4 may have a different workflow.

DRracer (approved):
Conditionally approved, just check the one functional change marked in my review. Otherwise ok

Firmware/mmu2.cpp
3d-gussner: @gudnimg Did you test this?

gudnimg: No I don’t think I tested it specifically. We can drop this change if you think its risky.

PR #4565, opened 2024-01-11
Move all _i(" to messages.cpp/.h

3d-gussner:

PR #4650, opened 2024-04-03
LCD_Menu_cleanup

### LINKED ISSUE #3361: [BUG] Settings/Calibrate/Calibrate Z should not have a "right arrow"

3d-gussner:
New version of https://github.com/prusa3d/Prusa-Firmware/pull/4543 with minor changes

Change to `MENU_ITEM_FUNCTION_P`

- [X] Calibrate Z
- [X] Thermal Model cal.

Fixes #3361

This PR will break after https://github.com/prusa3d/Prusa-Firmware/pull/4565 but not a big issue.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -4582,7 +4582,7 @@ static void lcd_calibration_menu()
     // MK2
     MENU_ITEM_FUNCTION_P(_T(MSG_CALIBRATE_BED), lcd_mesh_calibration);
     // "Calibrate Z" with storing the reference values to EEPROM.
-    MENU_ITEM_SUBMENU_P(_T(MSG_HOMEYZ), lcd_mesh_calibration_z);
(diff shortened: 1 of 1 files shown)


gudnimg (approved):
Looks good to me

3d-gussner:
@gudnimg Thanks for the review and approval

Issue #3008, opened 2021-02-01
[BUG] Cold extrude not saved during Power Panic


wavexx:
**Printer type** - MK2.5+
**Printer firmware version** - 3.9+
**MMU Upgrade** - N/A
**SD card or USB/Octoprint** - N/A

**Describe the bug**
A print started using the cold extrude feature (M302) will fail to resume printing after a power-panic, since the limit temperature is not saved.
  
**To Reproduce**
Start a print with cold-extrusion (any temp below 175):

```
M302 S150
M104 S150
```

Trigger PP. As soon as extrusion is started the printer will stop due to cold extrusion prevention.

**Expected behavior**
Printer should continue printing as usual.

This is unlikely to cause significant issues, since the current cold-extrusion limit is pretty low.

3d-gussner:
@gudnimg Can you please add `M302` value to PowerPanic

3d-gussner:
Fixed with #4656

PR #4656, opened 2024-04-09
Add Extrude mintemp save and restore during power panic

3d-gussner:
Thanks to @wavexx in #3008
Minor documentation changes
Tested with MK404
- Set `M302 S150`
- Start print at 155°C
- PowerPanic during print
- Let it recover
- Check `D3 Ax0c7d C2` that the correct value has been saved to EEPROM during PP
- Set temp to 155°C which is below the default of 175°C
- Verify that no ` cold extrusion prevented` is shown

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -7449,7 +7449,7 @@ Sigma_Exit:
       set_extrude_min_temp(temp);
     }
     break;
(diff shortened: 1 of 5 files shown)


gudnimg (approved):
Looks good to me

wavexx (approved):
Looking good as well

PR #4659, opened 2024-04-11
PFW-1460 Update build scripts

3d-gussner:
Builds all firmware.

Tested on
- [x] Linux (Ubuntu 20.04 LTS)
  - [X] git clone
  - [X] download zip
- [X] Windows 10
  - [X]  git clone
  - [X] download zip

UPDATE 10 July:
It turns out that cmake on Windows and maybe also on MacOS has issues with wildcard usage, while Linux works fine.
Created https://github.com/prusa3d/Prusa-Firmware/pull/4708 to solve this issue.

### DIFF
--- a/utils/build.py
+++ b/utils/build.py
@@ -1,0 +1,389 @@
+#!/usr/bin/env python3
+import argparse
+import os
+import platform
+import random
+import re
+import shutil
+import subprocess
+import sys
+import xml.etree.ElementTree as ET
+from abc import ABC, abstractmethod, abstractproperty
+from copy import deepcopy
+from enum import Enum
+from functools import lru_cache
+from pathlib import Path
+from typing import Dict, List, Optional
+from uuid import uuid4
+
+try:
+    from tqdm import tqdm
+except ModuleNotFoundError:
+
+    def tqdm(iterable, *args, **kwargs):
+        return iterable
+
+    if os.isatty(sys.stdout.fileno()) and random.randint(0, 10) <= 1:
+        print('TIP: run `pip install -m tqdm` to get a nice progress bar')
+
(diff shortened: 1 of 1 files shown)


3d-gussner:
While testing Windows 10 the `build.py` script fails during compressing, while the Linux version is working correctly.
Looking into that.

sarusani:
The problem is not in the build.py. I get the same error when I run the "ALL_MULTILANG" CMake directly from VSCode. (Windows)

All multilang builds for MK25 & MK25S variants fail.
(English only builds work fine, so do all builds for MK3 & MK3S variants)

I'll have a closer look on the weekend.

3d-gussner:
I think something goes wrong here on windows https://github.com/prusa3d/Prusa-Firmware/blob/MK3/CMakeLists.txt#L594-L595
Can you check on Windows if build\release\release is created which would be https://github.com/prusa3d/Prusa-Firmware/blob/MK3/CMakeLists.txt#L594

Or maybe it takes bit longer on Windows to create the directory and the next line is executed too early?

sarusani:
Windows doesn't like the `${variant_name}_${FN_VERSION_SUFFIX}_en*.hex`.
The wildcard doesn't work. If you (just for testing) change it to `${variant_name}_${FN_VERSION_SUFFIX}_en-ro.hex` the build works.

PS:
I think the problem is that in linux the filepath is resolved by the shell and then passed to cmake. In windows this doesn't seem to be the case...
(Not sure about this 😄)

3d-gussner:
@sarusani Please try https://github.com/prusa3d/Prusa-Firmware/pull/4708 + this PR.

DRracer (approved):

PR #4661, opened 2024-04-14
PFW-1557 MMU: Improve unload handling when print is aborted

gudnimg:
- Don't allow re-heating the nozzle or the MMU to unload the filament if the print was aborted by a temperature error
- Allow MMU to unload filament if there is no temperature error, and there is no fan error when Fan check is enabled
- When Fan check is disabled, the MMU may unload if there is no temperature error

Change in memory:
Flash: +14 bytes
SRAM: 0 bytes

PFW-1557

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -5716,23 +5716,26 @@ void lcd_print_stop_finish()
 
         // restore the auto hotend state
         hotendDefaultAutoFanState();
-    }
 
-    if (MMU2::mmu2.Enabled() && MMU2::mmu2.FindaDetectsFilament())
-    {
-        // The print was aborted while when the nozzle was cold:
-        //     1. in a paused state                      => a partial backup in RAM is always available
-        //     2. after a recoverable thermal/fan error had paused the print => only extruder temperature is saved to RAM
-        if (printingIsPaused())
-        {
-            // Restore temperature saved in ram after pausing print
-            restore_extruder_temperature_from_ram();
-        }
+        if (MMU2::mmu2.Enabled() && MMU2::mmu2.FindaDetectsFilament()
+#ifdef FANCHECK
+            && fan_check_error != EFCE_REPORTED
+#endif //FANCHECK
+        ) {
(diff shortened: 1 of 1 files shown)


Firmware/ultralcd.cpp
```diff
@@ -5716,23 +5716,26 @@ void lcd_print_stop_finish()
...
+                restore_extruder_temperature_from_ram();
+            }
 
-        // If the pause state was cleared previously or the target temperature is 0°C in the case
-        // of an unconditional stop. In that scenario we do not want to unload.
-        if (target_temperature[0] >= extrude_min_temp) {
-            MMU2::mmu2.unload(); // M702
+            // If the pause state was cleared previously or the target temperature is 0°C in the case
+            // of an unconditional stop. In that scenario we do not want to unload.
+            if (target_temperature[0] >= extrude_min_temp) {
```
3d-gussner, line 5736: Shouldn't this be
`if (target_temperature[0] >= extrude_min_temp && (int)degHotend(active_extruder) > extrude_min_temp) {`
or
`if ((int)degHotend(active_extruder) > extrude_min_temp) {` 

as the `target_temperature` can be set BUT the current temperature isn't reached yet?

3d-gussner: The `restore_extruder_temperature_from_ram` before will bring up the temp to the saved temperature, so there is not a big difference.
But most if statements comparing `extrude_min_temp` check against the `(int)degHotend(active_extruder)` which makes sense.

gudnimg: The unload functions makes sure to wait: https://github.com/prusa3d/Prusa-Firmware/blob/MK3/Firmware/mmu2.cpp#L479

gudnimg: The reason I added the if statement `if (target_temperature[0] >= extrude_min_temp)` is simply to check if heating is disabled. An example of this is when a print is stopped by TMC2130 Crash detection or Unconditional stop. And then prevent the unload from happening.

3d-gussner (approved):

PR #4662, opened 2024-04-15
Stop print when SDcard is removed while printing.

3d-gussner:
Stop print when SD card is removed
- [X] Tested on MK404

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -7417,6 +7417,9 @@ void menu_lcd_lcdupdate_func(void)
 		}
 		else
 		{
+			if(GetPrinterState() == PrinterState::IsSDPrinting) {
+				lcd_print_stop();
+			}
 			card.release();
 			LCD_MESSAGERPGM(_T(MSG_SD_REMOVED));
 		}


mkbel:
Does it really stop the print or pause the print?

3d-gussner:
@mkbel It stops the print. To ensure that the printer doesn't stay heated above the print. 

Pausing the print needs way more investigation.
- Is the re-inserted SD the same? 
- Is the file unchanged? 
- and more

Firmware/ultralcd.cpp
```diff
@@ -7417,6 +7417,9 @@ void menu_lcd_lcdupdate_func(void)
 		}
 		else
 		{
+			if(GetPrinterState() == PrinterState::IsSDPrinting) {
+				lcd_print_stop();
```
gudnimg, line 7421: While I'm sure this stop the print fine, I'd be very careful though with doing this inside `menu_lcd_lcdupdate_func`. If `lcd_print_stop` is triggering lcd_update anywhere then you may be creating a recursion to `menu_lcd_lcdupdate_func`.

I don't see any calls to LCD update by looking at the code.

mkbel:
I don't think you need to check anything (if the file is the same or not) upon re-insertion. If the user modifies the file it is his fault.
In Buddy firmware print is paused when USB flash drive is removed during the print. And I don't think Buddy firmware checks anything when the print is resumed.
It would be nice to have the same behaviour across our product range.

Patronics:
> I don't think you need to check anything (if the file is the same or not) upon re-insertion. If the user modifies the file it is his fault.

I was just recently thinking that this could actually be a useful feature to specifically allow, I've had a project where I realized partway through I'd made a mistake near the top of the print, but no changes were needed to the lower portion. I'd speculated as to whether I could remove the SD card (or flash drive, in that case, it was a Prusa Mini), reslice the file (being careful to keep the position, support, slicing settings, etc unchanged), and reinsert it to allow it to resume with the fix.

Having that officially supported (at the user's own risk, of course), would be really useful to save or fix long prints sometimes!

3d-gussner:
Closing as it may cause more issues than solving.

PR #4689, opened 2024-05-16
PFW-964: Skip filament unload/load after Shipping factory reset

3d-gussner:
when MMU is enabled. With MMU enabled no filament should be loaded while idle.


leptun (approved):
I don't see any issues with this. Just as a check that I understood the code, this is the unload between xyz cal and thermal model cal?

3d-gussner:
> I don't see any issues with this. Just as a check that I understood the code, this is the unload between xyz cal and thermal model cal?

It just skips the unload/load when MMU is enabled, which is normally a case after someone enabled the MMU and then Shipping reset the printer.
@leptun Thanks for the review and approval.

PR #4715, opened 2024-07-13
Beep on M601 pause

alexiri:
Make some noise when pausing a print so the user knows the printer is waiting for them.

(Tested on MK3S only)

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -7972,6 +7972,8 @@ Sigma_Exit:
             cmdqueue_pop_front(); //trick because we want skip this command (M601) after restore
             lcd_pause_print();
         }
+
+        Sound_MakeSound(e_SOUND_TYPE_StandardPrompt);
     }
     break;
 


gudnimg:
Hi @alexiri sorry for late review. I am a bit unsure about this change because usually the printer doesnt beep when paused.

For host printing (Octoprint, Prusa Link) the user should see the paused status remotely. Do I understand correctly your proposed change is mainly aimed at SD card printing?

I wonder if using a M300 g-code prior to M601 is a better option?

alexiri:
Hi @gudnimg, thanks for taking a look.

You're right, the printer doesn't beep when paused, that's precisely why I'd like to introduce this change. :smile: The printer _does_ [beep with an M600](https://github.com/prusa3d/Prusa-Firmware/blob/MK3/Firmware/Marlin_main.cpp#L11253) color change to get the user's attention, why should an M601 pause be any different? In both cases, you want to get the user's attention and let them know they're supposed to do something (change the filament, insert a magnet/nut, whatever).

This isn't just for SD printing (in fact all the printing I do is host printing).

gudnimg:
What do you think about this? I feel that using `sound_wait_for_user();` / `sound_wait_for_user_reset();` is a more configurable approach. It is consistent with M600 and MMU error screens and it also respects the configurable sound settings (Loud, Once, Assist, and Silent).

Both M602 and resuming the print via LCD will stop the beeping (if not in 'Once' and 'Silent' mode).

[sound_M601_example.txt](https://github.com/user-attachments/files/16404609/sound_M601_example.txt)

```
diff --git a/Firmware/Marlin_main.cpp b/Firmware/Marlin_main.cpp
index ddf70b8d3..45d26410b 100644
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -7971,6 +7971,9 @@ Sigma_Exit:
             ClearToSend(); //send OK even before the command finishes executing because we want to make sure it is not skipped because of cmdqueue_pop_front();
             cmdqueue_pop_front(); //trick because we want skip this command (M601) after restore
             lcd_pause_print();
+
+            // Start beeping to get the user's attention
+            sound_wait_for_user();
         }
     }
     break;
diff --git a/Firmware/ultralcd.cpp b/Firmware/ultralcd.cpp
index 08ca782ad..3d92f4d42 100644
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -5028,6 +5028,9 @@ bool resume_print_checks() {
 //! as restore_print_from_ram_and_continue() calls lcd_update() internally.
 void lcd_resume_print()
 {
+    // Printer may be beeping if it was paused via M601, stop the beeping
+    sound_wait_for_user_reset();
+
     // reset lcd and ensure we can resume first
     if (!resume_print_checks()) return;
 
```

alexiri:
If you think that's best, it's fine for me. Note that `Sound_MakeSound` [also respects the sound settings](https://github.com/prusa3d/Prusa-Firmware/blob/MK3/Firmware/sound.cpp#L75), though it only beeps once.

gudnimg:
@alexiri I'd like to discuss this with the other collaborators/maintainers and hear their opinions. Added this PR to the 3.14.1 milestone as a reminder. :)

3d-gussner:
@alexiri Sorry for the late reply. We will not merge this PR as you can update your PruisaSlicer profile by modifying the Printers -> Custom G-code -> Pause Print G-code and adding there a [`M300`](https://reprap.org/wiki/G-code#M300:_Play_beep_sound) gcode in front of `M601`

PR #4717, opened 2024-07-15
Fix an issue in manual fan check where prior knob clicks are not consumed

gudnimg:
The issue can appear when the knob is clicked at the "Self test start" screen which is hardcoded to stay on the screen for 2 seconds. Any knob clicks during this interval will go unconsumed.

Issue was introduced in 3.13.0, likely by https://github.com/prusa3d/Prusa-Firmware/commit/a7e9ccfb57e436311cb28b5636b8f66fe7982909

Issue mainly affects MK2.5S.

Steps to reproduce:
1. Run selftest
2. When you see “Self test start” on the LCD, click the knob. 
3. Wait for fan self test to run
4. Observe error: The printer doesnt give any choice to select whether the fan is spinning or not.


3d-gussner (approved):
Tested on MK404 and the accidentally pressed knob issue is solved.

PR #4727, opened 2024-07-25
PFW-1559 Add filament presence check at start of print

### LINKED ISSUE #4380: FW 3.13.0-3.13.1 filament detection not working at print start
**Printer type**  MK3S+
**Printer firmware version** -  3.13.1

**Describe the bug**
I normally send a print request from PrusaSlicer without filament loaded. After mesh bed leveling, the print head goes to the front left to print the lead in line and detects there is no filament. Then starts the load filament routine.

After updating to FW 3.13.1 the printer starts printing with no filament loaded.

I confirmed the filament sensor is not indicating 1. All filament sensor settings are ON.

Down graded the FW and printed same file again and lack of filament is recognized.

### LINKED ISSUE #4719: [BUG] Printing starts without filament
MK3S+

**Printer firmware version** -  [e.g. 3.8.1, 3.8.1-RC1, ...]
3.14.0

**Describe the bug**
Since firmware 3.13.2 my MK3S+ tries to start printing even when there is no filament loaded. With previous firmware versions the printer stopped after the bed level test and asked to load filament, but now the printer ignores the missing filament and just starts printing. I reported this issue back in November 2023 to Prusa support, but nothing happened :-(

**To Reproduce**
Unload any filament and start printing from SD card or USB. After bed level test, printer will start printing without filament.

**Expected behavior**
The firmware should check the filament sensor after bed level test to make sure that filament is loaded. If filament is missing, the user should be asked to load filament.

gudnimg:
Add filament presence check at start of print job. This PR aims to resolve issues where a user starts a print with no filament loaded. The idea is to use the M862 warning dialogs instead of relying on filament sensor runout. Relying on filament runout will require some fragile logic which may not be very reliable and risks adding new issues.

A new menu setting is added to configure how strict the check should be.

- None: No check is performed
- Warning **(default)**: The user is warned about missing filament and is prompted to continue with `Yes`/`No`. If `No` is selected, the print is aborted. If no user input is given (e.g. from host printing) then the warning will expire in 30 seconds and the printer assumes the `Yes` option was selected.
- Strict: If the filament is not detected when a print is started, it is immediately canceled with a message saying the filament is missing.

The check is currently performed on G-codes: M24 and M75. Starting a SD print and Host print respectively.

### DIFF
--- a/Firmware/eeprom.h
+++ b/Firmware/eeprom.h
@@ -246,9 +246,7 @@ static_assert(sizeof(Sheets) == EEPROM_SHEETS_SIZEOF, "Sizeof(Sheets) is not EEP
 | ^           | ^       | ^                                     | 01h 1        | ^                     | MMU cutter: __enabled__                           | ^            | ^
 | ^           | ^       | ^                                     | 02h 2        | ^                     | MMU cutter: __always__                            | ^            | ^
 | 0x0DAE 3502 | uint16  | EEPROM_UVLO_MESH_BED_LEVELING_FULL    | ???          | ff ffh 65535          | Saved MBL points                                  | Power Panic  | D3 Ax0dae C288
-| 0x0DAD 3501 | uint8   | EEPROM_CHECK_FILAMENT                 | 01h 1        | ffh 255               | Check mode for filament is: __warn__              | LCD menu     | D3 Ax0dad C1
-| ^           | ^       | ^                                     | 02h 2        | ^                     | Check mode for filament is: __strict__            | ^            | ^
-| ^           | ^       | ^                                     | 00h 0        | ^                     | Check mode for filament is: __none__              | ^            | ^
+| 0x0DAD 3501 | uint8   | _EEPROM_FREE_NR9_                     | ???          | ffh 255               | _Free EEPROM space_                               | _free space_ | D3 Ax0dad C1
 | 0x0DAC 3500 | bool    | EEPROM_MBL_MAGNET_ELIMINATION         | 01h 1        | ffh 255               | Mesh bed leveling does: __ignores__ magnets       | LCD menu     | D3 Ax0dac C1
 | ^           | ^       | ^                                     | 00h 0        | ^                     | Mesh bed leveling does: __NOT ignores__ magnets   | ^            | ^
 | 0x0DAB 3499 | uint8   | EEPROM_MBL_POINTS_NR                  | 03h 3        | ffh 255               | Mesh bed leveling points: __3x3__                 | LCD menu     | D3 Ax0dab C1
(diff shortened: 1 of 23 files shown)


Firmware/eeprom.h
```diff
@@ -245,7 +245,9 @@ static_assert(sizeof(Sheets) == EEPROM_SHEETS_SIZEOF, "Sizeof(Sheets) is not EEP
 | ^           | ^       | ^                                     | 01h 1        | ^                     | MMU cutter: __enabled__                           | ^            | ^
 | ^           | ^       | ^                                     | 02h 2        | ^                     | MMU cutter: __always__                            | ^            | ^
 | 0x0DAE 3502 | uint16  | EEPROM_UVLO_MESH_BED_LEVELING_FULL    | ???          | ff ffh 65535          | Saved MBL points                                  | Power Panic  | D3 Ax0dae C288
-| 0x0DAD 3501 | uint8   | _EEPROM_FREE_NR9_                     | ???          | ffh 255               | _Free EEPROM space_                               | _free space_ | D3 Ax0dad C1
+| 0x0DAD 3501 | uint8   | EEPROM_CHECK_FILAMENT                 | 01h 1        | ffh 255               | Check mode for filament is: __warn__              | LCD menu     | D3 Ax0dad C1
```
leptun: I don't think it's a good idea to reuse this. We can only reuse eeprom slots if we know for sure that the old value is ok or not initialized. This can only be done feasibly when a feature needs multiple eeprom bytes or when we don't care about the old value, such as in power panic.
So I suggest you consume a new eeprom byte for this.

Firmware/messages.cpp
```diff
@@ -151,6 +151,8 @@ const char MSG_GCODE_NEWER_FIRMWARE_CONTINUE[] PROGMEM_I1 = ISTR("G-code sliced
 const char MSG_GCODE_NEWER_FIRMWARE_CANCELLED[] PROGMEM_I1 = ISTR("G-code sliced for a newer firmware. Please update the firmware. Print cancelled."); ////MSG_GCODE_NEWER_FIRMWARE_CANCELLED c=20 r=8
 const char MSG_GCODE_DIFF_CONTINUE[] PROGMEM_I1 = ISTR("G-code sliced for a different level. Continue?"); ////MSG_GCODE_DIFF_CONTINUE c=20 r=3
 const char MSG_GCODE_DIFF_CANCELLED[] PROGMEM_I1 = ISTR("G-code sliced for a different level. Please re-slice the model again. Print cancelled."); ////MSG_GCODE_DIFF_CANCELLED c=20 r=8
+const char MSG_MISSING_FILAMENT_CONTINUE[] PROGMEM_I1 = ISTR("There is no filament loaded. Continue?"); ////MSG_MISSING_FILAMENT_CONTINUE c=20 r=3;
+const char MSG_MISSING_FILAMENT_CANCELLED[] PROGMEM_I1 = ISTR("There is no filament loaded. Print cancelled."); ////MSG_MISSING_FILAMENT_CANCELLED c=20 r=8
```
leptun, line 158: I'm not too fond of the duplicated messages. This is a good place to save some flash in the future.

Firmware/eeprom.h
gudnimg: Done

Firmware/messages.cpp
gudnimg: I think we should not tackle that in this PR as we discussed earlier today. But FYI the strings I have at the moment can be considered placeholders. I just typed something to create the initial PR. We can change the wording if we want.

gudnimg:
The following videos demonstrate the changes more visually 😃 


This video shows the default setting (Warning). I attempt to start a print with no filament loaded.
https://github.com/user-attachments/assets/334b5e05-1906-4891-9bea-38dbfcf5865a

This video shows the "Strict" setting. It will always cancel the print if filament is not loaded.
https://github.com/user-attachments/assets/d5a4fcf2-1716-4241-b88a-de5b2e829f15

leptun (approved):
Tested on MK3 (PAT9125). Works really nicely!

gudnimg:
Synced the PR with latest MK3 branch. Also got rid of the "fixup" commit to make the history a bit cleaner. (No code has been changed)

gudnimg:
Please do not merge this yet. I want to remove the mmu2.h include in utils.cpp. To make unit testing easier to begin with. 

To check if the MMU is enabled, I think it is ok to read the EEPROM value. The printer check shouldnt care if the MMU is communicating or not, it only cares if it is _enabled_ at all.

gudnimg:
Should be ready for review again

3d-gussner (changes requested):
Please check https://github.com/gudnimg/Prusa-Firmware/pull/8

3d-gussner (approved):

PR #4728, opened 2024-07-26
Fix regression introduced in 3.13.0 where `mFilamentPrompt` menu times out after 30 seconds

### LINKED ISSUE #4718: Unload Filament Command Only Gives 30 Seconds to Comply
In previous versions of the firmware, choosing "Unload Filament" on an MK3s+, and then the filament type, would start a pre-heating process. Once it was up to the necessary temperature, the printer would sit and wait on a screen that said something to the effect of "click to unload." And it would just sit there for a LONG time. Eventually it would time out...after maybe close to 10 minutes, and give the heating disabled due to safety timer.

In recent changes to the firmware, now when you click the filament type to unload, it again heats up to temperature, but then gives only 30 seconds to click to unload. If you don't click it in that small window, it goes back to the main menu, forcing you to scroll and click "Unload Filament" again. I timed it, and I think it was 30 seconds exactly, so it seems like this value was changed intentionally.

This change makes very little sense. If you only have a single printer, that means you have to stand there and watch it. You can't work on anything else, change filament colors, etc. This problem is even worse if you run a print farm. Previously I would tell all my printers to unload, then I could do all my filament color prepping, and then each printer would be waiting asking me to click to unload. I could go through all of them so fast. Click click click click and unload a bunch of them.

Now with this 30 second timeout, I have to scroll and click unload on every single printer. I know that doesn't sound awful, but why the 30 second timeout at all? If a user wants to access anything else like changing sheet type, they can do that before or after the filament unload.

Love PrusaSlicer, but this specific change has been a step backwards for me personally. I don't know if this change was introduced with the 3.13 series of firmware, but I'd love for it to go back to sitting and waiting for the user. After all...if the user clicks unload filament, that's what they want to do! 🙂

gudnimg:
Fixes https://github.com/prusa3d/Prusa-Firmware/issues/4718

This PR fixes a regression introduced in 3.13.0 here: https://github.com/prusa3d/Prusa-Firmware/commit/a7e9ccfb57e436311cb28b5636b8f66fe7982909#diff-5fbfdca43c192573e31d8e78d255c510fb6e213b118cf38173277ec2a2163880R7462

Before 3.13.0, the menu never timed out until the safety timer turned off the heaters
After 3.13.0, the menu times out after 30 seconds.

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -1815,6 +1815,7 @@ void lcd_print_target_temps_first_line() {
 
 static void mFilamentPrompt()
 {
+    lcd_timeoutToStatus.stop();
 lcd_print_target_temps_first_line();
 lcd_puts_at_P(0,1, _T(MSG_PRESS_KNOB));
 lcd_set_cursor(0,2);
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -1813,61 +1813,59 @@ void lcd_print_target_temps_first_line() {
     }
 }
 
-static void mFilamentPrompt()
-{
+static void mFilamentPrompt() {
     lcd_timeoutToStatus.stop();
-lcd_print_target_temps_first_line();
-lcd_puts_at_P(0,1, _T(MSG_PRESS_KNOB));
-lcd_set_cursor(0,2);
-switch(eFilamentAction)
-     {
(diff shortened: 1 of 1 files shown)


sarusani:
Tested and works as expected. (No timeout until safety timer kicks in)

3d-gussner (approved):

Issue #4703, opened 2024-06-12
FIrmware should check filament sensor before nozzle change


dangerweenie:
MK3

I often forget to unload filament before beginning nozzle change heatup, which leads to clogs due to the 280c heatup.

Would be very very simple for FW to simply check filament sensor and/or remind user to unload filament before heating up for nozzle swap.

3d-gussner:
@dangerweenie Thanks for the idea. Can't promise anything but mark it for FW 3.14.1

3d-gussner:
@dangerweenie FYI @sarusani PR has been merged and will be in FW 3.14.1

PR #4736, opened 2024-07-28
Remove recursion workaround in Nozzle Change

gudnimg:
This workaround isn't needed today. At the time this workaround was added, `lcd_show_multiscreen_message_yes_no_and_wait_P` was calling `lcd_update()` via `lcd_update_enable(true)` which caused all sorts of problems like first layer calibration being inaccessible.

When testing this PR, I found one more place where `lcd_update()` was being called: [7b70f6a](https://github.com/prusa3d/Prusa-Firmware/pull/4736/commits/7b70f6a5922fca082dd6dcf61a32e766fd3e2e15) which has now been eliminated. This change is required to fully remove the workaround.

Steps to test:

1. Power on printer
2. Run Nozzle Change from Hardware setup menu
3. **Test:** at no point does the LCD seem to _freeze_ or become _unresponsive_.

### DIFF
--- a/Firmware/lcd.h
+++ b/Firmware/lcd.h
@@ -111,18 +111,18 @@ extern void lcd_buttons_update(void);
 //! When constructed (on stack), original state state of lcd_update_enabled is stored
 //! and LCD updates are disabled.
 //! When destroyed (gone out of scope), original state of LCD update is restored.
-//! It has zero overhead compared to storing bool saved = lcd_update_enabled
-//! and calling lcd_update_enable(false) and lcd_update_enable(saved).
+//! Do not call lcd_update_enable() to prevent calling lcd_update() in sensitive code.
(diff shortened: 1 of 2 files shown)


gudnimg:
Note: this PR is related to https://github.com/prusa3d/Prusa-Firmware/issues/4703

We should not be adding more “hacks” like this to add the unload filament dialog :)

gudnimg:
Opening this PR up for review. Nozzle change works well on my MK3S+. There may be regressions introduced by [7b70f6a](https://github.com/prusa3d/Prusa-Firmware/pull/4736/commits/7b70f6a5922fca082dd6dcf61a32e766fd3e2e15) but I think this is a good change.

3d-gussner (approved):
Tested on MK404

- [X] MK3S nozzle change
- [X] REVO MK3S nozzle change

Issue #4724, opened 2024-07-22
[FEATURE REQUEST] After 'Set Ready', go back to main info screen


jrytio:
The new 'Set Ready' feature works great, but it's awkward that it stays in the menu after clicking it.  Then you can't see if a new job was picked up via the LCD message or by watching the temp targets change.

Would be great if after clicking 'Set Ready' you're exited out of the menu and put back to the main info screen.

sarusani:
Am I assuming correctly, that you're using it in combination with with OctoPrint? The current behavior is optimized for use with PrusaLink.

jrytio:
Yep, with OctoPrint.  I looked through your plugin and mimicked into my own.  How does PrusaLink handle that differently?   When I receive the action:ready command, I start a print in OctoPrint quite similar to yours.   I'd just like the printer LCD to go back to the main screen when that happens.    The only way I can see it started a print is if the 'Tune' menu pops up, I'd rather see my M117 messages come through on the main screen to confirm the print is starting. 

  I assumed this was a firmware thing, but if there's something I can send back from the host to have the LCD go back to the main screen, that'd be just as good from my side.

sarusani:
PrusaLink uses "Set Ready" a bit different and answers with a `M72 S1`. This will change the LCD menu to show  "Set not Ready". So it makes sense to stay in the Menu. (So you can toggle between the 2 states)
This made no sense in OctoPrint, so my plugin doesn't send the `M72`...

Currently I don't have a proper solution for you.

In my opinion, when the host is OctoPrint, the best solution would be to replace the "Set Ready" menu entry completely with an "Start Host Print" entry that sends `action:start` (natively supported by OctoPrint) instead of `action:ready`. (And then of course closes the menu after it's been selected.)

Maybe that's something that could be discussed for a future release.

jrytio:
Fair enough, it still works for me, I may just fork the Firmware if it bothers me that much.   FWIW it still does toggle to 'Set Not Ready' even though I'm not sending back M72 S1.  

In our homemade system we'll put M117 messages onto the screen if a) there's no job waiting or b) the filament listed doesn't match the next job.   We dont actually load them up in Octoprint, the set ready goes out and polls for jobs waiting for that printer and starts one if all the conditions are right.   Prior to 'Set Ready' we just used the 'PETG Preheat' as our trigger.  If a user preheated to PETG temps and there was a job waiting, we'd start it.   Set Ready is definitely an improvement.    

[...]

PR #4737, opened 2024-07-30
Allow to replace Set Ready menu entry

sarusani:
The "Set Ready"/"Set not Ready" menu entries don't make sense if the host is not PrusaLink and `action:ready` & `action:notready` are not known outside of PrusaLink.

I suggest allowing to replace this entry via configuration file and replace the entry with "Print from host" which uses `action:start` instead. (`action:start` is natively supported by OctoPrint)


gudnimg:
Is this PR related to https://github.com/prusa3d/Prusa-Firmware/issues/4724 ?

sarusani:
It was inspired by #4724, but it won't solve the OP's problem. As far as I understood, #4724 is ready to be closed without any further actions.

(OP wants to keep using `action:ready`, but also close the menu.)

Firmware/variants/MK25-RAMBo10a.h
```diff
@@ -265,6 +265,7 @@
  *------------------------------------*/
 
 //#define HOST_SHUTDOWN              //Host supports "//action:shutdown" feature
+//#define REPLACE_SETREADY           //Replace "Set Ready"/"Set not Ready" menu entry with "Print from host" menu entry
```
gudnimg: I feel the description should say more what this option does. For example to answer the question "Why/When would I want to replace set ready?"

The description here is likely our only documentation :)

sarusani: I also updated the description for the shutdown feature.

gudnimg (approved):
Looks good to me. Can't comment on the translations though.

sarusani:
The translations need to be checked. I tried, but I'm sure some of them are really bad. 😄

3d-gussner:
Please verify and approve the new translations. Remember that the message is limited to 18 characters,
- [x] :czech_republic:  @DRracer
- [X] :de: @3d-gussner
- [ ] :es: 
- [x] :fr: @Ilovemyhous @ErwanAliasr1
- [x] :it: @wavexx
- [x] :poland: @arekm
- [x] :netherlands: @vintagepc @stelgenhof
- [x] :romania: @leptun @Hauzman
- [x] :hungary: @AttilaSVK @Hauzman
- [ ] :croatia: @Prime1910 @PRPA041
- [x] :slovakia: @ingbrzy @shatter136
- [ ] :sweden: @Painkiller56
- [ ] :norway: @OS-kar @trondkla

ingbrzy (approved):
SK is OK

stelgenhof:
NL is good

3d-gussner:
@stelgenhof @ingbrzy Thanks for the quick review, very appreciated.

AttilaSVK:
Hungarian should be “Nyomtatás gépről”
The string above (set not ready) is bad as well. That one should be “Nem áll készen”.

3d-gussner:
> Hungarian should be “Nyomtatás gépről” The string above (set not ready) is bad as well. That one should be “Nem áll készen”.

@AttilaSVK Thanks for the review and suggestion also to fix the other message :hugs: 
@sarusani Please add the changes.

arekm:
For PL - if "druk" as noun then it's ok. (if as verb then "Drukuj z hosta")

3d-gussner:
> For PL - if "druk" as noun then it's ok. (if as verb then "Drukuj z hosta")

@arekm Thanks. But we have also `Print from SD card` translated to `Druk z karty SD`
See https://github.com/prusa3d/Prusa-Firmware/blob/MK3/lang/po/Firmware_pl.po#L1664-L1667

In German is it `Drucken`, which is a verb, so I would assume that it should be `Drukuj` 
But I am not sure what is used in Buddy 32-bit firmware or PrusaConnect. Can you please check as I can't read "Polish".

ErwanAliasr1:
FR is :ok:

leptun:
RO has missing diacritics. It should be „Print. de pe gazdă”. The „Oprește gazda” translation has correct diacritics in that other context, so don't update that other translation.

lang/po/Firmware_cs.po
```diff
@@ -2573,6 +2573,11 @@ msgstr "Připravit"
 msgid "Set not Ready"
 msgstr "Zrušit Připravena"
 
+#. MSG_HOSTPRINT c=18
+#: ../../Firmware/messages.cpp:107 ../../Firmware/ultralcd.cpp:5285
+msgid "Print from host"
+msgstr "Tisk z hostitele"
```
DRracer: I'd vote for something like "Tisk ze serveru"
... the word "hostitel" is a direct translation of "host" and sounds weird when used in technical meaning.
The trouble with the EN word "host" is the fact, that the same word describes a "guest" in CZ. The same applies to SK...

leptun: Same applies to RO, but I can't think of another word to refer to the printer host.

DRracer: In that case "Tisk z hosta", even though it doesn't make sense to non-technical people.
May be even "Tisk z RPi" would make sense

leptun: Hmmm that is still not ideal. In other places in the fw where we say RPi we strictly mean the UART1 port on the back of the einsy, but here it could be USB as well and not limited to a raspberry pi. "Print from USB" is also not a good replacement as it has the same issues, but in reverse where USB doesn't make sense for the rpi port.

wavexx:
Italian is ok. I have the same comment regarding "host". I would have written "Print from remote", which would then translate to "Stampa da remoto" in italian, which is much more understandable.

Ilovemyhous (commented):
French has a green light from my side.

3d-gussner:
> For PL - if "druk" as noun then it's ok. (if as verb then "Drukuj z hosta")

@sarusani Please update to `Drukuj z hosta`

DRracer:
CZ is usable :wink:

3d-gussner (approved):

PR #4739, opened 2024-07-31
Remove a bit of unused code which is disabled at compile time

gudnimg:
This is a proposal to clean up the project a little bit...

Some of these sections haven't been used in over 6 years and don't compile anymore.

Reason to want this removed is to reduce the code we maintain and make the code more readable.

Note: this also removes a bit of references to `lcd_update_enable()` which I don't want to bother touching. (related to PFW-1452)


Firmware/Marlin_main.cpp
```diff
@@ -1382,10 +1357,6 @@ void setup()
 
 	setup_photpin();
 
-#if 0
-	servo_init();
```
leptun, line 1386: I'd also delete Servo.h/cpp. They are unusable on all boards that we support since most (and in some cases all) timers are used up for something else. That something else will surely break if Servos are attached. Better not have that old code anymore

Firmware/lcd.cpp
```diff
@@ -95,21 +95,6 @@ static const CustomCharacter Font[] PROGMEM = {
 #define CUSTOM_CHARACTERS_CNT (sizeof(Font) / sizeof(Font[0]))
 
 static void lcd_display(void);
-
-#if 0
```
leptun, line 99: since we are removing these, might as well remove the VT100 code as we don't need it anymore and takes tons of flash when enabled.
Still thinking what to do about the 8bit lcd code. I think I'll keep that one as it's not too intrusive.

leptun (approved):
I agree with the currently chosen cleanup. Please consider also the extra stuff I mentioned earlier.

Firmware/Marlin_main.cpp
gudnimg: Done 👍 Hopefully I didn't miss anything. [eb79694](https://github.com/prusa3d/Prusa-Firmware/pull/4739/commits/eb7969441a01dd51afa3a0b59d79a289cb76ecd9)

Firmware/lcd.cpp
gudnimg: Done 👍

gudnimg:
Ok, I'm gonna stop adding more stuff to this PR. This is probably enough for one PR :) It's so tempting to cleanup more.

PR #4741, opened 2024-08-03
PR 4739 follow-up (two things were missed)

gudnimg:
Follow up to https://github.com/prusa3d/Prusa-Firmware/pull/4739

I spotted two minor things I missed

1. `EEPROM_BABYSTEP_Z0` is no longer used anywhere (this was used in `G83` which is now removed)
2. `lcd_pick_babystep` function isn't defined anywhere or used. (this was used in `G85` which is now removed)

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -726,12 +726,6 @@ void softReset(void) {
     while(1);
 }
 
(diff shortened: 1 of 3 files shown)


leptun (approved):
Looks like old marlin code 👍

PR #4746, opened 2024-08-06
tests: test_run_all target must depend on tests executable

gudnimg:
Otherwise the tests and Catch2 isn't built


gudnimg:
The tests seem to run normally now. So this PR is open for review.

3d-gussner (approved):

PR #4748, opened 2024-08-08
planner/stepper: simplify `ultralcd.h` include into `lcd.h`

gudnimg:
Move `UserECoolEnabled()` from `ultralcd.cpp` into `eeprom.cpp` since this function has nothing to do with LCD specifically.
To avoid including `Prusa_farm.h` into `epprom.cpp`, simply replace `farm_mode` variable by reading the farm mode from EEPROM directly. This change allows us to remove `FarmOrUserECool()` function.

The goal is to avoid including `ultralcd.h` in general since it is a nightmare to unit test as is. Including `lcd.h` is less of a problem than `ultralcd.h` 

----

TODO:

- [ ] Unify `SILENT_MODE_*` and `TMC2130_MODE_*` preprocessor symbols. From EEPROM perspective these are the same.

Something like this:

```cpp
// On TMC2130 printers we have either Normal or Stealth mode.
// On non-TMC2130 printers we have Normal (Power), Stealth (Silent), and Auto mode.
enum StepperMode : uint8_t {
  NORMAL = 0,
  STEALTH = 1,
  AUTO = 2,
};
```

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -8319,10 +8319,8 @@ Sigma_Exit:
     }
     else if (code_seen('P'))
     {
-        uint8_t newMenuMode = (mcode_in_progress==914) ? SILENT_MODE_NORMAL : SILENT_MODE_STEALTH;
-
         // Update both EEPROM value and cached value in RAM
-        st_update_stepper_mode(newMenuMode);
+        st_update_stepper_mode(newMode);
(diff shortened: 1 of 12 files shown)


gudnimg:
I opened another can of worms. Marking as draft for now :) Need to review and do some further cleanup

gudnimg:
I will split this PR into two. The changes to the stealth mode caching need to be reviewed thoroughly. I will open a separate PR with only the changes to stealth mode when I find time.

Issue #3227, opened 2021-07-10
[BUG] building the firmware with SNMM enabled fails with multiple errors


gudnimg:
When I enable the SNMM and try to build the firmware it just fails. It seems to me the code just hasn't been updated in a while and now we're left with dead code.

Is there any reason to keep this code? Can we remove all SNMM code if it doesn't work anyways?

```
C:\Users\Gudni\OneDrive\Desktop\Prusa-Firmware\Firmware\mmu.cpp: In function 'void extr_adj(uint8_t)':
mmu.cpp:1045:2: error: 'lcdDrawUpdate' was not declared in this scope
  lcdDrawUpdate = 2;
  ^~~~~~~~~~~~~
C:\Users\Gudni\OneDrive\Desktop\Prusa-Firmware\Firmware\mmu.cpp:1045:2: note: suggested alternative: 'lcd_draw_update'
  lcdDrawUpdate = 2;
  ^~~~~~~~~~~~~
  lcd_draw_update
In file included from C:\Users\Gudni\OneDrive\Desktop\Prusa-Firmware\Firmware\Configuration.h:60:0,
                 from C:\Users\Gudni\OneDrive\Desktop\Prusa-Firmware\Firmware\Marlin_main.cpp:47:
C:\Users\Gudni\OneDrive\Desktop\Prusa-Firmware\Firmware\Marlin_main.cpp: In function 'void setup()':
Configuration_prusa.h:161:23: error: array must be initialized with a brace-enclosed initializer
 #define BOWDEN_LENGTH 433 //default total length for filament fast loading part; max length for extrusion is 465 mm!; this length can be adjusted in service menu
                       ^
C:\Users\Gudni\OneDrive\Desktop\Prusa-Firmware\Firmware\Marlin_main.cpp:1504:13: note: in expansion of macro 'BOWDEN_LENGTH'
    int _z = BOWDEN_LENGTH;
             ^~~~~~~~~~~~~
C:\Users\Gudni\OneDrive\Desktop\Prusa-Firmware\Firmware\ultralcd.cpp: In function 'void lcd_commands()':
ultralcd.cpp:919:16: error: 'count_e' was not declared in this scope
   float extr = count_e(0.2, width, length);
                ^~~~~~~
ultralcd.cpp:1169:24: error: cannot convert 'int' to 'LcdCommands' in assignment
    lcd_commands_type = 0;
                        ^
exit status 1
'lcdDrawUpdate' was not declared in this scope

```

[...]

PR #4749, opened 2024-08-09
Remove legacy Auto Bed Leveling code

gudnimg:
We discussed removing this code some time ago in 2021 (See https://github.com/prusa3d/Prusa-Firmware/issues/3227) and I want to propose again to remove auto bed leveling. The code has been unmaintained for many years and doesn't compile. This functionality has also not been tested / verified for many years.

The goal is to clean up the project so it's easier to maintain and further improve the project.


Firmware/Marlin_main.cpp
```diff
@@ -9321,12 +8618,7 @@ void clamp_to_software_endstops(float target[3])
 
     // Clamp the Z coordinate.
     if (min_software_endstops) {
-        float negative_z_offset = 0;
-        #ifdef ENABLE_AUTO_BED_LEVELING
-            if (Z_PROBE_OFFSET_FROM_EXTRUDER < 0) negative_z_offset = negative_z_offset + Z_PROBE_OFFSET_FROM_EXTRUDER;
-            if (cs.add_homing[Z_AXIS] < 0) negative_z_offset = negative_z_offset + cs.add_homing[Z_AXIS];
-        #endif
-        if (target[Z_AXIS] < min_pos[Z_AXIS]+negative_z_offset) target[Z_AXIS] = min_pos[Z_AXIS]+negative_z_offset;
+        if (target[Z_AXIS] < min_pos[Z_AXIS]) target[Z_AXIS] = min_pos[Z_AXIS];
```
gudnimg, line 8639: If you're wondering where 14 bytes of flash are saved. It is here, by removing `negative_z_offset` since it is always 0.

Firmware/Configuration.h
```diff
@@ -414,13 +303,6 @@ your extruder heater takes 2 minutes to hit the target on heating.
 
 // Custom M code points
 #define CUSTOM_M_CODES
```
leptun: can also be removed

Firmware/Marlin_main.cpp
```diff
@@ -3810,10 +3616,7 @@ extern uint8_t st_backlash_y;
 //!@n G11 - Retract recover
 //!@n G21 - Sets Units to Millimters
 //!@n G28 - Home all Axes one at a time
-//!@n G29 - Detailed Z-Probe
 //!@n G30 - Single Z Probe
```
leptun, line 3619: G30 is also deleted

leptun (approved):
Approved, but see some further cleanup I noted above.

Firmware/Marlin_main.cpp
gudnimg: There is still this one, not sure if it should be deleted.
image

Firmware/Configuration.h
gudnimg: Done

Firmware/Marlin_main.cpp
leptun: Oh didn't see this one. It looks to be enabled. As for how useful it is, no idea as I never had to use it. I would keep it as we are not that tight on resources.

PR #4750, opened 2024-08-09
Lower MIN_Z_FOR_UNLOAD

### LINKED ISSUE #4700: [BUG] Z axes is going up after Unload filament is selected
**Printer type** - [Any]
**Printer firmware version** -  [I think from the beginning]

Yet another Load/Unload issue. After press unload filament Z axess is going up +50.
* Harder to unload when using enclosure. No benefit with/without enclosure.
* After following failed unloads extruder reaches its Z limit and still trying to going up. Impossible to fix clogged filament in enclosure.
* When last print finished eg at Z=170 after pressing Unload then is completely failure, motoros trying to going up, very fard to unload filament in enclosure most probably resulting clog.

**To Reproduce**
Press "Unload filament" and wait until extruder is going +50

**Expected behavior**
After pressing Unload only setpoint heat should be applied and that's it. I see no reason to move extruder up to unload filament.

sarusani:
Lower the amount of Z lift on filament unload from 50mm to ~~10mm~~ 20mm.

This does not alter the behavior of M600. (M600 uses MIN_Z_FOR_SWAP)

Resolves #4700

### DIFF
--- a/Firmware/Configuration.h
+++ b/Firmware/Configuration.h
@@ -487,7 +487,7 @@ your extruder heater takes 2 minutes to hit the target on heating.
 // Try to maintain a minimum distance from the bed even when Z is
 // unknown when doing the following operations
 #define MIN_Z_FOR_LOAD    50 // lcd filament loading or autoload
-#define MIN_Z_FOR_UNLOAD  50 // lcd filament unloading
+#define MIN_Z_FOR_UNLOAD  20 // lcd filament unloading
 #define MIN_Z_FOR_SWAP    27 // filament change (including M600)
 #define MIN_Z_FOR_PREHEAT 10 // lcd preheat
 


gudnimg (approved):
This change looks good to me.

3d-gussner:
@sarusani @gudnimg Please see my comment https://github.com/prusa3d/Prusa-Firmware/issues/4700#issuecomment-2288313190

Please test the described issue as I couldn't reproduce it.

This PR changes the Unload height to +10mm after a start or reset, this is in my opinion too close to the bed.

sarusani:
I just assumed that what's he doing. You're right, it can only happen after a reset/reboot. I chose 10mm because that's also used for preheating, so I used the closest save distance. I personally use 25mm on my printer.

3d-gussner:
Looking at Buddy firmware they use 20mm https://github.com/search?q=repo%3Aprusa3d%2FPrusa-Firmware-Buddy%20Z_AXIS_UNLOAD_POS&type=code maybe we can "sync" that value.

sarusani:
Yes, I think that makes sense. 👍

3d-gussner (approved):

PR #4754, opened 2024-08-16
Add load filament option if autoload is disabled

### LINKED ISSUE #4752: [BUG] 3.14.0 When Autoload is turned off, inserting filament only gives you the "Unload" option, while you need "Load" in this case
This is due to the 3.14.0 UX improvement of combining Load and Unload in a single menu item.

When you have Autoload turned off the workflow to change filament is:
* press Unload filament (fine)
* pull out the filament
* insert the new filament
* press Load filament => Because you already inserted the filament you can only select "Unload filament" now

The workaround is to press Load filament before actually inserting the filament and then quickly insert it, which is non-optimal.

Note that I do want to keep the filament sensor enabled for filament run-out detection.

The easy fix would be to not combine the Load/Unload when Autoload is turned off (just as is the case when the filament sensor is disabled). I assume that people who turn off Autoload do this to gain more control, so would generally prefer to just have both explicit Load/Unload options. That is my preference in any case.

This is on a MK3s+ by the way.

sarusani:
~~Adds "Load filament" menu entry when autoload is disabled. Originally it was hidden as soon as you feed filament into the sensor. It is still hidden if autoload is enabled.~~

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -5363,6 +5363,7 @@ static void lcd_main_menu()
                             MENU_ITEM_SUBMENU_P(_T(MSG_AUTOLOAD_FILAMENT), lcd_menu_AutoLoadFilament);
                         } else {
                             MENU_ITEM_SUBMENU_P(_T(MSG_LOAD_FILAMENT), lcd_LoadFilament);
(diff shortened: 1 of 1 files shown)


sarusani:
Sorry, forgot how to use my brain 😄 See #4755

PR #4755, opened 2024-08-16
Add "load filament" if autoload is disabled

### LINKED ISSUE #4752: [BUG] 3.14.0 When Autoload is turned off, inserting filament only gives you the "Unload" option, while you need "Load" in this case

sarusani:
Adds "Load filament" menu entry when autoload is disabled. Originally it was hidden as soon as you feed filament into the sensor. It is still hidden if autoload is enabled.

Fixes https://github.com/prusa3d/Prusa-Firmware/issues/4752

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -5383,11 +5383,12 @@ static void lcd_main_menu()
             } else {
 #ifdef FILAMENT_SENSOR
                 if (fsensor.isEnabled()) {
+                    if (!fsensor.getAutoLoadEnabled()) {
+                        MENU_ITEM_SUBMENU_P(_T(MSG_LOAD_FILAMENT), lcd_LoadFilament);
+                    }
                     if (!fsensor.getFilamentPresent()) {
                         if (fsensor.getAutoLoadEnabled()) {
                             MENU_ITEM_SUBMENU_P(_T(MSG_AUTOLOAD_FILAMENT), lcd_menu_AutoLoadFilament);
-                        } else {
-                            MENU_ITEM_SUBMENU_P(_T(MSG_LOAD_FILAMENT), lcd_LoadFilament);
                         }
                     } else {
                         MENU_ITEM_SUBMENU_P(_T(MSG_UNLOAD_FILAMENT), lcd_unLoadFilament);


Firmware/ultralcd.cpp
```diff
@@ -5365,6 +5365,9 @@ static void lcd_main_menu()
                             MENU_ITEM_SUBMENU_P(_T(MSG_LOAD_FILAMENT), lcd_LoadFilament);
                         }
                     } else {
+                        if (!fsensor.getAutoLoadEnabled()) {
+                            MENU_ITEM_SUBMENU_P(_T(MSG_LOAD_FILAMENT), lcd_LoadFilament);
+                        }
```
gudnimg: I think this logic can be simplified to this:

```cpp
if (fsensor.isEnabled()) {

    if (!fsensor.getAutoLoadEnabled()) {
        MENU_ITEM_SUBMENU_P(_T(MSG_LOAD_FILAMENT), lcd_LoadFilament);
    }

    if (!fsensor.getFilamentPresent()) {
        if (fsensor.getAutoLoadEnabled()) {
            MENU_ITEM_SUBMENU_P(_T(MSG_AUTOLOAD_FILAMENT), lcd_menu_AutoLoadFilament);
        }
    } else {
        MENU_ITEM_SUBMENU_P(_T(MSG_UNLOAD_FILAMENT), lcd_unLoadFilament);
    }
}
```

Perhaps we can remove `lcd_menu_AutoLoadFilament`? That menu item doesn't really do anything. Then the logic could look like this:

```cpp
if (fsensor.isEnabled()) {

    if (!fsensor.getAutoLoadEnabled()) {
        MENU_ITEM_SUBMENU_P(_T(MSG_LOAD_FILAMENT), lcd_LoadFilament);
    }

    if (fsensor.getFilamentPresent()) {
        MENU_ITEM_SUBMENU_P(_T(MSG_UNLOAD_FILAMENT), lcd_unLoadFilament);
    }
}
```

gudnimg: @3d-gussner Is it possible for us to remove `lcd_menu_AutoLoadFilament`? I feel it should suffice for the user to open the filament sensor settings to view if autoloading is enabled or not. One motivation for removing this is we gain some resources for translations.

sarusani: I like the idea of removing it.
I added the removal as a separate commit, so you can choose if you want to merge it.

3d-gussner (approved):
Tested on MK404 and works correctly.

PR #4758, opened 2024-08-20
Abort nozzle change if filament is loaded

### LINKED ISSUE #4703: FIrmware should check filament sensor before nozzle change

sarusani:
Abort Nozzle change if filament sensor is enabled and filament is loaded.

Resolves #4703

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -4406,6 +4406,13 @@ static void sheets_menu()
 
 static void nozzle_change()
 {
+#ifdef FILAMENT_SENSOR
+    if (fsensor.isEnabled() && fsensor.getFilamentPresent()) {
+        lcd_show_fullscreen_message_and_wait_P(_T(MSG_UNLOAD_FILAMENT_REPEAT));
+        lcd_return_to_status();
+        return;
+    }
+#endif //FILAMENT_SENSOR
     lcd_commands_type = LcdCommands::NozzleCNG;
     lcd_return_to_status();
 }


sarusani:
Currently I check for both original and revo nozzles. Is unloading required for a revo nozzle change or should I only check for original?

Firmware/ultralcd.cpp
```diff
@@ -983,6 +983,15 @@ void lcd_commands()
                 lcd_commands_step = 3;
                 break;
             case 3:
+#ifdef FILAMENT_SENSOR
+                if (fsensor.isEnabled() && fsensor.getFilamentPresent()) {
+                    lcd_show_fullscreen_message_and_wait_P(_T(MSG_UNLOAD_FILAMENT_REPEAT));
+                    lcd_commands_step = 0;
+                    lcd_commands_type = LcdCommands::Idle;
+                    SetPrinterState(PrinterState::Idle);
```
gudnimg: Can we re-use case 1 to avoid duplicating code? It saves a little bit of flash memory.
```suggestion
                    lcd_commands_step = 1;
```

sarusani: I didn't use it because it exits to the hardware setup menu, so you would need to manually go up two menu levels to get to the unload option.

gudnimg: Fair enough. This isn't a blocker IMO.

I had another idea though I want to throw out here :)

Would it be simpler (code-wise) to check if the filament is present in the menu item instead?:

```cpp
static void nozzle_change()
{
    lcd_commands_type = LcdCommands::NozzleCNG;
    lcd_return_to_status();
}
```
Below is untested code:

```cpp
static void nozzle_change()
{
    if (fsensor.isEnabled() && fsensor.getFilamentPresent()) {
        lcd_show_fullscreen_message_and_wait_P(_T(MSG_UNLOAD_FILAMENT_REPEAT));
    } else {
        lcd_commands_type = LcdCommands::NozzleCNG;
    }
    lcd_return_to_status();
}
```

I don't remember if `lcd_show_fullscreen_message_and_wait_P` is still a problem within a menu item, but I think it will work today since there is no call to `lcd_update()`.

If this doesn't work, then please ignore my comment. It's just an idea.

sarusani: Changing it in the menu item works and saves some bytes. 👍

3d-gussner (approved):
Tested on MK404

- [X] MK3S without filament loaded, behaves the same as before
- [X] MK3S with filament loaded, shows Unload message and exits
- [X] MK3S + MMU3 without filament loaded, behaves the same as before
- [X] Mk3S + MMU3 with filament loaded, shows Unload message and exits

PR #4766, opened 2024-09-01
Remove sm4 line xyze UI

gudnimg:
Cleaning up my fork and found this branch. Changing the file from `.c` to `.cpp` was not building correctly in the Travis build, I never found out why but I suspect it was cache related. Now that we're using CMake it should run fine...

Creating this PR to check if CI passes

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -252,7 +252,7 @@ set(FW_SOURCES
     SdFatUtil.cpp
     SdFile.cpp
     SdVolume.cpp
-    sm4.c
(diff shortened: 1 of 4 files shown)


gudnimg:
Nice these changes now pass Github CI :) I guess we can review it then.

leptun (approved):
looks good 👍

sarusani:
👍 Builds work as expected in VSCode (CMake) windows.

3d-gussner (approved):
Works

PR #4767, opened 2024-09-01
optimisation: Simplify saving linearity correction

gudnimg:
Found this optimisation when cleaning up my fork. It looks like it's still relevant.

----

Keep in mind that the EEPROM addresses are in descending order. This means we can't use eeprom_read_block() or eeprom_update_block()

Address --> Axis
3831 --> X-axis
3830 --> Y-axis
3829 --> Z-axis
3828 --> E-axis

Change in memory:
Flash: -140 bytes
SRAM: 0 bytes

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -4495,20 +4495,17 @@ static void lcd_settings_menu()
 
 #ifdef TMC2130
 static void lcd_settings_linearity_correction_menu_save() {
-    bool changed = false;
     for (uint8_t axis = 0; axis < NUM_AXIS; axis++) {
-        // Constrain the value
-        if (tmc2130_wave_fac[axis] < TMC2130_WAVE_FAC1000_MIN) tmc2130_wave_fac[axis] = 0;
 
-        // Has the value changed?
-        changed |= (eeprom_read_byte((uint8_t*)EEPROM_TMC2130_WAVE_X_FAC - axis) != tmc2130_wave_fac[axis]);
+        if (tmc2130_wave_fac[axis] < TMC2130_WAVE_FAC1000_MIN) {
+            tmc2130_wave_fac[axis] = 0;
+        }
 
-        // If the value is changed, then write to EEPROM
         eeprom_update_byte_notify((uint8_t*)EEPROM_TMC2130_WAVE_X_FAC - axis, tmc2130_wave_fac[axis]);
     }
 
-    // If any of the values changed, then re-init the TMC2130 driver
-    if (changed) tmc2130_init(TMCInitParams(false, FarmOrUserECool()));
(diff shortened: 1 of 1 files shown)


Firmware/ultralcd.cpp
```diff
@@ -4494,28 +4494,20 @@ static void lcd_settings_menu()
...
+
+        // Has the value changed?
+        changed |= (eeprom_read_byte((uint8_t*)EEPROM_TMC2130_WAVE_X_FAC - axis) != tmc2130_wave_fac[axis]);
+
+        // If the value is changed, then write to EEPROM
+        eeprom_update_byte_notify((uint8_t*)EEPROM_TMC2130_WAVE_X_FAC - axis, tmc2130_wave_fac[axis]);
+    }
+
+    // If any of the values changed, then re-init the TMC2130 driver
     if (changed) tmc2130_init(TMCInitParams(false, FarmOrUserECool()));
```
leptun: Personally I don't see the point of only reinitializing the drivers if something changed. The reinit is quick and the printer is expected to be idle anyway. Not to mention eeprom_update_byte_notify _also_ does a read and compare for the notify. I would just get rid of the "changed" stuff and only call eeprom_update_byte_notify directly. Also the function is called only on menu exit in an obscure menu, I wouldn't care too much about how "inefficient" it is to save (it isn't).

gudnimg: Done :) saves 22 bytes of flash

gudnimg: I also checked if replacing `tmc2130_init` with `tmc2130_set_wave` saves any memory... unfortunately that's not the case.

leptun (approved):

3d-gussner (approved):
Tested with MK404 and behaves the same as before :+1: 
With `DEBUG_EEPROM_CHANGES` enabled

Test: `Settings -> Lin. correction` and change from XYZE all 0 to X 34 Y 33 Z 32 E 31

before
```
[2024-09-02 10:55:31.951287] RECV EEPROMChng b  3831 255 -> 34
[2024-09-02 10:55:31.953469] RECV EEPROMChng b  3830 255 -> 33
[2024-09-02 10:55:31.955096] RECV EEPROMChng b  3829 255 -> 32
[2024-09-02 10:55:31.956794] RECV EEPROMChng b  3828 255 -> 31
```
after factory reset and with this PR
```
[2024-09-02 11:09:35.037897] RECV EEPROMChng b  3831 255 -> 34
[2024-09-02 11:09:35.039573] RECV EEPROMChng b  3830 255 -> 33
[2024-09-02 11:09:35.041767] RECV EEPROMChng b  3829 255 -> 32
[2024-09-02 11:09:35.043484] RECV EEPROMChng b  3828 255 -> 31
```

PR #4771, opened 2024-09-14
PFW-1561 Fix MMU unload issue

gudnimg:
Steps to reproduce:

1. Check Settings -> Fan check [On]
2. Start MMU print
3. While printing stall the hotend fan until the printer has finished pausing the print job.
4. You can stop halting the fan now.
5. Now, sit back and wait until the extruder temperature is below `EXTRUDER_AUTO_FAN_TEMPERATURE` (50°C). This can take ~10 minutes.
6. Stop the print job via LCD menu

**Expected result:** MMU should **NOT** re-heat the nozzle and unload the filament

### DIFF
--- a/Firmware/ultralcd.cpp
+++ b/Firmware/ultralcd.cpp
@@ -5676,7 +5676,7 @@ void lcd_print_stop_finish()
 
         if (MMU2::mmu2.Enabled() && MMU2::mmu2.FindaDetectsFilament()
 #ifdef FANCHECK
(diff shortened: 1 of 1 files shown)


gudnimg:
@3d-gussner Good news, I can confirm this fixes the issue.

I could reproduce the issue on current MK3 branch (5aa67489dab067faa15a9fefd66314144d302367) and with this pull request the MMU no longer unloads when a print is aborted due to fan error, when the temperature is below 50°C.

PR #4805, opened 2024-10-29
Remove hardcoded MMU tool count

3d-gussner:
1. Remove hard coded tool values https://github.com/prusa3d/Prusa-Firmware/pull/4805/commits/cdb6635eac3b061276e2c722a7c9a3dd5d62a1fd
  1.1. Preload to MMU -> All
  1.2. Loading Test -> All
  1.3.  SpoolJoin

### DIFF
--- a/Firmware/SpoolJoin.cpp
+++ b/Firmware/SpoolJoin.cpp
@@ -55,7 +55,7 @@ uint8_t SpoolJoin::nextSlot()
     SERIAL_ECHOPGM("SpoolJoin: ");
     SERIAL_ECHO((int)currentMMUSlot);
 
-    if (currentMMUSlot >= 4) currentMMUSlot = 0;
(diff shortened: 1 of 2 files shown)


Firmware/mmu2.cpp
```diff
@@ -482,10 +482,6 @@ void MMU2::UnloadInner() {
         IncrementMMUFails();
     }
     MakeSound(Confirm);
-
-    // no active tool
-    SetCurrentTool(MMU2_NO_TOOL);
-    tool_change_extruder = MMU2_NO_TOOL;
 }
```
gudnimg: What exactly does this change fix? Steps to reproduce would help describe it. I am worried this change introduces bugs. If this is indeed a bug, then it's likely its on the MK4 side too.

3d-gussner: Added the steps to test it.

gudnimg (approved):
Looks good to me

Issue #4572, opened 2024-01-17
[ENHANCEMENT] MK3s update First Layer Calibration to run from front to back


handro123:
**Enter what type of printer or upgrade the enhancement applies to.**
MK3s+ with E3d Revo Six
MMU3 UltiMulti

**Is your enhancement related to a problem? Please describe.**
I am a long time Prusa Mini owner and recently got a used MK3s+ with MMU. On the Mini when doing First Layer Calibration it starts at the front and snakes to the back where the square is made in the rear left corner. This is good when running First Layer Calibration as you can easily see the filament, feel with your finger (carefully) and make adjustments as needed. 

With the MK3s+ it starts the First Layer Calibration in the rear and moves forward. You can't easily see the previous rows of extrusion as you have to reach around the X-axis and Extruder. 

  
**Describe the solution you'd like**
Update the MK3s+ First Layer Calibration to match the Prusa Mini / Buddy firmware, where it starts at the front and finishes at back of the print bed.

3d-gussner:
@handro123 Thanks for the suggestion. I can't promise we will change it but I will discuss this internally.

pedjas:
It may be useful if user can rotate calibration test in 90 degree steps.

3d-gussner:
@handro123 Implemented new First layer cal in #4811. If you want to compile the firmware yourself please un-comment `#define NEW_FIRST_LAYER_CAL`  in the variant you need.

Please consider to close this issue.

@pedjas Thanks for the idea. But adding rotation would increase the code size quite a bit, especially when you want to add some LCD menu.

3d-gussner:
Closing as it has been implemented for custom builds.

PR #4811, opened 2024-11-14
Optimze and new first layer calibration

3d-gussner:
- Reduced buffer and flash size by
  - changing to relative mode during the first layer calibration pattern
  - "compressing" the Gcode line by removing spaces 
- added NEW_FIRST_LAYER_CAL to perform 1st layer cal from front to rear as requested in https://github.com/prusa3d/Prusa-Firmware/issues/4572

By default the previous known first layer calibration pattern is used. You need to un-comment `//#define NEW_FIRST_LAYER_CAL` in the variant file to use the new pattern from front to rear.

The new first layer calibration pattern from front to rear with square ends in the middle of the bed.
- The calibration of the first layer at the center is most important
- The left, right, front rear sides can manually corrected via LCD -> Calibration -> Bed level correct menu
- The first layer calibration pattern continues directly after the purge to keep the pressure stable in the nozzle

New_first_layer_cal

### DIFF
--- a/Firmware/first_lay_cal.cpp
+++ b/Firmware/first_lay_cal.cpp
@@ -184,19 +184,16 @@ void lay1cal_meander_start(float layer_height, float extrusion_width)
     static const char fmt1[] PROGMEM = "G1Z%g";
     enquecommandf_P(fmt1, layer_height);
     enquecommand_P(feedrate_F1080);
+    enquecommand_P(MSG_G91); //enable relative XYZ
 #ifdef NEW_FIRST_LAYER_CAL
     enquecommandf_P(extrude_fmt_Y, short_length, count_e(layer_height, extrusion_width, short_length));
-#endif //_NEW_FIRST_LAYER_CAL
-
-    enquecommand_P(MSG_G91); //enable relative XYZ
-#ifndef NEW_FIRST_LAYER_CAL
+    enquecommandf_P(extrude_fmt_X, long_length*invert, count_e(layer_height, extrusion_width, long_length));
+    enquecommandf_P(extrude_fmt_Y, -short_length*invert, count_e(layer_height, extrusion_width, short_length));
+#else
     enquecommandf_P(extrude_fmt_X, 25.f*invert, count_e(layer_height, extrusion_width * 4.f, 25));
     enquecommandf_P(extrude_fmt_X, 25.f*invert, count_e(layer_height, extrusion_width * 2.f, 25));
     enquecommandf_P(extrude_fmt_X, 100.f*invert, count_e(layer_height, extrusion_width, 100));
     enquecommandf_P(extrude_fmt_Y, -20.f*invert, count_e(layer_height, extrusion_width, 20));
-#else
-    enquecommandf_P(extrude_fmt_X, long_length*invert, count_e(layer_height, extrusion_width, long_length));
-    enquecommandf_P(extrude_fmt_Y, -short_length*invert, count_e(layer_height, extrusion_width, short_length));
 #endif //_NEW_FIRST_LAYER_CAL
 }
 
--- a/Firmware/first_lay_cal.cpp
+++ b/Firmware/first_lay_cal.cpp
@@ -224,17 +224,16 @@ void lay1cal_meander(float layer_height, float extrusion_width)
 
 //! @brief Print square
 //!
-//! This function needs to be called 4 times with step of 0,4,8,12
+//! This function enqueues 4 lines of the square, so it needs to be called multiple times
 //!
 //! @param cmd_buffer character buffer needed to format gcodes
-//! @param i iteration
-void lay1cal_square(uint8_t step, float layer_height, float extrusion_width)
+void lay1cal_square(float layer_height, float extrusion_width)
 {
     const float Y_spacing = spacing(layer_height, extrusion_width);
     const float long_extrusion = count_e(layer_height, extrusion_width, square_width);
     const float short_extrusion = count_e(layer_height, extrusion_width, Y_spacing);
 
-    for (uint8_t i = step; i < step+4; ++i)
+    for (uint8_t i = 0; i < 4; i++)
     {
         enquecommandf_P(extrude_fmt_X, square_width*invert, long_extrusion);
         enquecommandf_P(extrude_fmt_Y, -Y_spacing*invert, short_extrusion);
--- a/Firmware/first_lay_cal.cpp
+++ b/Firmware/first_lay_cal.cpp
@@ -188,7 +188,7 @@ void lay1cal_meander_start(float layer_height, float extrusion_width)
     enquecommandf_P(extrude_fmt_Y, short_length, count_e(layer_height, extrusion_width, short_length));
 #endif //_NEW_FIRST_LAYER_CAL
 
-    enquecommand_P(PSTR("G91"));
+    enquecommand_P(MSG_G91); //enable relative XYZ
 #ifndef NEW_FIRST_LAYER_CAL
     enquecommandf_P(extrude_fmt_X, (float)25*invert, count_e(layer_height, extrusion_width * 4.f, 25));
     enquecommandf_P(extrude_fmt_X, (float)25*invert, count_e(layer_height, extrusion_width * 2.f, 25));
--- a/Firmware/first_lay_cal.cpp
+++ b/Firmware/first_lay_cal.cpp
@@ -51,6 +51,7 @@ static void lay1cal_common_enqueue_loop(const char * const * cmd_sequence, const
 static const char extrude_fmt_X[] PROGMEM = "G1X%gE%g";
 static const char extrude_fmt_Y[] PROGMEM = "G1Y%gE%g";
 static const char zero_extrusion[] PROGMEM = "G92E0";
+static const char feedrate_F1080[] PROGMEM = "G1F1080";
 #ifndef NEW_FIRST_LAYER_CAL
 int8_t invert = 1;
 const float short_length = 20;
@@ -120,7 +121,7 @@ void lay1cal_intro_line(bool extraPurgeNeeded, float layer_height, float extrusi
     static const char cmd_intro_mmu_8[] PROGMEM = "G1X240E25F2200";
     static const char cmd_intro_mmu_9[] PROGMEM = "G1Y-2F1000";
     static const char cmd_intro_mmu_10[] PROGMEM = "G1X200E8F1400";
-    static const char cmd_intro_mmu_11[] PROGMEM = "G1Z0.2F1000";
+    static const char cmd_intro_mmu_11[] PROGMEM = "G1Z0.2";
     static const char * const cmd_intro_mmu[] PROGMEM =
     {
         // first 2 items are only relevant if filament was not loaded - i.e. extraPurgeNeeded == true
@@ -144,6 +145,7 @@ void lay1cal_intro_line(bool extraPurgeNeeded, float layer_height, float extrusi
     }
     else
     {
+        enquecommand_P(feedrate_F1080); //fixed velocity for the intro line
         enquecommandf_P(extrude_fmt_X, (float)60, count_e(layer_height, extrusion_width * 4.f, 60));
         enquecommandf_P(extrude_fmt_X, (float)202.5, count_e(layer_height, extrusion_width * 8.f, 142.5));
     }
@@ -179,9 +181,9 @@ void lay1cal_meander_start(float layer_height, float extrusion_width)
 #ifndef NEW_FIRST_LAYER_CAL
     enquecommand_P(PSTR("G1X50Y155"));
 #endif //_NEW_FIRST_LAYER_CAL
-    static const char fmt1[] PROGMEM = "G1Z%gF7200";
+    static const char fmt1[] PROGMEM = "G1Z%g";
     enquecommandf_P(fmt1, layer_height);
-    enquecommand_P(PSTR("G1F1080"));
+    enquecommand_P(feedrate_F1080);
 #ifdef NEW_FIRST_LAYER_CAL
     enquecommandf_P(extrude_fmt_Y, short_length, count_e(layer_height, extrusion_width, short_length));
 #endif //_NEW_FIRST_LAYER_CAL
--- a/Firmware/first_lay_cal.cpp
+++ b/Firmware/first_lay_cal.cpp
@@ -146,8 +146,8 @@ void lay1cal_intro_line(bool extraPurgeNeeded, float layer_height, float extrusi
     else
     {
         enquecommand_P(feedrate_F1080); //fixed velocity for the intro line
-        enquecommandf_P(extrude_fmt_X, (float)60, count_e(layer_height, extrusion_width * 4.f, 60));
-        enquecommandf_P(extrude_fmt_X, (float)202.5, count_e(layer_height, extrusion_width * 8.f, 142.5));
+        enquecommandf_P(extrude_fmt_X, 60.f, count_e(layer_height, extrusion_width * 4.f, 60));
+        enquecommandf_P(extrude_fmt_X, 202.5f, count_e(layer_height, extrusion_width * 8.f, 142.5));
     }
 }
 
@@ -190,10 +190,10 @@ void lay1cal_meander_start(float layer_height, float extrusion_width)
 
     enquecommand_P(MSG_G91); //enable relative XYZ
 #ifndef NEW_FIRST_LAYER_CAL
-    enquecommandf_P(extrude_fmt_X, (float)25*invert, count_e(layer_height, extrusion_width * 4.f, 25));
-    enquecommandf_P(extrude_fmt_X, (float)25*invert, count_e(layer_height, extrusion_width * 2.f, 25));
-    enquecommandf_P(extrude_fmt_X, (float)100*invert, count_e(layer_height, extrusion_width, 100));
-    enquecommandf_P(extrude_fmt_Y, (float)-20*invert, count_e(layer_height, extrusion_width, 20));
+    enquecommandf_P(extrude_fmt_X, 25.f*invert, count_e(layer_height, extrusion_width * 4.f, 25));
+    enquecommandf_P(extrude_fmt_X, 25.f*invert, count_e(layer_height, extrusion_width * 2.f, 25));
+    enquecommandf_P(extrude_fmt_X, 100.f*invert, count_e(layer_height, extrusion_width, 100));
+    enquecommandf_P(extrude_fmt_Y, -20.f*invert, count_e(layer_height, extrusion_width, 20));
 #else
     enquecommandf_P(extrude_fmt_X, long_length*invert, count_e(layer_height, extrusion_width, long_length));
     enquecommandf_P(extrude_fmt_Y, -short_length*invert, count_e(layer_height, extrusion_width, short_length));
--- a/Firmware/first_lay_cal.cpp
+++ b/Firmware/first_lay_cal.cpp
@@ -210,7 +210,9 @@ void lay1cal_meander(float layer_height, float extrusion_width)
         enquecommandf_P(extrude_fmt_Y, invert * -short_length, short_extrusion);
     }
 #ifdef NEW_FIRST_LAYER_CAL
-    enquecommandf_P(extrude_fmt_X, -(long_length/2-square_width/2), long_extrusion); //~Middle of bed X125
+    constexpr float mid_length = 0.5f * long_length - 0.5f * square_width;
+    const float mid_extrusion = count_e(layer_height, extrusion_width, mid_length);
+    enquecommandf_P(extrude_fmt_X, -mid_length, mid_extrusion); //~Middle of bed X125
     enquecommandf_P(extrude_fmt_Y, short_length, short_extrusion); //~Middle of bed Y105
 #endif //NEW_FIRST_LAYER_CAL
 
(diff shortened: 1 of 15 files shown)


3d-gussner:
First build

 All values in bytes. Δ Delta to base
| Target| ΔFlash | ΔSRAM | Used Flash | Used SRAM | Free Flash | Free SRAM |
| ------| ------ | ----- | -----------| --------- | ---------- | --------- |
| `MK3S_MULTILANG` | -668 | 0 | 247634 | 5654 | 6318 | 2538 |
| `MK3_MULTILANG` | -662 | 0 | 247010 | 5663 | 6942 | 2529 |

 That is more than expected. Running a rebase.

Still more than I expected, but I compiled mostly the new first layer cal pattern version
All values in bytes. Δ Delta to base

| Target| ΔFlash | ΔSRAM | Used Flash | Used SRAM | Free Flash | Free SRAM |
| ------| ------ | ----- | -----------| --------- | ---------- | --------- |
| `MK3S_MULTILANG` | -412 | 0 | 247890 | 5654 | 6062 | 2538 |
| `MK3_MULTILANG` | -412 | 0 | 247260 | 5663 | 6692 | 2529 |

Firmware/first_lay_cal.cpp
```diff
@@ -210,41 +224,37 @@ void lay1cal_meander(float layer_height, float extrusion_width)
...
-            cmd_cal_finish_4,
-            cmd_cal_finish_5
+            cmd_cal_finish_4
     };
 
+    enquecommand_P(MSG_G90);
     lay1cal_common_enqueue_loop(cmd_cal_finish, (sizeof(cmd_cal_finish)/sizeof(cmd_cal_finish[0])));
 
     if (mmu_enabled) enquecommand_P(MSG_M702); //unload from nozzle
+    disable_heater();
```
gudnimg: Note that MMU M702 needs the heater to be on to complete the unload. The heaters are already turned off above with  `M140 S0` (See `cmd_cal_finish_2`) and `M104 S0` (See `cmd_cal_finish_3`)

We can then also avoid including `#include "temperature.h"`

Now that I look at the code, the `M702` g-code should actually come before `M104 S0`

Firmware/first_lay_cal.cpp
```diff
@@ -96,65 +107,61 @@ bool lay1cal_load_filament(uint8_t filament)
...
 {
-    static const char cmd_intro_mmu_3[] PROGMEM = "G1 X55 E29 F1073";
-    static const char cmd_intro_mmu_4[] PROGMEM = "G1 X5 E29 F1800";
-    static const char cmd_intro_mmu_5[] PROGMEM = "G1 X55 E8 F2000";
-    static const char cmd_intro_mmu_6[] PROGMEM = "G1 Z0.3 F1000";
-    static const char cmd_intro_mmu_8[] PROGMEM = "G1 X240 E25  F2200";
-    static const char cmd_intro_mmu_9[] PROGMEM = "G1 Y-2 F1000";
-    static const char cmd_intro_mmu_10[] PROGMEM = "G1 X55 E25 F1400";
-    static const char cmd_intro_mmu_11[] PROGMEM = "G1 Z0.2 F1000";
-    static const char cmd_intro_mmu_12[] PROGMEM = "G1 X5 E4 F1000";
```
gudnimg, line 107: Is `cmd_intro_mmu_12` removal intentional? Also `cmd_intro_mmu_10` is changed, why?

Firmware/first_lay_cal.cpp
```diff
@@ -96,65 +107,61 @@ bool lay1cal_load_filament(uint8_t filament)
...
+            enquecommand_P(static_cast<char*>(pgm_read_ptr(&cmd_intro_mmu[i])));
         }
     }
     else
     {
-        static const char fmt1[] PROGMEM = "G1 X%d E%-.3f F1000";
-        enquecommandf_P(fmt1, 60, count_e(layer_height, extrusion_width * 4.f, 60));
-        enquecommandf_P(fmt1, 100, count_e(layer_height, extrusion_width * 8.f, 40));
+        enquecommandf_P(extrude_fmt_X, (float)60, count_e(layer_height, extrusion_width * 4.f, 60));
+        enquecommandf_P(extrude_fmt_X, (float)202.5, count_e(layer_height, extrusion_width * 8.f, 142.5));
```
gudnimg: This new format string removed the feedrate settings, previously it was 1000 mm/min or 16.67 mm/s

This previous g-code is setting 1000 mm/min so in this case it's OK. But if that string is changed, then this g-code will also change with it.

Firmware/first_lay_cal.cpp
3d-gussner: Yes both with and without MMU purge line end at ~same position to continue with the new first layer cal pattern directly after the purge to keep the nozzle pressure up.

Firmware/first_lay_cal.cpp
3d-gussner: If no feedrate is provided the last set feedrate is used. Saves few bytes and the printed lines look the same as before.

Firmware/first_lay_cal.cpp
3d-gussner: Changing to `disable_heater();` saves few additional bytes.

Firmware/first_lay_cal.cpp
leptun: I'm not sure if reducing the MMU purge line will have any unintended side effects. It could be a problem is switching between material types, but I suppose the user should pay attention when doing that anyway. Color bleed isn't really of significance on the first layer calibration, so I think this should be ok. Still I propose testing play around with this with material changes with the reduced purge.

Firmware/first_lay_cal.cpp
leptun: I agree with @gudnimg that we should probably set the feedrate to something known here once before the non-mmu purge line. No clue if something before it sets the feedrate to 1000. And yea, not specifying the F every command is fine, but we should still set the F before these two lines just for consistency at the expense of ~30B of flash.
From what I see quickly looking through the code, nothing before this sets the feedrate, so these two lines would run at whatever feedrate was set before this, which could be reasonable (default feedrate) or something ridiculously high if something else executed before it. Please enqueue a `G1F<something>` before enqueueing any other commands in the non-mmu intro.

Firmware/first_lay_cal.cpp
```diff
@@ -96,65 +107,61 @@ bool lay1cal_load_filament(uint8_t filament)
...
             zero_extrusion,
-            cmd_pre_meander_2,
+            MSG_G90,
             MSG_M83, // use relative distances for extrusion
+#ifndef NEW_FIRST_LAYER_CAL
             cmd_pre_meander_4,
             cmd_pre_meander_5,
+#endif //NEW_FIRST_LAYER_CAL
             cmd_pre_meander_6,
-            cmd_pre_meander_7,
```
leptun, line 157: The travel move will be at F7200, but I agree with this change. I always felt the travel is too slow for the first layer cal, too much oozing can happen.

Firmware/first_lay_cal.cpp
```diff
@@ -163,42 +170,49 @@ void lay1cal_before_meander()
...
-    enquecommandf_P(extrude_fmt,  75, 155, count_e(layer_height, extrusion_width * 4.f, 25));
-    enquecommandf_P(extrude_fmt, 100, 155, count_e(layer_height, extrusion_width * 2.f, 25));
-    enquecommandf_P(extrude_fmt, 200, 155, count_e(layer_height, extrusion_width, 100));
-    enquecommandf_P(extrude_fmt, 200, 135, count_e(layer_height, extrusion_width, 20));
+    enquecommand_P(PSTR("G1F1080"));
+#ifdef NEW_FIRST_LAYER_CAL
+    enquecommandf_P(extrude_fmt_Y, short_length, count_e(layer_height, extrusion_width, short_length));
+#endif //_NEW_FIRST_LAYER_CAL
+
+    enquecommand_P(PSTR("G91"));
```
leptun: maybe extract this to messages.h/cpp? - done

Firmware/first_lay_cal.cpp
```diff
@@ -163,42 +170,49 @@ void lay1cal_before_meander()
...
+    static const char fmt1[] PROGMEM = "G1Z%gF7200";
     enquecommandf_P(fmt1, layer_height);
-
-    enquecommand_P(PSTR("G1 F1080"));
-
-    enquecommandf_P(extrude_fmt,  75, 155, count_e(layer_height, extrusion_width * 4.f, 25));
-    enquecommandf_P(extrude_fmt, 100, 155, count_e(layer_height, extrusion_width * 2.f, 25));
-    enquecommandf_P(extrude_fmt, 200, 155, count_e(layer_height, extrusion_width, 100));
-    enquecommandf_P(extrude_fmt, 200, 135, count_e(layer_height, extrusion_width, 20));
+    enquecommand_P(PSTR("G1F1080"));
```
leptun: interesting choice for F. Could we maybe have a global G1F1000 that we can reuse here? Saves a few more bytes of flash and addresses the issue I mentioned earlier.

Firmware/first_lay_cal.cpp
leptun: Also, shouldn't this line be above the first meander move for the new calibration? This feels slightly misplaced.

Firmware/first_lay_cal.cpp
leptun: I agree with @gudnimg. We should not disable the heaters before the M702. Moreover, directly calling disable_heater bypasses the cmdqueue, so stuff ends up executing out of order. Please switch back to M104 and M140, but after the M702 stuff. I know it will add back some bytes, but I would rather have that than mix and match cmdqueue and non-cmdqueue actions in the automation and cause issues with the M702 unload.

~~Also grrrrrr the MK3SMMU3 prusaslicer profiles also do this out of order :/~~ I can't read. Ignore that :>

Firmware/first_lay_cal.cpp
```diff
@@ -210,41 +224,37 @@ void lay1cal_meander(float layer_height, float extrusion_width)
 //! @param i iteration
 void lay1cal_square(uint8_t step, float layer_height, float extrusion_width)
```
leptun: step parameter is now useless. Please remove it and fix the for loop to 4 cycles. Update comment above also and save some more flash by reducing parameter count :>

leptun (changes requested):

Firmware/first_lay_cal.cpp
leptun: I'd like to see the changes from #4813 in this PR to address this issue.

gudnimg: I’ll prepare a PR on @3d-gussner ‘s fork soon

Firmware/first_lay_cal.cpp
leptun: done, saved 74B of flash

Firmware/first_lay_cal.cpp
leptun: Merged. Should be resolved now. @3d-gussner please check this in the sim.

Firmware/first_lay_cal.cpp
leptun: Pushed, please double check.

Firmware/first_lay_cal.cpp
leptun: Changed to F1080 in multiple places and reused the same command in progmem

Firmware/ultralcd.cpp
```diff
@@ -857,16 +857,16 @@ void lcd_commands()
...
-                lay1cal_square(1, layer_height, extrusion_width);
+                lay1cal_square(layer_height, extrusion_width);
                 break;
             case 4:
-                lay1cal_square(2, layer_height, extrusion_width);
+                lay1cal_square(layer_height, extrusion_width);
                 break;
             case 3:
-                lay1cal_square(3, layer_height, extrusion_width);
+                lay1cal_square(layer_height, extrusion_width);
```
gudnimg: Surely this can be optimised now :)? `layer_height` and `extrusion_width` are defined as constants.

Prehaps:


```c
case 6:
case 5:
case 4:
case 3:
    lay1cal_square(layer_height, extrusion_width);
    break;
```

Firmware/first_lay_cal.cpp
```diff
@@ -37,12 +37,30 @@ static constexpr float spacing(float layer_height, float extrusion_width, float
...
-static const char zero_extrusion[] PROGMEM = "G92 E0";
+static const char extrude_fmt_X[] PROGMEM = "G1X%gE%g";
+static const char extrude_fmt_Y[] PROGMEM = "G1Y%gE%g";
+static const char zero_extrusion[] PROGMEM = "G92E0";
+static const char feedrate_F1080[] PROGMEM = "G1F1080";
+#ifndef NEW_FIRST_LAYER_CAL
+int8_t invert = 1;
+const float short_length = 20;
+#else
+int8_t invert = -1;
```
gudnimg: Is there any point having the same variable with the same value here regardless if `NEW_FIRST_LAYER_CAL` is defined or not.

Can `int8_t invert = 1;` be moved out of the `ifdef`? It doesn't save any memory, but its shorter code.

Firmware/first_lay_cal.cpp
gudnimg: Another minor nitpick.

You can change:

`(float)60` to `60.f`
`(float)202.5` to` 202.5f`

Slightly shorter code.

Firmware/first_lay_cal.cpp
```diff
@@ -163,88 +178,93 @@ void lay1cal_before_meander()
...
+#ifdef NEW_FIRST_LAYER_CAL
+    enquecommandf_P(extrude_fmt_Y, short_length, count_e(layer_height, extrusion_width, short_length));
+#endif //_NEW_FIRST_LAYER_CAL
+
+    enquecommand_P(MSG_G91); //enable relative XYZ
+#ifndef NEW_FIRST_LAYER_CAL
+    enquecommandf_P(extrude_fmt_X, (float)25*invert, count_e(layer_height, extrusion_width * 4.f, 25));
+    enquecommandf_P(extrude_fmt_X, (float)25*invert, count_e(layer_height, extrusion_width * 2.f, 25));
+    enquecommandf_P(extrude_fmt_X, (float)100*invert, count_e(layer_height, extrusion_width, 100));
+    enquecommandf_P(extrude_fmt_Y, (float)-20*invert, count_e(layer_height, extrusion_width, 20));
```
gudnimg: I think dropping the float cast would be nice:

`(float)25*invert` can be change to `invert * 25.f`

Not critical, just nitpicking :)

Firmware/first_lay_cal.cpp
```diff
@@ -163,88 +178,93 @@ void lay1cal_before_meander()
...
+        enquecommandf_P(extrude_fmt_X, -x_pos*invert, long_extrusion);
 
-        x_pos += long_length;
+        x_pos = -x_pos;
 
-        long_length = -long_length;
+        enquecommandf_P(extrude_fmt_Y, -short_length*invert, short_extrusion);
     }
+#ifdef NEW_FIRST_LAYER_CAL
+    enquecommandf_P(extrude_fmt_X, -(long_length/2-square_width/2), long_extrusion); //~Middle of bed X125
```
gudnimg: Consider using multiplication instead:

`long_length/2` -> `long_length * 0.5f`

It may be the compiler is smart enough to do this change automatically, but lets be explicit. Multiplication is more efficient than division.

Firmware/ultralcd.cpp
leptun: I think the plan is to have it account for the nozzle size. Layer height is indeed always 0.2mm but the extrusion width is changed based on the hardware setup.

Firmware/first_lay_cal.cpp
```diff
@@ -163,88 +178,93 @@ void lay1cal_before_meander()
...
+#endif //_NEW_FIRST_LAYER_CAL
+
+    enquecommand_P(MSG_G91); //enable relative XYZ
+#ifndef NEW_FIRST_LAYER_CAL
+    enquecommandf_P(extrude_fmt_X, (float)25*invert, count_e(layer_height, extrusion_width * 4.f, 25));
+    enquecommandf_P(extrude_fmt_X, (float)25*invert, count_e(layer_height, extrusion_width * 2.f, 25));
+    enquecommandf_P(extrude_fmt_X, (float)100*invert, count_e(layer_height, extrusion_width, 100));
+    enquecommandf_P(extrude_fmt_Y, (float)-20*invert, count_e(layer_height, extrusion_width, 20));
+#else
+    enquecommandf_P(extrude_fmt_X, long_length*invert, count_e(layer_height, extrusion_width, long_length));
```
leptun, line 198: this line is overextruded since a shorter length is used with the same E count. Needs adjusting

image

Firmware/first_lay_cal.cpp
leptun: Woops, commented on the wrong thing.

This line is overextruded since a shorter length is used with the same E count. Needs adjusting

image

Firmware/first_lay_cal.cpp
3d-gussner: The values are `-1`  vs `1` to invert the direction

Firmware/first_lay_cal.cpp
3d-gussner: @leptun Good point. the extrusion needs to be 0.5 of the long_length

Firmware/first_lay_cal.cpp
gudnimg: Sorry I thought I saw both as -1 😅 false alarm.

Firmware/ultralcd.cpp
gudnimg: Indeed. But we can assume the nozzle diameter wont change during first layer calibration.

3d-gussner: No it doesn't change.

gudnimg: Made a PR: https://github.com/3d-gussner/Prusa-Firmware/pull/97

Firmware/first_lay_cal.cpp
```diff
@@ -37,12 +37,30 @@ static constexpr float spacing(float layer_height, float extrusion_width, float
...
+            continue;
+        }
+
+        enquecommand_P(static_cast<char*>(pgm_ptr));
     }
 }
 
-static const char extrude_fmt[] PROGMEM = "G1 X%d Y%d E%-.5f";
-static const char zero_extrusion[] PROGMEM = "G92 E0";
+static const char extrude_fmt_X[] PROGMEM = "G1X%gE%g";
```
gudnimg: Does G1 support scientific notation? `g` means `Use the shortest representation: %e or %f`. Is something like 3.9265e+2 parsed correctly?

leptun: nope. If we end up scientific notation then this will fail for sure. Maybe just a simple %f might be better suited. That doesn't pad the end with zeros, no?

Firmware/ultralcd.cpp
leptun: merged

Firmware/first_lay_cal.cpp
leptun: Are you sure? Feels to me like it would be better but still not 100% correct as that would still slightly overextrude. Remember that the line is (half of long_length - half of the square width). So just setting it to 0.5x will not account for the square size

Firmware/first_lay_cal.cpp
3d-gussner: I don't think that the avr will produce any `e+2` and also never seen anything longer than 7 chars `"G1Y23.6111E0.74031"`

leptun: Can you try to use %f and see if it generates the same commands? I think it will and will definitely be safer to use as it should always work.

Firmware/first_lay_cal.cpp
leptun: > Consider using multiplication instead:
> 
> `long_length/2` -> `long_length * 0.5f`
> 
> It may be the compiler is smart enough to do this change automatically, but lets be explicit. Multiplication is more efficient than division.

All divisions are done inside constexpr or on constant compile time computed variables, so I am not worried about these.

Firmware/first_lay_cal.cpp
leptun: @gudnimg fixed nitpick

Firmware/first_lay_cal.cpp
leptun: fixed

Firmware/first_lay_cal.cpp
leptun: Fixed @3d-gussner

Firmware/first_lay_cal.cpp
leptun: @3d-gussner fixed overextrusion with new pattern. Please check

gudnimg: Reminder for myself: check if that division is done at compile time. If not, is the division done once or twice.

leptun: It was. I extracted that to a constexpr and no size change happened. After that I also switched to 0.5f*. Changed in https://github.com/prusa3d/Prusa-Firmware/pull/4811/commits/7a47bf30be79c6b9e9c042dbd6384eccad28a592

gudnimg (approved):
Approving, I'd like to see @leptun approve as well. Or at least 2nd approval :)

leptun (approved):
Tested both calibration versions in MK404 with fancy extrusion and colored extrusion based on width. Looks good to me. Only tested with 0.4mm nozzle

3d-gussner:
@gudnimg @leptun Thanks for the feedback and review.

jedisct1:
Been running my MK3S+ with `NEW_FIRST_LAYER_CAL` for a while, and it works perfectly.

PR #4813, opened 2024-11-16
Fix an issue in first layer calibration where MMU unloads with disabled hotend heater

gudnimg:
Minor issue I found while reviewing https://github.com/prusa3d/Prusa-Firmware/pull/4811

Change in memory:
Flash: -16 bytes
SRAM: 0 bytes


leptun (approved):
This does feel a bit hacky, but it also simplifies lay1cal_finish quite a bit. So I think we should go forward with this, but not directly. I would move these changes on top of #4811 and close this PR.
Also really nice touch by disabling the bed while unloading the filament👍

leptun:
superseded by #4811 and merged into that.

Issue #4716, opened 2024-07-15
[FEATURE REQUEST] Don't cool down nozzle right away on print pause (M601)


alexiri:
Please, before you create a new feature request, please make sure you searched in open and closed issues and couldn't find anything that matches.

If it makes sense, enter what type of printer or upgrade the feature request applies to.
**Printer type** - all
**MMU Upgrade** - N/A

**Is your feature request related to a problem? Please describe.**
  I'm always frustrated when I pause a print (to add a nut, magnet, etc.) and the nozzle immediately starts to cool down. When resuming a few seconds later, I have to wait for the nozzle to warm back up.

**Describe the solution you'd like**
  After pausing, keep the nozzle hot for a few minutes, in case the user resumes the print quickly enough.

3d-gussner:
@alexiri Thanks for the request, we had similar requests before and there are two sides on that.
1. If you stand next to the printer when it is going to pause, then I agree that the nozzle cool down could be delayed
2. But often you have the pause in the gcode and don't babysit the printer, which means that you come back after some time and then the filament is "cooking".

So the question is what is a good way to improve it for point 1 while preventing the filament being cooked and degrade.

I assign this request to myself and discuss that with the team and filament department.

Can't promise anything, let's see if and what we can do.

Cooling down is a safe way to prevent cooking/degradation of the filament.

alexiri:
How long does the filament take to "cook"? Waiting 60 seconds before cooling might already be good enough.

3d-gussner:
@alexiri Just an update: I got some feedback about the cooking. After looking at the PrusaXL Prusa Slicer Profiles most Idle temperatures, where the tool is waiting, are around 70 -100 °C.
One feedback was that "some" filaments tend to overcook if not extruded but couldn't get any specific values.

We will discuss the idea to give the pause a short time period before cooling down. Maybe set some Idle temperatures similar to the XL profiles to reduce the heat-up time from cold/idle to print temperatures during pauses.

vincele:
Maybe only do a partial cooldown to avoid filament cooking, but stay warm enough to be able to start printing again quickly.

vincele:
Any news on that ? I often pause for ~30secs and the temp gets lowered a lot, which is annoying.

Another idea would be (don't know if it's possible) to understand that it's an "interactive pause" (F.E.: clicking the "pause" button in octoprint).

PR #4815, opened 2024-11-22
:construction: WIP: Add HEATER_LOWER_TEMP 50 

3d-gussner:
to lower instead of turining off the hotend temperature during a pause. Thanks for the request #4716 by @alexiri

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -9877,7 +9877,11 @@ void long_pause() //long pause print
 
     // Stop heaters
     heating_status = HeatingStatus::NO_HEATING;
+#ifdef HEATER_LOWER_TEMP
+    setTargetHotend(degHotend(active_extruder)- HEATER_LOWER_TEMP);
+#else
     setTargetHotend(0);
+#endif //HEATER_LOWER_TEMP
 
     // Lift z
     raise_z(pause_position[Z_AXIS]);
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -9878,7 +9878,7 @@ void long_pause() //long pause print
     // Stop heaters
     heating_status = HeatingStatus::NO_HEATING;
 #ifdef HEATER_LOWER_TEMP
-    setTargetHotend(degHotend(active_extruder)- HEATER_LOWER_TEMP);
+    setTargetHotend(target_temperature[active_extruder]- HEATER_LOWER_TEMP);
 #else
     setTargetHotend(0);
 #endif //HEATER_LOWER_TEMP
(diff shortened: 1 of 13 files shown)


Firmware/variants/MK25-RAMBo10a.h
```diff
@@ -184,6 +184,9 @@
 #endif
 #define BED_MAXTEMP 125
 
+//Lowertemp
+#define HEATER_LOWER_TEMP 50 //lower hotend temperature by 50°C
```
gudnimg, line 188: I think this new setting should be off by default so that we do not change the default behavior. Users who know what they are doing can then change this setting on their own in my opinion.

----
I feel this description could be more detailed. This doesn't really tell me what the setting does, and I would have look at the source code to understand. They way I see it, the comment/description here is basically our documentation.

Suggested description:
```
Uncomment to change the maximum allowed temperature drop while a print is paused. The default behavior sets the target temperature to 0°C.
Example: HEATER_LOWER_TEMP = 10 and target temperature is 215°C when the print was paused. The target temperature will then be set to 205°C instead of 0°C.
```

Firmware/Marlin_main.cpp
```diff
@@ -9877,7 +9877,11 @@ void long_pause() //long pause print
 
     // Stop heaters
     heating_status = HeatingStatus::NO_HEATING;
+#ifdef HEATER_LOWER_TEMP
+    setTargetHotend(target_temperature[active_extruder]- HEATER_LOWER_TEMP);
```
gudnimg, line 9881: I think we need some protection here to cover edges cases. This can produce negative values: `HEATER_LOWER_TEMP > target_temperature[active_extruder]`

What happens if we are coming from a temperature error? `ThermalStop()` can sometimes pause prints.

gudnimg (changes requested):

Firmware/Marlin_main.cpp
3d-gussner: Very good point, thanks. Fan check, uvlo and ThermalStop call `lcd_pause_print` --> `LcdCommands::LongPause` -> long_pause`

leptun: Might want to subtract only if the target temperature is >= HEATER_LOWER_TEMP. Otherwise you might underflow the target temperature in even more scenarios where the target is a small number

Firmware/variants/MK25-RAMBo10a.h
leptun: Also agree with off by default. Was a bit confused about the feature until I read the source code for the PR (recall discussion from last meeting). I like the description provided by @gudnimg 👍

PR #4839, opened 2025-01-24
Align MIN_Z raise for unload & load

### LINKED ISSUE #4838: [BUG] After powering on, Z axis moves up twice during filament change
### Printer model

MK3S/+

### MMU model

No-MMU

### Firmware versions

3.14.1

### Upgrades and modifications

none

### Printing from...

SD Card

### Describe the bug

Minor annoyance: After turning on the printer, the Z axis moves up twice while I'm changing the filament: once when unloading, once when loading. The second one isn't really necessary, and can result in the Z axis topping out if it's already near the top.

Again, not a big deal but hopefully an easy fix. And it only seems to happen after turning the power on. I thought performing an auto-home after startup would be a viable workaround (assumption: firmware doesn't know actual Z height until homed) but it does it in that case too.

### How to reproduce

1. Power on the printer
2. Select "unload filament" (printhead heats up)
3. Press knob to eject filament (printhead moves up and ejects filament)
4. Insert new filament (printhead pulls in filament, then moves up _again_)

### Expected behavior

1. Power on the printer
2. Select "unload filament" (printhead heats up)
3. Press knob to eject filament (printhead moves up and ejects filament)
4. Insert new filament (printhead pulls in filament, with no additional Z movement)

### Files

_No response_

sarusani:
Align values for MIN_Z raise, to prevent movement between unload & load operations.

Fixes #4838

### DIFF
--- a/Firmware/Configuration.h
+++ b/Firmware/Configuration.h
@@ -358,8 +358,8 @@ your extruder heater takes 2 minutes to hit the target on heating.
 
 // Try to maintain a minimum distance from the bed even when Z is
 // unknown when doing the following operations
-#define MIN_Z_FOR_LOAD    50 // lcd filament loading or autoload
-#define MIN_Z_FOR_UNLOAD  20 // lcd filament unloading
+#define MIN_Z_FOR_LOAD    35 // lcd filament loading or autoload (values for load and unload have been unified to prevent movement between unload & load operations!)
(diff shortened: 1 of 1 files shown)


3d-gussner (changes requested):
As `MIN_Z_FOR_SWAP` is also filament related maybe it makes sense also to allign that one as well

3d-gussner (approved):

PR #4846, opened 2025-02-11
M300 set 3.5s duration limit

3d-gussner:
Solves https://github.com/prusa3d/Prusa-Firmware/issues/2045

**Post release tasks**
- [ ]  Update doxygen documentation
- [ ] Update RepRap Gcode wiki

### DIFF
--- a/Firmware/Marlin_main.cpp
+++ b/Firmware/Marlin_main.cpp
@@ -6515,11 +6515,11 @@ void process_commands()
 
     #### Parameters
     - `S` - frequency in Hz. Not all firmware versions support this parameter
-    - `P` - duration in milliseconds
+    - `P` - duration in milliseconds max 3500ms
     */
     case 300: // M300
     {
-      uint16_t beepP = code_seen('P') ? code_value() : 1000;
+      uint16_t beepP = code_seen('P') ? min(code_value(), 3500) : 1000;
       uint16_t beepS;
       if (!code_seen('S'))
           beepS = 0;


leptun (commented):
Choosing 4000ms might be a bit risky given that it exactly matches what the watchdog timer is specified to be running at. However, this may still be fine given the following:
- According to avrlibc, we set a timeout of 4 seconds, but that is not entirely correct. The datasheet specifies the watchdog crystal at 128KHz (that is 128000Hz) and the prescaler that we use to get 4s of timeout is /512K (that is 524288, not 512000!!), so the timeout is in fact set to 4.096s which is slightly bigger than the max beep time chosen. Still cutting it close at only ~96ms to spare. This obviously assumes the timer is running at the nominal speed, but that is never the case, which brings us to the next note I have.
- The watchdog timer speed is not only influenced by VCC voltage, but also by temperature. Here are the speeds of the timer under various conditions from the datasheet:
image
As can be seen there, the speed is always less than 128KHz, so we have a bit of extra margin at all times, so this may be good enough as a solution.

3d-gussner:
@leptun I tested on MK404 and at `M300 P4505 S0` it resets the printer `P4504` survives. That is why I have chosen `4000` instead of `4500`

3d-gussner:
@leptun Which value would you prefer? 3500 3750?

leptun:
3500 sounds good 👍

kimtwio (approved):
Sorry, wrong account ... :)

sarusani (approved):
Tested on MK3S

PR #4832, opened 2025-01-08
Steel sheet check

3d-gussner:
Thanks @jfestrada for the idea in #4569. 

This is a different approach for this request.

We have the https://help.prusa3d.com/materials (which is missing the PP powder coated steel sheet) and we could use it to warn users when they start to print:

- [X] Add new parameter Type to the `M850`  gcode and `HW Setup -> Sheets` to set the type of sheet
  - [X] Smooth
  - [X] Textured
  - [X] Satin
  - [X] NylonPA
  - [X] PP
  - [X] Custom

  - [X] Reuse and add eeprom values needed
    - The Bed and PINDA temp values within the sheets eeprom aren't / have never been used. Replace these Sheet type
previous values are `>10` as the MINtemp for Bed would not allow any lower values so resetting these with `255` or default values `0-6` is not an issue

  - [X] Also add `M862.7` to check the compatible sheets with parameter `P`. In the PrusaSlicer Filament profiles we could add it to the custom G-code
As the materials help article shows compatibility :heavy_check_mark:  and warnings :grey_exclamation:and not to use :x: I would suggest a bit based value
bit 0 = Smooth compatible
bit 1 = Textured compatible
bit 2 = Satin compatible
bit 3 = NylonPA compatible
bit 4 = PP compatible
bit 5 = Custom sheet compatible (like G10 sheets see https://www.youtube.com/watch?v=g0PK4oXbJT8)

- [X] Add an additional Check state `Always` or `Wait` which is the same as `Warn` but doesn't timeout and HAS to be confirmed on LCD
  - [X] With `Always` any `M115` Gcode will trigger the LCD message to be confirmed.

- [X] Of course the default setting for the Sheet type check would be warn (with timeout)

**ToDos:**
  - [x] Bit check `M862.7 P` 
  - [x] And an additional `M862.7` parameter `Warn` with same bit based value
       - Example:
       - PLA : M862.7 P7 W0 PLA is compatible on Smooth, Textured and Satin,  no additional warnings BUT NOT compatible on NylonPA and PP
       - ASA: M862.7 P4 W3 ASA is compatible on Satin, warning to use glue stick on Smooth and Textured and BUT NOT compatible on NylonPA

- [x] In case there is a warning or stop print send also an `//action:notification` to inform host users.

In PrusaConnect we already have some sheet information

- [ ] Another check we could add is in the Preheat menu, so users which use that will be warned before the starting a print or 1st layer calibration.

- [ ] Update Materials help page as PP sheet is missing
[...]

PR #4879, opened 2025-09-06
Add M88 G-code to align Z-axis to full-step position

### LINKED ISSUE #2876: [ENHANCEMENT] Should align Z axes to a fullstep when idling/poweroff
**Printer type** - MK3S, MK3 (partially MK2.5S and earlier)
**MMU Upgrade** - N/A

We should ensure the Z motors are disengaged in a full step, when possible.

If the motors happen to be halfway between a fullstep, there's a good chance that two motors will snap to opposing fullsteps, skewing the axis. This is not usually a problem, since a single fullstep of difference with the leadscrew doesn't lead to significant changes (and these changes get compensated by the MBL anyway), however if this keeps happening it will create additional strain on the POM nuts, eventually leading to backlash.
 
The printer should re-align Z to a fullstep before powering off Z when possible and sensible to do so.
Note that this is already implemented when PowerPanic kicks in, in order to preserve an accurate position after the power-cycle. It just needs to be re-used.

On the MK3S we can detect when the power is switced off by the button. In this case, and if the Z motors are still powered, alignment should be performed. On the MK3 this is only possible if the power chord is pulled, but the same reasoning applies.

It's harder to say what should be done with a M84 (release steppers) and with the inactivity shutdown timer (M85), since this causes an unexpected move which is not requested. I feel that M84 should release the steppers immediately, but M85 could be allowed to re-align all axes before release, as long as we return to the original position when powering them on again.

A good interim solution IMHO would be to add a new M-code to perform the stepper alignment. Such code should be called by the printer end-gcode section just prior to releasing the steppers. This is simple, easy, and avoids skew build-up in the most common case.

MathiasGyldenberg:
This pull request introduces a new G-code command `M88` that moves the Z-axis to the nearest full-step position. This enhancement addresses [Issue #2876](https://github.com/prusa3d/Prusa-Firmware/issues/2876), which proposes a method to reduce backlash and improve Z-axis reliability when steppers are disabled.

### Details

- New and unused G-code M88.
- Logic copied from power panic feature.
- Ensures the target position is within Z bounds before executing the move.
- Adds a new LCD menu item labeled `"Fullstep Z axis"` to trigger `M88` manually.
- Message string defined in `messages.cpp` and declared in `messages.h`.

### Benefits

- Reduces Z-axis skew and backlash by aligning Z motors to nearest full-step.
- Provides users with manual control via LCD menu as well as g-code command.

### Notes

- Tested on MK3S+ in both menu and end G-code.

---

### Related Issue

Fixes [#2876](https://github.com/prusa3d/Prusa-Firmware/issues/2876)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,8 +8,6 @@ on:
     branches: [ MK3, MK3_* ]
     tags:
     - 'v*'
-    - 't*'
-    - 'c*'
 
 env:
   GH_ANNOTATIONS: 1
@@ -43,7 +41,7 @@ jobs:
         fetch-depth: 0
 
     - name: Cache Dependencies
-      uses: actions/cache@v4
+      uses: actions/cache@v4.0.0
       id: cache-pkgs
       with:
           path: ".dependencies"
@@ -65,17 +63,16 @@ jobs:
 
     - name: Upload artifacts
       if: ${{ !github.event.pull_request }}
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@v3.1.1
       with:
         name: Firmware
         path: build/*.hex
 
     - name: RELEASE THE KRAKEN
-      if: startsWith(github.ref, 'refs/tags/v') || startsWith(github.ref, 'refs/tags/t') || startsWith(github.ref, 'refs/tags/c')
+      if: startsWith(github.ref, 'refs/tags/v')
       uses: "marvinpinto/action-automatic-releases@latest"
       with:
         repo_token: "${{ secrets.GITHUB_TOKEN }}"
-        automatic_release_tag: ${{ github.ref_name }}
         draft: true
         files: |
           ${{ github.workspace }}/build/release/*.hex
@@ -109,7 +106,7 @@ jobs:
         fetch-depth: 0
 
     - name: Cache Dependencies
-      uses: actions/cache@v4
+      uses: actions/cache@v4.0.0
       id: cache-pkgs
       with:
           path: ".dependencies"
@@ -157,7 +154,7 @@ jobs:
         fetch-depth: 0
 
     - name: Cache Dependencies
-      uses: actions/cache@v4
+      uses: actions/cache@v4.0.0
       id: cache-pkgs
       with:
           path: ".dependencies"
--- a/.github/workflows/pr-size.yml
+++ b/.github/workflows/pr-size.yml
@@ -26,7 +26,7 @@ jobs:
       uses: actions/checkout@v4
 
     - name: Cache Dependencies
-      uses: actions/cache@v4
+      uses: actions/cache@v4.0.0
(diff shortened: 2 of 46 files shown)


3d-gussner:
@MathiasGyldenberg Thanks for the PR.
Looking at the https://reprap.org/wiki/G-code I would suggest to look for another gcode that might do something similar and add a parameter. This would reduce the introduction of new gcodes.

Copy the part from the power panic also uses more resources. It would probably save some bytes by making this part a new process and use it in both `uvlo` and the gcode part.

MathiasGyldenberg:
Thanks for your feedback @3d-gussner

I agree that adding this as a parameter to an existing G-code would simplify things. I’ve reviewed the list, but from my perspective, none of the existing commands align well with the intended purpose of aligning Z steppers to nearest fullstep.

I’m open to suggestions if there’s a better fit than introducing a dedicated command.

I’ll also refactor the logic into a reusable function so it can be called both from the G-code and power-panic.

3d-gussner:
`G88` has been reserved and has no function yet. As it is close to other step related gcodes. This would be a good one.

MathiasGyldenberg:
Good one with G88 - Haven't found out what it was intented to in the first place.
I've updated the commit to replace M88 with G88 as suggested. The related logic has now been refactored into a dedicated function for better clarity and reuse.

Fully tested on MK3S+

3d-gussner:
Please rebase your fork.
- `git remote -v` to see if you have prusa3d as upstream repository
- `git fetch --all` to get changes from the prusa3d repo
- `git rebase upstream/MK3` to update your branch `Fullstep-Z` 

there are too many changes in this PR

3d-gussner:
Other thing would be to update the translations. See https://github.com/prusa3d/Prusa-Firmware/blob/MK3/lang/README.md

I usually make a backup of the existing `pot` and `po` files to `pot_bak` and `po_bak` before running the python scripts
- then rename the new `pot` and `po` files to `pot_new` and `po_new`
- rename the `_bak` files back to original 
- and use `Meld` to add the new messages

I know it sounds complicated BUT the translation change review is then only limited to the new messages. Here an example https://github.com/prusa3d/Prusa-Firmware/pull/4832/files#diff-14fedb8dfdce511a0bdcf251bbe0fc1c53d69fcec094186519cc6f57826b4b1b

Before a release when everything has been locked down I run the translation scripts to update everything.

But if you need some help I can do it for you and create a PR into your branch.

MathiasGyldenberg:
Thanks. I'm glad you mentioned if I need some help. My experience using GitHub is very limited, so any help is highly appriciated.

MathiasGyldenberg:
Again thank you for your guide and help – well described.

I've updated the commit with translations, with a little translation-help from Copilot. Let me know if anything is needed from my side, and I'll be happy to assist further or make any adjustments.

